PR #23, opened 2017-01-17
Add `dig` to dnsperf image, somehow was missing

bowei:


MrHohn:
`go get ./...` is broken though.

gmarek:
Yeah - it is #26

PR #25, opened 2017-01-27
Copied compare tool from k8s.io/contrib and updated its Godeps

shyamjvs:
Copied the compare tool to perf-tests repo from contrib, as it makes more sense to have this tool here.
Updated its dependencies (mainly k8s.io/*) to newer revisions and added/removed used/unused packages.
Also migrated the dependencies from Godeps/_workspace to vendor, since _workspace support would be deprecated starting from go1.8.

cc @k8s-sig-testing-misc @wojtek-t @gmarek


gmarek:
Travis is unhappy.

shyamjvs:
Well, it seems like Travis was never too happy -> https://github.com/kubernetes/perf-tests/pulls. I opened issue #26 for this.
And for this PR, after some 1:1s with misery, finally got the godeps right. You can check this by doing the following:
- Start with a clean `GOPATH` (don't necessarily have to do this unless you are super keen about correctness)
- `git clone` this PR
- `godep restore`
- `rm -rf Godep/ vendor/`
- godep save ./...
- `git diff` (to check no dependencies have been newly added/removed)
- `make` (also `make test` if you want to)
- `./build/compare --left-build-number=582 --right-build-number=581` eg. command to test the binary

godep_is_coming2

shyamjvs:
@gmarek You might have to manually merge this, since travis is failing.

Issue #26, opened 2017-01-28
Travis is unhappy


shyamjvs:
Seems like Travis is failing always due to `go get ./...` and `go test ./..` commands run against the repo.
This is because we use glide as our dep manager instead of godep in `clusterloader` which works differently.
We need to fix this by doing one of the following:
- Use glide's get and test equivalents for clusterloader and go for all the other directories
- Move clusterloader to godep, so we have uniformity across the whole repo

cc @kubernetes/sig-scalability-misc @gmarek @spxtr

gmarek:
@sjug - it's failing on Godeps for cluster loader. Can you fix that?

sjug:
@gmarek `clusterloader` has it's own vendor directory in `./clusterloader` that seems to be the issue.

I'm not sure why the package has issues I'm taking a look into it now, but in the meantime maybe: https://github.com/golang/go/issues/11659#issuecomment-171678025
```
go get $(go list ./... | grep -v /vendor/)
go test $(go list ./... | grep -v /vendor/)
```

timothysc:
godep smashing e2e staging dep-tree of doooom -
https://github.com/kubernetes/kubernetes/issues/43246#issuecomment-291532473

sjug:
@timothysc even with your solution I'm still seeing errors around containernetworking, specifically:

```
# github.com/kubernetes/perf-tests/clusterloader/vendor/github.com/containernetworking/cni/pkg/invoke
vendor/github.com/containernetworking/cni/pkg/invoke/exec.go:64: cannot use "github.com/containernetworking/cni/pkg/types".Result as type "github.com/kubernetes/perf-tests/clusterloader/vendor/github.com/containernetworking/cni/pkg/types".Result in return argument:
        "github.com/containernetworking/cni/pkg/types".Result does not implement "github.com/kubernetes/perf-tests/clusterloader/vendor/github.com/containernetworking/cni/pkg/types".Result (wrong type for GetAsVersion method)
                have GetAsVersion(string) ("github.com/containernetworking/cni/pkg/types".Result, error)
                want GetAsVersion(string) ("github.com/kubernetes/perf-tests/clusterloader/vendor/github.com/containernetworking/cni/pkg/types".Result, error)
```

gmarek:
@sjug - after @wojtek-t fixes it seems that godep issue is gone. The thing that's wrong now is a failing test.

gmarek:
[...]

PR #40, opened 2017-04-24
Dependency nightmare

sjug:
All this to try and make [Travis happy](https://github.com/kubernetes/perf-tests/issues/26).


sjug:
@gmarek this is as good as it's going to get. If you want Travis to pass recursive go test, then you'll have to ask https://github.com/kubernetes/kubernetes how they did it.

gmarek:
They did it by flattening the repos (as @timothysc tool should have done).

```
clusterloader/vendor/k8s.io/client-go/rest/request.go:21:2: cannot find package "context" in any of:
	/home/travis/gopath/src/k8s.io/perf-tests/clusterloader/vendor/context (vendor tree)
	/home/travis/.gimme/versions/go1.6.linux.amd64/src/context (from $GOROOT)
	/home/travis/gopath/src/context (from $GOPATH)
```

sjug:
Which tool is that?

timothysc:
@sjug this isn't godep, this is the main repo doing unholy "staging" things that absolutely 0 projects do.  As a result there needs to be some script hackery to bypass.  My script flattens the dep-chain.

sjug:
@timothysc what/where is your script?

timothysc:
In your repo you are trying to update: 

``` 
rm -rf Godeps vendor
export KPATH=$GOPATH/src/k8s.io/kubernetes
export GOPATH=$GOPATH:$KPATH/staging

$KPATH/hack/generate-bindata.sh
godep save ./...
cp -L -R $KPATH/vendor/k8s.io/api* $KPATH/vendor/k8s.io/client-go ./vendor/k8s.io/
cp $KPATH/test/e2e/generated/bindata.go ./vendor/k8s.io/kubernetes/test/e2e/generated/bindata.go
```

sjug:
@timothysc I ran all that to generate this PR, of course.

wojtek-t:
@gmarek @sjug - so what is the reason behind having a separate "vendor" directory under clusterloader/ ?
Why can't we have just a single top-level vendor/ directory?

gmarek:
We tried that in contrib, and it doesn't work. Imagine that you have an old *thing* that uses version X of something. Then you write a *another thing* which uses version X+1 of the same something. If you have a single vendor it's just pain and tears (even more than having separate ones).

wojtek-t:
I see - thanks.

That said, it's clear that something is still missing in this PR. It seems that the code doesn't see the golang "context" package at all.

sjug:
I don't understand the purpose of running `go get` and `go test` on vendor.

I have used @timothysc commands to create vendor and this is the end result. How is it possible that it is unable to find a built-in package?

wojtek-t:
@sjug - I never knew how to use godeps. so I'm not really able to help with this problem.

gmarek:
I copied it from somewhere, but I agree - no point of running `go get/test` on vendor. Feel free to send a PR that skips them.

sjug:
@gmarek Funnily enough Travis is now failing on the same issue with other packages. Did you want me to fix them in this PR or leave it for the original author?

wojtek-t:
I'm fine with merging this as is, though if you could fix them in a separate PR it would be awesome (so that we have passing travis).

sjug:
@wojtek-t If we bring up golang from 1.6 then I won't need the last commit fix, nor will the other package so that works too :smile:

wojtek-t:
@gmarek - it seems that after bumping to Go 1.7 it is worse, a bunch of other things doesn't compile, e.g.:

# k8s.io/perf-tests/clusterloader/vendor/github.com/Microsoft/go-winio
clusterloader/vendor/github.com/Microsoft/go-winio/file.go:45: undefined: syscall.Overlapped

gmarek:
The same thing broke in the benchmark. I'm kind of curious what this packet actually is...

shyamjvs:
Even in general, it might be a good practice (not 100% sure) to check just if all dependency packages are downloadable when we are running go get. Whether the tools inside the repo actually compile is a headache that we should delegate to the makefiles inside those repos.

sjug:
So for the missing packages according to `go test`:
"github.com/elazarl/goproxy"
"github.com/stretchr/testify/assert"

These are not pulled in by `godep`, should I add them manually to vendor?

shyamjvs:
@sjug Not sure if I understand it totally, but something's wrong if you're needing to vendor manually. Because this would mean godep failed to identify those dependencies during `godep save ./...`.

wojtek-t:
@sjug - we are finally at the state where the errors seem reasonable:
```
# k8s.io/perf-tests/clusterloader
clusterloader/cluster_loader.go:78: cannot use f (type *"k8s.io/perf-tests/clusterloader/vendor/k8s.io/kubernetes/test/e2e/framework".Framework) as type *"github.com/kubernetes/perf-tests/clusterloader/vendor/k8s.io/kubernetes/test/e2e/framework".Framework in argument to "github.com/kubernetes/perf-tests/clusterloader/framework".CreatePods
clusterloader/cluster_loader.go:78: cannot use config.Spec (type "k8s.io/perf-tests/clusterloader/vendor/k8s.io/kubernetes/pkg/api/v1".PodSpec) as type "github.com/kubernetes/perf-tests/clusterloader/vendor/k8s.io/kubernetes/pkg/api/v1".PodSpec in argument to "github.com/kubernetes/perf-tests/clusterloader/framework".CreatePods
```

Can you please take a look why we have those?

sjug:
Taking a look at this, it's very strange because I don't get any errors building at all?

wojtek-t:
> Why does the error message lead with # k8s.io/perf-tests/clusterloader? That's not the path... 

Sorry, I don't udnerstand. The repo is k8s.io/perf-tests, main directory is clusterloader, so it seems that it is the path.

> Are you guys able to reproduce it locally?

Didn't try to patch your PR and I'm afraid I won't have time to patch it in the next few days.

wojtek-t:
OK I'm going to merge this now, since the state with this PR is better than without it.

But we should fix this ASAP.
@sjug - will be able to look into it?

shyamjvs:
So I tried running `godep restore` with this change and followed it up with a `godep save ./...` to check the integrity of the migration. And the save failed with the following:
```
godep: Package (github.com/aws/aws-sdk-go/private/endpoints) not found
```

I guess something's wrong with the way godeps have been created. Besides I created godeps from glide deps with the wrong paths fixed last week (https://github.com/shyamjvs/perf-tests/commit/fcba424962cad452c96a7ac3c4afd1b7cfba1df7) and the size of my `/vendor` dir was 17MB smaller. Did you do any manual tweaking with the godep @sjug ?

sjug:
@wojtek-t but the path is set to `github.com/perf-tests/clusterloader`?

@shyamjvs I don't think it's simply restorable as vendoring in k8s is not that simple, I had to use this script https://github.com/kubernetes/perf-tests/pull/40#issuecomment-297142963

wojtek-t:
> @wojtek-t but the path is set to github.com/perf-tests/clusterloader?

As I wrote somewhere already, I don't understand godeps. But we are refering to all k8s repos as:
k8s.io/<repo>/<dir> and changing it here fixed the problem. And it works now, while it didn't work (also for me locally) before.

PR #44, opened 2017-04-27
Make 'go get' in travis script only download packages

shyamjvs:
Ref https://github.com/kubernetes/perf-tests/pull/40

This should hopefully fix the issue with `github.com/Microsoft/go-winio` dependency for benchmark, compare and clusterloader.

cc @gmarek @wojtek-t @sjug


shyamjvs:
@gmarek Well at least you should've waited for Travis to finish :)

gmarek:
It'd have failed because of cluster-loader.

shyamjvs:
But we could've checked if it fixed the part it needed to :)

sjug:
You realize it still breaks all the same for go test, right?

```
# k8s.io/perf-tests/benchmark/vendor/github.com/Microsoft/go-winio
benchmark/vendor/github.com/Microsoft/go-winio/file.go:45: undefined: syscall.Overlapped
# k8s.io/perf-tests/benchmark/vendor/github.com/seccomp/libseccomp-golang
benchmark/vendor/github.com/seccomp/libseccomp-golang/seccomp.go:25:22: fatal error: seccomp.h: No such file or directory
compilation terminated.
``

shyamjvs:
Yes that's true. That is why I was suggesting delegating to Makefiles of individual directories. It might not be the best solution (as we'd have to iterate over top level diretories), but I guess works good enough until we flatten godeps across the entire repo. For the size of this repo this might be ok to do (no?)

PR #49, opened 2017-05-05
Add mechanism for benchmark to filter metric samples with low request count

shyamjvs:
Ref https://github.com/kubernetes/kubernetes/issues/44701#issuecomment-297537285
The default is to filter those latency samples with less than 10 api requests.

cc @gmarek @wojtek-t


shyamjvs:
The travis `go build` step works fine for me:

```
$ time go build $(go list ./... | grep -v "/vendor/")

real	1m36.218s
user	13m37.788s
sys	0m52.288s
```

It says timeout after waiting for 10m. Let's try again and see if it fails (which could mean it counts user time as opposed to real time).

shyamjvs:
@gmarek I've fixed the unit tests and the travis timeout issue. Can you PTAL?

PR #50, opened 2017-05-05
[Do not merge] Check no. of cores available on travis

shyamjvs:
Ref #49 
Seems like `go build $(go list ./... | grep -v "/vendor/")` is taking too long to execute (> 10m) and hence timing out.
Let's check if we have enough cores available on jenkins for parallel execution.

cc @gmarek @wojtek-t

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ install:
     - export PATH=$GOPATH/bin:$PATH
 
 script:
+    - grep -c ^processor /proc/cpuinfo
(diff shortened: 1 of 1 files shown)


shyamjvs:
Ok, so it seems like travis is having only 2 cores available while my local machine has 12 cores. I guess that explains why travis is timing out for #49 .

shyamjvs:
Travis wait command fixed the issue. Closing this.

PR #51, opened 2017-05-05
Add travis wait to 'go build' command for avoiding timeout

shyamjvs:
Ref #49 #50 

Following instruction here: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received.
This should make travis print something to stdout once every minute to ensure longer run.

cc @gmarek @wojtek-t


wojtek-t:
Let's see if that helps.

/lgtm

PR #52, opened 2017-05-09
Basic compile & run script

sjug:
Matches the CI filename, just build and run cluster loader.

The `/config/test.yaml` is not very useful. I will either edit this PR with the new config or have another one that follows shortly.

The run command assumes that k8s has the unencrypted port open (8080) as well as the kubeconfig at the default location of `~/.kube/config`

Nothing is parameterized at this point.

### DIFF
--- a/clusterloader/run-e2e.sh
+++ b/clusterloader/run-e2e.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+cd ${KUBE_ROOT}/e2e/ && go test -c -o e2e.test
+./e2e.test --host="${KUBE_MASTER_URL}" --ginkgo.v=true --ginkgo.focus="Cluster Loader" --kubeconfig="${HOME}/.kube/config" --viper-config=../config/test
+


clusterloader/run-e2e.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
wojtek-t: You need this copyright boilerplate, as e.g. in:
https://github.com/kubernetes/perf-tests/blob/master/verify/test.sh

clusterloader/run-e2e.sh
```diff
@@ -0,0 +1,6 @@
...
+set -ex
```
wojtek-t: What's that?

clusterloader/run-e2e.sh
```diff
@@ -0,0 +1,6 @@
...
+
+cd e2e/ && go test -c -o e2e.test
```
wojtek-t: We shouldn't rely on the fact that the script is being run from the current directory.
Instead we should do something similar as we are doing in:
https://github.com/kubernetes/perf-tests/blob/master/verify/test.sh

which is:
KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
cd ${KUBE_ROOT}

clusterloader/run-e2e.sh
```diff
@@ -0,0 +1,6 @@
...
+./e2e.test --host="127.0.0.1:8080" --ginkgo.v=true --ginkgo.focus="Cluster Loader" --kubeconfig="${HOME}/.kube/config" --viper-config=../config/test
```
wojtek-t: I think --host=127.0.0.1:8080 is not correct, because in general in jenkins the test is not running on the master machine

sjug: How can we locate the api server?

wojtek-t: I would try to replicate the logic from:
https://github.com/kubernetes/kubernetes/blob/master/hack/ginkgo-e2e.sh

It seems ${KUBE_MASTER_URL} should do the job.

BTW - there might be still some missing flags here, but I guess we will figure it out once this is submitted and we will look into logs from jenkins runs - I don't know how to easily do it differently.

clusterloader/run-e2e.sh
sjug:
```
-e  Exit immediately if a command exits with a non-zero status.
-x  Print commands and their arguments as they are executed.
```

clusterloader/run-e2e.sh
sjug: Is `${KUBE_ROOT}` the root directory of `perf-tests`?

clusterloader/run-e2e.sh
sjug: Done

clusterloader/run-e2e.sh
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+cd ${KUBE_ROOT}/e2e/ && go test -c -o e2e.test
+./e2e.test --host="${KUBE_MASTER_URL}" --ginkgo.v=true --ginkgo.focus="Cluster Loader" --kubeconfig="${HOME}/.kube/config" --viper-config=../config/test
```
sjug: Kubeconfig is probably not in the default location either

wojtek-t: That one I don't now - it seems that the script from main repo is not setting it at all (so maybe it is?)

wojtek-t:
/lgtm

Will merge on green tests and we will see what will happen. We can iterate on this if this won't work.

PR #53, opened 2017-05-11
Adding dockerization for benchmark tool

shyamjvs:
First step towards adding a CI job for the tool.

cc @kubernetes/sig-scalability-misc @wojtek-t @gmarek


shyamjvs:
@gmarek Removed the latest tag for the image as discussed offline. PTAL.

PR #55, opened 2017-05-15
RCs added, resize working, labels exposed

sjug:
PR to add some of the functionality as per our design discussions a few weeks ago.

First of all, there is now a tuningset to support a delay between creation of each new defined project.
We also support creation of RCs and modification of RCs. To do this we had to be able to recognize existing namespaces, as well as existing RCs, so there's less blind creation going on now. The generic side effect is that we can now define different "projects" with the same name and then have different objects deployed in each.

The RC makes use of the image field which was not utilized for pod definitions. 
Labels have also been exposed to the config so the user can now set their own.

### DIFF
--- a/clusterloader/cluster_loader.go
+++ b/clusterloader/cluster_loader.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 The Kubernetes Authors.
+Copyright 2017 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,14 +17,11 @@ limitations under the License.
 package clusterloader
 
 import (
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strconv"
-	"time"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
@@ -57,31 +54,63 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
 		// TODO sjug: add concurrency
 		for _, p := range project {
 			// Find tuning if we have it
-			tuning := getTuningSet(tuningSets, p.Tuning)
-			framework.Logf("Our tuning set is: %v", tuning)
+			tuning := clusterloaderframework.TuningSets(tuningSets).Get(p.Tuning)
+
+			framework.Logf("Tuning set is: %+v", tuning)
 			for j := 0; j < p.Number; j++ {
-				// Create namespaces as defined in Cluster Loader config
+				// Create namespaces as defined in the config
 				nsName := appendIntToString(p.Basename, j)
-				ns, err := f.CreateNamespace(nsName, nil)
-				framework.ExpectNoError(err)
-				framework.Logf("%d/%d : Created new namespace: %v", j+1, p.Number, nsName)
-				namespaces = append(namespaces, ns)
+				ns, err := clusterloaderframework.CreateNSIfNotExists(f, nsName)
+				if err != nil {
+					framework.Failf("Error creating NS: %v", err)
+				}
+				// Keep track of all the namespaces we have created, not too useful currently
+				namespaces = appendUnique(namespaces, ns)
 
 				// Create templates as defined
 				for _, template := range p.Templates {
-					createTemplate(template.Basename, ns, mkPath(template.File), template.Number, tuning)
+					if err = createTemplate(template.Basename, ns, clusterloaderframework.MakePath(template.File), template.Number, tuning); err != nil {
+						framework.Failf("Error creating template, %v", err)
+					}
+				}
+				// RCs are a thing as well
+				for _, RC := range p.RCs {
+					config, err := RC.ParseConfig()
+					if err != nil {
+						framework.Failf("Error parsing config, %v", err)
+					}
+					label, err := RC.ConvertToLabelSet()
+					if err != nil {
+						framework.Failf("Error creating Labels, %v", err)
+					}
+					if err = clusterloaderframework.CreateRC(f, RC.Basename, ns.Name, label, config.Spec, RC.Number); err != nil {
+						framework.Failf("Error creating RC, %v", err)
+					}
 				}
 				// This is too familiar, create pods
 				for _, pod := range p.Pods {
-					config := parsePods(mkPath(pod.File))
-					labels := map[string]string{"purpose": "test"}
-					clusterloaderframework.CreatePods(f, pod.Basename, ns.Name, labels, config.Spec, pod.Number, tuning)
+					config, err := pod.ParseConfig()
+					if err != nil {
+						framework.Failf("Error parsing config, %v", err)
+					}
+					label, err := pod.ConvertToLabelSet()
+					if err != nil {
+						framework.Failf("Error creating Labels, %v", err)
+					}
+					clusterloaderframework.CreatePods(f, pod.Basename, ns.Name, label, config.Spec, pod.Number, tuning)
 				}
+			}
+			// Only sleeps for each new project defined in the config
+			// need to move up to sleep for every copy
+			// TODO: Consider if we want sleeps between each iteration
+			if tuning != nil {
+				tuning.Project.Delay()
 			}
 		}
 
-		// Wait for pods to be running
+		// Wait for pods to be running in all new namespaces
 		for _, ns := range namespaces {
+			// TODO If created namespace doesn't have a pod with matching label we will hang
 			label := labels.SelectorFromSet(labels.Set(map[string]string{"purpose": "test"}))
 			err := testutils.WaitForPodsWithLabelRunning(c, ns.Name, label)
 			if err != nil {
@@ -93,60 +122,41 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
 	})
 })
 
-// getTuningSet matches the name of the tuning set defined in the project and returns a pointer to the set
-func getTuningSet(tuningSets []clusterloaderframework.TuningSetType, podTuning string) (tuning *clusterloaderframework.TuningSetType) {
-	if podTuning != "" {
-		// Iterate through defined tuningSets
-		for _, ts := range tuningSets {
-			// If we have a matching tuningSet keep it
-			if ts.Name == podTuning {
-				tuning = &ts
-				return
-			}
+// appendUnique appends new namespace pointers to a slice
+func appendUnique(allNS []*v1.Namespace, newNS *v1.Namespace) []*v1.Namespace {
+	for _, NS := range allNS {
+		if NS.Name == newNS.Name {
+			return allNS
 		}
-		framework.Failf("No pod tuning found for: %s", podTuning)
 	}
-	return nil
-}
-
-// mkPath returns fully qualfied file location as a string
-func mkPath(file string) string {
-	// Handle an empty filename.
-	if file == "" {
-		framework.Failf("No template file defined!")
-	}
-	// TODO: We should enable passing this as a flag instead of hardcoding.
-	return filepath.Join(os.Getenv("GOPATH"), "src/k8s.io/perf-tests/clusterloader/content/", file)
+	return append(allNS, newNS)
 }
 
 // createTemplate does regex substitution against the template file, then creates the template
-func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObjects int, tuning *clusterloaderframework.TuningSetType) {
+func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObjects int, tuning *clusterloaderframework.TuningSet) error {
 	// Try to read the file
 	content, err := ioutil.ReadFile(configPath)
 	if err != nil {
-		framework.Failf("Error reading file: %s", err)
+		return err
 	}
 
 	// ${IDENTIFER} is what we're replacing in the file
-	regex, err := regexp.Compile("\\${IDENTIFIER}")
-	if err != nil {
-		framework.Failf("Error compiling regex: %v", err)
-	}
+	regex := regexp.MustCompile("\\${IDENTIFIER}")
 
 	for i := 0; i < numObjects; i++ {
 		result := regex.ReplaceAll(content, []byte(strconv.Itoa(i)))
 
 		tmpfile, err := ioutil.TempFile("", "cl")
 		if err != nil {
-			framework.Failf("Error creating new tempfile: %v", err)
+			return err
 		}
 		defer os.Remove(tmpfile.Name())
 
 		if _, err := tmpfile.Write(result); err != nil {
-			framework.Failf("Error writing to tempfile: %v", err)
+			return err
 		}
 		if err := tmpfile.Close(); err != nil {
-			framework.Failf("Error closing tempfile: %v", err)
+			return err
 		}
 
 		framework.RunKubectlOrDie("create", "-f", tmpfile.Name(), getNsCmdFlag(ns))
@@ -154,32 +164,18 @@ func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObj
 
 		// If there is a tuning set defined for this template
 		if tuning != nil {
-			if tuning.Templates.RateLimit.Delay != 0 {
-				framework.Logf("Sleeping %d ms between template creation.", tuning.Templates.RateLimit.Delay)
-				time.Sleep(time.Duration(tuning.Templates.RateLimit.Delay) * time.Millisecond)
+			if err := tuning.Templates.Delay(); err != nil {
+				return err
 			}
 			if tuning.Templates.Stepping.StepSize != 0 && (i+1)%tuning.Templates.Stepping.StepSize == 0 {
-				framework.Logf("We have created %d templates and are now sleeping for %d seconds", i+1, tuning.Templates.Stepping.Pause)
-				time.Sleep(time.Duration(tuning.Templates.Stepping.Pause) * time.Second)
+				framework.Logf("We have created %d templates; sleep for %v", i+1, tuning.Templates.Stepping.Pause)
+				if err := tuning.Templates.Pause(); err != nil {
+					return err
+				}
 			}
 		}
 	}
-}
-
-// parsePods unmarshalls the json file defined in the CL config into a struct
-func parsePods(jsonFile string) (configStruct v1.Pod) {
-	configFile, err := ioutil.ReadFile(jsonFile)
-	if err != nil {
-		framework.Failf("Cant read config file. Error: %v", err)
-	}
-
-	err = json.Unmarshal(configFile, &configStruct)
-	if err != nil {
-		framework.Failf("Unable to unmarshal pod config. Error: %v", err)
-	}
-
-	framework.Logf("The loaded config file is: %+v", configStruct.Spec.Containers)
-	return
+	return nil
 }
 
 func getNsCmdFlag(ns *v1.Namespace) string {
--- a/clusterloader/config/rctest.yaml
+++ b/clusterloader/config/rctest.yaml
@@ -1,0 +1,37 @@
+ClusterLoader:
+  delete: true
+  projects:
+    - num: 1
+      basename: clusterproject
+      tuning: default
+      RCs:
+        - num: 5
+          image: gcr.io/google_containers/pause-amd64:3.0
+          basename: pausepods
+          label: purpose=test
+    - num: 2
+      basename: zip
+      tuning: default
+      RCs:
+        - num: 5
+          image: gcr.io/google_containers/pause-amd64:3.0
+          basename: pausepods
+    - num: 2
+      basename: zip
+      tuning: default
+      RCs:
+        - num: 1
+          image: gcr.io/google_containers/pause-amd64:3.0
+          basename: grumble
+          label: app=grumble
+  tuningsets:
+    - name: default
+      pods:
+        stepping:
+          stepsize: 10
+          pause: 30s
+        ratelimit:
+          delay: 100ms
+      project:
+        ratelimit:
+          delay: 10s
--- a/clusterloader/framework/pod_utils.go
+++ b/clusterloader/framework/pod_utils.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"fmt"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+const maxRetries = 5
+
+// CreatePods creates pods in a user defined namspace with user configurable tuning sets
+func CreatePods(f *framework.Framework, name, namespace string, labels labels.Set, spec v1.PodSpec, maxCount int, tuning *TuningSet) error {
+	for i := 0; i < maxCount; i++ {
+		framework.Logf("%v/%v : Creating pod", i+1, maxCount)
+		podObj := newPod(name, namespace, i, labels, spec)
+		if _, err := createNewPodWithRetries(f, namespace, podObj); err != nil {
+			return err
+		}
+		if tuning == nil {
+			continue
+		}
+		// If a rate limit has been defined we wait for N ms between creation
+		if err := tuning.Pods.Delay(); err != nil {
+			return err
+		}
+		// If a stepping tuningset has been defined in the config, we wait for the step of pods to be created, and pause
+		if tuning.Pods.Stepping.StepSize != 0 && (i+1)%tuning.Pods.Stepping.StepSize == 0 {
+			verifyRunning := f.NewClusterVerification(
+				&v1.Namespace{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: namespace,
+					},
+				},
+				framework.PodStateVerification{
+					Selectors:   labels,
+					ValidPhases: []v1.PodPhase{v1.PodRunning},
+				},
+			)
+
+			duration, err := time.ParseDuration(tuning.Pods.Stepping.Timeout)
+			if err != nil {
+				framework.Logf("Timeout not set: %v", err)
+			}
+			pods, err := verifyRunning.WaitFor(i+1, duration)
+			if err != nil {
+				return err
+			} else if len(pods) < i+1 {
+				return fmt.Errorf("Only got %v out of %v", len(pods), i+1)
+			}
+
+			framework.Logf("We have created %d pods; sleep %v", i+1, duration)
+			if err := tuning.Pods.Pause(); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// createNewPodWithRetries uses polling to retry pod creation
+func createNewPodWithRetries(f *framework.Framework, namespace string, podObj *v1.Pod) (pod *v1.Pod, err error) {
+	for retryCount := 0; retryCount < maxRetries; retryCount++ {
+		pod, err = f.ClientSet.Core().Pods(namespace).Create(podObj)
+		if err == nil {
+			break
+		}
+	}
+	return
+}
+
+// newPod creates a new Pod config object
+func newPod(name, namespace string, num int, labels map[string]string, spec v1.PodSpec) *v1.Pod {
+	return &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf(name+"-pod-%v", num),
+			Namespace: namespace,
+			Labels:    labels,
+		},
+		Spec: spec,
+	}
+}
--- a/clusterloader/framework/rc_utils.go
+++ b/clusterloader/framework/rc_utils.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	kutils "k8s.io/kubernetes/test/utils"
+)
+
+// CreateRC will create a new RC if it does not exist, or it will update an existing RC with a new replica count if it does exist
+func CreateRC(f *framework.Framework, name, namespace string, label labels.Set, spec v1.PodSpec, replicas int) error {
+	_, err := f.ClientSet.Core().ReplicationControllers(namespace).Get(name, metav1.GetOptions{})
+	// If RC is not found, Get() will return a NotFound error
+	if errors.IsNotFound(err) {
+		newRC := newRC(name, int32(replicas), label, spec)
+		if _, err := createNewRCWithRetries(f, name, namespace, newRC); err != nil {
+			return err
+		}
+	} else {
+		updateReplicas := func(update *v1.ReplicationController) {
+			x := int32(replicas)
+			update.Spec.Replicas = &x
+		}
+		if _, err := framework.UpdateReplicationControllerWithRetries(f.ClientSet, namespace, name, updateReplicas); err != nil {
+			return err
+		}
+	}
+
+	// Wait for pods running matching label using podstore
+	// does not take replica count into effect, nor owner reference
+	return kutils.WaitForPodsWithLabelRunning(f.ClientSet, namespace, labels.SelectorFromSet(label))
+}
+
+// createNewRCWithRetries uses polling to retry RC creation
+func createNewRCWithRetries(f *framework.Framework, name, namespace string, rcObj *v1.ReplicationController) (rc *v1.ReplicationController, err error) {
+	for retryCount := 0; retryCount < maxRetries; retryCount++ {
+		if rc, err = f.ClientSet.Core().ReplicationControllers(namespace).Create(rcObj); err == nil {
+			framework.Logf("Created replication controller %q", name)
+			break
+		}
+	}
+	return
+}
+
+// newRC creates a new ReplicationController config object
+func newRC(rsName string, replicas int32, rcPodLabels map[string]string, spec v1.PodSpec) *v1.ReplicationController {
+	return &v1.ReplicationController{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: rsName,
+		},
+		Spec: v1.ReplicationControllerSpec{
+			Replicas: func(i int32) *int32 { return &i }(replicas),
+			Template: &v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: rcPodLabels,
+				},
+				Spec: spec,
+			},
+		},
+	}
+}
--- a/clusterloader/framework/tuning_utils.go
+++ b/clusterloader/framework/tuning_utils.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"time"
+
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// Delay acts on a TuningSetObject and will sleep for RateLimit.Delay
+func (tuning *TuningSetObject) Delay() error {
+	return sleep(tuning.RateLimit.Delay)
+}
+
+// Pause acts on a TuningSetObject and will sleep for Stepping.Pause
+func (tuning *TuningSetObject) Pause() error {
+	return sleep(tuning.Stepping.Pause)
+}
+
+func sleep(d string) error {
+	if d != "" {
+		duration, err := time.ParseDuration(d)
+		if err != nil {
+			return err
+		}
+		framework.Logf("Waiting for %v", duration)
+		time.Sleep(duration)
+	}
+	return nil
+}
+
+// Get matches the name of the tuning set defined in the project and returns a pointer to the set
+func (ts TuningSets) Get(name string) *TuningSet {
+	if name == "" {
+		return nil
+	}
+	// Iterate through defined tuningSets
+	for _, v := range ts {
+		// If we have a matching tuningSet keep it
+		if v.Name == name {
+			return &v
+		}
+	}
+	framework.Logf("No tuning found for %q", name)
+	return nil
+}
--- a/clusterloader/framework/util.go
+++ b/clusterloader/framework/util.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2016 The Kubernetes Authors.
+Copyright 2017 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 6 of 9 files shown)


clusterloader/framework/util.go
```diff
@@ -21,19 +21,55 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/kubernetes/pkg/api/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
 const maxRetries = 5
 
+// CreateNS creates a namespace if it is new, otherwise it will return the existing namespace pointer
```
gmarek: As I expect this file to grow a lot, it's probably worth splitting it in front. Please create pod_utils.go, rc_utils.go and leave only generic utils in 'utils.go'. Hopefully that will be enough.

clusterloader/framework/util.go
```diff
@@ -21,19 +21,55 @@ import (
...
+}
+
 // CreatePods creates pods in user defined namspaces with user configurable tuning sets
 func CreatePods(f *framework.Framework, appName string, ns string, labels map[string]string, spec v1.PodSpec, maxCount int, tuning *TuningSetType) {
 	for i := 0; i < maxCount; i++ {
 		framework.Logf("%v/%v : Creating pod", i+1, maxCount)
 		// Retry on pod creation failure
 		for retryCount := 0; retryCount < maxRetries; retryCount++ {
 			_, err := f.ClientSet.Core().Pods(ns).Create(&v1.Pod{
+				//f.ClientSet.Core().ReplicationControllers(ns).Create(&v1.ReplicationController{})
```
gmarek: what's this?

clusterloader/framework/util.go
```diff
@@ -80,3 +113,86 @@ func CreatePods(f *framework.Framework, appName string, ns string, labels map[st
 		}
 	}
 }
+
+func CreateRC(f *framework.Framework, namespace, name, imageName string, label map[string]string, replicas int) error {
```
gmarek: It's better to split this into multiple functions: one that creates a RC object, and one that does the creation (retries), and one that waits until it's running. Then it'll be actually possible to write unit tests for, at least some of, those functions. (Yes 'framework' doesn't follow those rules, but it doesn't mean that it's OK).

clusterloader/cluster_loader.go
```diff
@@ -58,30 +59,43 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
 		for _, p := range project {
 			// Find tuning if we have it
 			tuning := getTuningSet(tuningSets, p.Tuning)
-			framework.Logf("Our tuning set is: %v", tuning)
+			framework.Logf("Our tuning set is: %+v", tuning)
```
wojtek-t: s/Our tuning.../Tuning.../

clusterloader/framework/util.go
```diff
@@ -80,3 +113,86 @@ func CreatePods(f *framework.Framework, appName string, ns string, labels map[st
...
+		framework.Logf("NEW RC: %+v", rc)
+	case 1:
+		rc, err := framework.UpdateReplicationControllerWithRetries(f.ClientSet, namespace, name, func(update *v1.ReplicationController) {
+			x := int32(replicas)
+			update.Spec.Replicas = &x
+		})
+		if err != nil {
+			return err
+		}
+		framework.Logf("UPDATED RC: %+v", rc)
```
gmarek: Add a default case

clusterloader/cluster_loader.go
```diff
@@ -93,22 +107,46 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
 	})
 })
 
+// appendUnique appends new namespace pointers to a slice
+func appendUnique(allNS []*v1.Namespace, newNS *v1.Namespace) []*v1.Namespace {
```
wojtek-t: Instead of it - can we use "sets.String" and List() on it?

clusterloader/cluster_loader.go
```diff
@@ -93,22 +107,46 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
+			return allNS
+		}
+	}
+	return append(allNS, newNS)
+}
+
 // getTuningSet matches the name of the tuning set defined in the project and returns a pointer to the set
-func getTuningSet(tuningSets []clusterloaderframework.TuningSetType, podTuning string) (tuning *clusterloaderframework.TuningSetType) {
-	if podTuning != "" {
+func getTuningSet(tuningSets []clusterloaderframework.TuningSetType, tuningName string) (tuning *clusterloaderframework.TuningSetType) {
```
wojtek-t: In general I'm against of having "Type" in the type name definition - we already know this is type, so this is useless.

That said, once we switch it to "TunningSet", I suggest introducing "TunningSetList" (or sth like that) type and making this a method on it.

clusterloader/config/rctest.yaml
```diff
@@ -0,0 +1,38 @@
+provider: local
```
wojtek-t: i think we shouldn't hardcode provider in config yamls. What is it used for?

clusterloader/framework/util.go
wojtek-t: +1

clusterloader/framework/util.go
```diff
@@ -21,19 +21,55 @@ import (
...
+	fullNamespace := getNamespace(f, namespaceName)
+
+	if fullNamespace == "" {
+		ns, err = f.CreateNamespace(namespaceName, nil)
+		if err != nil {
+			return nil, err
+		}
+		framework.Logf("Created new namespace: %s", namespaceName)
+	} else {
+		ns, err = f.ClientSet.CoreV1().Namespaces().Get(fullNamespace, metav1.GetOptions{})
```
wojtek-t: Since it is getting namespace in case it already exists, CreateNS is not the best name for this function, I guess.

BTW - what's the usecase for it? When we already have a namespace?

clusterloader/framework/util.go
wojtek-t: Please remove

clusterloader/framework/util.go
```diff
@@ -80,3 +113,86 @@ func CreatePods(f *framework.Framework, appName string, ns string, labels map[st
 		}
 	}
 }
+
+func CreateRC(f *framework.Framework, namespace, name, imageName string, label map[string]string, replicas int) error {
+	nodeSelector := fields.OneTermEqualSelector("metadata.name", name)
+	list, _ := f.ClientSet.Core().ReplicationControllers(namespace).List(metav1.ListOptions{FieldSelector: nodeSelector.String()})
```
wojtek-t: I don't understand this line - "FieldSelector: nodeSelector.String()" looks pretty bad to me.

sjug: Do you mean the variable names look bad? Or what do you mean?

clusterloader/config/rctest.yaml
sjug: e2e wants a provider, it has some other fn's in `e2e.go` that are dependent on it afaik. I can remove this.

clusterloader/framework/util.go
sjug: Yes when we have an existing namespace and we want to do something inside it (like modify an existing RC in an existing NS).

clusterloader/framework/util.go
sjug: Do you have a preference to which case if default? (new vs update)

gmarek: It depends on what behavior for it to have if you'll get 2+ results.

clusterloader/framework/util.go
wojtek-t: Yeah - nodeSelector is something that selects nodes in kubernetes nomenclature. Here we are selecting RCs.

clusterloader/config/rctest.yaml
wojtek-t: I see. So that should be passed from the script running test, not being part of config which should be provider-independent.

clusterloader/framework/util.go
wojtek-t: Makes sense. So we should probably call it "CreateIfNotExists" or sth like that.

wojtek-t:
@sjug @gmarek - BTW, we finally have passing cluster-loader e2e tests set up here:
https://k8s-testgrid.appspot.com/perf-tests#cluster-loader

clusterloader/framework/util.go
sjug: Fair enough, will change thanks!

sjug:
@wojtek-t That's excellent, I see it is running successfully in the logs. I'll push up more relevant configs for it to run soon.

wojtek-t:
> @wojtek-t That's excellent, I see it is running successfully in the logs. I'll push up more relevant configs for it to run soon.

I wouldn't focus on configs that much now - I would just like them to test the existing functionality for now.

The other thing is to enable having multiple different tests, i.e. have different sets of configs but reuse the existing test to run them. It's mostly a matter of flags etc. I will file a dedicated issue for it with more cleaner description of what I want.

sjug:
@wojtek-t 
I think the most valuable outcome will be to have a run to run comparison to find possible regressions like the go 1.8 issue. How do we pass/fail the tests based on actual resource utilization and or run speed?

wojtek-t:
> I think the most valuable outcome will be to have a run to run comparison 

comparison with what?

> to find possible regressions like the go 1.8 issue. How do we pass/fail the test based on actual resource utilization and or run speed?

I would like to have to have the full functionality, and then hopefully rewrite existing load and density tests to clusterloader framework. But that's a second step. First we need to have a framework.

sjug:
@wojtek-t Between each run on the test-grid. Other than just having pass fail built on the test literally failing to run. It shouldn't be hard to create some more verbose logging or creating an artifact with each run. But then how do we compare it to the previous?

wojtek-t:
> @wojtek-t Between each run on the test-grid. Other than just having pass fail built on the test literally failing to run. It shouldn't be hard to create some more verbose logging or creating an artifact with each run. But then how do we compare it to the previous?

We aren't doing comparison between runs in any of our runs. We just have some static thresholds hardcoded in the test and we are testing against them.

wojtek-t:
But we are mostly testing our SLOs, which is api-call latencies and pod startup times. I would like to migrate our tests to clusterloader framework, and get rid of the existing ones, but we're not there yet with the framework.

clusterloader/cluster_loader.go
sjug: @wojtek-t I'm not sure what you mean, you want me to make a set?

wojtek-t: oh sorry, I think misread it. You can probably ignore this comment.

clusterloader/framework/util.go
sjug: k8s doesn't support having duplicate named objects so it would only return 0 or 1. Will change 1 to default.

clusterloader/framework/util.go
sjug: Broken up

clusterloader/framework/util.go
sjug: Removed

clusterloader/framework/util.go
sjug: @gmarek Let me know if you want me to break it up further

clusterloader/cluster_loader.go
sjug: Done

clusterloader/cluster_loader.go
sjug: Done

clusterloader/config/rctest.yaml
sjug: Removed

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
wojtek-t: 2017

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+const maxRetries = 5
+
+// CreatePods creates pods in user defined namspaces with user configurable tuning sets
```
wojtek-t: s/namspaces/namespaces/

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
...
+func CreatePods(f *framework.Framework, appName string, ns string, labels map[string]string, spec v1.PodSpec, maxCount int, tuning *TuningSet) {
```
wojtek-t: Can we change appName to just name? We are generally using everywhere just name and namespace.

Also:
s/appName string, ns string/name, namespace string/

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
...
+	for i := 0; i < maxCount; i++ {
+		framework.Logf("%v/%v : Creating pod", i+1, maxCount)
+		// Retry on pod creation failure
+		for retryCount := 0; retryCount < maxRetries; retryCount++ {
```
wojtek-t: Please extract this loop into a separate function.

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
...
+			// If a rate limit has been defined we wait for N ms between creation
+			tuning.Pods.Sleep()
+			// If a stepping tuningset has been defined in the config, we wait for the step of pods to be created, and pause
+			if tuning.Pods.Stepping.StepSize != 0 && (i+1)%tuning.Pods.Stepping.StepSize == 0 {
+				verifyRunning := f.NewClusterVerification(
+					&v1.Namespace{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: ns,
+						},
+						Status: v1.NamespaceStatus{},
```
wojtek-t: Why do we need this line?

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
...
+					},
+					framework.PodStateVerification{
+						Selectors:   labels,
+						ValidPhases: []v1.PodPhase{v1.PodRunning},
+					},
+				)
+
+				pods, err := verifyRunning.WaitFor(i+1, tuning.Pods.Stepping.Timeout*time.Second)
```
wojtek-t: tunning.Pods.Stepping.Timeout should already be set correctly.
We shouldn't need to multiply it by "time.Second".

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,79 @@
...
+					Labels:    labels,
+				},
+				Spec: spec,
+			})
+			if err == nil {
+				break
+			}
+			framework.ExpectNoError(err)
+		}
+		if tuning != nil {
```
wojtek-t: Let's change it to:
```
if tuning == nil {
  cotninue
}
...
```

Then the code is less branched and easier to follow.

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// CreateRC will create a new RC if it does not exist, or it will update an existing RC with a new replica count if it does exist
+func CreateRC(f *framework.Framework, namespace, name, imageName string, label map[string]string, replicas int) error {
+	fieldSelector := fields.OneTermEqualSelector("metadata.name", name)
+	list, _ := f.ClientSet.Core().ReplicationControllers(namespace).List(metav1.ListOptions{FieldSelector: fieldSelector.String()})
```
wojtek-t: Hmm - this line is strange. You're basically trying to get an RC with a given name. Why you just don't do:
```
f.ClientSet.Core().ReplicationControllers(namespace).Get(name)
```
?

Then also the switch wouldn't have to be switch.

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// createNewRCWithRetries uses polling to retry RC creation
+func createNewRCWithRetries(f *framework.Framework, name, namespace string, rcObj *v1.ReplicationController) (*v1.ReplicationController, error) {
+	var rc *v1.ReplicationController
+	var updateErr error
+	pollErr := wait.PollImmediate(10*time.Millisecond, 1*time.Minute, func() (bool, error) {
+		var err error
+		if rc, err = f.ClientSet.Core().ReplicationControllers(namespace).Create(rcObj); err == nil {
+			framework.Logf("Creating replication controller %q", name)
```
wojtek-t: s/Creating/Created/

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,130 @@
...
+	fieldSelector := fields.OneTermEqualSelector("metadata.name", name)
+	list, _ := f.ClientSet.Core().ReplicationControllers(namespace).List(metav1.ListOptions{FieldSelector: fieldSelector.String()})
+	switch len(list.Items) {
+	case 0:
+		newRC := newRC(name, int32(replicas), label, name, imageName)
+		rc, err := createNewRCWithRetries(f, name, namespace, newRC)
+		if err != nil {
+			return err
+		}
+		framework.Logf("NEW RC: %+v", rc)
```
wojtek-t: Remove this line - the fact that it is created is already logged by "createNewRCWithRetries"

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,130 @@
...
+		return false, nil
+	})
+	if pollErr == wait.ErrWaitTimeout {
+		pollErr = fmt.Errorf("Unable to create new RC %q: %v", name, updateErr)
+	}
+	return rc, pollErr
+}
+
+// waitForRCPodsRunning will verify that a list of pods will all be running
+func waitForRCPodsRunning(f *framework.Framework, podList *v1.PodList, podName, namespace string, replicas int) error {
```
wojtek-t: This function is extremely strange to me.
It doesn't wait for pods to be running - it just checks something on the current list of pods.

In my opinion we shouldn't do it like that. I see two options:
1. just really wait until all pods are running here (I see drawbacks of it)
2. spin up a go-routine that will wait for all pods to be running (with some timeout).

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,130 @@
...
+	framework.Logf("Ensuring all pods for ReplicationController %q are running", podName)
+	running := 0
+	for _, pod := range podList.Items {
```
wojtek-t: The other thing is that it shouldn't be "poll-based". I mean you shouldn't periodially list pods (in general, you shouldn't list pods). Instead of that you should watch for them.

We already have a simple utility for it - see this:
https://github.com/kubernetes/kubernetes/blob/master/test/utils/pod_store.go#L37

clusterloader/framework/util.go
```diff
@@ -25,58 +25,59 @@ import (
...
+	existingNamespaces, _ := f.ClientSet.Core().Namespaces().List(metav1.ListOptions{})
+	for _, value := range existingNamespaces.Items {
+		if value.GenerateName == fmt.Sprintf("e2e-tests-%v-", baseName) {
+			return value.Name
+		}
+	}
+	return ""
+}
+
+// Sleep acts on a TuningSetObject and will sleep for the specified duration
```
wojtek-t: These functions should be moved to framework/tunning_set.go (or sth like that) - these aren't test-specific.

clusterloader/framework/util.go
```diff
@@ -25,58 +25,59 @@ import (
...
+	if tuning.RateLimit.Delay != 0 {
+		framework.Logf("Waiting for %d ms", tuning.RateLimit.Delay)
+		time.Sleep(tuning.RateLimit.Delay * time.Millisecond)
+	}
+}
 
-				framework.Logf("We have created %d pods and are now sleeping for %d seconds", i+1, tuning.Pods.Stepping.Pause)
-				time.Sleep(tuning.Pods.Stepping.Pause * time.Second)
+// SelectTuning matches the name of the tuning set defined in the project and returns a pointer to the set
+func (tuningSets TuningSets) SelectTuning(tuningName string) (tuning *TuningSet) {
```
wojtek-t:
```
func(ts *TunningSets) Get(name string) (*TunningSet, error) {
  if name == "" {
    return nil, nil
  }
  ...
}
```

sjug: TuningSets is just a type that contains a slice, so having a pointer to it doesn't really help? There's no way there would be an error? If there is no tuning set that matches the name then `*TuningSet` is just nil?

clusterloader/framework/pod_utils.go
sjug: tunng.Pods.Stepping.Timeout is type `time.Duration`, don't we need to convert that?

clusterloader/framework/rc_utils.go
sjug: Sure, I can use `.Get()` as well. If there won't have to be a `switch` there will have to be some other control, like an `if` so I'm not sure it really makes much difference.

clusterloader/framework/pod_utils.go
wojtek-t: We shouldn't need to.

clusterloader/framework/rc_utils.go
wojtek-t: Get() is much more efficient than list.

clusterloader/framework/util.go
wojtek-t: I'm fine dropping pointer.

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/pod_utils.go
sjug: Created using polling instead of for loop `createNewPodWithRetries`

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/pod_utils.go
sjug: Removed

clusterloader/framework/rc_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Done here and for update

clusterloader/framework/rc_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Got rid of this entire function

clusterloader/framework/rc_utils.go
sjug: Using pod store now

clusterloader/framework/pod_utils.go
sjug: I don't know how to do it without conversion?
https://golang.org/pkg/time/#Sleep

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+	return nil
+}
+
+// createNewPodWithRetries uses polling to retry RC creation
+func createNewPodWithRetries(f *framework.Framework, name, namespace string, podObj *v1.Pod) (*v1.Pod, error) {
+	var pod *v1.Pod
+	var updateErr error
+	pollErr := wait.PollImmediate(10*time.Millisecond, 1*time.Minute, func() (bool, error) {
```
wojtek-t: I think we should change it to just simple for loop (< maxRetries).

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	kutils "k8s.io/kubernetes/test/utils"
+)
+
+// CreateRC will create a new RC if it does not exist, or it will update an existing RC with a new replica count if it does exist
+func CreateRC(f *framework.Framework, namespace, name, imageName string, label map[string]string, replicas int) error {
+	rc, _ := f.ClientSet.Core().ReplicationControllers(namespace).Get(name, metav1.GetOptions{})
+	if rc.Name == "" {
```
wojtek-t: This isn't correct - this would never be the case.
Basically Get() will either return a correct RC (which has to have name) or an error "notexisting" (you can look for it in main repo probably).

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,103 @@
...
+		newRC := newRC(name, int32(replicas), label, name, imageName)
+		_, err := createNewRCWithRetries(f, name, namespace, newRC)
```
wojtek-t: if ...; err != nil {

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		_, err := framework.UpdateReplicationControllerWithRetries(f.ClientSet, namespace, name, func(update *v1.ReplicationController) {
```
wojtek-t: Can you please change to:

updateReplicas := func(...) {
 ....
}
if _, err := framework.UpdateRepl...(...); err != nil {
  return err
}
```

File: clusterloader/framework/rc_utils.go [hunk 29]
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+	return nil
+}
+
+// createNewRCWithRetries uses polling to retry RC creation
+func createNewRCWithRetries(f *framework.Framework, name, namespace string, rcObj *v1.ReplicationController) (*v1.ReplicationController, error) {
+	var rc *v1.ReplicationController
+	var updateErr error
+	pollErr := wait.PollImmediate(10*time.Millisecond, 1*time.Minute, func() (bool, error) {
```
wojtek-t: Same here about for loop up to maxRetries.

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,103 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: rsName,
+		},
+		Spec: v1.ReplicationControllerSpec{
+			Replicas: func(i int32) *int32 { return &i }(replicas),
+			Template: &v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: rcPodLabels,
+				},
+				Spec: v1.PodSpec{
```
wojtek-t: Shouldn't we also pass PodSpec here (the same way we are passing it to newPod)?

clusterloader/framework/tuning_utils.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil
+	}
+	// Iterate through defined tuningSets
+	for _, v := range ts {
+		// If we have a matching tuningSet keep it
+		if v.Name == name {
+			return &v
+		}
+	}
+	framework.Failf("No tuning found for %q", name)
```
wojtek-t: Instead of framework.Failf, just return such error.

clusterloader/framework/pod_utils.go
sjug: I removed it at first then I checked other instances of `.WaitFor` and they all have the time conversion so I added it back.

wojtek-t: That isn't correct in my opinion. Here is the main problem:
```
x := time.Second * time.Second
fmt.Println(x)

277777h46m40s
```

So if you set duration correctly, multiplying it by time.Second breaks everrything.

If this is also like that in other places, we should fix all of them.

clusterloader/framework/rc_utils.go
sjug: It is the case, if it is not found `.Get()` returns a pointer to nil RC. However, I can change it to check the error if that's what you prefer.

clusterloader/framework/rc_utils.go
sjug: The downside is that then we have another external file dependency on the pod file. I will put a condition in both to check if the user has specified a podfile or just an imagename, and then create the appropriate struct.

clusterloader/framework/rc_utils.go
sjug: Do you generally prefer for the error check to be done in line with the function call, as I can keep this in mind for the future if that's the case? If not, what do you base your decision on. Thanks!

clusterloader/framework/rc_utils.go
sjug: Should I generally try to keep the error check on the same line as the function call?

clusterloader/framework/rc_utils.go
wojtek-t: It will return just nil. Doing rc.Name will cause a panic then.

clusterloader/framework/rc_utils.go
wojtek-t: If the only result from the function you need is an error, then generally yes.

clusterloader/framework/rc_utils.go
wojtek-t: That sgtm.

clusterloader/cluster_loader.go
```diff
@@ -93,20 +108,28 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
+	for _, NS := range allNS {
+		if NS.Name == newNS.Name {
+			return allNS
+		}
+	}
+	return append(allNS, newNS)
+}
+
+// stringToMapString takes a string in the format "key=value,key2=value2" and returns a corresponding map
+func stringToMapString(input string) map[string]string {
```
wojtek-t: We already have function that is doing this:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/labels/labels.go#L157

We should use it instead of creating another one.

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Changed to check the error instead

clusterloader/framework/rc_utils.go
sjug: Changed in several places

clusterloader/framework/rc_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Done, see new method `ParseConfig()`

clusterloader/framework/tuning_utils.go
sjug: `Failf` is useless here, changed to `Logf`

clusterloader/cluster_loader.go
sjug: Created new method `ConvertToLabelSet()` using existing function `ConvertSelectorToLabelsMap` but adding nil check for label (need default value).

clusterloader/cluster_loader.go
```diff
@@ -57,31 +55,51 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
-					createTemplate(template.Basename, ns, mkPath(template.File), template.Number, tuning)
+					createTemplate(template.Basename, ns, clusterloaderframework.MakePath(template.File), template.Number, tuning)
+				}
+				// RCs are a thing as well
+				for _, RC := range p.RCs {
+					config, err := RC.ParseConfig()
+					if err != nil {
+						framework.Failf("Error parsing config, %v", err)
+					}
+					label, _ := RC.ConvertToLabelSet()
```
wojtek-t: Please don't ignore error

clusterloader/cluster_loader.go
```diff
@@ -57,31 +55,51 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
 				// This is too familiar, create pods
 				for _, pod := range p.Pods {
-					config := parsePods(mkPath(pod.File))
-					labels := map[string]string{"purpose": "test"}
-					clusterloaderframework.CreatePods(f, pod.Basename, ns.Name, labels, config.Spec, pod.Number, tuning)
+					config, err := pod.ParseConfig()
+					if err != nil {
+						framework.Failf("Error parsing config, %v", err)
+					}
+					label, _ := pod.ConvertToLabelSet()
```
wojtek-t: Please don't ignore errors.

clusterloader/cluster_loader.go
```diff
@@ -57,31 +55,51 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
+					clusterloaderframework.CreatePods(f, pod.Basename, ns.Name, label, config.Spec, pod.Number, tuning)
 				}
 			}
+			// Only sleeps for each new project defined in the config, need to move up to sleep for every copy
```
wojtek-t: I would say let's move then? WDYT?

clusterloader/cluster_loader.go
```diff
@@ -156,32 +158,16 @@ func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObj
 		if tuning != nil {
 			if tuning.Templates.RateLimit.Delay != 0 {
 				framework.Logf("Sleeping %d ms between template creation.", tuning.Templates.RateLimit.Delay)
-				time.Sleep(time.Duration(tuning.Templates.RateLimit.Delay) * time.Millisecond)
+				time.Sleep(tuning.Templates.RateLimit.Delay * time.Millisecond)
```
wojtek-t: I don't think we should multiply by time.Millisecond.

clusterloader/cluster_loader.go
```diff
@@ -156,32 +158,16 @@ func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObj
...
 			}
 			if tuning.Templates.Stepping.StepSize != 0 && (i+1)%tuning.Templates.Stepping.StepSize == 0 {
 				framework.Logf("We have created %d templates and are now sleeping for %d seconds", i+1, tuning.Templates.Stepping.Pause)
-				time.Sleep(time.Duration(tuning.Templates.Stepping.Pause) * time.Second)
+				time.Sleep(tuning.Templates.Stepping.Pause * time.Second)
```
wojtek-t: Same here - don't multiply by time.Second

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,93 @@
...
+			)
+
+			pods, err := verifyRunning.WaitFor(i+1, tuning.Pods.Stepping.Timeout*time.Second)
+			if err != nil {
+				framework.Failf("Error in wait... %v", err)
+			} else if len(pods) < i+1 {
+				framework.Failf("Only got %v out of %v", len(pods), i+1)
+			}
+
+			framework.Logf("We have created %d pods and are now sleeping for %d seconds", i+1, tuning.Pods.Stepping.Pause)
```
wojtek-t: s/pods and are now sleeping, for %d second/pods; sleep %ds/

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,93 @@
...
+			time.Sleep(tuning.Pods.Stepping.Pause * time.Second)
```
wojtek-t: Don't multiply by time.Second.

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	kutils "k8s.io/kubernetes/test/utils"
+)
+
+// CreateRC will create a new RC if it does not exist, or it will update an existing RC with a new replica count if it does exist
+func CreateRC(f *framework.Framework, name, namespace string, label labels.Set, spec v1.PodSpec, replicas int) error {
+	_, err := f.ClientSet.Core().ReplicationControllers(namespace).Get(name, metav1.GetOptions{})
+	// If RC is not found, Get() will return a NotFound error
+	if _, ok := err.(*errors.StatusError); ok {
```
wojtek-t: StatusError doesn't have to be "NotFound".
```
if errors.IsNotFound(err) {
```

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+	return nil
+}
+
+// createNewRCWithRetries uses polling to retry RC creation
+func createNewRCWithRetries(f *framework.Framework, name, namespace string, rcObj *v1.ReplicationController) (rc *v1.ReplicationController, err error) {
+	for retryCount := 0; retryCount < maxRetries; retryCount++ {
+		if rc, err = f.ClientSet.Core().ReplicationControllers(namespace).Create(rcObj); err == nil {
+			framework.Logf("Created replication controller %q", name)
```
wojtek-t: Please remove this log.

clusterloader/framework/util.go
```diff
@@ -17,66 +17,97 @@ limitations under the License.
...
+			TerminationGracePeriodSeconds: &zero,
+			Containers: []v1.Container{
+				{
+					Name:  cl.Basename,
+					Image: cl.Image,
+				},
+			},
+		}
+	} else {
+		// If we don't have either the file nor imagename, we have issues
```
wojtek-t: Remove this comment - this doesn't bring any value.

clusterloader/framework/util.go
```diff
@@ -17,66 +17,97 @@ limitations under the License.
...
 
-				pods, err := verifyRunning.WaitFor(i+1, tuning.Pods.Stepping.Timeout*time.Second)
-				if err != nil {
-					framework.Failf("Error in wait... %v", err)
-				} else if len(pods) < i+1 {
-					framework.Failf("Only got %v out of %v", len(pods), i+1)
-				}
+// ParseConfig unmarshalls the json file defined in the CL config into a struct
+func (cl *ClusterLoaderObject) ParseConfig() (*v1.Pod, error) {
+	configStruct := &v1.Pod{}
```
wojtek-t: Can we rename configStruct to pod or sth like that?

clusterloader/cluster_loader.go
sjug: I'm not sure how I feel about this one... I think the functionality you guys wanted with the delay was to be between new projects based on basename, not iterations of the same. I feel like maybe we should have both possible, but then we need a new config parameter.

clusterloader/cluster_loader.go
sjug: Default duration is nanoseconds [time.go](https://github.com/golang/go/blob/master/src/time/time.go#L611-L612) ([play](https://play.golang.org/p/Dk7uEJu03T))

clusterloader/framework/rc_utils.go
sjug: Don't we need some sort of output to track the progress, and timing of the test?

clusterloader/cluster_loader.go
wojtek-t: OK - I don't have strong opinion about it. We can leave as is. Please split the line into two then making it a TODO: Consider ...

clusterloader/cluster_loader.go
wojtek-t: I know - but if you are setting "duartion: 1" it is what you are asking for.
No matter what it is, if I'm passing "time.Duration" to a function, I'm expecting it to operation on a given duration not on duration multipled by something.

sjug: I don't think expecting the user to enter time variables in nanoseconds is reasonable.

wojtek-t: Then let's not use duration in the config file.
I'm not going to agree on multiplying a duration from config by some constants (second, millisecond, ...)

sjug: If we use `int` we'll still need to convert it to `duration * time.Const` though?

wojtek-t: If you make the field in the config like "timeout_seconds" (or something explicitly mentioning units), then conversion is expected.
I just don't want to multiply duration by duration to get duration - this is really bad in my opinion.

But to be honest, I think that you should be able to set field with type duration to things like "1s", "10m" or sth like that (and this is the way we should proceed in my opinion).

sjug: > But to be honest, I think that you should be able to set field with type duration to things like "1s", "10m" or sth like that (and this is the way we should proceed in my opinion).

I'll work on making a way to parse this

wojtek-t: FYI - in k8s components we are using this flags.DurationVar(...) which supports it.

clusterloader/cluster_loader.go
sjug: Done

clusterloader/cluster_loader.go
sjug: Done

clusterloader/cluster_loader.go
sjug: Done

clusterloader/cluster_loader.go
sjug: Created two new functions in `util.go` to handle conversion of `string` to correct `time.Duration`

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Nice thanks, done

clusterloader/framework/util.go
sjug: Renamed to `pod`, done

clusterloader/framework/util.go
sjug: Removed

clusterloader/cluster_loader.go
```diff
@@ -154,34 +164,26 @@ func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObj
 
 		// If there is a tuning set defined for this template
 		if tuning != nil {
-			if tuning.Templates.RateLimit.Delay != 0 {
-				framework.Logf("Sleeping %d ms between template creation.", tuning.Templates.RateLimit.Delay)
-				time.Sleep(time.Duration(tuning.Templates.RateLimit.Delay) * time.Millisecond)
+			if tuning.Templates.RateLimit.Delay != "" {
+				duration, err := clusterloaderframework.TimeSplit(tuning.Templates.RateLimit.Delay)
```
wojtek-t: Instead of introducing a new custom function, you can just use:
https://golang.org/pkg/time/#ParseDuration

clusterloader/cluster_loader.go
```diff
@@ -154,34 +164,26 @@ func createTemplate(baseName string, ns *v1.Namespace, configPath string, numObj
...
+				time.Sleep(duration)
 			}
 			if tuning.Templates.Stepping.StepSize != 0 && (i+1)%tuning.Templates.Stepping.StepSize == 0 {
-				framework.Logf("We have created %d templates and are now sleeping for %d seconds", i+1, tuning.Templates.Stepping.Pause)
-				time.Sleep(time.Duration(tuning.Templates.Stepping.Pause) * time.Second)
+				duration, err := clusterloaderframework.TimeSplit(tuning.Templates.Stepping.Pause)
+				if err != nil {
+					framework.Failf("Unable to get tuning: $v", err)
+				}
+				framework.Logf("We have created %d templates and are now sleeping for %v", i+1, duration)
```
wojtek-t: s/templates and are now sleeping for %v/template; sleep for %v/

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,103 @@
...
+		podObj := newPod(name, namespace, i, labels, spec)
+		if _, err := createNewPodWithRetries(f, namespace, podObj); err != nil {
+			return err
+		}
+		if tuning == nil {
+			continue
+		}
+		// If a rate limit has been defined we wait for N ms between creation
+		if err := tuning.Pods.Sleep(); err != nil {
+			framework.Failf("Unable to get tuning: %v", err)
```
wojtek-t: I would like to avoid "framework.Failf" in all helper functions. Instead just return an error here.

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,103 @@
...
+				},
+				framework.PodStateVerification{
+					Selectors:   labels,
+					ValidPhases: []v1.PodPhase{v1.PodRunning},
+				},
+			)
+
+			duration, err := TimeSplit(tuning.Pods.Stepping.Timeout)
+			if err != nil {
+				framework.Logf("Unable to get tuning: %v", err)
```
wojtek-t: return an error

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			pods, err := verifyRunning.WaitFor(i+1, duration)
+			if err != nil {
+				framework.Failf("Error in wait... %v", err)
```
wojtek-t: Return error. Same below.

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,81 @@
...
+		updateReplicas := func(update *v1.ReplicationController) {
+			x := int32(replicas)
+			update.Spec.Replicas = &x
+		}
+		if _, err := framework.UpdateReplicationControllerWithRetries(f.ClientSet, namespace, name, updateReplicas); err != nil {
+			return err
+		}
+	}
+
+	// Wait for pods running matching label using podstore, does not take replica count into effect, nor owner reference
```
wojtek-t: nit: can you please split this line into two?

clusterloader/framework/rc_utils.go
```diff
@@ -0,0 +1,81 @@
...
+	if err = kutils.WaitForPodsWithLabelRunning(f.ClientSet, namespace, labels.SelectorFromSet(label)); err != nil {
```
wojtek-t: you can just do
```
return ktuils.WaitFor...
```

clusterloader/framework/rc_utils.go
wojtek-t: OK let's leave. We can tune later.

clusterloader/framework/pod_utils.go
```diff
@@ -0,0 +1,103 @@
...
+			} else if len(pods) < i+1 {
+				framework.Failf("Only got %v out of %v", len(pods), i+1)
+			}
+
+			duration, err = TimeSplit(tuning.Pods.Stepping.Pause)
```
wojtek-t: Instead of those lines, can we introduce Pause() method on tuning.Pods (similarly to tuning.Pods.Sleep())?

clusterloader/framework/util.go
```diff
@@ -17,66 +17,130 @@ limitations under the License.
...
+// ConvertToLabelSet will convert the string label to a set, while also setting a default value
+func (cl *ClusterLoaderObject) ConvertToLabelSet() (labels.Set, error) {
+	if cl.Label == "" {
+		cl.Label = "purpose=test"
+	}
+	label, err := labels.ConvertSelectorToLabelsMap(cl.Label)
+	return label, err
+}
+
+// TimeSplit will take the user readable time string and convert it into corresponding time.Duration
```
wojtek-t: As I wrote above - remove this function and replace with ParseDuration:
https://golang.org/pkg/time/#ParseDuration

clusterloader/framework/pod_utils.go
sjug: The timeout is an optional field, that's why I didn't `Failf`

wojtek-t: Sorry, let's ignore this comment.

clusterloader/framework/pod_utils.go
sjug: I agree, I'd like a clean way to do this. Will try something out.

clusterloader/cluster_loader.go
sjug: That seems like it would be good to have in the std lib, thanks

clusterloader/cluster_loader.go
sjug: Changed

clusterloader/framework/pod_utils.go
sjug: Agreed, changed

clusterloader/framework/pod_utils.go
sjug: Done

clusterloader/framework/pod_utils.go
sjug: Done, two separate methods `Delay` and `Pause`, `Delay` is just a sleep between each, where `Pause` is specific to steppings.

clusterloader/framework/rc_utils.go
sjug: Done

clusterloader/framework/rc_utils.go
sjug: Good idea, changed

clusterloader/framework/util.go
sjug: Done

clusterloader/cluster_loader.go
```diff
@@ -93,93 +120,63 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
 
 		framework.RunKubectlOrDie("create", "-f", tmpfile.Name(), getNsCmdFlag(ns))
 		framework.Logf("%d/%d : Created template %s", i+1, numObjects, baseName)
 
 		// If there is a tuning set defined for this template
 		if tuning != nil {
-			if tuning.Templates.RateLimit.Delay != 0 {
-				framework.Logf("Sleeping %d ms between template creation.", tuning.Templates.RateLimit.Delay)
-				time.Sleep(time.Duration(tuning.Templates.RateLimit.Delay) * time.Millisecond)
+			if tuning.Templates.RateLimit.Delay != "" {
```
wojtek-t: Just one more idea. Maybe we should hide this whole if in the Delay method?

Then here the code would look like:
```
if err := tuning.Templates.Delay(); err != nil {
  return err
}
```

And the Delay method would look like:
```
func (...) Delay() {
  if ....Delay != "" { 
    framework.Logf("Sleeping for %v", ...)
    // What currently is in the Delay function.
  }
}
```

clusterloader/cluster_loader.go
```diff
@@ -93,93 +120,63 @@ var _ = framework.KubeDescribe("Cluster Loader [Feature:ManualPerformance]", fun
...
+				framework.Logf("Sleeping %v between template creation.", tuning.Templates.RateLimit.Delay)
+				if err := tuning.Templates.Delay(); err != nil {
+					return err
+				}
 			}
 			if tuning.Templates.Stepping.StepSize != 0 && (i+1)%tuning.Templates.Stepping.StepSize == 0 {
-				framework.Logf("We have created %d templates and are now sleeping for %d seconds", i+1, tuning.Templates.Stepping.Pause)
-				time.Sleep(time.Duration(tuning.Templates.Stepping.Pause) * time.Second)
+				framework.Logf("We have created %d templates; sleep for %v", i+1, tuning.Templates.Stepping.Pause)
```
wojtek-t: And same here.

clusterloader/cluster_loader.go
sjug: I do the `nil`/`""` check in the `sleep()` function that `Delay()` calls. 
The issue is that if we don't have the `tuning.X.RateLimit.Delay` defined then we're returning an error, but it's not necessarily mandatory. Need to soften the error handling of the `Delay()`.

clusterloader/cluster_loader.go
sjug: Wouldn't work here since the check is to do with the step sizes and not just the presence of the wait time string.

clusterloader/cluster_loader.go
sjug: If we have a nil string then we're returning a nil error in `sleep()` so that will fail silently, removed the extra conditional check.

sjug:
Squashed commits

wojtek-t:
@sjug - thanks. Let's merge and we can iterate on it later if/when needed.

/lgtm

sjug:
Great thanks Wojciech, we should discuss what functionality you guys want
next

wojtek-t:
> Great thanks Wojciech, we should discuss what functionality you guys want
next

I would say, ability to delete all the stuff (similarly to how we create it) is the next thing.

sjug:
Deletion based on label or base name, or what field/selector were you
thinking?

wojtek-t:
I would like to be able to e.g. delete all replications controller that were created before.

PR #56, opened 2017-05-15
Fix run-e2e.sh script in clusterloader

wojtek-t:
@sjug


sjug:
Does `ginkgo.focus` not handle whitespace properly anymore?
Otherwise, lgtm.

wojtek-t:
> Does ginkgo.focus not handle whitespace properly anymore?

It never did - we had problems with it starting 2 years ago and it still lasts.

gmarek:
\lgtm

PR #61, opened 2017-05-16
Replace dockerization logic for benchmark with a runner script instead

shyamjvs:
Seems like docker image for running the tool is not needed. All we need is a runner for dockerized compilation and then running the binary with the right args.
This seems to be the pattern for running simple binaries in CI jobs, where anyway the running of bootstrap.py is dockerized. (Eg: Job ci-cadvisor-node-kubelet (https://github.com/kubernetes/test-infra/blob/master/jobs/config.json#L10) simply calls the runner script in the cadvisor repo (https://github.com/google/cadvisor/blob/master/build/jenkins_e2e.sh))

cc @wojtek-t @gmarek

### DIFF
--- a/benchmark/Makefile
+++ b/benchmark/Makefile
@@ -16,11 +16,7 @@ BINARY := build/benchmark
 SOURCES := $(shell find pkg -name '*.go' | grep -v "_test.go")
 TESTS := $(shell find pkg -name '*_test.go')
 
-REGISTRY := gcr.io/k8s-testimages
-IMAGE := $(REGISTRY)/benchmark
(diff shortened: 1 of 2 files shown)


shyamjvs:
@gmarek Removed the quotes around "${BENCHMARK_ARGS}" and verified that it works now. Thanks for pointing it.

shyamjvs:
Removed the dockerized build as it is unnecessary, because `make benchmark` will anyway use the right golang/goarch/... based on the go installation available (both locally or inside a docker container). And then we run the compiled binary on the same machine, which should work.

PR #65, opened 2017-05-30
Benchmark tool prints results in a classified and sorted way

shyamjvs:
The results printed by the tool would now be split by latency percentiles (50, 90, 99), and sorted by the avg ratio within each of those individual tables.

cc @wojtek-t

### DIFF
--- a/benchmark/cmd/benchmark.go
+++ b/benchmark/cmd/benchmark.go
@@ -114,6 +114,24 @@ func compare(jobComparisonData *util.JobComparisonData) {
 	}
 }
 
+// Pretty print results of the comparison.
+func printResults(jobComparisonData *util.JobComparisonData) {
+	glog.Infof("Metric-wise results of comparison for 99th percentile:")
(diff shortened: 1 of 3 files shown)


benchmark/cmd/benchmark.go
```diff
@@ -114,6 +114,24 @@ func compare(jobComparisonData *util.JobComparisonData) {
...
+		return k.Percentile != "Perc50"
+	})
+	glog.Infof("")
+	glog.Infof("Metric-wise results of comparison for 90th percentile:")
+	jobComparisonData.PrettyPrintWithFilter(func(k util.MetricKey, d util.MetricComparisonData) bool {
+		return k.Percentile != "Perc90"
+	})
+	glog.Infof("")
+	glog.Infof("Metric-wise results of comparison for 99th percentile:")
+	jobComparisonData.PrettyPrintWithFilter(func(k util.MetricKey, d util.MetricComparisonData) bool {
```
gmarek: I think the order should be reversed - 99th percentile is more important, hence it should be on top.

shyamjvs: Done

PR #78, opened 2017-10-09
Use namely/perfest as Docker image

bobbytables:
I've updated the image that this project uses to include the latest changes and published it to my companies Docker hub. It now uses the min mtu set to 96 and increments it by 40 each time. This change was made a while ago but no docker image was published for it.

I also removed a few of the tests because they rely on services that this project does not create. Only the `w2` service is actually created by this tool.


gmarek:
cc @bowei

bowei:
I see the image change, but there are also a bunch of param changes. Can you put why the changes are there in the commit + github PR description?

clkao:
Does it make sense to configure an automated build on dockerhub to track this directory?

PR #85, opened 2017-11-23
Fixed bug in handlePulledImageEvent

kozl:
Fixed bug in handlePulledImageEvent function. It leads to negative startup time for pods, that pull images from docker registry. As a result they are not accounted.


kozl:
Signed the CLA

gmarek:
@kozl - I can't verify CLA. Can you please make sure it's correctly signed.

wojtek-t:
This is inactive for almost half an year now - closing.

PR #86, opened 2017-12-08
Fix netperf README nits

mozhuli:


network/benchmarks/netperf/README.md
```diff
@@ -6,7 +6,7 @@ A standardized benchmark to measure Kubernetes networking performance on multipl
 ## Implementation and Test Methodology
 
 The benchmark can be executed via a single Go binary invocation that triggers all the automated testing located in the orchestrator and worker pods as seen below. The test uses a custom docker container that has the go binary and iperf3 and other tools built into it.
-The orchestrator pod coordinates the worker pods to run tests in serial order for the 4 scenarios described below, at MTUs (MSS tuning for TCP and direct packet size tuning for UDP).
```
bowei: I would prefer just saying "for the scenarios described below" and then we don't have to keep these in sync

PR #89, opened 2018-01-16
Make SLO monitor publish full pod startup latency including image pul…

gmarek:
…l time

@hulkholden @shyamjvs @wojtek-t

### DIFF
--- a/slo-monitor/src/monitors/pod_monitor.go
+++ b/slo-monitor/src/monitors/pod_monitor.go
@@ -40,6 +40,8 @@ import (
 const (
 	// PodE2EStartupLatencyKey is a key for pod startup latency monitoring metric.
 	PodE2EStartupLatencyKey = "slomonitor_pod_e2e_startup_latency_seconds"
+	// PodFullStartupLatencyKey is a key for pod startup latency monitoring metric including pull times.
+	PodFullStartupLatencyKey = "slomonitor_pod_full_startup_latency_seconds"
 )
 
 var (
@@ -51,6 +53,15 @@ var (
 			Buckets: prometheus.LinearBuckets(0.5, 0.25, 50),
 		},
 	)
+
+	// PodFullStartupLatency is a prometheus metric for monitoring pod startup latency including image pull times.
+	PodFullStartupLatency = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    PodFullStartupLatencyKey,
+			Help:    "Pod e2e startup latencies in milliseconds, with image pull times",
+			Buckets: prometheus.LinearBuckets(0.5, 0.5, 100),
+		},
+	)
 )
 
 var registerMetrics sync.Once
@@ -59,6 +70,7 @@ var registerMetrics sync.Once
 func Register() {
 	registerMetrics.Do(func() {
 		prometheus.MustRegister(PodE2EStartupLatency)
+		prometheus.MustRegister(PodFullStartupLatency)
 	})
 }
 
@@ -104,7 +116,7 @@ func NewPodStartupLatencyDataMonitor(c clientset.Interface, purgeAfter time.Dura
 }
 
 // Run starts a PodStartupLatencyDataMonitor: it creates all watches, populates PodStartupData and updates
-// PodE2EStartupLatency metric
+// PodE2EStartupLatency and PodFullStartupLatency metrics
 func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
 	controller := NewWatcherWithHandler(
 		&cache.ListWatch{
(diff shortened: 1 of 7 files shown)


slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -51,6 +52,14 @@ var (
 			Buckets: prometheus.LinearBuckets(0.5, 0.25, 50),
 		},
 	)
+
+	PodFullStartupLatency = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    PodFullStartupLatencyKey,
+			Help:    "Pod e2e startup latencies in milliseconds, without image pull times",
```
shyamjvs: Please fix this description. RIght now it's saying the same as the other metric.

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -51,6 +52,14 @@ var (
...
+			Buckets: prometheus.LinearBuckets(0.5, 0.5, 100),
```
shyamjvs: Do we really need those many buckets with that width (compared to the other metric)?
I *might* be wrong but IIRC from a past discussion, histogram with more buckets costs server-side computations more for calculating %iles.

gmarek: It's additional 800 bytes. I really wouldn't worry about that.

slo-monitor/src/monitors/pod_monitor.go
gmarek: Done.

slo-monitor/src/monitors/pod_monitor.go
shyamjvs: Well, if you say so :)

gmarek: There's probably some overhead, but we can live with few orders of magnitude of overhead in this case;). After all it's just 100 64-bit floats + mutex (or several).

shyamjvs:
LGTM.
Will defer the final lgtm to other reviewers for any comment they may have.

shyamjvs:
Oops... seems like you've assigned only me. Feel free to lgtm unless you want others to review too.

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -279,12 +291,14 @@ func (pm *PodStartupLatencyDataMonitor) updateMetric(key string, data *PodStartu
...
 		data.accountedFor = true
-		startupTime := data.observedRunning.Sub(data.created) - data.finishedPulling.Sub(data.startedPulling)
+		fullStartupTime := data.observedRunning.Sub(data.created)
+		startupTime := fullStartupTime - data.finishedPulling.Sub(data.startedPulling)
 		if startupTime < 0 {
 			glog.Warningf("Saw negative startup time for %v: %v", key, data)
 			startupTime = 0
 		}
 		PodE2EStartupLatency.Observe(float64(startupTime / time.Second))
+		PodFullStartupLatency.Observe(float64(fullStartupTime / time.Second))
```
crassirostris: `fullStartupTime.Seconds()`?

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -51,6 +53,15 @@ var (
 			Buckets: prometheus.LinearBuckets(0.5, 0.25, 50),
 		},
 	)
+
+	// PodFullStartupLatency is a prometheus metric for monitoring pod startup latency including image pull times.
+	PodFullStartupLatency = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    PodFullStartupLatencyKey,
+			Help:    "Pod e2e startup latencies in milliseconds, with image pull times",
```
crassirostris: `Pod e2e startup latencies in SECONDS, with image pull times.`?

slo-monitor/src/monitors/pod_monitor.go
gmarek: Really? Finally:)

slo-monitor/src/monitors/pod_monitor.go
gmarek: Oops. Thanks.

slo-monitor/src/monitors/pod_monitor.go
shyamjvs: Yup, memory is one thing. But there's also 2x more computation involved in computing %ile with 100 buckets as compared to 50 buckets (in the worst case) - so it can be slower. Though I'm not sure how much it'll affect in practice - you probably can neglect it.

gmarek: It would be true it it were Summary. But that's Histogram, so no computation is happening.

shyamjvs: On client-side yes, but on server-side there is computation happening for percentiles (and that's a bit heavy). Ref -  https://prometheus.io/docs/practices/histograms/#quantiles

gmarek: I know nothing about that.

PR #91, opened 2018-02-02
Make SLO monitor publish full pod startup latency including image pul…

gmarek:
…l time

Parts I forgot to push previously


shyamjvs:
/lgtm
Btw - I remember I had this comment about changing the metric description to 'seconds' but I guess I forgot to post it :)

PR #93, opened 2018-02-19
Increase benchmark comparison threshold for more variance

shyamjvs:
/cc @wojtek-t


shyamjvs:
The failure with `go get` seems unrelated - I guess travis is broken for the repo. Merging this manually.

PR #100, opened 2018-03-29
[dns/py] Fixed identation, added requirements.txt and other small changes

migueleliasweb:
- Chaged the identation to follow pep8 patters: https://www.python.org/dev/peps/pep-0008/#indentation
- Added requirements.txt to facilitate local builds and tests
- Fixed most of warnings from `pylint`
- Python3

How to replicate:

**Master branch**
```bash
$ git checkout master 
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
From github.com:kubernetes/perf-tests
 * branch              master     -> FETCH_HEAD
Already up-to-date.
$ make pylint 
cd py && pylint --rcfile=../pylintrc *.py
Using config file perf-tests/dns/pylintrc
************* Module data
C:176, 0: Wrong continued indentation (remove 3 spaces).
                       if r.name in results['data'] else None
                    |  ^ (bad-continuation)
E: 99,16: Undefined variable 'reduce' (undefined-variable)
C: 19, 0: standard import "import re" should be placed before "import numpy" (wrong-import-order)
C: 20, 0: standard import "import sqlite3" should be placed before "import numpy" (wrong-import-order)
************* Module data_test
W: 32, 4: Using deprecated method assertEquals() (deprecated-method)
W: 39, 4: Using deprecated method assertEquals() (deprecated-method)
************* Module params
C:232, 9: Do not use `len(SEQUENCE)` to determine if a sequence is empty (len-as-condition)
R:231, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
************* Module params_test
W: 60, 4: Using deprecated method assertEquals() (deprecated-method)
W: 68, 4: Using deprecated method assertEquals() (deprecated-method)
W: 72, 4: Using deprecated method assertEquals() (deprecated-method)
W: 97, 4: Using deprecated method assertEquals() (deprecated-method)
W:108, 4: Using deprecated method assertEquals() (deprecated-method)
W:110, 4: Using deprecated m
```

**This PR**
```bash
$ git checkout feature/py-dns-small-improvements 
Switched to branch 'feature/py-dns-small-improvements'
$ make pylint 
cd py && pylint --rcfile=../pylintrc *.py
perf-tests/dns/pylintrc
************* Module params
C:236,15: Do not use `len(SEQUENCE)` to determine if a sequence is empty (len-as-condition)
R:235, 8: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
************* Module runner
[...]

PR #102, opened 2018-04-03
[dns/py] Major improvements on dns perf tests

migueleliasweb:
A superset of changes from #100 .

In order to make the DNS perf tests work I had to change quite a bit o code that's why the "full PR" was broken into two.

This PR adds:

- Improved error reporting on boilerplate verification
- Integrated reporting with reporting.py (dns/py/report.py)
- Dockerfile for running python script for dns-perf


bowei:
Can you avoid changing the indentation and whitespace changes? I'm seeing a lot of lines changed due to this.

migueleliasweb:
@bowei Sure ! Should I create a separate PR for this or should I just forget it ?

bowei:
I would prefer to not reindent the files as that obscures git history without much benefit, unless the code is currently in an inconsistent state.

migueleliasweb:
@bowei Basically what happened is that I ended up changing quite a bit o stuff. Sadly the code for the dnsperf from the master branch is not working at all, as seen from my description on #100 .

If you don't mind the amount of lines changed, it would be very interesting to see this PR's code on the master branch since this fixes the whole dns perf test and also adds some interesting features to the module.

The identation althought not essencial, helps the code to be in sync with the PEP8. Once again, not critical but a good practice.

If the identation is a major problem, have a look at this other PR with way less changes due to the PEP8.

See: #103

PR #104, opened 2018-04-06
Pass a configmap to the netperf orch

mkoderer:
Mount a configmap volume to the netperf-orch pod in order to
control the scenarios that are executed. This makes it more
flexible and doesn't need any recompile + image push.


mkoderer:
@bowei can I get your feedback on that?

bowei:
Can we add a command line flag to explicitly enable use of the config map?

PR #107, opened 2018-04-06
Netperf: Add argument for namespace cleanup

mkoderer:


network/benchmarks/netperf/launch.go
```diff
@@ -57,6 +57,7 @@ var (
 	tag            string
 	kubeConfig     string
 	netperfImage   string
+	cleanup_only   bool
```
gonzolino: You should rename `cleanup_only` to `cleanupOnly` in order to make the linter happy.

network/benchmarks/netperf/launch.go
```diff
@@ -73,6 +74,8 @@ func init() {
 	flag.StringVar(&netperfImage, "image", "sirot/netperf-latest", "Docker image used to run the network tests")
 	flag.StringVar(&kubeConfig, "kubeConfig", "",
 		"Location of the kube configuration file ($HOME/.kube/config")
+	flag.BoolVar(&cleanupOnly, "cleanup", false,
```
bowei: Let change the description to be more explicit: "Run the cleanup resources phase only (use this flag to clean up orphaned resources from a test run)"

PR #109, opened 2018-04-10
Updating go version.

krzysied:
Go version 1.7 -> 1.9


wojtek-t:
@krzysied - format enforce by "go lint" changed between those versions.

You need to run:
hack/update-gofmt.sh
to fix that.

[Though please make the output of that as a separate commit of this PR.]

PR #110, opened 2018-04-11
Changing deployment config

krzysied:
Increasing memory limit for the deployment.
Decreasing web page refresh frequency.

### DIFF
--- a/perfdash/deployment.yaml
+++ b/perfdash/deployment.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: default
   labels:
     app: perfdash
-    version: "0.15"
+    version: "0.16"
 spec:
   selector:
     matchLabels:
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: perfdash
-        image: k8s.gcr.io/perfdash:0.15
+        image: k8s.gcr.io/perfdash:0.16
         command:
           - /perfdash
           -   --www=true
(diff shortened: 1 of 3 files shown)


perfdash/deployment.yaml
```diff
@@ -37,8 +37,8 @@ spec:
         resources:
           requests:
             cpu: 600m
-            memory: 1200Mi
+            memory: 2400Mi
```
wojtek-t: Let's not be very conservative with it - I don't want to bump that every time we add something.

Can we increase it to 500Mi or maybe even more?

krzysied: Do you mean to increase to 5000Mi or by 500Mi?

wojtek-t: Good question - I don't really know.
I just don't want to be close the limit - if we know that we need ~2000Mi of memory and believe this is a valid usage, let's leave significantly more space there to avoid bumping it every single time.

I think 5000Mi may be enough.
OTOH, this may not fit into nodes if they are small. So we need to balance those two.

I'm just saying we shouldn't be that close to the actual usage with limit - I will let you figure out the correct solution for it.

krzysied: Based on my observation, perfdash uses ~2000Mi. I'm increasing limit to 4000Mi.

wojtek-t:
OK - sorry. Now I know - my login there is bad. Will fix.

PR #111, opened 2018-04-12
Fix wojtek-t`s login in OWNERS file

wojtek-t:


shyamjvs:
Crap.. we reached a deadlock. I'm not an owner of perf-tests :p

shyamjvs:
Merging manually due to bootstrapping problem (Marek is OOO).

PR #112, opened 2018-04-12
Keeping selected test name

krzysied:
Refresh shouldn't change selected test name (if possible).


shyamjvs:
/lgtm
(we were still discussing abt it offline:)

Thanks for this cool change.
Btw @wojtek-t - this should no longer trigger that refresh issue.

PR #117, opened 2018-04-25
undefined values workaround

krzysied:
Workaround for undefined values.
Allows test phase plots to be displayed.


shyamjvs:
Could you also create the new image and update the version as a separate commit in the same PR?

krzysied:
>Could you also create the new image and update the version as a separate commit in the same PR

Done.

PR #120, opened 2018-05-10
ClusterLoader design

wojtek-t:
@kubernetes/sig-scalability-feature-requests 
@sjug @countspongebob @shyamjvs


clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+	Measurements []string
+}
+```
+
+
+### Top level config
+To define a test scenario, user will be providing the following config:
+
+```
+struct Config {
```
sjug: This is the biggest question mark, as the goal is to make running these tests accessible to users. While the data structures are robust in allowing CRUD operations as well as multiple steps, the configuration will be nested and elaborate.

The current configuration is much more declarative in nature, as we define:
```
ClusterLoader:
  projects:
    - num: 2
      basename: nginx-explorer
      tuning: default
      RCs:
        - num: 10
          file: nginx.yaml
        - num: 1
          file: explorer-pod.yaml
      pods:
        - num: 50
          image: k8s.gcr.io/pause-amd64:3.0
          basename: pausepods
          file: pod-pause.json
```
This is roughly equivalent to:
```
Config:
  Namespaces: 2
  Steps:
    Phases:
      Creates:	
      - NamespaceRange:
          min: 1
          max: 2
	ObjectType: RC
	ObjectTemplate: /path/nginx.yaml
	ReplicasPerNamespace: 10
	TuningSet: default
      - NamespaceRange:
          min: 1
          max: 2
	ObjectType: RC
	ObjectTemplate: /path/explorer-pod.yaml
	ReplicasPerNamespace: 1
	TuningSet: default
      - NamespaceRange:
          min: 1
          max: 2
	ObjectType: Pod
	ObjectTemplate: /path/pod-pause.json
	ReplicasPerNamespace: 50
	TuningSet: default
```

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+be fairly simple to extend these types.
+
+
+#### PhaseDelete
+
+PhaseDelete defines a set of objects that should be deleted within a single phase.
+We define it as following:
+
+```
+struct PhaseDelete {
```
sjug: Is there some way to simply clean up all objects created?

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+	TuningSet string
+}
+```
+
+This describes that objects of a given type from a given namespaces selected by a
+given selector should be deleted and those updates should be spread over time using
+a referenced tuning set.
+
+
+### Measurements
```
sjug: I like building in the measurement capture into Cluster Loader..

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+```
+
+The remaining question is how to incorporate them into config. To avoid
+introducing another fields to “Phase” type (and making it more confusing to users),
+we introduce a concept of a “test step” where a test step can be either a
+number of phases or a number of measurement actions. We define it as following:
+
+```
+struct Step {
+	// Exactly one of these slices should be non-empty.
```
sjug: Why can a step only have a phase or a measurement?

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+	// TBD: Way to access cluster.
+
+	// Number of namespace in the test.
+	Namespaces int32
+	// Steps of the test.
+	Steps []Step
+}
+```
+
+With a test being defined by a single json-file, it should be pretty simple to
```
sjug: Will yaml be compatible, as it is far more human readable?

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+
+```
+struct TuningSet {
+	Name string
+	// Exactly one of the following should be set.
+	QpsLoad        *QpsLoad
+	RandomizedLoad *RandomizedLoad
+}
+
+// QpsLoad defines a uniform load with a given QPS.
```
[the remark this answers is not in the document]
wojtek-t: Not sure I understand this comment.
QPS - is "queries per second" - I don't understand the second part.

clusterloader/doc/vision.md
wojtek-t: Not now at least. For now the easiest way would be to:
- for every object type, create a phase delete selecting all pods (label.Everything()) from all namespace (by appropriate range)

I don't think it's very painful. WDYT?

clusterloader/doc/vision.md
wojtek-t: Thanks!

clusterloader/doc/vision.md
wojtek-t: What else do you suggest?

clusterloader/doc/vision.md
wojtek-t: 1. With the current definition, it's not possible to test Updates. So we need a concept of phase to achieve that (and that's pretty crucial).

2. Yes - it's roughly that (we may consider building-in the "namespace" deeper into that).
But after thinking much more about it in the meantime, I think that exposing explicitly some resources (e.g. pods, RCs etc. simply doesn't scale). and I want to be it generic so that all resources are treated roughly the same.

clusterloader/doc/vision.md
wojtek-t: Yes - we should make it work for both json and yaml.

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+As a result, we will introduce the following rules:
+- In the top-level test definition, we will define the number of namespaces that will
+  be used by a test.
+- All the namespaces will have names of a form “namespace-<number>” for number
+  in range 1..N (where N is the number of namespaces in a test)
+
+To make it usable, we introduce the following type:
+
+```
+struct Range {
```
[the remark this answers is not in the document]
wojtek-t: In general, we need to make it possible to make some namespace different than others. So we need to make it possible to create something in just subset of namespaces.

clusterloader/doc/vision.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Tunning Set
+
+Since we would like to be fully load even very big clusters, we need to be able to
+create a number of “similar” objects. A “Tuning Set” concept will allow us to
+spread those operations across some time.
+We define Tuning Set as following:
+
+```
+struct TuningSet {
```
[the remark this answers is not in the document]
wojtek-t: I'm fine with making another tuning set (e.g. sth like:
```
type StepLoad struct {
  BurstSize int32
  Pause  time.Duration
}
```

Will that address your concern?

[In general, I think that we may need more in the future. I don't consider the stepping one as crucial initially, but if you really think it's needed, I'm fine with that.]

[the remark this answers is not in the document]
sjug: Yeah that works, we use it very regularly.

clusterloader/doc/vision.md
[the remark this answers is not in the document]
sjug: Is that for consistent addressing of the namespaces as well? namespace-min to namespace-max?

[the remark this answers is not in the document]
wojtek-t: Yes - exactly.

clusterloader/doc/vision.md
sjug: Just not sure why they cant both happen in one step? I guess one would have to happen before the other if they were both defined?

clusterloader/doc/vision.md
sjug: Sounds good

clusterloader/doc/vision.md
sjug: The problem would be if we have many object types the delete phase would be rather large.

clusterloader/doc/vision.md
wojtek-t: Ordering was my main motivation. I would like to really make sure that we are able to start measurement before we start tests (or some of its part) and stop after everything is done.
I didn't want to introduce ordering within a phase to not overcomplicate things.

clusterloader/doc/vision.md
wojtek-t: We may consider making ObjectType to be slice of types.

Would you be ok with that?

clusterloader/doc/vision.md
sjug: It could also be possible to add states to the existing config type having a parent slice of projects, that way one can redefine the "state" at various points. Of course there would have to be additional logic to decide of create/update/delete between the states.

clusterloader/doc/vision.md
wojtek-t: I made that explicit in the doc.

clusterloader/doc/vision.md
wojtek-t: As discussed during SIG meeting yesterday - I agree that more declarative approach makes sense.
I refined the proposal a bit to accommodate that.

Also, I made it very clear what the logic for deciding create/update/delete should be to make it as simple as possible and understandable to users.

clusterloader/doc/vision.md
wojtek-t: Done

clusterloader/doc/vision.md
wojtek-t: After thinking about it a bit more (and changing the proposal a bit to make it declarative), I think this is no longer a problem.
1. Delete is pretty declarative now.
2. If you don't care about deletes as a test owner, you may simply don't care about it.
ClusterLoader itself at the end (as "AfterAllSteps" step) will be deleting all namespaces and that would automatically delete all objects from them too.

I made that explicit in the doc.

wojtek-t:
@krzysied @wasylkowski-a

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,322 @@
...
+We make them separate to ensure correct ordering between those two types of
+actions. It's very important to ensure that proper measurements are started
+before we start given actions and they are stopped when all actions are done.
+
+### Top level config
+To define a test scenario, user will be providing the following config:
+
+```
+struct Config {
+	// TBD: Way to access cluster.
```
shyamjvs: I think the cannonical way we're doing this today in our testing is [passing path to](https://github.com/kubernetes/kubernetes/blob/master/hack/ginkgo-e2e.sh#L78) kubeconfig file. That's probably easier than passing address, cert, etc. Maybe let's use that?

wojtek-t: I'm fine with that (it's a bit of an implementation detail though).

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,322 @@
...
+
+The remaining question is how to incorporate them into config. To avoid
+introducing another fields to “Phase” type (and making it more confusing to users),
+we introduce a concept of a “test step” where a test step can be either a
+number of phases or a number of measurement actions. We define it as following:
+
+```
+struct Step {
+	// Only one of Phases or Measurements can be non-empty.
+	Phases       []Phase
```
shyamjvs: Is there a reason why you need to make these fields arrays? We're already having an array of Steps at the higher-level right? Unnecessary nesting can complicate our json.

clusterloader/docs/vision.md
shyamjvs: In that case, maybe we shouldn't be mentioning it here and then add a TBD? :)

wojtek-t: It should be mentioned here, because it will be here. I added "path to kubeconfig" as possible solution here.

clusterloader/docs/vision.md
wojtek-t: yeah - good point. Changed to pointers.

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,322 @@
...
+
+
+### Phase
+Since phase will consist of a number of create, update and delete operations (potentially
+on many different object types). So at the high level description of phase will be
+defined as:
+
+```
+struct Phase {
+	Declarations []PhaseDeclare
```
shyamjvs: The same here.. You can simply have `Step = { *PhaseDeclare or *Measurement }` instead of one more struct and array.

shyamjvs:
@wojtek-t This lgtm at a high-level. While I continue to review, a suggestion from my side -

It would be much easier to comprehend the doc if you structure it in a top-down manner IMO. Personally, it was a bit hard for me to keep track of each individual structs that you were introducing and then slowly trying to relate them. Instead if you started with defining what a Testconfig is (i.e a series of steps) and then what a step is (i.e a measurement or a phase) and so on, it'll be easier to understand.

clusterloader/docs/vision.md
wojtek-t: That isn't true - this would reduce what can be expressed.
Having here []PhaseDeclare allows you to e.g. in parallel create one set of object, create the second set of objects and delete the third of objects.
So basically, PhaseDeclare`s within a single phase are being executed in parallel.

wojtek-t: That said, we can probably get rid of `Phase` here and change Step to be either []PhaseDeclare or Measurement.

I will change that and make explicit in the doc.

shyamjvs: > I will change that and make explicit in the doc.

Yes, thanks.

And wrt your parallelization argument, let me think a bit more if this is the way we want to formulate it.

wojtek-t: Actually, I simplified it a bit. Now we only have a concept of "Step", "Phase" and "Measurement" and I got rid of "PhaseDeclare" at all.

PTAL

shyamjvs:
> you can read it from the end to beginning :-)

Trust me.. This is what I ended up doing (to save time).

> understand the building blocks and then I understand what the final thing is

This might be really good approach when you're coming up with the design. But as someone (like me) who's reading the doc and isn't already familiar with your line of thought when you were writing it, it's harder to reason why you're doing it that way.. until when I reach the end and the suspense is finally revealed :)

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,320 @@
...
+	Name string
+	// Exactly one of the following should be set.
+	QpsLoad        *QpsLoad
+	RandomizedLoad *RandomizedLoad
+	SteppedLoad    *SteppedLoad
+}
+
+// QpsLoad defines a uniform load with a given QPS.
+struct QpsLoad {
+	InitialDelay time.Duration
```
shyamjvs: I sense that InitialDelay is a characteristic that can be common across the strategies.
Do you think we should move it up into TuningSet?

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,320 @@
...
+	Qps          float
+}
+
+// RandomizedLoad defines a load that is spread randomly
+// across a given total time.
+struct RandomizedLoad {
```
shyamjvs: Can you help me understand how this is different from QpsLoad?

clusterloader/docs/vision.md
wojtek-t: Yes - done.

clusterloader/docs/vision.md
wojtek-t: QPSLoad is that we generate exactly "QPS" queries per second (so one query every 1/QPS seconds).
RandomizedLoad is that we randomize the load over such period of time.

They are pretty similar in general.

[As an example:
- RandomizedLoad is how we spread over time things in load test
- QpsLoad is how we spread over time creations of latency pods]

shyamjvs: I see.. thanks.

In that case, I believe it'll be much better to have `TimeSpread` instead of `AverageQps` (as you probably don't want to calculate it's value based on TimeSpread and #calls).

wojtek-t: I'm not sure I understand it.
I think that average qps is exactly what I would like to pass.

IIUC, `TimeSpread` you're proposing correctly, would mean that I would like to create everything over a given time - this for me would be a separate policy (which we may want to add in the future, but for the purpose of migrating our existing e2e tests it won't be needed)

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,318 @@
...
+#### Phase
+
+Phase declaratively defines state of objects we should reach in the underlying
+Kuberneces cluster. A single declaration may result in a number of create, update
+and delete operation depending on the current state of the cluster.
+
+We define the phase as following:
+
+```
+struct Phase {
```
shyamjvs: I can't think of a way in which we can use this way of defining Phase to for e.g doing sth like scale RCs from a given template to different values of `replicas` (like how we're doing in our load test).

So in the current form, IIUC this either means having 16k templates i.e one per RC (which I don't think we want to do) or scaling all RCs using a given template to the same no. of replicas (which can potentially change scenario of our load test).

shyamjvs: > which can potentially change scenario of our load test

Maybe that's fine, but I'm not sure.

clusterloader/docs/vision.md
shyamjvs: Ok.. let me instead ask this question. What do you think should be the averageQps for the scaling operation in our load test?

clusterloader/docs/vision.md
wojtek-t: > I can't think of a way in which we can use this way of defining Phase to for e.g doing sth like scale RCs from a given template to different values of replicas (like how we're doing in our load test).

You can - the template will contain:
```
"numReplicas: "3+Rand()%5"
```

We don't need to change the scenario of our tests :)

clusterloader/docs/vision.md
wojtek-t: Currently we have exactly that - it's by default set to 10.

I don't want to use timespread, because then if you say change the size of the cluster - they you only need to change the number of objects and not change the tunning set.

clusterloader/docs/vision.md
shyamjvs: That's a cool idea! One thing to note however is this might mean evaluating the template 16k times (which is probably fine).

wojtek-t: Yes - that would require mutliple evaluations. I think it's fine, but it's a good point that it requires a note - adding.

clusterloader/docs/vision.md
wojtek-t: > That's the expected throughput we want for pod churn. I believe what we want here is the rate at which we're going to change the RC objects as those are what would be involved in the Phase operations IIUC.

Yes - that was kind of shortcut from my side.
So let's assume that we want to have 10pods/s churn on average.
Then, if small RCs are 1/2 of all pods, then for these we want to generate 5pods/s. Given they have numReplicas=5, we would like to generate 1 per s on average.
if medium RCs are 1/4 of all pods, then we want to generate 2.5pods/s. Given they have numReplicas=25 (IIRC), that would mean an average load of 0.1 per s.
and so on

Yes - it requires some match to be done by someone creating a test. We may consider changing our tests in the future. But I think the steps should be:
- migrate to cluster loader
- then start changing tests if we want
We shouldn't do the two steps at the same time.

clusterloader/docs/vision.md
wojtek-t: Added to caveats section.

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,326 @@
...
+The remaining question is how to incorporate them into config.
+We will do that by extending `Step` definition, where each step can be
+either a number of phases or a number of measurement actions.
+We define it as following:
+
+```
+struct Step {
+	// Only one of these can be non-empty.
+	Phases       []Phase
+	Measurements []string
```
shyamjvs: This is still the same. Aren't we moving it to single entry?

wojtek-t: I reverted. The reasons are:
- I got rid of "DeclarativePhase": Phase is now effectively what previously was "DeclarativePhase".
 We need to have a slice of those, if we want to run couple different things in parallel.
- if we need a slice for phases, let's make it consistent for measurements :)

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Step
+
+Each step will consist of a number of create, update and delete operations (potentially
+on many different object types). At the high level, single step contains with a number
+of `Phase` definitions, which all are being run in parallel:
+
+```
+struct Step {
+	Phases []Phase
```
shyamjvs: This is a bit confusing. You've redefined the same struct later in the document differently (with Measurements).

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,326 @@
...
+Kuberneces cluster. A single declaration may result in a number of create, update
+and delete operation depending on the current state of the cluster.
+
+We define the phase as following:
+
+```
+struct Phase {
+	// Set of namespace in which objects should be reconciled.
+	// If null, objects are assumed to be cluster scoped.
+	NamespaceRange *Range
```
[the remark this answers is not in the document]
shyamjvs: Sg.. thanks!

clusterloader/docs/vision.md
shyamjvs: Ok, I think we can live with it.

shyamjvs:
Btw - I still suggest having top-down approach. It's very easy (at least for me) to form a mental picture if you can begin with saying "test config is a list of steps" and then "a step is a set of measurements or phases done in parallel". In the current form, I'm having to read so many details about phases, measurements, etc even before reaching the foundational point of what overall test looks like.

shyamjvs:
We can probably verify this hypothesis with a different reader (as I might be biased) :)
@wasylkowski-a @krzysied

clusterloader/docs/vision.md
wojtek-t: Done

wojtek-t:
> It seems you feel extremely strongly about that.
> Since I don't have that strong opinion here, I will change that tomorrow.

This is now changed to be more top-down. PTAL if that's fine for you.

Also, I modified the `Phase` a bit to make it more flexible.

shyamjvs:
Thanks for making the change - will TAL soon.

> PTAL if that's fine for you.

Like we were discussing offline, not having definition of 'Config' at the top (i.e above the definition of 'Step') is making it a bit hard to follow. Maybe you could fix that?

wojtek-t:
> Like we were discussing offline, not having definition of 'Config' at the top (i.e above the definition of 'Step') is making it a bit hard to follow. Maybe you could fix that?

Done - PTAL

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+
+Author: wojtek-t
+
+Last update time: 9th May 2018
+
+## Background
+
+As of 31/03/2018, all our scalability tests, are regular e2e tests written in Go.
+It makes them really unfriendly for developers not working on scalability who just
+want to load tests Kubernetes features they are working on. Doing so in many
```
shyamjvs: load test

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+situation requires understanding how those tests really work, modify their code
```
shyamjvs: situations

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+to test the new feature and only then run and debug tests (or create load test for
```
shyamjvs: You could move this long sentence inside brackets to new sentence?

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+your feature on your own, which may be easier, but on the other hand may not
+exercise important metrics that our performance test check). This workflow is far
```
shyamjvs: tests

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+measurements are started before we start given actions and they are stopped when
+all actions are done.
+
+Also note that all `Phases` and `Measurements` within a single `Step` will be
+run in parallel - a `Step` ends when all its `Phases` or `Measurements` finish.
+
+### Phase
+
+Phase declaratively defines state of objects we should reach in the underlying
+Kuberneces cluster. A single declaration may result in a number of create, update
```
shyamjvs: Kubernetes

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+and delete operation depending on the current state of the cluster.
```
shyamjvs: operations

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+from optimal.
+
+That said, long time ago we came up with the idea that users should be able to
+just bring their own objects definitions in JSON format, potentially annotate them
+with some metadata to describe how the load should be generated from these and
+testing infrastructure should do everything else automatically.
+
+In early 2017 the prototype of “Cluster Loader” has been created. It proved that
+configuring tests with json/yaml files is possible. But its functionality is very
+limited and it was very far from enabling migration any of existing scalability
```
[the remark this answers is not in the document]
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+
+Phase declaratively defines state of objects we should reach in the underlying
+Kuberneces cluster. A single declaration may result in a number of create, update
+and delete operation depending on the current state of the cluster.
+
+We define the phase as following:
+
+```
+struct Phase {
+	// Set of namespace in which objects should be reconciled.
```
shyamjvs: namespaces

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+## Vision
+
+At the high level, a single test will consist of a number of steps. If we have
+really good usecase, we may consider creating a DAG of steps, but for now we
+don't have any rationale for it. In each of steps we will be creating, updating
+and/or deleting a number of different objects.
+Additionally, we will introduce a set of predefined operations (that user will be
+able to use as phases). They will allow users to monitor/measure performance impact
+of user-defined phases.
+The following subsections will describe this in a bit more details.
```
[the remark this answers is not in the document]
wojtek-t: Interesting - I didn't know that.

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+We would like to get back to that and build fully functional Cluster Loader and
+use it as a framework to run all our scalability tests. This doc is describing
+high-level vision of how this will work.
+
+
+## Vision
+
+At the high level, a single test will consist of a number of steps. If we have
+really good usecase, we may consider creating a DAG of steps, but for now we
+don't have any rationale for it. In each of steps we will be creating, updating
```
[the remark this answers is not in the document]
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+	// Steps of the test.
+	Steps []Step
+}
+```
+
+With a test being defined by a single json/yaml file, it should be pretty simple
+to modify scenarios, and fork them to new ones.
+
+Note that before running any steps, ClusterLoader will create all the necessary
+namespace and after running all of them will delete them (together with all
```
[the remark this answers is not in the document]
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,335 @@
...
+objects that remained undeleted after running the test). Namespace are described
```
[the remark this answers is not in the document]
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,334 @@
...
+	// Set of namespaces in which objects should be reconciled.
+	// If null, objects are assumed to be cluster scoped.
+	NamespaceRange *NamespaceRange
+	// Number of instances of a given object to exist in each
+	// of referenced namespaces.
+	ReplicasPerNamespace int32
+	// Name of TuningSet to be used.
+	TuningSet string
+	// A set of objects that should be reconciled.
+	Objects []Object
```
shyamjvs: I'm not sure I understand why this should be an array?
This would iiuc mean that we use same `ReplicasPerNamespace` to be applied for all the objects in the array (which I can't see much use of).

shyamjvs: Also same applies wrt other configs like TuningSet and NamespaceRange.

shyamjvs: Ok... I see you have an explanation below. Let me read it.

shyamjvs: Read it and I still have some confusion. So IIUC:

- we already have steps that are serial
- within a step we have phases that are parallel
- with a phase we have objects that are again serially processed

What I don't understand is why it isn't enough to have just one serial construct and one parallel construct? Maybe we can discuss this f2f.

shyamjvs:
So @wojtek-t and me discussed offline about this and agreed after discussing some use-cases that the current design is a reasonably good trade-off of usability and flexibility. I'm going to lgtm this (neglecting typos by will:) once explanations of few design choices are made more clearer. Also, we'd follow this up with a user-guide demonstrating how to use the tool for different use-cases.

clusterloader/docs/vision.md
wojtek-t: Discussed f2f and added rationale below.

wojtek-t:
> So @wojtek-t and me discussed offline about this and agreed after discussing some use-cases that the current design is a reasonably good trade-off of usability and flexibility. I'm going to lgtm this (neglecting typos by will:) once explanations of few design choices are made more clearer. Also, we'd follow this up with a user-guide demonstrating how to use the tool for different use-cases.

Added some explanations. PTAL if that is satisfying for you and if not let me know which part should be clarified.

clusterloader/doc/vision.md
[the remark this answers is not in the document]
jeremyeder: It's a basic requirements. Allows us to drill down into efficiency metrics and prom data at particular load levels, often when we see a noticeable change  in relevant metrics

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+
+We would like to get back to that and build fully functional Cluster Loader and
+use it as a framework to run all our scalability tests. This doc is describing
+high-level vision of how this will work.
+
+
+## Vision
+
+At the high level, a single test will consist of a number of steps. If we have
+really good usecase, we may consider creating a DAG of steps, but for now we
```
shyamjvs: Btw - I think this DAG argument can now be removed with the newer design. It's possible to achieve even a DAG of steps now with just a sequence of steps by merging nodes of the DAG by BFS level number. And by merging nodes, I mean joining their []Phases. Wdyt?

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+}
+```
+
+We make `Phases` and `Measurements` separate concepts to ensure correct ordering
+between those two types of actions. It's very important to ensure that proper
+measurements are started before we start given actions and they are stopped when
+all actions are done.
+
+Also note that all `Phases` and `Measurements` within a single `Step` will be
+run in parallel - a `Step` ends when all its `Phases` or `Measurements` finish.
```
shyamjvs: Are we also mentioning somewhere that the 'steps' themselves run serially?

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+	// Base name from which names of objects will be created.
+	// Names of objects will be "basename-0", "basename-1", ...
+	Basename string
+	// A file path to object definition.
+	ObjectTemplatePath string
+}
+```
+
+The semantic of the above structure will be as following:
+- `Phases` within a single `Step` will be run in parallel
```
shyamjvs: Ok.. So here's a good place to clarify that the 'steps' are run serially (as I suggested in above comment).

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+single `Phase` operations for a given `Object` may only be of a single type
+(create, update of delete).
+
+To make it more explicit:
+- if `ReplicasPerNamespace` is different than it previously was, we will create
+  or delete a number of objects to reach expected cluster state
+- if `ReplicasPerNamespace` is the same as it previously was, we will update all
+  objects to the referenced template.
+
+An appropriate validation will be added to cluster loader to ensure that for a
```
shyamjvs: s/that/the above/

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+given input config.
+
+Note that (namespace number, object type, basename) tuple defines a set of object.
```
shyamjvs: 'a set of replicated objects' ?

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+	// Names of objects will be "basename-0", "basename-1", ...
+	Basename string
+	// A file path to object definition.
+	ObjectTemplatePath string
+}
+```
+
+The semantic of the above structure will be as following:
+- `Phases` within a single `Step` will be run in parallel
+- `Object` within a single `Phase` will be reconciled in serial for a given
```
shyamjvs: Objects

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+  (namespace, replica number pair). For different (namespace, replica number)
```
shyamjvs: (namespace, replica number) pair

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+  pairs they will be spread using a given tuning set.
+
+The rationale for having such structure is the following:
+- Slice of `Object` allows for making a unit of work to be greater than
```
shyamjvs: I'd suggest rewording this sentence as:

`Objects` represents a collection of k8s objects that can be logically thought of as a unit of workload (for e.g a web application that's comprised of a service, a deployment and a volume). Conceptually, this collection is our unit for replication. The replication itself is done according to the TuningSet and ReplicasPerNamespace parameters of the `Phase`. Also note that we process the `Objects` slice serially which allows ordering while processing objects of a unit (e.g create service before creating deployment).

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+```
+struct ObjectType {
+	APIGroup   string
+	APIVersion string
+	Kind       string
+}
+```
+
+Using this will allow us to easily use dynamic client in most of the places
+which may significantly simply Cluster Loader itself.
```
shyamjvs: simplify

clusterloader/docs/vision.md
shyamjvs: Makes sense.. thx for explaining.

clusterloader/docs/vision.md
```diff
@@ -0,0 +1,344 @@
...
+```
+struct Phase {
+	// Set of namespaces in which objects should be reconciled.
+	// If null, objects are assumed to be cluster scoped.
+	NamespaceRange *NamespaceRange
+	// Number of instances of a given object to exist in each
+	// of referenced namespaces.
+	ReplicasPerNamespace int32
+	// Name of TuningSet to be used.
+	TuningSet string
```
shyamjvs: You are using name of tuningSet instead of the struct itself. Where are we defining those tuningSets then? Are they defined globally (I can't find them in `Config`)?

clusterloader/docs/vision.md
wojtek-t: Yeah - I think it no longer makes much sense. Removed.

clusterloader/docs/vision.md
wojtek-t: Yes - it was part of rationale. But I also made that explicit here.

clusterloader/docs/vision.md
wojtek-t: Good point - yes - they should be part of config. Added.

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: Slightly modified - but thanks for making it clear.

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

clusterloader/docs/vision.md
wojtek-t: done

PR #122, opened 2018-05-15
Update README for comparison tool

wojtek-t:


shyamjvs:
I guess we're confusing 'compare' tool and 'benchmark' tool here (the latter is what I called performance comparison tool).

shyamjvs:
Also, I've already linked the design doc for the benchmark tool from it's README - https://github.com/kubernetes/perf-tests/blob/master/benchmark/README.md.
IMO we should keep that readme short and crisp with the design doc linked there (instead of putting design doc in readme). Wdyt?

wojtek-t:
> I guess we're confusing 'compare' tool and 'benchmark' tool here (the latter is what I called performance comparison tool).

Yeah sorry for that.

> Also, I've already linked the design doc for the benchmark tool from it's README - 

I made the docs directory for that.

PR #125, opened 2018-05-24
Perfdash label refresh

krzysied:
This change will prevent from changing selected label values due to data refresh.
Labels will be reset only if needed.


shyamjvs:
/lgtm

Thanks for this much-needed fix!

PR #142, opened 2018-06-26
ClusterLoader - Adding api validation

krzysied:
Adding validation for the new clusterloader api structures. Tests included.

### DIFF
--- a/clusterloader2/api/validation/validation.go
+++ b/clusterloader2/api/validation/validation.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import (
+	"os"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does a simple validation for config and all included structures.
+// Returns list of all not satisfied conditions.
+func VerifyConfig(c *api.Config) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if c.AutomanagedNamespaces < 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("automanagedNamespaces"), c.AutomanagedNamespaces, "must be non-negative"))
+	}
+	for i := range c.TuningSets {
+		allErrs = append(allErrs, verifyTuningSet(&c.TuningSets[i], field.NewPath("tuningSets").Index(i))...)
+	}
+	for i := range c.Steps {
+		allErrs = append(allErrs, verifyStep(&c.Steps[i], field.NewPath("steps").Index(i))...)
+	}
+	return allErrs
+}
+
+func verifyStep(s *api.Step, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for i := range s.Phases {
+		allErrs = append(allErrs, verifyPhase(&s.Phases[i], fldPath.Child("phases").Index(i))...)
+	}
+	for i := range s.Measurements {
+		allErrs = append(allErrs, verifyMeasurement(&s.Measurements[i], fldPath.Child("measurements").Index(i))...)
+	}
+	isPhasesEmpty := len(s.Phases) == 0
+	isMeasurementsEmpty := len(s.Measurements) == 0
+	if isPhasesEmpty && isMeasurementsEmpty {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("measurements"), s.Measurements, "can't be empty with empty phases"))
+	}
+	if !isPhasesEmpty && !isMeasurementsEmpty {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("measurements"), s.Measurements, "can't be non-empty with non-empty phases"))
+	}
+	return allErrs
+}
+
+func verifyPhase(p *api.Phase, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if p.NamespaceRange != nil {
+		allErrs = append(allErrs, verifyNamespaceRange(p.NamespaceRange, fldPath.Child("namespaceRange"))...)
+	}
+	if p.ReplicasPerNamespace < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("replicasPerNamespace"), p.ReplicasPerNamespace, "must be non-negative"))
+	}
+	for i := range p.ObjectBundle {
+		allErrs = append(allErrs, verifyObject(&p.ObjectBundle[i], fldPath.Child("objectBundle").Index(i))...)
+	}
+	return allErrs
+}
+
+func verifyObject(o *api.Object, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	errs := validation.IsDNS1123Subdomain(o.Basename)
+	for i := range errs {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("basename"), o.Basename, errs[i]))
+	}
+	if !fileExists(o.ObjectTemplatePath) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("objectTemplatePath"), o.ObjectTemplatePath, "file must exists"))
+	}
+	return allErrs
+}
+
+func verifyNamespaceRange(nr *api.NamespaceRange, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if nr.Min < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("min"), nr.Min, "must be non-negative"))
+	}
+	if nr.Max < nr.Min {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("max"), nr.Max, "must be greater than min"))
+	}
+	return allErrs
+}
+
+func verifyTuningSet(ts *api.TuningSet, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	tuningSetsNumber := 0
+	if ts.QpsLoad != nil {
+		tuningSetsNumber++
+		allErrs = append(allErrs, verifyQpsLoad(ts.QpsLoad, fldPath.Child("qpsLoad"))...)
+	}
+	if ts.RandomizedLoad != nil {
+		if tuningSetsNumber > 0 {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("randomizedLoad"), "may not specify more than 1 tuning set type"))
+		} else {
+			tuningSetsNumber++
+			allErrs = append(allErrs, verifyRandomizedLoad(ts.RandomizedLoad, fldPath.Child("randomizedLoad"))...)
+		}
+	}
+	if ts.SteppedLoad != nil {
+		if tuningSetsNumber > 0 {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("steppedLoad"), "may not specify more than 1 tuning set type"))
+		} else {
+			tuningSetsNumber++
+			allErrs = append(allErrs, verifySteppedLoad(ts.SteppedLoad, fldPath.Child("steppedLoad"))...)
+		}
+	}
+	return allErrs
+}
+
+func verifyMeasurement(_ *api.Measurement, _ *field.Path) field.ErrorList {
+	return nil
+}
+
+func verifyQpsLoad(ql *api.QpsLoad, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if ql.Qps <= 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("qps"), ql.Qps, "must have positive value"))
+	}
+	return allErrs
+}
+
+func verifyRandomizedLoad(rl *api.RandomizedLoad, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if rl.AverageQps <= 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("averageQps"), rl.AverageQps, "must have positive value"))
+	}
+	return allErrs
+}
+
+func verifySteppedLoad(sl *api.SteppedLoad, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if sl.BurstSize <= 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("burstSize"), sl.BurstSize, "must have positive value"))
+	}
+	return allErrs
+}
+
+func fileExists(path string) bool {
+	_, err := os.Stat(path)
+	return !os.IsNotExist(err)
+}
--- a/clusterloader2/api/validation/validation_test.go
+++ b/clusterloader2/api/validation/validation_test.go
@@ -1,0 +1,281 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import (
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+func isValid(errList field.ErrorList) bool {
+	return len(errList) == 0
+}
+
+func TestVerifyQpsLoad(t *testing.T) {
+	for i, test := range []struct {
+		input    api.QpsLoad
+		expected bool
+	}{
+		{
+			input: api.QpsLoad{
+				Qps: 10,
+			},
+			expected: true,
+		},
+		{
+			input: api.QpsLoad{
+				Qps: -1,
+			},
+			expected: false,
+		},
+	} {
+		got := isValid(verifyQpsLoad(&test.input, field.NewPath("")))
+		if test.expected != got {
+			t.Errorf("Verify nr %d = %v, want %v", i, got, test.expected)
+		}
+	}
+}
+
+func TestVerifyRandomizedLoad(t *testing.T) {
+	for i, test := range []struct {
+		input    api.RandomizedLoad
+		expected bool
+	}{
+		{
+			input: api.RandomizedLoad{
+				AverageQps: 10,
+			},
+			expected: true,
+		},
+		{
+			input: api.RandomizedLoad{
+				AverageQps: -1,
+			},
+			expected: false,
+		},
+	} {
+		got := isValid(verifyRandomizedLoad(&test.input, field.NewPath("")))
+		if test.expected != got {
+			t.Errorf("Verify nr %d = %v, want %v", i, got, test.expected)
+		}
+	}
+}
+
+func TestVerifySteppedLoad(t *testing.T) {
+	for i, test := range []struct {
+		input    api.SteppedLoad
+		expected bool
+	}{
+		{
+			input: api.SteppedLoad{
+				BurstSize: 10,
+				StepDelay: time.Duration(10),
+			},
+			expected: true,
+		},
+		{
+			input: api.SteppedLoad{
+				BurstSize: -10,
+				StepDelay: time.Duration(10),
+			},
+			expected: false,
+		},
+	} {
+		got := isValid(verifySteppedLoad(&test.input, field.NewPath("")))
+		if test.expected != got {
+			t.Errorf("Verify nr %d = %v, want %v", i, got, test.expected)
+		}
+	}
+}
+
+func TestVerifyTuningSet(t *testing.T) {
+	for i, test := range []struct {
+		input    api.TuningSet
+		expected bool
+	}{
+		{
+			input: api.TuningSet{
+				QpsLoad: &api.QpsLoad{
+					Qps: 10,
+				},
+			},
+			expected: true,
+		},
+		{
+			input: api.TuningSet{
+				QpsLoad: &api.QpsLoad{
+					Qps: 10,
+				},
+				RandomizedLoad: &api.RandomizedLoad{
+					AverageQps: 10,
+				},
+			},
+			expected: false,
+		},
+	} {
+		got := isValid(verifyTuningSet(&test.input, field.NewPath("")))
+		if test.expected != got {
+			t.Errorf("Verify nr %d = %v, want %v", i, got, test.expected)
+		}
+	}
+}
+
+func TestVerifyNamespaceRange(t *testing.T) {
+	for i, test := range []struct {
+		input    api.NamespaceRange
+		expected bool
+	}{
+		{
+			input: api.NamespaceRange{
+				Min: 0,
+				Max: 10,
+			},
+			expected: true,
+		},
+		{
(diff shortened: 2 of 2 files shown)


clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does simple validatio for config and all included structures.
+// Returns error is any condition is not satisfied.
+func VerifyConfig(c *api.Config) error {
+	if c.AutomanagedNamespaces < 1 {
+		return fmt.Errorf("Expected one or more namespaces!")
```
wojtek-t: Not necessary  - I can imagine scenarios where users don't want to have any automanaged scenarios at all and want to handle everything themselves.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,158 @@
...
+			return err
+		}
+	}
+	for _, measurement := range s.Measurements {
+		if err := verifyMeasurement(&measurement); err != nil {
+			return err
+		}
+	}
+	if (len(s.Phases) == 0) == (len(s.Measurements) == 0) {
+		return fmt.Errorf("Empty Phases and Measurements!")
```
wojtek-t: errors should start with small letter (everywhere here)

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,158 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does simple validatio for config and all included structures.
+// Returns error is any condition is not satisfied.
+func VerifyConfig(c *api.Config) error {
```
wojtek-t: At the high level, maybe we should make the validation more similar to what we do in k8s?

In particular, we can vendor:
k8s.io/apimachinery/pkg/util/validation/...
package, and then make validation methods more like:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L330
(take fldPath as argument,  return ErrorList, use those field.Required, field.Invalid etc error)

clusterloader2/api/validation/validation.go
krzysied: Changed to >= 0 required.

clusterloader2/api/validation/validation.go
krzysied: done

clusterloader2/api/validation/validation.go
krzysied: I didn't know about k8s.io/apimachinery/pkg/util/validation/... - it looks quite useful.
I've change the code, so it uses ErrorList.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does simple validation for config and all included structures.
+// Returns error is any condition is not satisfied.
+func VerifyConfig(c *api.Config) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldPath := field.NewPath(".")
+	if c.AutomanagedNamespaces < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "AutomanagedNamespaces", "must be zero or more"))
```
wojtek-t: non-negative

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	for i, tuningSet := range c.TuningSets {
+		allErrs = append(allErrs, verifyTuningSet(&tuningSet, fldPath.Child("TuningSets").Index(i))...)
```
wojtek-t: all "Child" methods reference to json representation of a field. So it should start with small letter (it.s json tag value)

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// verifyPhase requires ReplicasPerNamespace to have positive value.
+func verifyPhase(p *api.Phase, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if p.NamespaceRange != nil {
+		allErrs = append(allErrs, verifyNamespaceRange(p.NamespaceRange, fldPath.Child("NamespaceRange"))...)
+	}
+	if p.ReplicasPerNamespace < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "ReplicasPerNamespace", "must be positive"))
```
wojtek-t: non-negative

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+// but never both at the same time.
+func verifyStep(s *api.Step, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for i, phase := range s.Phases {
+		allErrs = append(allErrs, verifyPhase(&phase, fldPath.Child("phases").Index(i))...)
+	}
+	for i, measurement := range s.Measurements {
+		allErrs = append(allErrs, verifyMeasurement(&measurement, fldPath.Child("measurements").Index(i))...)
+	}
+	if (len(s.Phases) == 0) == (len(s.Measurements) == 0) {
```
wojtek-t: While it is correct, I think the following is more readable.
```
numPhases := len(s.Phases)
numMeasurements := len(s.Measurement)
if numPhases + numMeasurements == 0 || (numPhases > 0 && numMeasurements > 0) {
  ....
```

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+	for i, tuningSet := range c.TuningSets {
+		allErrs = append(allErrs, verifyTuningSet(&tuningSet, fldPath.Child("tuningSets").Index(i))...)
+	}
+	for i, step := range c.Steps {
+		allErrs = append(allErrs, verifyStep(&step, fldPath.Child("steps").Index(i))...)
+	}
+	return allErrs
+}
+
+// verifyStep requires to either Phases or Measurements be non-empty,
```
wojtek-t: This is private function and this comment describes what the code does (which is a bad practice if the code is simple enough, which is the case here).

Please drop this comment (as well as other such comments for private functions).

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+		allErrs = append(allErrs, field.Invalid(fldPath, "max", "must be non-negative"))
+	}
+	if nr.Max < nr.Min {
+		allErrs = append(allErrs, field.Invalid(fldPath, "min, max", "Max must be greater than Min"))
+	}
+	return allErrs
+}
+
+func verifyObjectType(ot *api.ObjectType, fldPath *field.Path) field.ErrorList {
+	return field.ErrorList{}
```
wojtek-t: That does unnecessary memory allocation.

return nil

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// verifyTuningSet requires exactly one of tuning set type structure
+// (QpsLoad, RandomizedLoad or SteppedLoad) to be non-emtpy.
+func verifyTuningSet(ts *api.TuningSet, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	setTypes := 0
+	if ts.QpsLoad != nil {
+		allErrs = append(allErrs, verifyQpsLoad(ts.QpsLoad, fldPath.Child("qpsLoad"))...)
+		setTypes += 1
```
wojtek-t: setTypes++

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+
+func verifyObjectType(ot *api.ObjectType, fldPath *field.Path) field.ErrorList {
+	return field.ErrorList{}
+}
+
+// verifyTuningSet requires exactly one of tuning set type structure
+// (QpsLoad, RandomizedLoad or SteppedLoad) to be non-emtpy.
+func verifyTuningSet(ts *api.TuningSet, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	setTypes := 0
```
wojtek-t: tuningSetsNumber ?

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+		setTypes += 1
+	}
+	if setTypes != 1 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "qpsLoad, randomizedLoad, steppedLoad", "exactly one type of TuningSet should be set"))
+	}
+	return allErrs
+}
+
+func verifyMeasurement(m *api.Measurement, fldPath *field.Path) field.ErrorList {
+	return field.ErrorList{}
```
wojtek-t: return nil (same everywhere below in such case)

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+		allErrs = append(allErrs, verifySteppedLoad(ts.SteppedLoad, fldPath.Child("steppedLoad"))...)
+		setTypes += 1
+	}
+	if setTypes != 1 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "qpsLoad, randomizedLoad, steppedLoad", "exactly one type of TuningSet should be set"))
+	}
+	return allErrs
+}
+
+func verifyMeasurement(m *api.Measurement, fldPath *field.Path) field.ErrorList {
```
wojtek-t: If you don't use argument, please make it explicit in the function declaration:
```
verifyMeasurement(_ *api.Measurement, _ *field.Path)
```

Same everywhere else where needed.

krzysied: done

clusterloader2/api/validation/validation.go
krzysied: done

clusterloader2/api/validation/validation.go
krzysied: done

clusterloader2/api/validation/validation.go
krzysied: done

clusterloader2/api/validation/validation.go
krzysied: done

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+func VerifyConfig(c *api.Config) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldPath := field.NewPath(".")
+	if c.AutomanagedNamespaces < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "automanagedNamespaces", "must be non-negative"))
+	}
+	for i, tuningSet := range c.TuningSets {
+		allErrs = append(allErrs, verifyTuningSet(&tuningSet, fldPath.Child("tuningSets").Index(i))...)
+	}
+	for i, step := range c.Steps {
```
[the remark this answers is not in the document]
krzysied: done

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+
+// VerifyConfig does simple validation for config and all included structures.
+// Returns error is any condition is not satisfied.
+func VerifyConfig(c *api.Config) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldPath := field.NewPath(".")
+	if c.AutomanagedNamespaces < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "automanagedNamespaces", "must be non-negative"))
+	}
+	for i, tuningSet := range c.TuningSets {
```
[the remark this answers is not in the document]
krzysied: done

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does simple validation for config and all included structures.
+// Returns error is any condition is not satisfied.
+func VerifyConfig(c *api.Config) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldPath := field.NewPath(".")
```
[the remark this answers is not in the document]
krzysied: done

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+
+package validation
+
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does simple validation for config and all included structures.
+// Returns error is any condition is not satisfied.
```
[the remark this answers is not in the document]
krzysied: done

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,148 @@
...
+*/
+
+package validation
+
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does simple validation for config and all included structures.
```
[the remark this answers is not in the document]
krzysied: done

clusterloader2/api/validation/validation.go
wojtek-t: This is not done.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,147 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// VerifyConfig does a simple validation for config and all included structures.
+// Returns list of all not satisfied conditions.
+func VerifyConfig(c *api.Config) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if c.AutomanagedNamespaces < 0 {
+		allErrs = append(allErrs, field.Invalid(nil, "automanagedNamespaces", "must be non-negative"))
```
wojtek-t: why nil? this should be NewPath("automanagedNamespaces")

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,147 @@
...
+// but never both at the same time.
+func verifyStep(s *api.Step, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for i := range s.Phases {
+		allErrs = append(allErrs, verifyPhase(&s.Phases[i], fldPath.Child("phases").Index(i))...)
+	}
+	for i := range s.Measurements {
+		allErrs = append(allErrs, verifyMeasurement(&s.Measurements[i], fldPath.Child("measurements").Index(i))...)
+	}
+	if (len(s.Phases) == 0) == (len(s.Measurements) == 0) {
```
wojtek-t: Copying from previous review:

While it is correct, I think the following is more readable.
```
numPhases := len(s.Phases)
numMeasurements := len(s.Measurement)
if numPhases + numMeasurements == 0 || (numPhases > 0 && numMeasurements > 0) {
  ....
```

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// verifyPhase requires ReplicasPerNamespace to have positive value.
+func verifyPhase(p *api.Phase, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if p.NamespaceRange != nil {
+		allErrs = append(allErrs, verifyNamespaceRange(p.NamespaceRange, fldPath.Child("namespaceRange"))...)
+	}
+	if p.ReplicasPerNamespace < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath, "replicasPerNamespace", "must be non-negative"))
```
wojtek-t: This should be fldPath.Child("replicasPerNamespace").

There are also a number of other similar places - please fix.

clusterloader2/api/validation/validation.go
wojtek-t: Also, the second argument to field.Invalid is the actual value of the field, not it's name - please fix here and in a number of toher places.

krzysied: For some reason, I though that requires (path to structure, field name, text). I stand corrected.
Fixing other places.

clusterloader2/api/validation/validation.go
krzysied: Comments removed.

clusterloader2/api/validation/validation.go
krzysied: I've decided to use "is<smth>Empty" to describe the condition. It should be even clearer.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+	allErrs := field.ErrorList{}
+	for i := range s.Phases {
+		allErrs = append(allErrs, verifyPhase(&s.Phases[i], fldPath.Child("phases").Index(i))...)
+	}
+	for i := range s.Measurements {
+		allErrs = append(allErrs, verifyMeasurement(&s.Measurements[i], fldPath.Child("measurements").Index(i))...)
+	}
+	isPhasesEmpty := len(s.Phases) == 0
+	isMeasurementsEmpty := len(s.Measurements) == 0
+	if (isPhasesEmpty && isMeasurementsEmpty) || (!isPhasesEmpty && !isMeasurementsEmpty) {
```
wojtek-t: Let's split this into two separate if's (the error is different in both):
```
if isPhasesEmpty && isMeasurementsEmpty {
  allErrs = append(allErrs, field.Invalid(fldPath.Child("measurement"), s.Measurements, "can't be empty with empty phases")
}
if !isPhasesEmpty && !isMeasurementsEmpty {
   ...
}
```

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("replicasPerNamespace"), p.ReplicasPerNamespace, "must be non-negative"))
+	}
+	for i := range p.ObjectBundle {
+		allErrs = append(allErrs, verifyObject(&p.ObjectBundle[i], fldPath.Child("objectBundle").Index(i))...)
+	}
+	return allErrs
+}
+
+func verifyObject(o *api.Object, fldPath *field.Path) field.ErrorList {
+	return verifyObjectType(&o.ObjectType, fldPath.Child("objectType"))
```
wojtek-t: We should valid Basename too.
it probably should be "NameIsDNSSubdomain" check.

wojtek-t: Maybe the same for template path?

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func verifyNamespaceRange(nr *api.NamespaceRange, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if nr.Min < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("min"), nr.Min, "must be non-negative"))
+	}
+	if nr.Max < 0 {
```
wojtek-t: This check is not needed if min >=0, then it's enough to check if this is greater than min.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("max"), nr.Max, "must be non-negative"))
+	}
+	if nr.Max < nr.Min {
+		allErrs = append(allErrs, field.Invalid(fldPath, "", "max must be greater than min"))
```
wojtek-t: field.Invalid(fldPath, nr.Max, "must be greater than min")

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	return allErrs
+}
+
+func verifyObjectType(_ *api.ObjectType, _ *field.Path) field.ErrorList {
+	return nil
```
wojtek-t: Those should probably be also DNS subdomains at least.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+func verifyTuningSet(ts *api.TuningSet, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	tuningSetsNumber := 0
+	if ts.QpsLoad != nil {
+		allErrs = append(allErrs, verifyQpsLoad(ts.QpsLoad, fldPath.Child("qpsLoad"))...)
+		tuningSetsNumber++
+	}
+	if ts.RandomizedLoad != nil {
+		allErrs = append(allErrs, verifyRandomizedLoad(ts.RandomizedLoad, fldPath.Child("randomizedLoad"))...)
+		tuningSetsNumber++
```
wojtek-t: Instead of that, you should do (for consistency) the same what we're doing with validating VolumeSounrce for example:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L428

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+	return allErrs
+}
+
+func verifyMeasurement(_ *api.Measurement, _ *field.Path) field.ErrorList {
+	return nil
+}
+
+func verifyQpsLoad(ql *api.QpsLoad, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if ql.Qps < 0 {
```
wojtek-t: <= (qps= doesn't make much sense too)

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,139 @@
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("qps"), ql.Qps, "must have positive value"))
+	}
+	return allErrs
+}
+
+func verifyRandomizedLoad(rl *api.RandomizedLoad, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if rl.AverageQps < 0 {
```
wojtek-t: Same here and below.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,180 @@
...
+	"fmt"
+	"os"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/kubernetes/pkg/api/legacyscheme"
+	"k8s.io/perf-tests/clusterloader2/api"
+
+	_ "k8s.io/kubernetes/pkg/apis/core/install"
```
wojtek-t: I don't like it - why do you even added it?

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	for i := range p.ObjectBundle {
+		allErrs = append(allErrs, verifyObject(&p.ObjectBundle[i], fldPath.Child("objectBundle").Index(i))...)
+	}
+	return allErrs
+}
+
+func verifyObject(o *api.Object, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if o.ObjectType.Kind == "Service" {
```
wojtek-t: Please don't do anything like that - such validation will be done by underlying k8s cluster.
The goal of cluster-loader project is not to copy validation from k8s.

All of this should be removed.

clusterloader2/api/validation/validation.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	if nr.Max < nr.Min {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("max"), nr.Max, "must be greater than min"))
+	}
+	return allErrs
+}
+
+func verifyObjectType(objectType *api.ObjectType, fldPath *field.Path) field.ErrorList {
+	fmt.Printf("all: %v\n", legacyscheme.Scheme.AllKnownTypes())
+	for t := range legacyscheme.Scheme.KnownTypes(schema.GroupVersion{Group: objectType.APIGroup, Version: objectType.APIVersion}) {
```
wojtek-t: See my comment here: https://github.com/kubernetes/perf-tests/pull/148/files#r200572077

I'm now pretty convinced that this is actually not needed and whole ObjectType should be removed.

clusterloader2/api/validation/validation.go
krzysied: it is required for the object type verification. Due to ObjectType being removed, this part will also be removed.

clusterloader2/api/validation/validation.go
krzysied: I'm changing it to IsDNS1123Subdomain verification. This one should verify names correctly.

clusterloader2/api/validation/validation.go
krzysied: I'm removing object type verification.

wojtek-t:
Closing in favor of https://github.com/kubernetes/perf-tests/pull/1694

PR #144, opened 2018-06-27
ClusterLoader - File unmarshaling

krzysied:
Adding function that can unmarshal json or yaml file into go structure

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -63,6 +63,8 @@ type Phase struct {
 
 // Object is a structure that defines the object managed be the tests.
 type Object struct {
+	// TODO(krzysied): possibly ObjectType will be removed. All data can be
+	// acquired from the template.
 	// ObjectType is a type for a given object.
 	ObjectType ObjectType `json: objectType`
 	// Basename is a string from which names of objects will be created.
--- a/clusterloader2/config/codec.go
+++ b/clusterloader2/config/codec.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/viper"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// ReadConfig creates config from file specified by the given path.
+func ReadConfig(path string) (*api.Config, error) {
+	// This must be done after common flags are registered, since Viper is a flag option.
+	base := filepath.Base(path)
+	ext := filepath.Ext(base)
+	viper.SetConfigName(strings.TrimSuffix(base, ext))
+	viper.AddConfigPath(filepath.Dir(path))
+	if err := viper.ReadInConfig(); err != nil {
+		return nil, fmt.Errorf("read config failed: %v", err)
+	}
+	var config api.Config
+	if err := viper.Unmarshal(&config); err != nil {
+		return nil, fmt.Errorf("unmarshaling failed: %v", err)
+	}
+	return &config, nil
+}


clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// UnmarshalFileInto unmarshals json or yaml file specified by the given path.
+func UnmarshalFileInto(path string, out interface{}) error {
+	ext := strings.ToLower(filepath.Ext(path))
+	raw, err := ioutil.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("reading file error: %v", err)
+	}
+	if ext == ".yaml" {
```
wojtek-t: I don't like relying on file extensions.
We already have recognizing decoder in kubernetes repo and we should use that.

pkg/api/legacyscheme.Codecs.universalDecoder is what you're looking for.

Also, I would like to see a test for it.

krzysied: Shouldn't pkg/api/legacyscheme.Codecs.UniversalDecoder be only used for kubernetes configs?
Is it possible to decode any json file with this?

wojtek-t: What you're going to read with this function?

krzysied: I wanted function that could unmarshal api.Config and object template if possible.
I see that pkg/api/legacyscheme.Codecs.UniversalDecoder is perfect for object template, but I need something more generic for config structure.

wojtek-t: Then let's not reinvent the wheel and let's use viper the same way the original cluster loader was doing that.
clusterloader/framework/context.go

wojtek-t:
Please split vendor to a separate PR.

Also - do we really need all of that in vendor? How did you generate that?
[BTW - given how painful vendor is, it might be useful to have a readme how you generated that.]

clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,60 @@
...
+	return &config, nil
+}
+
+// ReadTemplate creates object tempalte from file specified by the given path.
+func ReadTemplate(path string) (runtime.Object, error) {
+	raw, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading file error: %v", err)
+	}
+	obj, err := runtime.Decode(legacyscheme.Codecs.UniversalDecoder(), raw)
```
wojtek-t: After looking into that, I started wondering if we would really need ObjectType at all?
I have a feeling now that it may not be needed.
Would be useful to add a TODO to types.go about that.

clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubernetes/pkg/api/legacyscheme"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// ReadConfig creates config from file specified by the given path.
+func ReadConfig(path string) (*api.Config, error) {
```
wojtek-t: Since what you need is name of config - let's pass a name here. If needed we can have a wrapper to translate from path.

clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,60 @@
...
+	// This must be done after common flags are registered, since Viper is a flag option.
+	base := filepath.Base(path)
+	ext := filepath.Ext(base)
+	viper.SetConfigName(strings.TrimSuffix(base, ext))
+	viper.AddConfigPath(os.Getenv("GOPATH") + "/src/k8s.io/perf-tests/clusterloader2/" + filepath.Dir(path))
```
wojtek-t: make this "src/..." a constant.

clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, fmt.Errorf("read config failed: %v", err)
+	}
+	var config api.Config
+	if err := viper.Unmarshal(&config); err != nil {
+		return nil, fmt.Errorf("unmarshaling failed: %v", err)
+	}
+	return &config, nil
+}
+
+// ReadTemplate creates object tempalte from file specified by the given path.
```
wojtek-t: template

clusterloader2/util/unmarshaling.go
krzysied: I think that passing path is better that passing only a name. I assume that user will pass path to config, which can be anywhere.

clusterloader2/util/unmarshaling.go
krzysied: I've decided to remove this "scr/.." part. At this moment I don't feel it is needed.

clusterloader2/util/unmarshaling.go
krzysied: Object type data can be retrieved from object template, if I understand correctly - every template has always metadata. 
TODO added.

wojtek-t:
@krzysied - as I already wrote above:


Please split vendor to a separate PR.Also - do we really need all of that in vendor? How did you generate that?[BTW - given how painful vendor is, it might be useful to have a readme how you generated that.]

krzysied:
@ wojtek-t I've just realized that I don't need explicitly legacyshema imported; I can use schema and codecs from apimachinery. This way there is no kubernetes dependency needed, thus no vendor required.

clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,61 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+var codecs = serializer.NewCodecFactory(runtime.NewScheme())
```
wojtek-t: I started wondering if that will work in our case. it depends on the compiled in (thus initialized) api-groups.
This may or may not work in our purpose - so I will hold-on with lgtm-ing this one.

krzysied: Probably there will be import to legacyschema added (and all apis). I just realized that initialized scheme is also required for validation PR.

wojtek-t: Validation? You're not validating objects so far..

krzysied: I meant pull request #142 - validation for test configs. I expect that if we want to have validation for ObjectType, we have to have scheme initialized.
It is just another argument for adding initialization for api groups...

wojtek-t: It isn't obvious to me we want to do that - this means that adding tests for newly added object APi requires rebuliding cluster loader. It's even more important when we start thinking about testing stuff for validation before it's merged.

Maybe it's actually enough to read it to runtime.Unstructured, where we don't really need to know the schema.

And for detecting the type of the object - parse it to e.g. metav1.PartialObjectMetadata

wojtek-t: So I would say - let's for now remove it completely, and merge this PR just with parsing config merged.

clusterloader2/util/unmarshaling.go
```diff
@@ -0,0 +1,61 @@
+/*
```
wojtek-t: I don't think this should live in util/unmarshaling.go

Maybe something like:
config/codec.go

clusterloader2/util/unmarshaling.go
krzysied: Ok. I'm moving ReadTemplate method to different PR.

PR #147, opened 2018-06-29
Fixing verify_golint.sh for go1.10+ versions

krzysied:
Removing go version check - everyone uses 1.5+ version as default.

### DIFF
--- a/verify/verify-golint.sh
+++ b/verify/verify-golint.sh
@@ -20,13 +20,6 @@ set -o pipefail
 
 KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
 
-GO_VERSION=($(go version))
-# golint only works for golang 1.5+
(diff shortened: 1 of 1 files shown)


verify/verify-golint.sh
```diff
@@ -22,7 +22,7 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
 
 GO_VERSION=($(go version))
 # golint only works for golang 1.5+
-if [[ -n $(echo "${GO_VERSION[2]}" | grep -E 'go1.1|go1.2|go1.3|go1.4') ]]; then
+if [[ -n $(echo "${GO_VERSION[2]}" | grep -E '(go1.1|go1.2|go1.3|go1.4)(\.|$)') ]]; then
```
wojtek-t: Let's remove this check completely instead.
We're already on much higher go version everywhere.

krzysied: done

PR #148, opened 2018-07-03
ClusterLoader - Adding framework

krzysied:
Adding cluster managing framework.

### DIFF
--- a/clusterloader2/framework/config/clientconfig.go
+++ b/clusterloader2/framework/config/clientconfig.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+const (
+	contentType = "application/vnd.kubernetes.protobuf"
+	qps         = 100
+	burst       = 200
+)
+
+// PrepareConfig creates and initializes client config.
+func PrepareConfig(path string) (*restclient.Config, error) {
+	config, err := loadConfig(path)
+	if err != nil {
+		return nil, err
+	}
+	initializeWithDefaults(config)
+	return config, nil
+}
+
+func restclientConfig(path string) (*clientcmdapi.Config, error) {
+	c, err := clientcmd.LoadFromFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("error loading kubeconfig: %v", err)
+	}
+	return c, nil
+}
+
+func loadConfig(path string) (*restclient.Config, error) {
+	c, err := restclientConfig(path)
+	if err != nil {
+		return nil, err
+	}
+	return clientcmd.NewDefaultClientConfig(*c, &clientcmd.ConfigOverrides{}).ClientConfig()
+}
+
+func initializeWithDefaults(config *restclient.Config) {
+	config.ContentType = contentType
+	config.QPS = qps
+	config.Burst = burst
+}
--- a/clusterloader2/framework/framework.go
+++ b/clusterloader2/framework/framework.go
@@ -1,0 +1,104 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/framework/client"
+	"k8s.io/perf-tests/clusterloader2/framework/config"
+)
+
+const (
+	automanagedNamespaceName = "namespace"
+)
+
+// Framework allows for interacting with Kubernetes cluster via
+// official Kubernetes client.
+type Framework struct {
+	automanagedNamespaceCount int
+	clientSet                 clientset.Interface
+	dynamicClient             dynamic.Interface
+}
+
+// NewFramework creates new framework based on given kubeconfig.
+func NewFramework(path string) (*Framework, error) {
+	conf, err := config.PrepareConfig(path)
+	if err != nil {
+		return nil, fmt.Errorf("config prepare failed: %v", err)
+	}
+	clientSet, err := clientset.NewForConfig(conf)
+	if err != nil {
+		return nil, fmt.Errorf("creating clientset failed: %v", err)
+	}
+	dynamicClient, err := dynamic.NewForConfig(conf)
+	if err != nil {
+		return nil, fmt.Errorf("creating dynamic config failed: %v", err)
+	}
+	return &Framework{0, clientSet, dynamicClient}, nil
+}
+
+// CreateAutomanagedNamespaces creates automanged namespaces.
+func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
+	if f.automanagedNamespaceCount != 0 {
+		return fmt.Errorf("automanaged namespaces already created")
+	}
+	for i := 0; i < namespaceCount; i++ {
+		name := fmt.Sprintf("%v-%d", automanagedNamespaceName, i)
+		if err := client.CreateNamespace(f.clientSet, name); err != nil {
+			return err
+		}
+		f.automanagedNamespaceCount++
+	}
+	return nil
+}
+
+// DeleteAutomanagedNamespaces deletes all automanged namespaces.
+func (f *Framework) DeleteAutomanagedNamespaces() error {
+	for i := 0; i < f.automanagedNamespaceCount; i++ {
+		name := fmt.Sprintf("%v-%d", automanagedNamespaceName, i)
+		if err := client.DeleteNamespace(f.clientSet, name); err != nil {
+			return err
+		}
+	}
+	f.automanagedNamespaceCount = 0
+	return nil
+}
+
+// CreateObject creates object base on given object description.
+func (f *Framework) CreateObject(namespace string, name string, obj *unstructured.Unstructured) error {
+	return client.CreateObject(f.dynamicClient, namespace, name, obj)
+}
+
+// UpdateObject updates object with given name using given object description.
+func (f *Framework) UpdateObject(namespace string, name string, obj *unstructured.Unstructured) error {
+	return client.UpdateObject(f.dynamicClient, namespace, name, obj)
+}
+
+// DeleteObject deletes object with given name and group-version-kind.
+func (f *Framework) DeleteObject(gvk schema.GroupVersionKind, namespace string, name string) error {
+	return client.DeleteObject(f.dynamicClient, gvk, namespace, name)
+}
+
+// GetObject retrieves object with given name and group-version-kind.
+func (f *Framework) GetObject(gvk schema.GroupVersionKind, namespace string, name string) (*unstructured.Unstructured, error) {
+	return client.GetObject(f.dynamicClient, gvk, namespace, name)
+}
--- a/clusterloader2/framework/client/objects.go
+++ b/clusterloader2/framework/client/objects.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client
+
+import (
+	"fmt"
+	"time"
+
+	apiv1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+const (
+	// Parameters for retrying with exponential backoff.
+	retryBackoffInitialDuration = 100 * time.Millisecond
+	retryBackoffFactor          = 3
+	retryBackoffJitter          = 0
+	retryBackoffSteps           = 6
+)
+
+// RetryWithExponentialBackOff a utility for retrying the given function with exponential backoff.
+func RetryWithExponentialBackOff(fn wait.ConditionFunc) error {
+	backoff := wait.Backoff{
+		Duration: retryBackoffInitialDuration,
+		Factor:   retryBackoffFactor,
+		Jitter:   retryBackoffJitter,
+		Steps:    retryBackoffSteps,
+	}
+	return wait.ExponentialBackoff(backoff, fn)
+}
+
+// IsRetryableAPIError verifies whether the error is retryable.
+func IsRetryableAPIError(err error) bool {
+	// These errors may indicate a transient error that we can retry in tests.
+	if apierrs.IsInternalError(err) || apierrs.IsTimeout(err) || apierrs.IsServerTimeout(err) ||
+		apierrs.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) {
+		return true
+	}
+	// If the error sends the Retry-After header, we respect it as an explicit confirmation we should retry.
+	if _, shouldRetry := apierrs.SuggestsClientDelay(err); shouldRetry {
+		return true
+	}
+	return false
+}
+
+func retryFunction(f func() error, isAllowedError func(error) bool) wait.ConditionFunc {
+	return func() (bool, error) {
+		err := f()
+		if err == nil || (isAllowedError != nil && isAllowedError(err)) {
+			return true, nil
+		}
+		if IsRetryableAPIError(err) {
+			return false, nil
+		}
+		return false, err
+	}
+}
+
+// CreateNamespace creates a single namespace with given name.
+func CreateNamespace(c clientset.Interface, namespace string) error {
+	createFunc := func() error {
+		_, err := c.CoreV1().Namespaces().Create(&apiv1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
+		return err
+	}
+	return RetryWithExponentialBackOff(retryFunction(createFunc, apierrs.IsAlreadyExists))
+
+}
+
+// DeleteNamespace deletes namespace with given name.
+func DeleteNamespace(c clientset.Interface, namespace string) error {
+	deleteFunc := func() error {
+		return c.CoreV1().Namespaces().Delete(namespace, nil)
+	}
+	return RetryWithExponentialBackOff(retryFunction(deleteFunc, apierrs.IsNotFound))
+}
+
+// CreateObject creates object based on given object description.
+func CreateObject(dynamicClient dynamic.Interface, namespace string, name string, obj *unstructured.Unstructured) error {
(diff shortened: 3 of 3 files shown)


clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,67 @@
...
+func convertObjectTypeToGVK(objectType *api.ObjectType) schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group:   objectType.APIGroup,
+		Version: objectType.APIVersion,
+		Kind:    objectType.Kind,
+	}
+}
+
+type Framework struct {
+	frameworkClient *framework.Framework
```
wojtek-t: loading framework means loading half of the world...
Can we avoid that?

IIUC, the only usecase for that as of now is the "CreateNamespace" method. 
If so, maybe we can start with implementing this ourselves (we don't need majority of it at the beginning and leave a TODO to consider reusing that in the future if we end-up vendoring whole framework anyway?)

@shyamjvs for thoughts too

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) ([]*apiv1.Namespace, error) {
+	return util.CreateNamespaces(f.frameworkClient, automanagedNamespaceName, namespaceCount)
+}
+
+func (f *Framework) CreateObject(namespace string, objectType *api.ObjectType, name string, obj *unstructured.Unstructured) error {
```
wojtek-t: The order of arguments is strange - please keep namespace and name together.
Same elsewhere.

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+		namespaces = append(namespaces, namespace)
+	}
+	return namespaces, nil
+}
+
+func CreateObject(dynamicClient dynamic.Interface, namespace string, gvk schema.GroupVersionKind, name string, obj *unstructured.Unstructured) error {
+	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
+	obj.SetName(name)
+	_, err := dynamicClient.Resource(gvr).Namespace(namespace).Create(obj)
```
wojtek-t: This should be retried - same in other places.

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func GetObject(dynamicClient dynamic.Interface, namespace string, gvk schema.GroupVersionKind, name string) (*unstructured.Unstructured, error) {
+	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
+	return dynamicClient.Resource(gvr).Namespace(namespace).Get(name, metav1.GetOptions{
+		IncludeUninitialized: true,
+	})
+}
+
+// TODO(krzysied): Add update object function
```
wojtek-t: Can we do that together - how is that different from create and delete?

clusterloader2/framework/framework.go
krzysied: I'm removing all references to "kubernetes...framework" (at least from framework).

clusterloader2/framework/util/util.go
krzysied: Adding retries - the same used in e.g. https://github.com/kubernetes/kubernetes/blob/d65039c56ce4de5f2efdc38aa1284eeb95f89169/test/utils/create_resources.go#L43

clusterloader2/framework/util/util.go
krzysied: If I'm not wrong, kubernetes has different implementation of update. In most cases Delete -> Create is used when this default method fails.

I will add Delete->Create here, but at some point I want to update this code, so it uses correct update call.

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	if err := RetryWithExponentialBackOff(getFunc); err != nil {
+		return nil, err
+	}
+	return obj, nil
+}
+
+// TODO(krzysied): Change it to legit update call
+func UpdateObject(dynamicClient dynamic.Interface, namespace string, name string, gvk schema.GroupVersionKind, obj *unstructured.Unstructured) error {
+	if err := DeleteObject(dynamicClient, namespace, name, gvk); err != nil {
```
wojtek-t: I don't understand this. You should use regular Update call.

krzysied: At this moment, I have no idea how to do this. I mean I know that update method exists. It requires unstructured.Unstructured object which is mostly that same as the one from get function, however it allows only some fields to be overwritten and I haven't figure out yet how to do this.

I would like to proceed  with this solution, cause in my opinion it is not crucial to have a legit update call at this moment. I will fix it later (there is TODO).

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,108 @@
...
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/framework/util"
+)
+
+const (
+	automanagedNamespaceName = "Namespace"
+)
+
+func convertObjectTypeToGVK(objectType *api.ObjectType) schema.GroupVersionKind {
```
wojtek-t: Can we move this function to some util file in this package?

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,108 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/framework/util"
+)
+
+const (
+	automanagedNamespaceName = "Namespace"
```
wojtek-t: "namespace" (instead of "Namespace")

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,108 @@
...
+
+func convertObjectTypeToGVK(objectType *api.ObjectType) schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group:   objectType.APIGroup,
+		Version: objectType.APIVersion,
+		Kind:    objectType.Kind,
+	}
+}
+
+// Framework allows to execute cluster related actions
```
wojtek-t: // Framework allows for interacting with Kubernetes cluster via
// official Kubernetes client.

or sth like that
- you don't perform any actions - you just send requests changing the desired state
- [also namespace is also an objects, it's not special concept]

clusterloader2/framework/util/kubeconfig.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	return c, nil
+}
+
+// LoadConfig loads kubeconfig from given location.
+func LoadConfig(path string) (*restclient.Config, error) {
+	c, err := restclientConfig(path)
+	if err != nil {
+		return restclient.InClusterConfig()
+	}
```
wojtek-t: I'm pretty sure we will need to tune some stuff.
Things like:
- QPS limits
- use protobufs as content-type

BTW - I also think we may want to use multiple clients (though that is less clear for now).

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if err := RetryWithExponentialBackOff(getFunc); err != nil {
+		return nil, err
+	}
+	return obj, nil
+}
+
+// UpdateObject updates with given name, group, version and kind based on given object description.
+func UpdateObject(dynamicClient dynamic.Interface, namespace string, name string, gvk schema.GroupVersionKind, obj *unstructured.Unstructured) error {
+	// TODO(krzysied): Change it to legit update call
```
wojtek-t: You will never convince me that this is right and I will never let it go.

As an example: I would like to update deployment to change number of replicas from 5 to 10.
That is significantly different action that deleting that deployment (meaning removing the existing 5 pods) and then creating new one with 10.

That is just bad.

Re your comment somwhere else - yes, some update actions are not allowed (e.g. some fields are immutable). And that is fine that you can't change them.

shyamjvs: +1. I was pointing the exact same example in our meeting today. It seems this is "too strong" a hack to proceed on with a TODO :)

I think the best place to look for an e.g of how to update an object is kubectl (`kubectl apply -f <file>` does just this).

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+	automanagedNamespaceName = "namespace"
+	contentType              = "application/vnd.kubernetes.protobuf"
+	qps                      = 100
+	burst                    = 200
+)
+
+// Framework allows for interacting with Kubernetes cluster via
+// official Kubernetes client.
+type Framework struct {
+	// TODO(krzysied): Change this to a set of FrameworkClients.
```
wojtek-t: I don't understand this todo - what you would like to achieve?

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+	clientSet     clientset.Interface
+	dynamicClient dynamic.Interface
+}
+
+// CreateAutomanagedNamespaces creates automanged namespaces.
+func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
+	for i := 0; i < namespaceCount; i++ {
+		if err := util.CreateNamespace(f.clientSet, fmt.Sprintf("%v-%d", automanagedNamespaceName, i)); err != nil {
```
wojtek-t: for better readability, maybe:
```
name := fmt.Sprintf(...)
if err := util.CreateNamespace(f.clientSet, name); err != nil {
  return err
}
```

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// DeleteAutomanagedNamespaces deletes automanged namespaces.
```
wojtek-t: all automanaged

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+func (f *Framework) DeleteAutomanagedNamespaces(namespaceCount int) error {
+	for i := 0; i < namespaceCount; i++ {
+		if err := util.DeleteNamespace(f.clientSet, fmt.Sprintf("%v-%d", automanagedNamespaceName, i)); err != nil {
```
wojtek-t: Same here as for create

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// CreateObject creates object base on given object description.
+func (f *Framework) CreateObject(namespace string, name string, objectType *api.ObjectType, obj *unstructured.Unstructured) error {
```
wojtek-t: I would suggest slightly different arguments order:

objectType, namespace, name, obj

These reflects how we define object (first what kind of object, then what namespace, then its name and then body)

Same everywhere else.

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// GetObject retrieves object with given name and object type.
+func (f *Framework) GetObject(namespace string, name string, objectType *api.ObjectType) (*unstructured.Unstructured, error) {
+	gvk := convertObjectTypeToGVK(objectType)
+	return util.GetObject(f.dynamicClient, namespace, name, gvk)
+}
+
+// UpdateObject updates object with given name and object type using given object description.
+func (f *Framework) UpdateObject(namespace string, name string, objectType *api.ObjectType, obj *unstructured.Unstructured) error {
```
wojtek-t: Move update between create and delete.

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,107 @@
...
+	gvk := convertObjectTypeToGVK(objectType)
+	return util.UpdateObject(f.dynamicClient, namespace, name, gvk, obj)
+}
+
+// NewFramework creates new framework based on given kubeconfig.
+func NewFramework(path string) (*Framework, error) {
```
wojtek-t: Move this after framework struct definition

clusterloader2/framework/util/kubeconfig.go
```diff
@@ -0,0 +1,42 @@
...
+
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func restclientConfig(path string) (*clientcmdapi.Config, error) {
+	c, err := clientcmd.LoadFromFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("error loading KubeConfig: %v", err)
```
wojtek-t: s/KubeConfig/kubeconfig/

clusterloader2/framework/util/kubeconfig.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	return c, nil
+}
+
+// LoadConfig loads kubeconfig from given location.
```
wojtek-t: from a given

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,201 @@
...
+
+// UpdateObject updates with given name, group, version and kind based on given object description.
+func UpdateObject(dynamicClient dynamic.Interface, namespace string, name string, gvk schema.GroupVersionKind, obj *unstructured.Unstructured) error {
+	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
+	obj.SetName(name)
+	currentObj, err := GetObject(dynamicClient, namespace, name, gvk)
+	if err != nil {
+		return err
+	}
+	current, err := currentObj.MarshalJSON()
```
wojtek-t: I don't get this - we don't want to create patch and send patch requests. Patch may be a separate command - we want to use Update.

DynamicClient exposes Update method - there are places where this is used, e.g.:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/garbagecollector/operations.go#L76

This method should look exactly the same way as create and delete.

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// GetObject retrieves object with given name, group, version and kind.
+func GetObject(dynamicClient dynamic.Interface, namespace string, name string, gvk schema.GroupVersionKind) (*unstructured.Unstructured, error) {
+	var obj *unstructured.Unstructured
+	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
+	getFunc := func() (bool, error) {
+		var err error
+		obj, err = dynamicClient.Resource(gvr).Namespace(namespace).Get(name, metav1.GetOptions{
+			IncludeUninitialized: true,
```
wojtek-t: It's not clear to me that we want to return uninitialized objects. In fact, I think we shouldn't.

So let's remove this and add a TODO to investigate if we have a need to include uninitialized objects.

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,201 @@
...
+		return false, fmt.Errorf("Failed to create namespace with non-retryable error: %v", err)
+	}
+	return RetryWithExponentialBackOff(createFunc)
+
+}
+
+// DeleteNamespace deletes namespace with given name.
+func DeleteNamespace(c clientset.Interface, namespace string) error {
+	deleteFunc := func() (bool, error) {
+		err := c.CoreV1().Namespaces().Delete(namespace, nil)
```
wojtek-t: There is significant conde duplication between those methods here. I think we can refactor them to avoid this.
```
func retryableFunc() func() (bool, error) {
  err := baseFunc()
  if err == nil || isAllowedError(err) {
    return true, nil
  }
  if IsRetryableError(err) {
    return false, nil
  }
  return false, err
}
```

and then:
```
func CreateNamespace(c clientset.Interface, namespace string) error {
  createFunc := func() error {
    _, err := c.CoreV1().Namespaces().Create(...)
   return err
  }
  return RetryWithExponentialBackOff(retryableFunc(baseFunc, apierrs.IsAlreadyExists))
```

[Alternatively, you may consider changing the last line into:
err := Retryable...
if err != nil {
  return fmt.Errorf("failed to create namespace: %v", err)
}
return nil

But I don't think it's really needed.

clusterloader2/framework/framework.go
krzysied: I want to add TODO for multiple clients. Will be rephrased...

clusterloader2/framework/framework.go
krzysied: I'm removing namespaces count parameter. Automaneged namespaces count will be part of framework state.
Delete will remove all of automanaged namespaces. Addition Create is assumed to not be used without Delete called first.

clusterloader2/framework/framework.go
krzysied: done

clusterloader2/framework/util/util.go
krzysied: done

clusterloader2/framework/util/util.go
krzysied: TODO added.

clusterloader2/framework/util/util.go
krzysied: update reimplemented.

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,117 @@
...
+	var namespaces []string
+	for i := 0; i < namespaceCount; i++ {
+		name := fmt.Sprintf("%v-%d", automanagedNamespaceName, i)
+		if err := util.CreateNamespace(f.clientSet, name); err != nil {
+			return nil, err
+		}
+		namespaces = append(namespaces, name)
+		f.automanagedNamespaceCount++
+	}
+	return namespaces, nil
```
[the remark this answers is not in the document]
krzysied: For every object related operation there is namespace name required. I though that passing list of names is better solution than making automaneged namepsace names by convention (in other packages than framework).

Other solution could be making public function for generating automanaged namespaces name.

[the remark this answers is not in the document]
wojtek-t: Those will be coming with NamespaceRange - so it will be needed in exactly one other place - defaulting function for NamespaceRange. I don't think that returning names here gives us somethimg.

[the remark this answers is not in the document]
krzysied: Ok. I'm removing namespaces names returning.

I guess at some point in the future we will need to make automanagedNamespaceName const public then, but for now it is not required.

clusterloader2/framework/framework.go
```diff
@@ -0,0 +1,114 @@
...
+	dynamicClient             dynamic.Interface
+}
+
+// NewFramework creates new framework based on given kubeconfig.
+func NewFramework(path string) (*Framework, error) {
+	config, err := util.LoadConfig(path)
+	if err != nil {
+		return nil, fmt.Errorf("config load failed: %v", err)
+	}
+	config.ContentType = contentType
```
wojtek-t: We should remove it from here. What you should do is:
- make LoadConfig function private one
- add one more function (sth like PrepareConfig(path) or sth like that) that will:
(a) call loadConfig
(b) set all additional stuff that we want.

clusterloader2/framework/util/kubeconfig.go
```diff
@@ -0,0 +1,42 @@
+/*
```
wojtek-t: please rename to clientconfig.go

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,154 @@
+/*
```
wojtek-t: I'm not huge fan of "util.go" files, because they hugely become a mix of everything.

Those are functions that implement client operations on objects. So maybe:
client/objects.go
?

Or sth like that?

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,154 @@
...
+		return true
+	}
+	// If the error sends the Retry-After header, we respect it as an explicit confirmation we should retry.
+	if _, shouldRetry := apierrs.SuggestsClientDelay(err); shouldRetry {
+		return true
+	}
+	return false
+}
+
+func retryOperation(operation func() error, isAllowedError func(error) bool) wait.ConditionFunc {
```
wojtek-t: I don't like "Operation" name - these aren't operations per. Operation in generally is something long-running that does many asynchronous things. Here, you are simply retrying some functions.

retryFunction is how this should be called.

clusterloader2/framework/util/util.go
```diff
@@ -0,0 +1,154 @@
...
+func DeleteNamespace(c clientset.Interface, namespace string) error {
+	deleteFunc := func() error {
+		return c.CoreV1().Namespaces().Delete(namespace, nil)
+	}
+	return RetryWithExponentialBackOff(retryOperation(deleteFunc, apierrs.IsNotFound))
+}
+
+// CreateObject creates object based on given object description.
+func CreateObject(dynamicClient dynamic.Interface, gvk schema.GroupVersionKind, namespace string, name string, obj *unstructured.Unstructured) error {
+	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
```
wojtek-t: I don't like it - the whole purpose of introducing this ObjectType in the api was to avoid doing any conversions and tricks to determine params that you need to pass in dynamic client.
So if that's not exactly what you need, we should change the API to be exactly what you need.

That said, I took a bit of deeper look whether we actually need that. And I'm now pretty strongly convinced that we don't need this at all.
Instead, we should get this information from the object itself. This seems exactly like a function that you would like to copy. So please create a PR to remove the ObjectType type from api completely and start getting that information from there.

krzysied: TBH, I'm not sure if there is other way to do this (, assuming that we are using dynamic client). To do any operation on the cluster, first you need to call Resource(gvr), which is using resource instead of kind. Because only kind is provided by the object, conversion needs to be made.

wojtek-t: Once we start relying on objects for that, maybe that's needed - but that's fine.

I just didn't want to introduce a dedicate type, that doesn't really buy us anything, because we need to convert it twice anyway.

clusterloader2/framework/config/clientconfig.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error loading kubeconfig: %v", err)
+	}
+	return c, nil
+}
+
+func loadConfig(path string) (*restclient.Config, error) {
+	c, err := restclientConfig(path)
+	if err != nil {
+		return restclient.InClusterConfig()
```
wojtek-t: I don't understand why we may want InClusterConfig in case of error.

krzysied: At first I thought that if config is not created from file, maybe there is other way to create it.
I going to remove this part - if function loads config from certain path, definitely shouldn't try to create config using other ways.

krzysied: done. PTAL.

wojtek-t:
OK - let's have this in and we can improve that later.

/lgtm
/approve

PR #151, opened 2018-07-06
ClusterLoader - unmarshaling templates

krzysied:


clusterloader2/config/codec.go
```diff
@@ -18,13 +18,18 @@ package config
...
 	"path/filepath"
 	"strings"
 
 	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/perf-tests/clusterloader2/api"
 )
 
+var decode = scheme.Codecs.UniversalDeserializer().Decode
```
shyamjvs: It seems like you only need it at one place - maybe you can directly call original function there?

clusterloader2/config/codec.go
```diff
@@ -41,3 +46,16 @@ func ReadConfig(path string) (*api.Config, error) {
 	}
 	return &config, nil
 }
+
+// ReadTemplate creates object template from file specified by the given path.
+func ReadTemplate(path string) (runtime.Object, error) {
+	raw, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("reading template error: %v", err)
```
shyamjvs: -> "error reading template file: "

clusterloader2/config/codec.go
krzysied: ok

clusterloader2/config/codec.go
krzysied: Should this type of error be returned  by ReadTemplate method or by ReadConfig method caller (as a wrapper for the error returned by ReadConfig)?

shyamjvs: Based on the scope of this error I would believe the former. In the outer function, the scope should be wider imo (for e.g sth like "couldn't fetch template: ")

krzysied: done

clusterloader2/config/codec.go
```diff
@@ -41,3 +44,16 @@ func ReadConfig(path string) (*api.Config, error) {
...
 	return &config, nil
 }
+
+// ReadTemplate creates object template from file specified by the given path.
+func ReadTemplate(path string) (runtime.Object, error) {
+	raw, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("error reading template file: %v", err)
+	}
+	obj, _, err := scheme.Codecs.UniversalDeserializer().Decode(raw, nil, nil)
```
wojtek-t: Question: if we won't compile in any api types (or in particular the type of the object) - what would that return?
Unstructured? Unknown? error?

krzysied: If object is not recognized, method will return error. By not recognized I mean the meta data has wrong structure or <api group, api version, kind> tuple is unknown.
If object has valid meta data, but spec is incorrect, Decode will return Unstructured with fields (for kind specified in meta data) initialized with whatever matched with template. This could possible be the problem, but I expect it to be also validate on server side.

BTW, k8s.io/client-go/kubernetes/scheme already has some structures added to scheme.

krzysied: > If object is not recognized, method will return error. By not recognized I mean the (...) <api group, api version, kind> tuple is unknown.

This case will apply to your scenario. Unknown - unknown to scheme.

wojtek-t: So that is tricky. You should assume that no schema is linked into cluster loader.

So my question now is: is it possible to easily convert Unknown to Unstructured what you use in all other places.

krzysied: > Unknown - unknown to scheme

By "Unknown" I meant word not structure. I haven't find any case when Unknown is returned.

> is it possible to easily convert Unknown to Unstructured what you use in all other places.

I think I found a solution for this. I can Decode template into Unstructured: https://github.com/kubernetes/kubernetes/blob/bc6fac95023e13d9f4ab542b484d26b05d12cad8/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructured.go#L41 .
Decode will require kind and apiVersion fields. If they are not provide it will return error. 
Decode will not check if <apiVersion, kind> is registered. Validation will be performed by the server.

clusterloader2/pkg/config/codec.go
```diff
@@ -41,3 +44,17 @@ func ReadConfig(path string) (*api.Config, error) {
...
 	return &config, nil
 }
+
+// ReadTemplate creates object template from file specified by the given path.
+func ReadTemplate(path string) (*unstructured.Unstructured, error) {
+	raw, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("error reading template file: %v", err)
+	}
+	obj := &unstructured.Unstructured{}
```
wojtek-t: And that is exactly what I wanted to see.

Issue #137, opened 2018-06-22
Take perf-dash to v1.0


shyamjvs:
We've made several improvements to [perf-dash](http://perf-dash.k8s.io/) in the past few months (thanks a lot @krzysied for the work!). I think it's time to make some final improvements and cut the first major release (1.0). Mainly, let's make using it easier (even for non-scalability folks - for e.g release team has been asking for it). Some work-items that come to my mind:

- Split the graphs based on job-name first (i.e gce-100, kubemark-500, etc) and then metric (i.e ApiserverResponsiveness, SchedulerMetrics, etc) so that navigation is easier
- Add units to the axes in the graphs (otherwise the graphs aren't understandable)
- Load individual graphs on demand rather than pre-loading everything at start (so we don't wait for tens of seconds before we can see anything)
- sth else?

/assign @krzysied 
cc @kubernetes/sig-scalability-misc @wojtek-t

krzysied:
What a coincident... #138

krzysied:
> Add units to the axes in the graphs (otherwise the graphs aren't understandable)

I've looked into this. The units are implemented, but there is a typo in https://github.com/kubernetes/perf-tests/blob/7881ce52d23d547e813e2ac6b6aa8870525e11c3/perfdash/www/script.js#L81 
I will add fix after #138 is merged.

krzysied:
With #149 being icing on the cake, I think we are ready to make 1.0 release.
@shyamjvs  Is there anything that should be added before making it 1.0?

shyamjvs:
Yes, it looks great - waaay better than before. Thanks @krzysied!
Could you make one final change of adding "Runs over time" label to the x-axis in the graphs? We can bump it to 1.0 after that.

shyamjvs:
Closing this with the final change merged. Thanks!
/close

PR #152, opened 2018-07-06
Perfdash: adding x-axis label

krzysied:
ref #137

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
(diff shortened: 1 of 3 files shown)


perfdash/Makefile
```diff
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 0.26
+TAG = 0.27
```
shyamjvs: Can you please change this to 1.0?

krzysied: Sure - assuming that this is the last required change (for 1.0 version).

PR #153, opened 2018-07-09
ClusterLoader - main

krzysied:
Main file for cluster loader.


clusterloader2/clusterloader.go
```diff
@@ -0,0 +1,71 @@
+/*
```
shyamjvs: Please move this to cmd/clusterloader.go

krzysied: Code structure has been changed:
- clusterloader.go moved to cmd/
- all other directories (except api and vendor) moved to pkg/

clusterloader2/README.md
```diff
@@ -1,6 +1,19 @@
 # ClusterLoader
 
-#### Vendor
+### Running ClusterLoader
+
+To run ClusterLoader type:
+```
+go run cmd/clusterloader.go --clientconfig=kubeConfig.yaml --testconfig=config.json
```
shyamjvs: Please rename this flag to `--kubeconfig` (to match the convention everywhere else).

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,71 @@
...
+	"flag"
+	"fmt"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/test"
+)
+
+var (
+	clientConfigPath = flag.String("clientconfig", "", "path to kuberentes client config")
```
shyamjvs: please change description to "Path to kubeconfig file"

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,71 @@
...
+	testConfigPath   = flag.String("testconfig", "", "path to test config")
```
shyamjvs: -> "Path to test config file"

clusterloader2/pkg/framework/framework.go
```diff
@@ -23,8 +23,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	clientset "k8s.io/client-go/kubernetes"
-	"k8s.io/perf-tests/clusterloader2/framework/client"
-	"k8s.io/perf-tests/clusterloader2/framework/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/config"
+
+	// ensure auth plugins are loaded
```
shyamjvs: I don't think this comment adds any value :)

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func validateFlags() []error {
+	errList := make([]error, 0)
```
shyamjvs: Since you seem to already import `k8s.io/apimachinery/pkg/util/errors` pkg, could you please use `errors.Aggregate` in that pkg instead of manually dealing with the list and printing it?

clusterloader2/pkg/test/test.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+)
+
+// RunTest runs test based on provided test configuration.
+func RunTest(f *framework.Framework, conf *api.Config) error {
+	defer f.DeleteAutomanagedNamespaces()
+	err := f.CreateAutomanagedNamespaces(int(conf.AutomanagedNamespaces))
```
shyamjvs: Please check that "conf != nil" before this.

clusterloader2/pkg/test/test.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return fmt.Errorf("automanaged namespaces creation failed: %v", err)
```
shyamjvs: I prefer error msgs in active form instead of passive form to make it crisp :)

So here -> "Failed to create automanaged namespaces: %v"

clusterloader2/README.md
krzysied: ok.

clusterloader2/cmd/clusterloader.go
krzysied: Sure. However it looks like this Aggregate is useful only for printing. I don't think it supports append() and len().

clusterloader2/pkg/framework/framework.go
krzysied: Necessary because of "_ import" policy. All "_ import"s are required to have a comment.

clusterloader2/pkg/test/test.go
krzysied: ok

clusterloader2/pkg/test/test.go
krzysied: I prefer to let it be as it is. In all other files passive form is used.

clusterloader2/cmd/clusterloader.go
krzysied: ok

clusterloader2/cmd/clusterloader.go
krzysied: ok

clusterloader2/README.md
```diff
@@ -1,6 +1,19 @@
 # ClusterLoader
 
-#### Vendor
+### Running ClusterLoader
+
+To run ClusterLoader type:
+```
+go run cmd/clusterloader.go --kubeconfig=kubeConfig.yaml --testconfig=config.json
+```
+Flags clientconfig and testconfig are necessary.
```
shyamjvs: -> kubeconfig

shyamjvs: It seems like this hasn't been changed at few places (including main).

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/test"
+)
+
+var (
+	clientConfigPath = flag.String("clientconfig", "", "path to the kubeconfig file")
```
shyamjvs: -> kubeconfig

clusterloader2/pkg/test/test.go
```diff
@@ -0,0 +1,40 @@
...
+
+// RunTest runs test based on provided test configuration.
+func RunTest(f *framework.Framework, conf *api.Config) error {
+	if f == nil {
+		return fmt.Errorf("framework must be provided")
+	}
+	if conf == nil {
+		return fmt.Errorf("test config must be provided")
+	}
+	defer f.DeleteAutomanagedNamespaces()
```
shyamjvs: We shouldn't be calling a deferred delete call before even creating. Please move this to after the "err != nil" block (so we don't try deleting un-created namespaces).

clusterloader2/README.md
krzysied: My bad. This time should be done.

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/test"
+)
+
+var (
+	clientConfigPath = flag.String("kubeconfig", "", "path to the kubeconfig file")
```
shyamjvs: Please change to kubeConfigPath :)

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,70 @@
...
+	if *testConfigPath == "" {
+		errList = append(errList, fmt.Errorf("no test config path specified"))
+	}
+	return errList
+}
+
+func main() {
+	flag.Parse()
+	if errList := validateFlags(); len(errList) > 0 {
+		fmt.Printf("Parsing flags error: %v\n", errors.NewAggregate(errList).Error())
```
shyamjvs: Could you replace fmt.Printf("foobar\n") with fmt.Println("foobar")? (here and everywhere)

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	f, err := framework.NewFramework(*clientConfigPath)
+	if err != nil {
+		fmt.Printf("Framework creation error: %v\n", err)
+		return
+	}
+	c, err := config.ReadConfig(*testConfigPath)
+	if err != nil {
+		fmt.Printf("Config materializing error: %v\n", err)
```
shyamjvs: "materializing error" sounds not too clear. I'd suggest replacing with "Config reading error".

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,70 @@
...
+		return
+	}
+
+	err = test.RunTest(f, c)
+	if err != nil {
+		fmt.Printf("Test execution failed: %v\n", err)
+		return
+	}
+
+	fmt.Printf("Test run successfully!\n")
```
shyamjvs: s/run/ran/

clusterloader2/pkg/test/test.go
shyamjvs: But I still don't understand reason why you would ever want to call it even before creating? It doesn't make sense trying to delete N uncreated objects (just because it's not leading to any error) :)

shyamjvs: Maybe I wasn't clear earlier - I was suggesting moving this one line below instead of moving it after `if err != nil`

clusterloader2/cmd/clusterloader.go
krzysied: done

clusterloader2/cmd/clusterloader.go
krzysied: done

clusterloader2/cmd/clusterloader.go
krzysied: done

clusterloader2/cmd/clusterloader.go
krzysied: done

clusterloader2/pkg/test/test.go
krzysied: > It doesn't make sense trying to delete N uncreated objects (just because it's not leading to any error) :)

It tries to delete only created ones.

> But I still don't understand reason why you would ever want to call it even before creating?

In this case it doesn't matter. Normally it depends on whether you want to execute code even if panic was risen or not. (assuming that I understand defer meaning correctly)

> Maybe I wasn't clear earlier - I was suggesting moving this one line below instead of moving it after if statement

This can be done :)

shyamjvs: > It tries to delete only created ones.

Doesn't matter even if the function is a no-op. What I'm saying here is "why should we make an unnecessary function call?"

> In this case it doesn't matter. Normally it depends on whether you want to execute code even if panic was risen or not. (assuming that I understand defer meaning correctly)

I'm not 100% sure, but defer wouldn't even be called if there's a panic before it - the binary crashes in such cases (unless panic handlers are registered).

> This can be done :)

Let's do that :)

shyamjvs: > Let's do that :)

Ok, it seems like you did make the change.

shyamjvs:
It seems like gofmt is failing in the presubmit:

```
!!! 'gofmt -s' needs to be run on the following files: 
./clusterloader2/cmd/clusterloader.go
```

Will lgtm once that's fixed.

clusterloader2/cmd/clusterloader.go
```diff
@@ -0,0 +1,70 @@
...
+	if *testConfigPath == "" {
+		errList = append(errList, fmt.Errorf("no test config path specified"))
+	}
+	return errList
+}
+
+func main() {
+	flag.Parse()
+	if errList := validateFlags(); len(errList) > 0 {
+		fmt.Println("Parsing flags error: %v", errors.NewAggregate(errList).Error())
```
wojtek-t: Please dont use fmt.Println - for consistency we should eveywhere use glog.

glog.Fatalf(...)

Same everywhere else.

krzysied: Done. Println replaced with Fatalf and Info.

PR #156, opened 2018-07-11
Perfdash: config for build ranges

krzysied:


perfdash/Makefile
```diff
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 1.0
+TAG = 1.01
```
shyamjvs: Why 1.01 instead of 1.1?

krzysied:
/cc mborsz
/cc shyamjvs

I definitely need suggestions for customized build ranges. Default is 100.
Btw, builds.json can also be changed during runtime.

perfdash/config/builds.json
```diff
@@ -0,0 +1,3 @@
+{
+	"gce-100Nodes": 100
```
shyamjvs: Like I mentioned earlier - let's not try to hardcode this config into perfdash (as then we'll need to rebuild and push everytime we want to change this).

Instead, let's pick this config from https://github.com/kubernetes/test-infra/blob/master/prow/config.yaml (from where we're already picking perfDashTag). You can add a new label there like `perfDashRange` or sth - and we can easily change this range then by just a single PR changing that value.

shyamjvs:
I've left above a high-level comment about this approach.

perfdash/Makefile
krzysied: For some reason I thought that we are always using two digits for minor version. I will change it to 1.1 if I assumed incorrectly.

perfdash/config/builds.json
krzysied: Good idea. I will change my code.

perfdash/Makefile
shyamjvs: It's usually just a serially increasing number for the minor version. (btw - naming it as 1.01 wouldn't let us have more than 99 minor versions in any decent way :)

perfdash/config/builds.json
krzysied: Done. `perfDashBuilds` is a new label for specifying builds.

perfdash/config.go
```diff
@@ -42,6 +43,7 @@ type TestDescription struct {
 type Tests struct {
 	Prefix       string
 	Descriptions map[string]TestDescription
+	Builds       int
```
shyamjvs: Maybe `NumBuilds` or `BuildsCount`?

perfdash/config.go
```diff
@@ -188,6 +190,19 @@ func getProwConfig() (Jobs, error) {
 				thisPeriodicConfig.Descriptions = jobTypeToDescriptions[jobType]
 				continue
 			}
+			if strings.HasPrefix(tag, "perfDashBuilds:") {
+				split := strings.SplitN(tag, ":", 2)
+				i, err := strconv.Atoi(strings.TrimSpace(split[1]))
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "warning: unparsable builds - %v\n", split[1])
+				}
+				if i < 1 {
```
shyamjvs: Shouldn't this be 0 (if you're excluding negative values)?

shyamjvs: Or change the error msg below.

perfdash/config.go
krzysied: Ok. Changing to `BuildsCount`.

perfdash/config.go
krzysied: You are right. Should be non-negative.

perfdash/config.go
```diff
@@ -188,6 +190,19 @@ func getProwConfig() (Jobs, error) {
 				thisPeriodicConfig.Descriptions = jobTypeToDescriptions[jobType]
 				continue
 			}
+			if strings.HasPrefix(tag, "perfDashBuilds:") {
```
shyamjvs: Also here perfDashBuildsCount

krzysied: done

PR #157, opened 2018-07-12
ClusterLoader - simple test implementation

krzysied:
Implementation for interfaces provided by `test/`.
Test execution logic will be implemented in the next PR.

### DIFF
--- a/clusterloader2/pkg/test/simpletestexecutor.go
+++ b/clusterloader2/pkg/test/simpletestexecutor.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+type simpleTestExecutor struct{}
+
+func createSimpleTestExecutor() TestExecutor {
+	return &simpleTestExecutor{}
+}
+
+// ExecuteTest executes test based on provided configuration.
+func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) error {
+	return nil
+}
+
+// ExecuteStep executes single test step based on provided step configuration.
+func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) error {
+	return nil
+}
+
+// ExecutePhase executes single test phase based on provided phase configuration.
+func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) error {
+	return nil
+}
+
+// ExecuteObject executes single test object operation based on provided object configuration.
+func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object) error {
+	return nil
+}
--- a/clusterloader2/pkg/test/test.go
+++ b/clusterloader2/pkg/test/test.go
@@ -27,11 +27,11 @@ import (
 var (
 	// CreateContext global function for creating context.
 	// This function should be set by Context implementation.
-	CreateContext CreatContextFunc
+	CreateContext = createSimpleContext
 
 	// Test is a singleton for test execution object.
 	// This object should be set by TestExecutor implementation.
-	Test TestExecutor
+	Test = createSimpleTestExecutor()
 )
 
 // RunTest runs test based on provided test configuration.
--- a/clusterloader2/pkg/test/simplecontext.go
+++ b/clusterloader2/pkg/test/simplecontext.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/state"
+)
+
+type simpleContext struct {
+	framework *framework.Framework
+	state     *state.NamespacesState
+}
+
+func createSimpleContext(f *framework.Framework, s *state.NamespacesState) Context {
+	return &simpleContext{
+		framework: f,
+		state:     s,
+	}
+}
+
+// GetFramework returns framework.
+func (sc *simpleContext) GetFramework() *framework.Framework {
+	return sc.framework
+}
+
+// GetState returns current test state.
+func (sc *simpleContext) GetState() *state.NamespacesState {
+	return sc.state
+}


clusterloader2/pkg/test/teststruct.go
```diff
@@ -0,0 +1,62 @@
...
+
+type testStruct struct {
+	framework *framework.Framework
+	state     *state.NamespacesState
+	done      <-chan struct{}
+	namespace string
+	prefix    string
+}
+
+func (ts *testStruct) setNamespace(ns string) *testStruct {
```
shyamjvs: I think you can remove this set functions. They're just called at one place.

clusterloader2/pkg/test/teststruct.go
```diff
@@ -0,0 +1,62 @@
...
+package test
+
+import (
+	"fmt"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/state"
+)
+
+type testStruct struct {
```
shyamjvs: We should give a better name to this. There are so many structs in the clusterloader code that are related to 'test' :)

clusterloader2/pkg/test/teststruct.go
krzysied: function removed.

clusterloader2/pkg/test/teststruct.go
krzysied: This should be named `context.go`, cause this structure is a context :)

clusterloader2/pkg/test/context.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/state"
+)
+
+// Context is a global test context.
+type Context interface {
```
shyamjvs: Cool.
Maybe rename to TestContext (also a name I remember from k8s code)?

clusterloader2/pkg/test/context.go
```diff
@@ -0,0 +1,135 @@
...
+	GetFramework() *framework.Framework
+	GetPath() string
```
shyamjvs: Get path to what? Name isn't clear.

clusterloader2/pkg/test/context.go
```diff
@@ -0,0 +1,135 @@
...
+// StepContext is a step level context.
+type StepContext interface {
+	GetPath() string
+	Phase(done <-chan struct{}, i int) PhaseContext
+}
+
+// PhaseContext is a phase level context.
+type PhaseContext interface {
+	GetPath() string
+	GetDone() <-chan struct{}
```
shyamjvs: This is also not much clear.

clusterloader2/pkg/test/context.go
```diff
@@ -0,0 +1,135 @@
+/*
```
shyamjvs: Can you split this file into two:

- context.go (keep all interface definitions here)
- simplecontext.go (move your implementation of the interface here)

So that it's more easier to understand/maintain.

clusterloader2/pkg/test/context.go
```diff
@@ -0,0 +1,135 @@
...
+// Context is a global test context.
+type Context interface {
+	GetFramework() *framework.Framework
+	GetPath() string
+	Step(i int) StepContext
+}
+
+// StepContext is a step level context.
+type StepContext interface {
+	GetPath() string
```
shyamjvs: Why do you need path for step?

clusterloader2/pkg/test/context.go
```diff
@@ -0,0 +1,51 @@
...
+
+// StepContext is a step level context.
+type StepContext interface {
+	GetExecutionPath() string
+	Phase(done <-chan struct{}, i int) PhaseContext
+}
+
+// PhaseContext is a phase level context.
+type PhaseContext interface {
+	GetExecutionPath() string
```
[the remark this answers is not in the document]
krzysied: I'm logging when program entered phase/step. Maybe it will be removed in final version, but I definitely need this during development.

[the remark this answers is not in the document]
shyamjvs: Yes, but what's the reason for needing this to be part of interface (it seems to be only some implementation detail)?
When you're defining an interface, you should think about the minimal and necessary methods to solve the problem at an abstract level.

clusterloader2/pkg/test/context.go
shyamjvs: See my response to https://github.com/kubernetes/perf-tests/pull/157#discussion_r202322149. You shouldn't pollute an interface with implementation-specific helpers, etc..

clusterloader2/pkg/test/context.go
shyamjvs: Yes, it's a common practice to have such a channel, but:

- For what are you using the done channel here?
- `GetDone` is giving a meaning to the method sth like 'get something done'

shyamjvs:
@krzysied Yesterday you were saying that this PR couldn't be split into further smaller parts. Now when I see it, seems like it can be - one part for adding TestContext and another implementing the test logic.

clusterloader2/pkg/test/context.go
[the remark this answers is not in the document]
krzysied: My assumption for this context is that it is "aware" of execution path. That's why there are interfaces for every config level (global, step, phase, object). That's why Path, Step and Object methods always take index (i) as an argument.
Assuming that there will be proper logging added to cluster loader someday, `GetExecutionPath` seems necessary.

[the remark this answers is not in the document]
wojtek-t: I don't think that's a good assumption. Execution path is debug details and shouldn't be part of API.
In order to run a given phase, you only need it's config.
[I can be persuaded that test state is another thing, though in theory you can grab the whole state from apiserver.]

Definitely a path is not something that is needed.

clusterloader2/pkg/test/test.go
```diff
@@ -31,10 +34,75 @@ func RunTest(f *framework.Framework, conf *api.Config) error {
...
 		return fmt.Errorf("automanaged namespaces creation failed: %v", err)
 	}
+
+	for i := range conf.Steps {
+		executeStep(ctx.Step(i), &conf.Steps[i])
+	}
 	return nil
 }
+
+func executeStep(ctx StepContext, step *api.Step) {
```
shyamjvs: Ideally, I'd make `Execute()` a method on the `Step` interface, because:

- this would make it possible to have different implementations of running a step (for e.g if someone wants to do something differently like "run only N phases of the step in parallel at a time")
- this would help you eliminate need for  ad-hoc methods like `GetExecutionStep()` that you've currently added to interface just for your implementation

shyamjvs:
So with https://github.com/kubernetes/perf-tests/pull/157 merged, I believe we can now modify this PR to make use of those interfaces properly.

krzysied:
@shyamjvs That's right. I've changed the PR so structures satisfy test interfaces.

clusterloader2/pkg/test/impl/install.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"k8s.io/perf-tests/clusterloader2/pkg/test"
+)
+
+func init() {
+	Install()
+}
+
+// Install assigns global context creation function and test executor.
+func Install() {
```
shyamjvs: If it's just a couple of lines, I'd suggest moving this directly into init() - to save a function call.

clusterloader2/pkg/test/impl/simplecontext.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func createSimpleContext(f *framework.Framework, s *state.NamespacesState) test.Context {
+	return &simpleContext{
+		framework: f,
+		state:     s,
+	}
+}
+
+// GetFramework returns framework client.
```
shyamjvs: 'client'? It's just framework right?

clusterloader2/pkg/test/impl/simplecontext.go
```diff
@@ -0,0 +1,45 @@
...
+func (sc *simpleContext) GetFramework() *framework.Framework {
+	return sc.framework
+}
+
+// GetState returns current cluster state.
```
shyamjvs: It's not cluster state - I'd call it test state.

clusterloader2/pkg/test/impl/simpletestexecutor.go
```diff
@@ -0,0 +1,48 @@
...
+package impl
+
+import (
+	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/pkg/test"
+)
+
+type simpleTestExecutor struct{}
+
+func createSimpleTestEcexutor() test.TestExecutor {
```
shyamjvs: fix typo

clusterloader2/pkg/test/test.go
```diff
@@ -27,11 +27,11 @@ import (
 var (
 	// CreateContext global function for creating context.
 	// This function should be set by Context implementation.
-	CreateContext CreatContextFunc
+	CreateContext = createSimpleContext
```
shyamjvs: If you're anyway hard-coding this value, is there a point in offering such extension?

krzysied: This is a default value. If other implementation is needed, It can always be replaced in runtime (e.g. using `init()` function from other implementation package).

shyamjvs: Ok, that sounds reasonable to me.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -0,0 +1,47 @@
...
+limitations under the License.
+*/
+
+package test
+
+import (
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+type simpleTestExecutor struct{}
```
wojtek-t: Should be called fakeTestExecutor (not simple)

@shyamjvs @krzysied

shyamjvs: I don't think so. IIUC this isn't a fake implementation, it was just left to be implemented later.
Anyway, this has been changed now I think.

shyamjvs: Krzysiek to confirm.

krzysied: In current version most of the test execution logic is implemented.
A lot of TODOs missing here, so it can be misleading.

PR #160, opened 2018-07-17
Perfdash: histogram

krzysied:


krzysied:
/hold
Possible changes.

perfdash/parser.go
```diff
@@ -228,3 +228,46 @@ func pareseSchedulingThroughput(data []byte, buildNumber int, testResult *BuildD
...
+	Buckets map[string]int    `json:"buckets"`
+}
+
+type HistogramVec []Histogram
+
+// This structure differs from the one from metric_util.go
+// Instead of explicit fields, we have a map.
+type EtcdMetrics map[string]HistogramVec
+
+func parseEtcdMetric(metricName string) func(data []byte, buildNumber int, testResult *BuildData) {
```
shyamjvs: Can you change this function so that we will be able to use this in general for all histograms (not just etcd)?

krzysied:
Here is an example of how the metrics will look like:
image
This represents the percentage of results being below the label value.

I think about inverting the graph:
image
This displays percentage of result above the label value. This way, graph behavior correlates with values changes (in previous example there is a counter-correlation).

Maybe there is more readable way for displaying this data. Any suggestions?

cc @shyamjvs

krzysied:
/hold cancel
I'm going to use original version (not inverted one).

shyamjvs:
/lgtm

Thanks.

PR #161, opened 2018-07-18
Perfdash: category level

krzysied:
Adding category level selector.

### DIFF
--- a/perfdash/www/index.html
+++ b/perfdash/www/index.html
@@ -11,10 +11,10 @@
         <!-- TODO: Need a better logo here -->
         <a href="https://k8s.io"><img src="k8s_logo.png" style="width:84; height: 84" /></a>
       </md-toolbar>
-      <div id="content" class="md-whiteframe-z2" style="padding: 20px">
+      <div id="content" class="md-whiteframe-z2" style="padding: 20px;">
         <div style="width: 200px; display: inline-block;">
           <md-input-container>
-            <md-select placeholder="JobName" ng-model="controller.jobName"  ng-change="controller.jobNameChanged()">
+            <md-select ng-model="controller.jobName" ng-change="controller.jobNameChanged()">
               <md-option ng-value="jobName" ng-repeat="jobName in controller.jobNames">
                 {{jobName}}
               </md-option>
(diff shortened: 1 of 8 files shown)


perfdash/www/index.html
```diff
@@ -11,17 +11,26 @@ <h1 class="md-display-2">Performance Dashboard</h1>
...
+            <md-select ng-model="controller.jobName" ng-change="controller.jobNameChanged()">
               <md-option ng-value="jobName" ng-repeat="jobName in controller.jobNames">
                 {{jobName}}
               </md-option>
             </md-select>
           </md-input-container>
         </div>
 	<div style="width: 200px; display: inline-block;">
+          <md-input-container>
+            <md-select placeholder="ComponentName" ng-model="controller.componentName"  ng-change="controller.componentNameChanged()">
```
shyamjvs: Could you rename this to be `MetricCategory` instead of `ComponentName` to be more generic?

krzysied: I will do this. However, I think `TestCategoryName` will be a better name - we are having `TestName` as the lowest level selector.

shyamjvs: I'm not sure because this is not the category for the test, but category for the metric.

shyamjvs: Sure, that makes sense.

perfdash/www/index.html
```diff
@@ -23,14 +23,23 @@ <h1 class="md-display-2">Performance Dashboard</h1>
         </div>
 	<div style="width: 200px; display: inline-block;">
           <md-input-container>
-            <md-select placeholder="TestName" ng-model="controller.testName"  ng-change="controller.testNameChanged()">
-              <md-option ng-value="testName" ng-repeat="testName in controller.testNames">
-                {{testName}}
+            <md-select placeholder="CategoryName" ng-model="controller.categoryName"  ng-change="controller.categoryNameChanged()">
```
shyamjvs: Sorry, I guess it wasn't clear earlier - I meant `MetricCategory` for this one (`CategoryName` doesn't say much by itself). So basically, JobName > MetricCategory > MetricName.

krzysied: done.

PR #164, opened 2018-07-18
ClusterLoader - Adding test, step, phase execution implementations

krzysied:
Simple implementations for ExecuteTest, ExecuteStep, ExecutePhase.

### DIFF
--- a/clusterloader2/pkg/test/interface.go
+++ b/clusterloader2/pkg/test/interface.go
@@ -34,8 +34,8 @@ type Context interface {
 
 // TestExecutor is an interface for test executing object.
 type TestExecutor interface {
-	ExecuteTest(ctx Context, conf *api.Config) error
-	ExecuteStep(ctx Context, step *api.Step) error
-	ExecutePhase(ctx Context, phase *api.Phase) error
-	ExecuteObject(ctx Context, object *api.Object) error
+	ExecuteTest(ctx Context, conf *api.Config) []error
+	ExecuteStep(ctx Context, step *api.Step) []error
+	ExecutePhase(ctx Context, phase *api.Phase) []error
+	ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32) []error
 }
--- a/clusterloader2/pkg/test/simpletestexecutor.go
+++ b/clusterloader2/pkg/test/simpletestexecutor.go
@@ -17,7 +17,15 @@ limitations under the License.
 package test
 
 import (
+	"fmt"
+	"sync"
+
+	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/state"
 )
 
 type simpleTestExecutor struct{}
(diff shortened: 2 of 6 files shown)


clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +33,72 @@ func createSimpleTestExecutor() TestExecutor {
 
 // ExecuteTest executes test based on provided configuration.
 func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) error {
+	err := ctx.GetFramework().CreateAutomanagedNamespaces(int(conf.AutomanagedNamespaces))
+	defer cleanupResources(ctx)
```
shyamjvs: Could you push this line to the top of function?

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +33,72 @@ func createSimpleTestExecutor() TestExecutor {
...
+	}
 	return nil
 }
 
 // ExecuteStep executes single test step based on provided step configuration.
 func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) error {
+	if len(step.Measurements) > 0 {
+		// TODO: handle measurments
+	} else {
+		var wg wait.Group
```
shyamjvs: Nice! I was having a comment in mind for your earlier PR to use waitgroup :)

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +33,72 @@ func createSimpleTestExecutor() TestExecutor {
...
+		defer wg.Wait()
+		for i := range step.Phases {
+			phase := &step.Phases[i]
+			wg.Start(func() {
+				ste.ExecutePhase(ctx, phase)
```
shyamjvs: I guess you can directly pass `&step.Phases[i]` here.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +33,72 @@ func createSimpleTestExecutor() TestExecutor {
...
+		}
+	}
 	return nil
 }
 
 // ExecutePhase executes single test phase based on provided phase configuration.
 func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) error {
+	if phase.NamespaceRange == nil {
+		for j := range phase.ObjectBundle {
+			innerExecuteObject(ctx, &phase.ObjectBundle[j], phase.ReplicasPerNamespace, "")
```
shyamjvs: You should somehow capture the error returned from `innerExecuteObject1 - else there's no point as you'll always return with nil.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +33,72 @@ func createSimpleTestExecutor() TestExecutor {
...
 // ExecuteStep executes single test step based on provided step configuration.
 func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) error {
+	if len(step.Measurements) > 0 {
+		// TODO: handle measurments
+	} else {
+		var wg wait.Group
+		defer wg.Wait()
+		for i := range step.Phases {
+			phase := &step.Phases[i]
+			wg.Start(func() {
```
shyamjvs: Same as below - you should somehow try to capture the error from executing individual phases.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +33,72 @@ func createSimpleTestExecutor() TestExecutor {
...
+			for j := range phase.ObjectBundle {
+				innerExecuteObject(ctx, &phase.ObjectBundle[j], phase.ReplicasPerNamespace, nsName)
+			}
+		}
+	}
 	return nil
 }
 
 // ExecuteObject executes single test object operation based on provided object configuration.
 func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object) error {
```
shyamjvs: Could you formulate (in normal text) what is it that you expect `ExecuteObject` to do? It'll help you implement this.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: Can't be done.
Passing `&step.Phases[i]` will make `i` to be called by inner function. `i` changes with every loop iteration, so `&step.Phases[i]` can have different values based on time which the call was made.

I can change it to `&step.Phases[j]`  and add `j :=i` instead of `phase := &step.Phases[i]`. Somewhere copy of `i` have to made so it will became "constant" for the inner function.

shyamjvs: You're right - I missed that sorry. Let's keep it as it is then.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: TBH, I want to implemented this part in separate PR and focus only on test, step and phase execution.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: In this implementation I though about always passing `nil`. Errors will be logged and not escalated.
On the other hand I need something for returning test states - success or fail.
Let me rethink this.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: done

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: I've added error propagation. I assume that this propagates only errors which are critical (execution should be stopped).

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: I've merge `ExecuteObject` and `innerExecuteObject` into one function. Interface has been changed.

clusterloader2/pkg/test/interface.go
```diff
@@ -37,5 +37,5 @@ type TestExecutor interface {
 	ExecuteTest(ctx Context, conf *api.Config) error
 	ExecuteStep(ctx Context, step *api.Step) error
 	ExecutePhase(ctx Context, phase *api.Phase) error
-	ExecuteObject(ctx Context, object *api.Object) error
+	ExecuteObject(ctx Context, namespace string, replicaCount int32, object *api.Object) error
```
shyamjvs: Could you move these two parameters to the right? (just to make the first 2 args consistent across all these functions)

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +34,80 @@ func createSimpleTestExecutor() TestExecutor {
...
+		var wg wait.Group
+		var lock sync.Mutex
+		for i := range step.Phases {
+			phase := &step.Phases[i]
+			wg.Start(func() {
+				if phaseErr := ste.ExecutePhase(ctx, phase); phaseErr != nil {
+					lock.Lock()
+					defer lock.Unlock()
+					if err == nil {
+						err = phaseErr
```
shyamjvs: Any reason why you're capturing only the first error and not all?

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +34,80 @@ func createSimpleTestExecutor() TestExecutor {
...
+		wg.Wait()
+	}
+	return err
 }
 
 // ExecutePhase executes single test phase based on provided phase configuration.
 func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) error {
+	if phase.NamespaceRange == nil {
+		for j := range phase.ObjectBundle {
+			ste.ExecuteObject(ctx, "", phase.ReplicasPerNamespace, &phase.ObjectBundle[j])
```
shyamjvs: TBH - I'm not sure what semantics ReplicasPerNamespace implies for non-namespaced objects. This is not a blocker for us (as I don't think we create any non-namespaced objects in our tests) - but sth we need to think about.

cc @wojtek-t

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,20 +34,80 @@ func createSimpleTestExecutor() TestExecutor {
...
+	} else {
+		nsBasename := framework.AutomanagedNamespaceName
+		if phase.NamespaceRange.Basename != nil {
+			nsBasename = *phase.NamespaceRange.Basename
+		}
+		for i := phase.NamespaceRange.Min; i <= phase.NamespaceRange.Max; i++ {
+			nsName := fmt.Sprintf("%v-%d", nsBasename, i)
+			for j := range phase.ObjectBundle {
+				if err := ste.ExecuteObject(ctx, nsName, phase.ReplicasPerNamespace, &phase.ObjectBundle[j]); err != nil {
+					return err
```
shyamjvs: Let's not fail fast (like we discussed earlier why we should try to avoid that in our scalability tests). Just record the error and skip creating any further objects in this particular bundle (we can still continue for other namespaces).

In fact, you can even continue at a lower granularity (i.e for other replicas of the objectBundle within the same namespace). And for that you'll need to change the `ExecuteObject` function to take a single replicaIndex and create only that one instance (instead of creating `replicaCount` many instances). Wdyt?

clusterloader2/pkg/test/interface.go
krzysied: OK.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: I have two assumptions:
* All errors will be logged.
* Only critical error will be returned from the function. After critical error, test should be terminated.

So, in case `ExecuteObject` fails, it will log error and return `nil`. This way test will not fail fast unless it is needed.

krzysied: On the other hand I can imagine other implementation that returns `[]error` from each `Execute...` function.
TBH, I'm not sure which one is better - both have pros and cons.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: All error will be logged, so no information will be missed.
In case of critical error the first one to occur should be the original error.

clusterloader2/pkg/test/simpletestexecutor.go
shyamjvs: What's your criteria to decide if something is a critical error?

clusterloader2/pkg/test/simpletestexecutor.go
shyamjvs: Could you change this to an errList and return that instead for completeness?
That'll be useful to generate an error report at the end of the test. For simplicity, you can merge all errors into a single one at the top-level (i.e in `ExecuteTest()`) for now.

krzysied: After thinking more about this, I decided to make every `Execute...` function return `[]error`.
Every object operation can fail, so when `n` replicas are created, there could be `n` errors return. Thus we need `[]error` being propagated from the bottom to the top.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: At this point I'm not sure what should be the criteria.
There is a function that decides whether given err list is critical (with TODO status).

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -27,21 +35,142 @@ func createSimpleTestExecutor() TestExecutor {
...
+	return errList
 }
 
 // ExecutePhase executes single test phase based on provided phase configuration.
-func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) error {
+func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []error {
+	var errList []error
+	nsList := createNamespacesList(phase.NamespaceRange)
+	for _, nsName := range nsList {
+		for j := len(phase.ObjectBundle) - 1; j >= 0; j-- {
```
shyamjvs: Why are you traversing the objectbundle in reverse direction?

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -27,21 +35,142 @@ func createSimpleTestExecutor() TestExecutor {
...
+			}
+			instances, exists := ctx.GetState().Get(nsName, id)
+			if !exists {
+				instances = &state.InstancesState{
+					ReplicaCount: 0,
+					Object:       phase.ObjectBundle[j],
+				}
+			}
+			currentInstancesCount := instances.ReplicaCount
+			instances.ReplicaCount = phase.ReplicasPerNamespace
```
shyamjvs: IIUC this logic is not correct because you're using instances.ReplicaCount as the indicator for 'existing no. of replicas of the object' (let's say 10) and then setting it to new value directly in the next line to the desired future state (let's say 100). So there can be problems like:

- If someone else accesses the state in the meanwhile, it's possible that they end up thinking the current no. of replicas is 100 instead of 10
- If going ahead to create the additional 90 replicas, you fail to create some - then you would end up in an inconsistent state (i.e reality != recorded state)

So IMO, you should increment it by 1 (using a lock) whenever you successfully create a replica. Sth like this:

```
err := CreateReplica()
if err == nil {
  lock.Lock()
  instances.ReplicaCount++
  lock.Unlock()
}
```

shyamjvs: Btw - Even with the above I think it's still possible to diverge recorded state from real state over time. But I think it's much safer than the above. We may want to fully solve the problem by listing objects at the start of this function to reconcile with correct state - but I'm fine neglecting that for now with a TODO.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -27,21 +35,142 @@ func createSimpleTestExecutor() TestExecutor {
...
+			ctx.GetState().Set(nsName, id, instances)
+			// Handling for remove objects
+			for replicaIndex := phase.ReplicasPerNamespace; replicaIndex < currentInstancesCount; replicaIndex++ {
+				if objectErrList := ste.ExecuteObject(ctx, &phase.ObjectBundle[j], nsName, replicaIndex); len(objectErrList) > 0 {
+					errList = append(errList, objectErrList...)
+					if isErrsCritical(objectErrList) {
+						return errList
+					}
+				}
```
shyamjvs: If indeed the case is of 'removing objects' why don't you just break from here? I think it's unnecessary to go to the update/create cases.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: I assumed that if objects are created in some specific order, then maybe they should be deleted in a reversed order. Are there problems like A can't be removed because it's used by B?

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: Let's assume that we have cluster with 20 A and 0 B objects. We want to have 10 x [A, B]. In this scenario we want to delete 10 A and create 10 B. As you can see both types of operation are used in one step.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: > If someone else accesses the state in the meanwhile, it's possible that they end up thinking the current no. of replicas is 100 instead of 10

There is an assumption that only one operation can be perform on certain object class (\<namespcae x basename x kind x group\>), so this kind of situation shouldn't ever happen.

> If going ahead to create the additional 90 replicas, you fail to create some - then you would end up in an inconsistent state (i.e reality != recorded state)

In this implementation, state = desired state not real cluster state. At some point we will use discovering cluster state, but not now.

clusterloader2/pkg/test/simpletestexecutor.go
shyamjvs: Yeah, that's true. However, what I meant above is that for a given object in objectbundle, you should need to do exactly one of the operations delete/create/update iiuc (and doesn't seem to be the case right now). So could you change code as follows:

```
for object in objectbundle {
 if desiredCount > currentCount {<code for create>}
 if desiredCount < currentCount {<code for delete>}
 if desiredCount == currentCount {<code for update>}
}
```

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -27,21 +35,164 @@ func createSimpleTestExecutor() TestExecutor {
...
+			}
+			instances.DesiredReplicaCount = phase.ReplicasPerNamespace
+			ctx.GetState().Set(nsName, id, instances)
+			instancesStates = append(instancesStates, instances)
+		}
+
+		var maxCurrentReplicaCount int32
+		for j := range instancesStates {
+			if instancesStates[j].CurrentReplicaCount > maxCurrentReplicaCount {
+				maxCurrentReplicaCount = instancesStates[j].CurrentReplicaCount
```
shyamjvs: Could you explain what you're trying to do here?

krzysied: This the handling for the situation where we have uneven number of object instances:
Let's assume that in cluster we have [A,B,C]x10 and  [A,B]x10. Then we got phase with [A,B,C] and replicasPerNamespace = 0. In this case we want to delete every bundle using [C -> B -> A] order. Because number of instances is different for different objects, we have to calculate max(|{A}|,|{B}|, |{C}| ) first.

shyamjvs: Ok, now I understand. I think that sounds reasonable.

Could you add a brief comment above each code block in this function so it's easier to understand?

krzysied: done.

PR #167, opened 2018-07-23
ClusterLoader - Adding read template with cache

krzysied:


clusterloader2/pkg/config/codec.go
```diff
@@ -58,3 +64,25 @@ func ReadTemplate(path string) (*unstructured.Unstructured, error) {
...
+// or uses cached object if available.
+func ReadTemplateWithCache(path string) (*unstructured.Unstructured, error) {
+	lock.RLock()
+	obj, exists := templateCache[path]
+	lock.RUnlock()
+	if !exists {
+		lock.Lock()
+		defer lock.Unlock()
+		// Recheck codititon.
+		obj, exists = templateCache[path]
```
shyamjvs: Why you're rechecking condition?

clusterloader2/pkg/config/codec.go
```diff
@@ -58,3 +64,25 @@ func ReadTemplate(path string) (*unstructured.Unstructured, error) {
 	}
 	return obj, nil
 }
+
+// ReadTemplateWithCache creates object template from file specified by the given path
+// or uses cached object if available.
+func ReadTemplateWithCache(path string) (*unstructured.Unstructured, error) {
+	lock.RLock()
```
shyamjvs: Guess you can just lock and defer unlock once at the top of the function to make is simple (as you're accessing multiple times).

clusterloader2/pkg/config/codec.go
krzysied: Common practice - it is assume that you want to run initialization only once.
It can happen that object is not in cache so thread waits for lock to initialize object. During this waiting time other thread can initialize object. So when first thread acquires lock condition has changed and it's not needed to perform all initialization again. That's why you want to almost always recheck conditions after entering the critical section.

clusterloader2/pkg/config/codec.go
krzysied: RLock is for read-only and can be acquire be multiple threads at the same time - will be used very often.
Lock blocks all the access to object - will be used only for initialization.

I want to separate this two cases and get better performance due to RLock being used.

clusterloader2/pkg/config/codec.go
```diff
@@ -21,13 +21,19 @@ import (
...
+	"sync"
 
 	"github.com/spf13/viper"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/perf-tests/clusterloader2/api"
 )
 
+var (
+	lock          sync.RWMutex
```
shyamjvs: Instead of defining global variables that are shared across the package, could you please move this into a separate TemplateCache struct? That way it'll also protect the purpose of the lock.

clusterloader2/pkg/config/codec.go
shyamjvs: Yes, that sounds quite reasonable with read-heavy load, which is the case here.

clusterloader2/pkg/config/codec.go
shyamjvs: Yes, that makes sense.

clusterloader2/pkg/config/codec.go
krzysied: By "move this into a separate TemplateCache struct" do you mean creating separate struct to have it encapsulated and still have singleton of this struct on global level?
Or do you want this struct to have something like create function so we can create different instances of cache struct?

shyamjvs: The latter.

krzysied: OK. So the consequence of this will be having cache stored in the test context - cache needs to be passed between different calls.

shyamjvs: Sure, that's better than having one globally usable instance of the package. Alternatively, if you want to avoid passing it across calls, you can instantiate a global variable of this struct type in that package.

krzysied: What about `TemplateProvider`? Because it's not only cache but also something that can read template from file system.

shyamjvs: I don't have any strong preference here :)

krzysied: OK. Then `TemplateProvider` it is.

PR #170, opened 2018-07-24
ClusterLoader - Adding execute object logic

krzysied:
Adding handling for create and delete operations. Update operation is not a part of this PR.

### DIFF
--- a/clusterloader2/pkg/test/interface.go
+++ b/clusterloader2/pkg/test/interface.go
@@ -25,6 +25,18 @@ import (
 // CreatContextFunc a type for function that creates Context based on given framework client and state.
 type CreatContextFunc func(f *framework.Framework, s *state.NamespacesState) Context
 
+// OperationType is a type of operation to be performed on an object.
+type OperationType int
+
+const (
+	// CREATE_OBJECT is create object operation.
+	CREATE_OBJECT = OperationType(0)
+	// UPDATE_OBJECT is update object operation.
+	UPDATE_OBJECT = OperationType(1)
+	// DELETE_OBJECT is delete object operation.
+	DELETE_OBJECT = OperationType(2)
+)
+
 // Context is an interface for test context.
 // Test context provides framework client and cluster state.
 type Context interface {
@@ -37,5 +49,5 @@ type TestExecutor interface {
 	ExecuteTest(ctx Context, conf *api.Config) []error
 	ExecuteStep(ctx Context, step *api.Step) []error
 	ExecutePhase(ctx Context, phase *api.Phase) []error
-	ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32) []error
+	ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32, operation OperationType) []error
 }
--- a/clusterloader2/pkg/test/simpletestexecutor.go
+++ b/clusterloader2/pkg/test/simpletestexecutor.go
@@ -80,6 +80,7 @@ func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) []error 
 
 // ExecutePhase executes single test phase based on provided phase configuration.
 func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []error {
+	// TODO: add tuning set
 	var errList []error
 	nsList := createNamespacesList(phase.NamespaceRange)
 	for _, nsName := range nsList {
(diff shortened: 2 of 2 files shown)


clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -152,12 +153,42 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
 
 // ExecuteObject executes single test object operation based on provided object configuration.
 func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32) []error {
+	var errList []error
```
shyamjvs: I would suggest here the approach of passing another parameter to this function for deciding create/update/delete (since you're anyway already calculating those cases in `ExecutePhase`). It'll make this function much simpler and stateless I hope.

krzysied: good idea. done.

clusterloader2/pkg/test/interface.go
```diff
@@ -25,6 +25,18 @@ import (
 // CreatContextFunc a type for function that creates Context based on given framework client and state.
 type CreatContextFunc func(f *framework.Framework, s *state.NamespacesState) Context
 
+// OperationType is a type of operation to be performed on an object.
+type OperationType int
+
+const (
+	// CREATE_OBJECT is create object operation.
+	CREATE_OBJECT = Operation(0)
```
shyamjvs: Shouldn't this be OperationType from above?

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -151,13 +161,34 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
...
 	} else {
-		// TODO: handle namespaces level object
+		objName := fmt.Sprintf("%v-%d", object.Basename, replicaIndex)
+		switch operation {
+		case CREATE_OBJECT:
+			if err := ctx.GetFramework().CreateObject(namespace, objName, template); err != nil {
+				errList = append(errList, err)
+			}
+		case UPDATE_OBJECT:
+			// TODO: add hanlding for update operation
```
shyamjvs: -> handling

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -151,13 +161,34 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
...
+		case DELETE_OBJECT:
+			if err := ctx.GetFramework().DeleteObject(gvk, namespace, objName); err != nil {
+				errList = append(errList, err)
+			}
+		default:
+			errList = append(errList, fmt.Errorf("Unsupported operation %v", operation))
```
shyamjvs: Please also try to include objName in the error msgs wherever relevant. Otherwise it's hard to trace back to particular operation.

clusterloader2/pkg/test/interface.go
krzysied: Should be. Last moment rename and it was not propagated properly :/

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: fixed

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: error massages changed.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -151,13 +161,34 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
...
+
 	if namespace == "" {
 		// TODO: handle cluster level object
 	} else {
-		// TODO: handle namespaces level object
+		objName := fmt.Sprintf("%v-%d", object.Basename, replicaIndex)
+		switch operation {
+		case CREATE_OBJECT:
+			if err := ctx.GetFramework().CreateObject(namespace, objName, template); err != nil {
+				errList = append(errList, fmt.Errorf("namespace %v object %v creation error: %v", namespace, objName, err))
```
shyamjvs: instead of recomputing the name string everytime in each error msg, could you just compute once at top and use that?

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -151,13 +161,34 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
 }
 
 // ExecuteObject executes single test object operation based on provided object configuration.
-func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32) []error {
+func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32, operation OperationType) []error {
```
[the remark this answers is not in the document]
krzysied: Nah... `operation` has better readability factor than `op`. I'm not scared of using long names.

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: I could, however only one of those cases will be used at the time and errors (should) rarely happen. If it's computed at the top, in most cases it will not be even used. I prefer to keep it as it is.

shyamjvs: You're right. It's computed only once, so it's ok.

PR #172, opened 2018-07-25
ClusterLoader - Object update

krzysied:
Updating object by creating patch based on already existing object and new object template.

### DIFF
--- a/clusterloader2/pkg/framework/client/objects.go
+++ b/clusterloader2/pkg/framework/client/objects.go
@@ -26,6 +26,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/jsonmergepatch"
+	"k8s.io/apimachinery/pkg/util/mergepatch"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
@@ -110,20 +113,19 @@ func CreateObject(dynamicClient dynamic.Interface, namespace string, name string
 
 // UpdateObject updates with given name, group, version and kind based on given object description.
 func UpdateObject(dynamicClient dynamic.Interface, namespace string, name string, obj *unstructured.Unstructured) error {
-	// TODO(krzysied): This method should be more generic and robust.
 	gvk := obj.GroupVersionKind()
 	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
-	currentObj, err := GetObject(dynamicClient, gvk, namespace, name)
-	if err != nil {
-		return err
-	}
-	spec, exists := obj.UnstructuredContent()["spec"]
-	if !exists {
-		return fmt.Errorf("Object without a spec")
-	}
-	unstructured.SetNestedField(currentObj.UnstructuredContent(), spec, "spec")
+	obj.SetName(name)
 	updateFunc := func() error {
-		_, err := dynamicClient.Resource(gvr).Namespace(namespace).Update(currentObj)
+		currentObj, err := dynamicClient.Resource(gvr).Namespace(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		patch, err := createPatch(currentObj, obj)
+		if err != nil {
+			return fmt.Errorf("creating patch diff error: %v", err)
+		}
+		_, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.StrategicMergePatchType, patch)
 		return err
 	}
 	return RetryWithExponentialBackOff(retryFunction(updateFunc, nil))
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/framework/client/objects.go
```diff
@@ -154,3 +157,17 @@ func GetObject(dynamicClient dynamic.Interface, gvk schema.GroupVersionKind, nam
...
+	if err != nil {
+		return []byte{}, err
+	}
+	modifiedJson, err := modified.MarshalJSON()
+	if err != nil {
+		return []byte{}, err
+	}
+	preconditions := []mergepatch.PreconditionFunc{mergepatch.RequireKeyUnchanged("apiVersion"),
+		mergepatch.RequireKeyUnchanged("kind"), mergepatch.RequireMetadataKeyUnchanged("name")}
+	return jsonmergepatch.CreateThreeWayJSONMergePatch(nil, modifiedJson, currentJson, preconditions...)
```
shyamjvs: Are you copying this logic over from somewhere?

shyamjvs: Since you're passing the first field of a 3-way merge (which is supposed to be the original object) as nil, that seems to be as good as just a 2-way merge with the new one being applied to the current one. Can't you instead just use `strategicpatch.CreateTwoWayMergePatch`?

shyamjvs: Actually I'm a bit unsure if both of them are the same. IIUC using TwoWayMergePatch means that whenever there's a conflict and we need to choose, the new one will simply override the old one. So we don't detect for conflicts before applying. However, ThreeWayMergePatch will throw an error with a conflict so we can get and retry again - but for that we need to pass original object I think?

@liggitt Could you give a lead on what's the right approach to follow here?

krzysied: CreateTwoWayMergePatch requires actual object structure to be passed. Currently we only use Unstructured which is not supported by this function.
TODO added.

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -110,20 +113,19 @@ func CreateObject(dynamicClient dynamic.Interface, namespace string, name string
...
-		return fmt.Errorf("Object without a spec")
+	obj.SetName(name)
+	patch, err := createPatch(currentObj, obj)
+	if err != nil {
+		return fmt.Errorf("creating patch diff error: %v", err)
 	}
-	unstructured.SetNestedField(currentObj.UnstructuredContent(), spec, "spec")
 	updateFunc := func() error {
-		_, err := dynamicClient.Resource(gvr).Namespace(namespace).Update(currentObj)
+		_, err := dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.StrategicMergePatchType, patch)
```
shyamjvs: Please move the above code block where you're computing the patch into this updateFunc. Otherwise I think the logic would be wrong, because you need to recalculate the patch each time you failed to update the object due to a conflict. RIght now you're re-using the same patch again and again.

clusterloader2/pkg/framework/client/objects.go
shyamjvs: Thanks for adding todo. Let's figure this out later if that's the right way.

clusterloader2/pkg/framework/client/objects.go
krzysied: Get object function and createPatch moved into retryable function.
I agree that's better approach, although previous one also worked due to the assumption that there are no external things operating on the same object.

shyamjvs: That's not a very good assumption to make in practice :)
Because there are so many interactions in the system and components that we aren't even aware of may potentially change the objects.

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -110,20 +113,19 @@ func CreateObject(dynamicClient dynamic.Interface, namespace string, name string
...
-		_, err := dynamicClient.Resource(gvr).Namespace(namespace).Update(currentObj)
+		currentObj, err := dynamicClient.Resource(gvr).Namespace(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		patch, err := createPatch(currentObj, obj)
+		if err != nil {
+			return fmt.Errorf("creating patch diff error: %v", err)
+		}
+		_, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.StrategicMergePatchType, patch)
```
wojtek-t: @krzysied @shyamjvs - this is no longer Update operation then. It should be called PatchObject or sth like that in that case.

shyamjvs: I agree. We should call this PatchObject based on what we're doing.

Also @krzysied, did you try using `Update()` instead? Did you face any problems with that?

krzysied: > Also @krzysied, did you try using Update() instead? Did you face any problems with that?

I've tried to use Update(). The problem was that it required to provided existing object with updated fields and I haven't found any support for this in k8s code.

So I've changed approach and I tried to implemented it the same way as `apply` is implemented in kubectl. It turned out `apply` is done by executing `patch` method.

> this is no longer Update operation then. It should be called PatchObject or sth like that in that case.

I have different opinion about this. I conciser patch as way of updating object - after executing this function the object will be (somehow) updated so the name seems valid.

I guess patch and update may have different meaning in k8s, so please let me know if that's the case.

wojtek-t: It's not about k8s - it's a more generic nomenclature. Yes - patching is a way of updating an object, but it's a different operation.
For Patch you provide a diff and that diff is applied to your object.
For Update, you provide a whole object to which the object should be updated.

I would still prefer using Update, but for now at the very least we should rename it to PatchObject.

shyamjvs: +1

> I would still prefer using Update

I thought about this a while ago and it wasn't very obvious to me how to generate updated object from current object and new json (for e.g how do we merge arrays?). That said, I guess we can continue to think about this with a TODO left for now.

krzysied: >For Patch you provide a diff and that diff is applied to your object.
For Update, you provide a whole object to which the object should be updated.

So if method takes whole object as a parameter, it should named 'update' according to your definition, right?

wojtek-t: What matters from our perspective (it's a framework for testing k8s) is what is send to k8s. And we are sending patch request to k8s.

krzysied: I've created #189 PR. It will change UpdateObject to PatchObject.

PR #176, opened 2018-07-30
ClusterLoader - Adding templating

krzysied:
Replacing object names + index template.

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -67,6 +67,8 @@ type Object struct {
 	Basename string `json: basename`
 	// ObjectTemplatePath specifies the path to object definition.
 	ObjectTemplatePath string `json: objectTemplatePath`
+	// TemplateFillMap specifies for each placeholder what value should it be replaced with.
+	TemplateFillMap map[string]interface{} `json: templateFillMap`
 }
 
 // NamespaceRange specifies the range of namespaces [Min, Max].
--- a/clusterloader2/pkg/config/codec.go
+++ b/clusterloader2/pkg/config/codec.go
@@ -51,8 +51,13 @@ func ReadTemplate(path string) (*unstructured.Unstructured, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading template file: %v", err)
 	}
+	return ConvertToObject(raw)
+}
+
+// ConvertToObject converts array of bytes into unstructured object.
+func ConvertToObject(raw []byte) (*unstructured.Unstructured, error) {
 	obj := &unstructured.Unstructured{}
-	_, _, err = scheme.Codecs.UniversalDeserializer().Decode(raw, nil, obj)
+	_, _, err := scheme.Codecs.UniversalDeserializer().Decode(raw, nil, obj)
 	if err != nil {
 		return nil, fmt.Errorf("unmarshaling error: %v", err)
 	}
--- a/clusterloader2/pkg/config/templateprovider.go
+++ b/clusterloader2/pkg/config/templateprovider.go
@@ -17,7 +17,11 @@ limitations under the License.
 package config
 
 import (
+	"bufio"
+	"bytes"
+	"fmt"
 	"sync"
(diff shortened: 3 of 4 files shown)


shyamjvs:
Couple of high-level comments:

- Don't reimplement logic for template evaluation - try to reuse the golang [template](https://golang.org/pkg/text/template/) library.
- Let's follow the approach we discussed offline to support templating and linking across the yamls. If at some places you have a trade-off between 'implementation complexity' and 'extra requirements imposed on users' - I'd suggest choosing the latter for now. That'll keep things simple and we can think about bettering usability later.

clusterloader2/api/types.go
```diff
@@ -67,6 +67,8 @@ type Object struct {
 	Basename string `json: basename`
 	// ObjectTemplatePath specifies the path to object definition.
 	ObjectTemplatePath string `json: objectTemplatePath`
+	// Mapping specifies for each placeholder what value should it be replaced with.
+	Mapping map[string]interface{} `json: mapping`
```
shyamjvs: Can you give some better name than just 'mapping'? Maybe TemplateFillMap?

clusterloader2/pkg/config/codec.go
```diff
@@ -47,12 +47,22 @@ func ReadConfig(path string) (*api.Config, error) {
...
-	raw, err := ioutil.ReadFile(path)
+	raw, err := ReadRaw(path)
 	if err != nil {
 		return nil, fmt.Errorf("error reading template file: %v", err)
 	}
+	return ConvertToTemplate(raw)
+}
+
+// ReadRaw reads a file specified by the given path.
+func ReadRaw(path string) ([]byte, error) {
```
shyamjvs: Do you really need a function for small one-liner? :)

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -171,7 +176,16 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
...
+	objName := fmt.Sprintf("%v-%d", object.Basename, replicaIndex)
+	var mapping map[string]interface{}
+	if object.Mapping == nil {
+		mapping = make(map[string]interface{})
+	} else {
+		mapping = object.Mapping
+	}
+	mapping[basenamePlaceholder] = objName
+	mapping[indexPlaceholder] = replicaIndex
+	template, err := ctx.GetTemplateProvider().GetTemplate(object.ObjectTemplatePath, mapping)
```
shyamjvs: Again here, I don't think you should be naming this variable 'template'. It's not a template here anymore iiuc, but the final object itself.

clusterloader2/pkg/config/templateprovider.go
```diff
@@ -26,35 +30,52 @@ import (
...
 		}
 	}
-	return obj.DeepCopy(), nil
+	return raw, nil
+}
+
+// GetTemplate creates object template from file specified by the given path
+// or uses cached object if available. Template's placeholders are replaced based
+// on provided mapping.
+func (tp *TemplateProvider) GetTemplate(path string, mapping map[string]interface{}) (*unstructured.Unstructured, error) {
```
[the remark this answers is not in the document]
shyamjvs: Actually, since you seem to return the final object itself (and not the evaluated template) - I instead suggest  sth like `TemplateToObject`.

shyamjvs:
Thanks. Lgtm modulo some comments on variable naming.
Also notice how crisp and simple this PR is now, after using golang template library :)

clusterloader2/api/types.go
shyamjvs: I'm still not convinced if that's a good name as 'mapping' field inside an 'object' struct could mean many different things.  If you `git grep "Mapping"` in the kubernetes code, you'll find several examples of mappings (portMapping, RestMapping, resourceMapping, etc).

krzysied: done

PR #177, opened 2018-08-01
ClusterLoader - Updating measurement api

krzysied:
Adding new fields to Measurement API:
- Identifier - is a identifier within method scope.
Pair <Method, Identifier> makes measurement unique at global scope. The new field is needed due to possible having multiple measurements  instances that uses the same measurement method.
- Params - map of <name, value>. Optional additional method parameters. Allows measurements to be more generic.

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -102,6 +102,10 @@ type TuningSet struct {
 type Measurement struct {
 	// Method is a name of a method registered in the ClusterLoader factory.
 	Method string
+	// Identifier is a string that differentiates measurement instances of the same method.
+	Identifier string
+	// Params is a map of {name: value} pairs which will be passed to the measurement method - allowing for injection of arbitrary parameters to it.
+	Params map[string]interface{}
 }
 
 // QpsLoad defines a uniform load with a given QPS.


clusterloader2/api/types.go
```diff
@@ -102,6 +102,10 @@ type TuningSet struct {
 type Measurement struct {
 	// Method is a name of a method registered in the ClusterLoader factory.
 	Method string
+	// Identifier is a string that differentiates measurement instances of the same method.
+	Identifier string
+	// Params is set of parameters (name: value). Interpretation of parameters type is handled by each measurement.
```
shyamjvs: Maybe sth like:

```
// Params is a map of {name: value} pairs which will be passed to the measurement method - allowing for injection of arbitrary parameters to it.
```

krzysied: done.

shyamjvs:
Since we started thinking about the precise details of the measurement API, I have some thoughts about how it should look like:

- We should change the Measurement interface to just have a single method in it called `Execute()` instead of two methods `Start()` and `Stop()`. This also would be similar to how we have `ExecutePhase` method for phases. The reason I say is the following:

  - Not all measurement operations need to have start and stop, because it may not make sense for certain operations (for e.g one-off things like `WaitForAllPodsReady`, `GatherMemoryProfile`, etc). IMO we shouldn't force user to implement 2 functions (start and stop) for such cases.
  - For operations where start and stop indeed make sense, you can still represent those through single `Execute` method - you'd just need to pass an additional start/stop as a parameter in `Params`

- The signature of the Execute function should look as follows - `func (m Measurement) Execute(ClientConfig, Params) error`. ClientConfig would contain `kubeClient` (to talk to the k8s API) and `cloudProvider` (to talk to cloud-provider API) - I can think of e.g.s why either of them would be needed. And Params of course is to allow injecting arbitrary inputs to the Execute method.

@wojtek-t

shyamjvs:
/lgtm

I'll send a PR to update the doc with the above thoughts. Let's see what Wojtek thinks.

PR #180, opened 2018-08-01
ClusterLoader - Adding measurement manager

krzysied:
Implementation of measurement factory and measurement manager.
CloudProvider will be add in future PR.

### DIFF
--- a/clusterloader2/pkg/measurement/factory.go
+++ b/clusterloader2/pkg/measurement/factory.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package measurement
+
+import (
+	"fmt"
+	"sync"
+)
+
+// Factory is a default global factory instance.
+var factory = newMeasurementFactory()
+
+func newMeasurementFactory() *measurementFactory {
+	return &measurementFactory{
+		createFuncs: make(map[string]createMeasurementFunc),
+	}
+}
+
+// measurementFactory is a factory that creates measurement instances.
+type measurementFactory struct {
+	lock        sync.RWMutex
+	createFuncs map[string]createMeasurementFunc
+}
+
+func (mc *measurementFactory) register(methodName string, createFunc createMeasurementFunc) error {
+	mc.lock.Lock()
+	defer mc.lock.Unlock()
+	_, exists := mc.createFuncs[methodName]
+	if exists {
+		return fmt.Errorf("measurement with method %v already exists", methodName)
+	}
+	mc.createFuncs[methodName] = createFunc
+	return nil
+}
+
+func (mc *measurementFactory) createMeasurement(methodName string) (Measurement, error) {
+	mc.lock.RLock()
+	defer mc.lock.RUnlock()
+	createFunc, exists := mc.createFuncs[methodName]
+	if !exists {
+		return nil, fmt.Errorf("unknown measurement method %s", methodName)
+	}
+	return createFunc(), nil
+}
+
+// Register registers create measurement function in measurement factory.
+func Register(methodName string, createFunc createMeasurementFunc) error {
+	return factory.register(methodName, createFunc)
+}
--- a/clusterloader2/pkg/measurement/interface.go
+++ b/clusterloader2/pkg/measurement/interface.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package measurement
+
+import (
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementConfig provides client and parameters required for the measurement execution.
+type MeasurementConfig struct {
+	// Clientset is a kubernetes client.
+	ClientSet clientset.Interface
+	// Params is a map of {name: value} pairs enabling for injection of arbitrary config
+	// into the Execute method.
+	Params map[string]interface{}
+	// TODO(krzysied): add CloudProvider.
+}
+
+// Measurement is an common interface for all measurements methods. It should be implemented by the user to
+// allow his/her measurement method to be registered in the measurement factory.
+// See https://github.com/kubernetes/perf-tests/blob/master/clusterloader/docs/design.md for reference.
+type Measurement interface {
+	Execute(config *MeasurementConfig) error
+}
+
+type createMeasurementFunc func() Measurement
--- a/clusterloader2/pkg/measurement/manager.go
+++ b/clusterloader2/pkg/measurement/manager.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package measurement
+
+import (
+	"sync"
+
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementManager manages all measurement executions.
+type MeasurementManager struct {
+	clientSet clientset.Interface
+
+	lock sync.Mutex
+	// map from method type and identifier to measurement instance.
+	measurements map[string]map[string]Measurement
+}
+
+// CreateMeasurementManager creates new instance of MeasurementManager.
+func CreateMeasurementManager(clientSet clientset.Interface) *MeasurementManager {
+	return &MeasurementManager{
+		clientSet:    clientSet,
+		measurements: make(map[string]map[string]Measurement),
+	}
+}
+
+// Execute executes measurement based on provided identifier, methodName and params.
+func (mm *MeasurementManager) Execute(methodName string, identifier string, params map[string]interface{}) error {
+	measurementInstance, err := mm.getMeasurementInstance(methodName, identifier)
+	if err != nil {
+		return err
+	}
+	config := &MeasurementConfig{
+		ClientSet: mm.clientSet,
+		Params:    params,
+	}
+	return measurementInstance.Execute(config)
+}
+
+func (mm *MeasurementManager) getMeasurementInstance(methodName string, identifier string) (Measurement, error) {
+	mm.lock.Lock()
+	defer mm.lock.Unlock()
+	if _, exists := mm.measurements[methodName]; !exists {
+		mm.measurements[methodName] = make(map[string]Measurement)
+	}
+	if _, exists := mm.measurements[methodName][identifier]; !exists {
+		measurementInstance, err := factory.createMeasurement(methodName)
+		if err != nil {
+			return nil, err
+		}
+		mm.measurements[methodName][identifier] = measurementInstance
+	}
+	return mm.measurements[methodName][identifier], nil
+}


clusterloader2/pkg/measurement/measurementmanager.go
```diff
@@ -0,0 +1,67 @@
+/*
```
shyamjvs: Since you're already inside measurement/ directory, I think you can simply name this file `manager.go` or sth.

clusterloader2/pkg/measurement/measurementfactory.go
```diff
@@ -0,0 +1,61 @@
+/*
```
shyamjvs: Since you're already inside measurement/ directory, I think you can simply name this file `factory.go` or sth.

clusterloader2/pkg/measurement/measurementmanager.go
```diff
@@ -0,0 +1,67 @@
...
+	lock         sync.Mutex
+	measurements map[string]map[string]Measurement
+	clientSet    clientset.Interface
+}
+
+// CreateMeasurementManager creates new instance of MeasurementManager.
+func CreateMeasurementManager(clientSet clientset.Interface) *MeasurementManager {
+	return &MeasurementManager{
+		measurements: make(map[string]map[string]Measurement),
+		clientSet:    clientSet,
```
shyamjvs: Move clientset above.

clusterloader2/pkg/measurement/measurementmanager.go
```diff
@@ -0,0 +1,67 @@
...
+	"sync"
+
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementManager manages all measurements calls and executions.
+type MeasurementManager struct {
+	lock         sync.Mutex
+	measurements map[string]map[string]Measurement
+	clientSet    clientset.Interface
```
shyamjvs: Move clientset above.

shyamjvs: Also I don't think it needs to be protected by the lock. It's only set once during initialization.

clusterloader2/pkg/measurement/measurementmanager.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"sync"
+
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementManager manages all measurements calls and executions.
+type MeasurementManager struct {
+	lock         sync.Mutex
+	measurements map[string]map[string]Measurement
```
shyamjvs: Please add a comment above this complex map explaining what it's storing.

clusterloader2/pkg/measurement/measurementfactory.go
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+package measurement
+
+import (
+	"fmt"
+	"sync"
+)
+
+var factory *measurmentsFactory
```
shyamjvs: spelling mistake

shyamjvs: Also change it to `measurementFactory` (i.e plural isn't needed).

clusterloader2/pkg/measurement/measurementmanager.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+	config := &MeasurementConfig{
+		ClientSet: mm.clientSet,
+		Params:    params,
+	}
+	return m.Execute(config)
+}
+
+func (mm *MeasurementManager) getMeasurement(identifier string, metricMethod string) (Measurement, error) {
```
shyamjvs: Why are you calling it metricMethod and not just method? (because it doesn't have to be a metric-related method)

clusterloader2/pkg/measurement/measurementmanager.go
krzysied: done

clusterloader2/pkg/measurement/measurementfactory.go
krzysied: done

clusterloader2/pkg/measurement/measurementmanager.go
krzysied: done

clusterloader2/pkg/measurement/measurementmanager.go
krzysied: done

clusterloader2/pkg/measurement/measurementmanager.go
krzysied: done

clusterloader2/pkg/measurement/measurementfactory.go
krzysied: done

clusterloader2/pkg/measurement/manager.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"sync"
+
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementManager manages all measurements calls and executions.
+type MeasurementManager struct {
+	clientSet    clientset.Interface
+	lock         sync.Mutex
```
shyamjvs: Please add newline above lock.

clusterloader2/pkg/measurement/manager.go
```diff
@@ -0,0 +1,69 @@
...
+
+package measurement
+
+import (
+	"sync"
+
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementManager manages all measurements calls and executions.
```
shyamjvs: -> measurement

shyamjvs: Also just executions is sufficient I guess.

clusterloader2/pkg/measurement/manager.go
```diff
@@ -0,0 +1,69 @@
...
+type MeasurementManager struct {
+	clientSet    clientset.Interface
+	lock         sync.Mutex
+	// measurements is a map: methodName -> identifier -> Measurement.
+	// First level divides measurements into method classes, second into instances.
```
shyamjvs: I think this comment is a bit redundant. Maybe simply say 'Map from method type and identifier to measurement instance'?

clusterloader2/pkg/measurement/factory.go
```diff
@@ -0,0 +1,61 @@
...
+package measurement
+
+import (
+	"fmt"
+	"sync"
+)
+
+var factory *measurementFactory
+
+func init() {
```
shyamjvs: Why do you need to have global instance? Like I mentioned earlier, it makes it hard to have 2 instances of this pkg.

clusterloader2/pkg/measurement/interface.go
```diff
@@ -0,0 +1,39 @@
...
+
+package measurement
+
+import (
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// MeasurementConfig provides client and paramters required for the measurement execution.
+type MeasurementConfig struct {
+	// Clientset is a kubernetes cloud client.
```
shyamjvs: Remove 'cloud'

clusterloader2/pkg/measurement/interface.go
```diff
@@ -0,0 +1,39 @@
...
+	ClientSet clientset.Interface
+	// Params is a map {name: value} pairs enabling for injection of arbitrary config
```
shyamjvs: map of

clusterloader2/pkg/measurement/interface.go
```diff
@@ -0,0 +1,39 @@
...
+	// TODO(krzysied): add CloudProvider.
+}
+
+// Measurement interface
+type Measurement interface {
+	Execute(config *MeasurementConfig) error
+}
+
+// CreateMeasurementFunc is a type of create function used by the measurement factory.
+type CreateMeasurementFunc func() Measurement
```
shyamjvs: I don't think this type needs to be public.

clusterloader2/pkg/measurement/manager.go
```diff
@@ -0,0 +1,69 @@
...
+		return err
+	}
+	config := &MeasurementConfig{
+		ClientSet: mm.clientSet,
+		Params:    params,
+	}
+	return m.Execute(config)
+}
+
+func (mm *MeasurementManager) getMeasurement(identifier string, methodName string) (Measurement, error) {
```
shyamjvs: -> getMeasurementObject? (because getMeasurement sounds ambiguous - it could even mean getting a real measurement)

shyamjvs: Nit - could you reverse the argument order of the function? (to match key order in the map)

clusterloader2/pkg/measurement/manager.go
krzysied: done

clusterloader2/pkg/measurement/manager.go
krzysied: done

clusterloader2/pkg/measurement/manager.go
krzysied: done

clusterloader2/pkg/measurement/factory.go
krzysied: I've added function that can create new instance of MeasurementFactory. Though, I'm still keeping global factory as default instance.

clusterloader2/pkg/measurement/interface.go
krzysied: done

clusterloader2/pkg/measurement/interface.go
krzysied: done

clusterloader2/pkg/measurement/manager.go
krzysied: `getMeasurementInstance` it is.
Order of parameters changed.

clusterloader2/pkg/measurement/manager.go
```diff
@@ -0,0 +1,71 @@
...
+
+	lock sync.Mutex
+	// measurements is a map from method type and identifier to measurement instance.
+	measurements map[string]map[string]Measurement
+}
+
+// CreateMeasurementManager creates new instance of MeasurementManager.
+func CreateMeasurementManager(factory *MeasurementFactory, clientSet clientset.Interface) *MeasurementManager {
+	return &MeasurementManager{
+		factory:      factory,
```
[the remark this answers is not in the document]
shyamjvs: This way users don't need to worry about managing 2 objects (i.e both factory and manager), but only the latter.

clusterloader2/pkg/measurement/manager.go
```diff
@@ -0,0 +1,71 @@
...
+		clientSet:    clientSet,
+		measurements: make(map[string]map[string]Measurement),
+	}
+}
+
+// Execute executes measurement based on provided identifier, methodName and params.
+func (mm *MeasurementManager) Execute(methodName string, identifier string, params map[string]interface{}) error {
+	mInstance, err := mm.getMeasurementInstance(methodName, identifier)
```
shyamjvs: You can as well name it `measurementInstance` instead of `mInstance` :)

clusterloader2/pkg/measurement/manager.go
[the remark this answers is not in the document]
shyamjvs: Infact, you could just merge factory and manager into single struct, as each of them pretty much holds 1 variable.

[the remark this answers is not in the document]
krzysied: > Infact, you could just merge factory and manager into single struct, as each of them pretty much holds 1 variable.

TL;DR: I will be very unhappy to merge those two structures.
Those two are two separated concept. Factory should be something on global level, that is aware of how to create measurement instances.
Manager exists in some context (test context in this case) - it uses factory to create measurement instances. Managers also stores states of instances.

> Initialize a new instance of factory in CreateMeasurementManager (so you don't need that global variable anymore)
Make Register() a public method on the MeasurementManager and make the Register() for the factory a private one, and call the latter from the former

Could you explain to me, how will measurement methods will register in factory? I still missing the point where Register is called by the measurement method implementation or measurement is added to manager.

With this implementation I try to do similar thing to how apis register to legacyscheme.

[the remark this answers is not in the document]
shyamjvs: If factory is something that is really independent of manager, why are you holding it as a field in the manager? So I think it makes sense to either:

- Keep the factory independent of manager and hold them separately, or
- Make manager manage everything. By this I don't mean merging both structs (that was just another suggestion), but I meant that from the cluster-loader perspective it only talks to manager

> Could you explain to me, how will measurement methods will register in factory?

How would it be done currently? It'll be just the same, but instead of calling Register() on factory, you'll call it on manager which will then be responsible for storing it.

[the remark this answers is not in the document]
krzysied: > If factory is something that is really independent of manager, why are you holding it as a field in the manager?

Factory is something independent. It's manager that depends on factory. Managers uses factory the same way it uses clientSet.
Plus, I can easily imagine having factory initialize once and then have multiple tests run with separate manager each that use this one factory. Therefore is think those two thing are separate concepts.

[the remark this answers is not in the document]
krzysied: > How would it be done currently? It'll be just the same, but instead of calling Register() on factory, you'll call it on manager which will then be responsible for storing it.

I'm keeping one instance of factory as global variable so measurements can register themselves. Does it mean that using your approach I should make one global manager? Or after creating manager should I explicitly register each measurement? - This part is still not clear for me :(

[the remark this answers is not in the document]
shyamjvs: - If factory is independent of manager's lifecycle, then why are your storing it as a field in the manager?
- If factory can exist without manager, why are you making createMeasurement() private? Because that's also something a user of factory would want besides Register()

> Therefore is think those two thing are separate concepts.

I agree and am not saying they're not. It's about what really do you want to expose to users of this package.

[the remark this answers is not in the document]
shyamjvs: > I'm keeping one instance of factory as global variable so measurements can register themselves. Does it mean that using your approach I should make one global manager?

Yes. The only change you should need to make is having a manager instance instead of a manager and a factory instance (whichever way you're defining them.. global or local or field ..)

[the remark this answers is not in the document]
krzysied: > If factory is independent of manager's lifecycle, then why are your storing it as a field in the manager?

Manager is using it therefore pointer is needed. In theory, someone can have separate factory instance. So there should be a way to choose which factory should be used.

TBH, I still prefer to have factory singleton in measurement module (as it was done in first place). This way there is no need to pass factory pointers etc.

> If factory can exist without manager, why are you making createMeasurement() private? Because that's also something a user of factory would want besides Register()

That's good observation. I should make createMeasurement public.

> Yes. The only change you should need to make is having a manager instance instead of a manager and a factory instance (whichever way you're defining them.. global or local or field ..)

Manager is something that should exist only on test level, thus can't be made global. On the other hand factory in independent of any test state so this one can be made global.

shyamjvs:
/lgtm
Thanks!

PR #182, opened 2018-08-03
ClusterLoader - Adding measurement manager to test context

krzysied:


clusterloader2/pkg/framework/framework.go
```diff
@@ -60,6 +60,16 @@ func NewFramework(path string) (*Framework, error) {
 	return &Framework{0, clientSet, dynamicClient}, nil
 }
 
+// GetClientSet returns clientSet client.
+func (f *Framework) GetClientSet() clientset.Interface {
+	return f.clientSet
+}
+
+// GetDynamicClient returns dynamic client.
+func (f *Framework) GetDynamicClient() dynamic.Interface {
```
shyamjvs: This is not called anywhere. Why have it then?

clusterloader2/pkg/test/simplecontext.go
```diff
@@ -19,23 +19,26 @@ package test
...
 
 type simpleContext struct {
-	framework        *framework.Framework
-	state            *state.NamespacesState
-	templateProvider *config.TemplateProvider
-	tickerFactory    ticker.TickerFactory
+	framework          *framework.Framework
+	state              *state.NamespacesState
+	templateProvider   *config.TemplateProvider
+	tickerFactory      ticker.TickerFactory
```
shyamjvs: This is oddly the only non-pointer type. Can you change it?

krzysied: This `TickerFactory` is an interface - it doesn't use * cause *type is already cast to this interface.

clusterloader2/pkg/framework/framework.go
krzysied: I thought that if there is need for getter to ClientSet than at some point will be need for DynamicClient also. It's just a getter.

shyamjvs: If we need it later, then let's add later. Otherwise this may potentially lie around as dead code forever.

clusterloader2/pkg/test/simplecontext.go
shyamjvs: Ok.

clusterloader2/pkg/framework/framework.go
krzysied: done.

PR #183, opened 2018-08-03
ClusterLoader - Adding measurement manger to test execution

krzysied:


krzysied:
/hold cancel
fixed.

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -62,12 +62,22 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) []erro
...
 	var errList []error
 	if len(step.Measurements) > 0 {
-		// TODO: handle measurements
+		for i := range step.Measurements {
+			// index is created to make i value unchangeable during thread execution.
+			index := i
+			wg.Start(func() {
+				if err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params); err != nil {
+					lock.Lock()
+					defer lock.Unlock()
```
shyamjvs: Maybe in a separate PR, but I think it makes sense to move this errList+lock combination into a struct (since you seem to use it at quite some places).

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -62,12 +62,22 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) []erro
...
+					errList = append(errList, fmt.Errorf("measurement call %s error: %v", step.Measurements[index].Method, err))
```
shyamjvs: Please also provide identifier in the error msg (otherwise we can't distinguish).

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: TODO added (line 66).

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: done

clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -62,12 +62,23 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) []erro
...
+	// TODO(krzysied): Consider moving lock and errList to separate structure.
+	var lock sync.Mutex
 	var errList []error
 	if len(step.Measurements) > 0 {
-		// TODO: handle measurements
+		for i := range step.Measurements {
+			// index is created to make i value unchangeable during thread execution.
+			index := i
+			wg.Start(func() {
+				if err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params); err != nil {
```
wojtek-t: @shyamjvs @krzysied - I started wondering again about that.

This suggests that Execute() is asynchrnous - it starts the measurements but doesn't wait for it to end.
But how is "verify" phase going to be implemented in with that interface?

Let's say that I want to measure pod startup time for some class of pods. So I call Execute on its measurement (with some params) and that starts measuring.

But at the end I would like to verify the metrics so I need some other (Verify or sth like that) method on Meausrement interface to do that (it can have default empty implementation since it may not be required for all measurements e.g. profiles).

Also, another question that I have is: I would like to stop all meausrements at the end of the test. How do I do that?

shyamjvs: > But how is "verify" phase going to be implemented in with that interface?

Couple options I can think of (assuming I understand your ask correctly):

- verify can be called part of the "stop" logic of Execute() (or just at the end of the function in cases where start/stop don't make sense)
- Execute() will write the results to let's say some json file. Verify could be another measurement (towards the end of the test) which does some post-processing of the metrics

> I would like to stop all meausrements at the end of the test. How do I do that?

Yeah, I see your point. One possibility is we ask the users to explicitly stop the measurements for which stopping makes sense. This is a bit imposing on users, but maybe it's ok?

PR #184, opened 2018-08-03
ClusterLoader - Adding measurement util

krzysied:
Adding functions for retrieving variables from map[string]interface{} that is used for measurement params.
This will be commonly used by measurement implementation due to map[string]interface{}  being  part of measurement interface.

### DIFF
--- a/clusterloader2/pkg/util/util.go
+++ b/clusterloader2/pkg/util/util.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+)
+
+// GetString tries to return value from map cast to string type. If value doesn't exist, error is returned.
+func GetString(dict map[string]interface{}, key string) (string, error) {
+	return getString(dict, key, "", fmt.Errorf("key %s not found", key))
+}
+
+// GetInt tries to return value from map cast to int type. If value doesn't exist, error is returned.
+func GetInt(dict map[string]interface{}, key string) (int, error) {
+	return getInt(dict, key, 0, fmt.Errorf("key %s not found", key))
+}
+
+// GetFloat64 tries to return value from map cast to float64 type. If value doesn't exist, error is returned.
+func GetFloat64(dict map[string]interface{}, key string) (float64, error) {
+	return getFloat64(dict, key, 0, fmt.Errorf("key %s not found", key))
+}
+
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/util/util.go
```diff
@@ -0,0 +1,91 @@
+/*
```
shyamjvs: Do you really need to create another 'util' directory for this file? Maybe move it to measurement/ ?

clusterloader2/pkg/measurement/util/util.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Types from interface{} create from json cannot be cast directly to int.
+	floatValue, ok := value.(float64)
+	if !ok {
+		return 0, fmt.Errorf("type assertion error: %v of is not an int", value)
+	}
+	return int(floatValue), nil
+}
+
+func getFloat64(dict map[string]interface{}, key string, defaultValue float64, defaultError error) (float64, error) {
```
shyamjvs: Do you really need one function for each type? Can't you make this a bit generic taking type as argument?

shyamjvs: I'm not sure if/how that can be done in golang though.

clusterloader2/pkg/measurement/util/util.go
krzysied: I've moved it to separate directory because this functionality is not necessary for measurement.
It's very convenient to have this kind of util, however measurement can be implemented without it.

clusterloader2/pkg/measurement/util/util.go
krzysied: As far as I know this thing can't be done in golang:
- You will need to have a way of setting returned varaible type during runtime. Templates are not supported by golang.
- You can make return type more generic like interface{}, but it will make no sens - this functions are made to cast variable into specific type.

clusterloader2/pkg/measurement/util/util.go
shyamjvs: If it's not a measurement specific util, I'd suggest moving it outside. Probably into pkg/util.

clusterloader2/pkg/measurement/util/util.go
```diff
@@ -0,0 +1,91 @@
...
+func GetIntOrDefault(dict map[string]interface{}, key string, defaultValue int) (int, error) {
+	return getInt(dict, key, defaultValue, nil)
+}
+
+// GetFloat64OrDefault tries to return value from map cast to float64 type. If value doesn't exist default value is used.
+func GetFloat64OrDefault(dict map[string]interface{}, key string, defaultValue float64) (float64, error) {
+	return getFloat64(dict, key, defaultValue, nil)
+}
+
+func getString(dict map[string]interface{}, key string, defaultValue string, defaultError error) (string, error) {
```
shyamjvs: Why do you need to take this defaultError? You seem to setting to mil anyway.

clusterloader2/pkg/measurement/util/util.go
shyamjvs: Ok, sgtm.

clusterloader2/pkg/measurement/util/util.go
krzysied: ok. done.

clusterloader2/pkg/measurement/util/util.go
krzysied: Sometimes it's nil, sometimes it's key not found error - depends on whether we allow default value or not.

PR #185, opened 2018-08-06
ClusterLoader - Adding timer measurement

krzysied:


clusterloader2/pkg/measurement/common/timer.go
```diff
@@ -0,0 +1,70 @@
...
+func createTimerMeasurment() measurement.Measurement {
+	return &timer{
+		initialized: false,
+	}
+}
+
+type timer struct {
+	initialized bool
+	startTime   time.Time
+}
```
shyamjvs: Do you really need initialized? Can't you just check if startTime is non-zero for that?

shyamjvs: (assuming time.Time has a default value we can compare for)

clusterloader2/pkg/measurement/common/timer.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Execute supports two actions. start - which start timer. stop - which stops timer
+// and prints to log time duration between start and stop.
+// Stop action requires label parameter to be provided.
+func (t *timer) Execute(config *measurement.MeasurementConfig) error {
+	action, err := util.GetString(config.Params, "action")
```
shyamjvs: You might want to import util package as _ (to a avoid using `util.`)

clusterloader2/pkg/measurement/common/timer.go
krzysied: done.

clusterloader2/pkg/measurement/common/timer.go
krzysied: I prefer explicit prefixes. I like to know which functions are external (just by looking).

PR #186, opened 2018-08-07
ClusterLoader  -  wait-for-pods measurement

krzysied:


clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,99 @@
...
+func createWaitForPodMeasurment() measurement.Measurement {
+	return &waitForPodMeasurment{}
+}
+
+type waitForPodMeasurment struct{}
+
+// Execute waits until desired number of pods is available or until timeout happens.
+// Pods can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+func (wfpm *waitForPodMeasurment) Execute(config *measurement.MeasurementConfig) error {
```
shyamjvs: I suggest wfpm -> w

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,99 @@
...
+	podCount, err := util.GetInt(config.Params, "podCount")
```
shyamjvs: Maybe `desiredPodCount`?

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")
+	if err != nil {
+		return err
+	}
+	fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")
+	if err != nil {
+		return err
+	}
+	timeout, err := util.GetIntOrDefault(config.Params, "timeout", 60)
```
shyamjvs: Can you make it timeoutSeconds or timeoutMinutes or sth like that (to make units clearer)? Even better if you can take a duration as the parameter (for e.g 12m30s)

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,99 @@
...
+
+func init() {
+	measurement.Register("WaitForPods", createWaitForPodMeasurment)
+}
+
+func createWaitForPodMeasurment() measurement.Measurement {
+	return &waitForPodMeasurment{}
+}
+
+type waitForPodMeasurment struct{}
```
shyamjvs: waitForPod what? For e.g it could be waitForPodRunning, waitForPodReady, etc. I suggest taking that as a parameter in the function and  deciding logic based on it.

shyamjvs: Also I suggest waitForPod's' - as we're waiting for multiple pods.

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		if namespace == metav1.NamespaceAll {
+			glog.Infof("WaitForPods: ready %d / %d", runningPodsCount, podCount)
+		} else {
+			glog.Infof("WaitForPods: %s: ready %d / %d", namespace, runningPodsCount, podCount)
+		}
+		if runningPodsCount < podCount {
+			continue
+		}
+		hasTimeout = false
```
shyamjvs: Please remove this `hasTimeout` variable and just return nil here.

clusterloader2/pkg/util/podstore.go
```diff
@@ -0,0 +1,86 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
shyamjvs: 2018

clusterloader2/pkg/util/podstore.go
```diff
@@ -0,0 +1,86 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file is copy of https://github.com/kubernetes/kubernetes/blob/master/test/utils/pod_store.go
+with slight changes regarding labelSelector and flagSelector applied.
```
shyamjvs: Is the only reason for copying over this file instead of vendoring is to make that change?

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: replace with `_` - structure is not used.

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: replaced.

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: I've added support for GetDuration(OrDefault) to util. It will parse stings like "12m30s", "2h".

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: Change to `waitForRunningPodsMeasurement`.
It don't want to allow (at least not now) passing pod status as parameter - handling for this doesn't seem trivial.

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: done.

clusterloader2/pkg/util/podstore.go
krzysied: done.
I thought I should keep previous value because is mostly a copy.

clusterloader2/pkg/util/podstore.go
krzysied: Not a only reason. I also didn't want to vendor whole `https://github.com/kubernetes/kubernetes/blob/master/test/utils` due to having dependencies to kubernetes repository.

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,94 @@
...
+	measurement.Register("WaitForRunningPods", createWaitForRunningPodsMeasurement)
+}
+
+func createWaitForRunningPodsMeasurement() measurement.Measurement {
+	return &waitForRunningPodsMeasurement{}
+}
+
+type waitForRunningPodsMeasurement struct{}
+
+// Execute waits until desired number of pods is available or until timeout happens.
```
shyamjvs: "desired number of pods are running"

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")
+	if err != nil {
+		return err
+	}
+	fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")
+	if err != nil {
+		return err
+	}
+	timeout, err := util.GetDurationOrDefault(config.Params, "timeout", time.Duration(60)*time.Second)
```
shyamjvs: `time.Second*60` isn't enough?

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,94 @@
...
+			continue
+		}
+		runningPodsCount := 0
+		for i := range pods {
+			if pods[i].Status.Phase == corev1.PodRunning {
+				runningPodsCount++
+			}
+		}
+		if namespace == metav1.NamespaceAll {
+			glog.Infof("WaitForPods: ready %d / %d", runningPodsCount, desiredPodCount)
```
shyamjvs: ready -> running (both mean different things in k8s)

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: I thought that is should be cast to time.Duration, but it turned out constant is also fine. done.

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: ok. done.

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,94 @@
...
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
```
shyamjvs: Please add newline after this import. We're usually seperating k8s imports from other vendor imports.

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,94 @@
...
+				runningPodsCount++
+			}
+		}
+		if namespace == metav1.NamespaceAll {
+			glog.Infof("WaitForPods: running %d / %d", runningPodsCount, desiredPodCount)
+		} else {
+			glog.Infof("WaitForPods: %s: running %d / %d", namespace, runningPodsCount, desiredPodCount)
+		}
+		if runningPodsCount < desiredPodCount {
+			continue
```
shyamjvs: Instead can you just negate this condition and `return nil` inside?

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: I have this style of import in every file `clisterloader2` files.
I agree we can change it to match k8s, but (because of many files that need to be changed) in separate PR.

clusterloader2/pkg/util/podstore.go
shyamjvs: Ok, that's reasonable for now.

clusterloader2/pkg/measurement/common/waitforpods.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		return nil
+	}
+
+	return fmt.Errorf("timeout while waiting for pods with labels %v and fields %v to be running", labelSelector, fieldSelector)
```
shyamjvs: Could you also include namespace, desiredCount and actual count in this error string? Sth like:

`Timeout while waiting for %v pods to be running in namespace '%v' with labels '%v' and fields '%v' - only %v found running.`

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: done

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: done

shyamjvs: Could you make it `namespace '%v' labels '%v' and fields '%v'` so that people will explicitly know they were empty instead of thinking there were some labels and fields

clusterloader2/pkg/measurement/common/waitforpods.go
shyamjvs: Sure.

clusterloader2/pkg/measurement/common/waitforpods.go
krzysied: done

PR #188, opened 2018-08-08
ClusterLoader - adding e2e test framework

krzysied:
Adding k8s test e2e framework to vendor.
Methods from e2e will be called by metric implementations.


clusterloader2/cmd/clusterloader.go
```diff
@@ -47,12 +48,18 @@ func validateFlags() []error {
 
 func main() {
 	defer glog.Flush()
+	gomega.RegisterFailHandler(failHandler)
+	k8sframework.RegisterCommonFlags()
+	k8sframework.RegisterClusterFlags()
+	k8sframework.RegisterStorageFlags()
```
shyamjvs: Do you need this? If not, please remove.

clusterloader2/cmd/clusterloader.go
```diff
@@ -47,12 +48,18 @@ func validateFlags() []error {
 
 func main() {
 	defer glog.Flush()
+	gomega.RegisterFailHandler(failHandler)
+	k8sframework.RegisterCommonFlags()
```
shyamjvs: Instead of registering all these flags yourself, could you just import `kubernetes/test/e2e` package which does it automatically as part of init?

shyamjvs: Or just call `framework.ViperizeFlags()`

PR #189, opened 2018-08-09
ClusterLoader - Renaming update to patch

krzysied:


clusterloader2/pkg/test/interface.go
```diff
@@ -34,8 +34,8 @@ type OperationType int
 const (
 	// CREATE_OBJECT is create object operation.
 	CREATE_OBJECT = OperationType(0)
-	// UPDATE_OBJECT is update object operation.
-	UPDATE_OBJECT = OperationType(1)
+	// PATCH_OBJECT is update object (using patch) operation.
+	PATCH_OBJECT = OperationType(1)
```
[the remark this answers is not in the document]
krzysied: Done.

About adding update: I implemented it some time ago. However, the problem was that to achieve this I had to create custom changes applying function (I haven't found any function capable of doing this in k8s code). So after discussion with Shyam we decided to use patch method instead - patch is used in kubctl so with this implementation we can use well-tested code instead of something custom made.
I will create separate PR with this update function - maybe we can figure out whether it can be change into proper solution.

clusterloader2/pkg/test/interface.go
```diff
@@ -34,10 +34,11 @@ type OperationType int
...
 	// DELETE_OBJECT is delete object operation.
 	DELETE_OBJECT = OperationType(2)
+	// TODO(krzysied): Figure out way to add and implement object update.
```
shyamjvs: Could you move this TODO below the PATCH_OBJECT comment? Also maybe rephrase it as `Figure out how to implement UPDATE_OBJECT operation`

krzysied: done

shyamjvs:
@cjwagner Seems like tide is stuck (even though presubmit it complains about has passed).
Will merge manually after giving it some more time.

cjwagner:
This looks like a problem with GitHub search itself that Jetstack is also encountering.
This query should contain this PR, but doesn't: 
https://github.com/search?q=is%3Apr+state%3Aopen+repo%3A%22kubernetes%2Fperf-tests%22+label%3A%22lgtm%22+label%3A%22approved%22

Not even this query shows the PR:
https://github.com/search?q=is%3Apr+state%3Aopen+repo%3A%22kubernetes%2Fperf-tests%22+label%3Aapproved&type=Issues

See the slack conversation here:
https://kubernetes.slack.com/archives/C09QZ4DQB/p1533817806000213

shyamjvs:
Ok, let me merge this manually then.

PR #190, opened 2018-08-09
ClusterLoader - API server latency

krzysied:


clusterloader2/pkg/measurement/common/apiserverlatency.go
```diff
@@ -0,0 +1,270 @@
+/*
```
shyamjvs: Please rename this file to apilatency.go (as apiserver latency can mean a thousand things).

clusterloader2/pkg/measurement/common/apiserverlatency.go
```diff
@@ -0,0 +1,270 @@
...
+	// as list response sizes are bigger in general for big clusters. We also use a higher threshold
+	// for list calls at cluster scope (this includes non-namespaced and all-namespaced calls).
+	apiListCallLatencyThreshold      time.Duration = 5 * time.Second
+	apiClusterScopeListCallThreshold time.Duration = 10 * time.Second
+
+	currentApiCallMetricsVersion = "v1"
+)
+
+func init() {
+	measurement.Register("ApiServerLatency", createApiServerLatencyMeasurement)
```
shyamjvs: Please rename to `APICallLatency`.

shyamjvs: Here and below.

clusterloader2/pkg/util/json.go
```diff
@@ -0,0 +1,36 @@
...
+package util
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// PrettyPrintJSON converts given data into formatted json.
+func PrettyPrintJSON(data interface{}) (string, error) {
```
shyamjvs: I suggest moving this into util.go file instead of having separate file for just this function (as I don't expect this to grow anyway).

clusterloader2/pkg/measurement/common/latencymetric.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type latencyMetric struct {
+	Perc50  time.Duration `json:"Perc50"`
+	Perc90  time.Duration `json:"Perc90"`
+	Perc99  time.Duration `json:"Perc99"`
+	Perc100 time.Duration `json:"Perc100"`
+}
+
+// Only 0.5, 0.9 and 0.99 quantiles are supported.
```
shyamjvs: If only those 3 are supported, why are you having `Perc100` field in the struct?

clusterloader2/pkg/measurement/common/apiserverlatency.go
```diff
@@ -0,0 +1,270 @@
...
+		glog.Infof("Resetting latency metrics in apiserver...")
+		body, err := config.ClientSet.CoreV1().RESTClient().Delete().AbsPath("/metrics").DoRaw()
+		if err != nil {
+			return err
+		}
+		if string(body) != "metrics reset\n" {
+			return fmt.Errorf("unexpected response: %q", string(body))
+		}
+		return nil
+	case "gather":
```
shyamjvs: Can you move this big code block below into separate function?

clusterloader2/pkg/measurement/common/apiserverlatency.go
```diff
@@ -0,0 +1,270 @@
...
+// - gather - Gathers and prints current api server latency data.
+func (*apiServerLatencyMeasurement) Execute(config *measurement.MeasurementConfig) error {
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return err
+	}
+
+	switch action {
+	case "reset":
+		glog.Infof("Resetting latency metrics in apiserver...")
```
shyamjvs: Can you move this code block into separate function called `ApiserverMetricsReset()` or sth?

clusterloader2/pkg/measurement/common/apiserverlatency.go
shyamjvs: Or even better apicalllatency.go.

clusterloader2/pkg/measurement/common/apiserverlatency.go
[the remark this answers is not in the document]
krzysied: TBH, I want to have clone of already existing metrics first and then apply any required changes. I would prefer to add TODO and take care of this later.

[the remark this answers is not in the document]
shyamjvs: @wojtek-t I agree with you. Though I think there are some details to be figured out before that (like what if the test runs only for few minutes (for e.g 100-node), how/if it is possible to get metric values from last 5 mins or do we need to reset every time, etc).

So I suggest TODO for now.

clusterloader2/pkg/measurement/common/apiserverlatency.go
krzysied: done

clusterloader2/pkg/measurement/common/apiserverlatency.go
krzysied: done

clusterloader2/pkg/util/json.go
krzysied: done

clusterloader2/pkg/measurement/common/latencymetric.go
krzysied: Perc100 removed.

clusterloader2/pkg/measurement/common/apiserverlatency.go
krzysied: done

clusterloader2/pkg/measurement/common/apiserverlatency.go
krzysied: done

clusterloader2/pkg/measurement/common/apiserverlatency.go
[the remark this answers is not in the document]
wojtek-t: I'm fine with TODO to unblock this PR, but we should do that sooner rather than later.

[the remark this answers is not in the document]
krzysied: TODO added.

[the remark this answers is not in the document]
shyamjvs: I think it's a bit more subtle than that. The new SLO definition is here - https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos . Maybe you could instead just change the TODO to `Migrate this to our new API latency SLO defined here - <above link>`

clusterloader2/pkg/util/prometheus.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
shyamjvs: Given that this prometheus stuff is specific to measurement, maybe you could move this file to `pkg/measurement`?

clusterloader2/pkg/measurement/common/latencymetric.go
```diff
@@ -0,0 +1,39 @@
+/*
```
shyamjvs: Since this is not a measurement itself (but more general util for measurements) and IIUC you're placing measurement implementations under `common/`, probably it makes sense to move this to somewhere like `measurement/util/`?

krzysied: `measurement/util` directory created. I moved there all files related to measurements that doesn't contain metric implementation by itself.

clusterloader2/pkg/util/prometheus.go
krzysied: Moved to `measurement/util`.

clusterloader2/pkg/measurement/common/apiserverlatency.go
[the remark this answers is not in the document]
krzysied: done.

shyamjvs:
One more thing (I've said this to you offline, but FTR) is before going further and implementing all the measurements, I think you should first start creating a very plain vanilla version of density (and probably load) test to clusterloader. You can begin by setting up a very simple CI job running a simple test (like create/delete few RCs) and then iterating on it. The job should be defined very similar to the pre-existing [clusterloader job](https://github.com/kubernetes/test-infra/blob/dd0ca22a53816efcc57563e8a0ad32c04c493fb3/jobs/config.json#L2166). Measurements can be added later at ease.

PR #191, opened 2018-08-09
ClusterLoader - object update

krzysied:


krzysied:
I created  `applyValuesMap` and `applyValuesList` for traversing structures used by Unstructured and applying all field values from `src` object. I assume that object retrieved by `GetObject` has all template's fields + some additional set by server, therefore all fields from new template should be copied to current object.

This implementation is working for simple cases - when fields change values.
The problem occurs  if some field has been removed - this situation is not handle at all. However, I think if I extend this to three way merge (using base template), handling deletion of fields should be doable.

@wojtek-t  Should I proceed with this implementation? Maybe have you seen something similar in k8s code which can be used here? I tried to find any use case for dynamic client update, but without success.

PR #192, opened 2018-08-13
ClusterLoader - Cluster config

krzysied:


krzysied:
After digging into metrics, it looks like additional cluster related information are needed by cluster loader. Cluster loader should operate based on two things:
- Test config, which specifies all test steps and measurements.
- Cluster config, which specifies all information related to the cluster the test is running on.
Currently we only use kubeconfig provided as a flag, but it's not enough for metric executions. I want cluster config to have kubeconfig path as one of it's parameters, with others being: provider, host, nodeCount... etc.

If we don't want to introduce cluster config, all parameters (used by metrics) can be provided by metric params. However, the problem is that cluster related parameters (provider, host, nodeCount...) will be specified in test config which will make test config cluster depended.

@wojtek-t @shyamjvs What is your opinion?

shyamjvs:
I would suggest keeping cluster config independent of test config (as I think we shouldn't need to change the latter whenever we change for e.g provider/nodecount/etc).

clusterloader2/api/cluster.go
```diff
@@ -0,0 +1,26 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+// ClusterConfig is a structure that represents cluster description.
+type ClusterConfig struct {
```
shyamjvs: I suggest moving this json to flags instead.

krzysied: Flags added.

clusterloader2/api/cluster.go
```diff
@@ -0,0 +1,26 @@
...
+	KubeConfigPath string `json: kubeConfigPath`
+	Provider       string `json: provider`
+	Host           string `json:host`
```
shyamjvs: json: host

krzysied: done.

clusterloader2/run-e2e.sh
```diff
@@ -20,11 +20,5 @@ set -o pipefail
...
 
-if [ $# -lt 1 ]
-then
-	echo 'Test path not provided!'
-	exit 1
-fi
-
 cd ${CLUSTERLOADER_ROOT}/ && go build -o clusterloader './cmd/'
-./clusterloader --kubeconfig="${HOME}/.kube/config" --testconfig=$1  --alsologtostderr
+./clusterloader --kubeconfig="${HOME}/.kube/config" "$@"
```
shyamjvs: Maybe just put `--alsologtostderr` in the flags.

clusterloader2/api/cluster.go
```diff
@@ -0,0 +1,26 @@
+/*
```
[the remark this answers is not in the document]
shyamjvs: I would assume this is in `cmd/` because these are just flags relating to binary.

clusterloader2/api/cluster.go
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+*/
+
+package api
+
+// ClusterConfig is a structure that represents cluster description.
+type ClusterConfig struct {
+	KubeConfigPath string `json: kubeConfigPath`
+	Provider       string `json: provider`
+	Host           string `json: host`
```
shyamjvs: If you're not using any variables at the moment, just remove them. You can add later as needed,

clusterloader2/cmd/clusterloader.go
```diff
@@ -31,38 +33,70 @@ import (
...
 var (
-	kubeConfigPath = flag.String("kubeconfig", "", "path to the kubeconfig file")
-	testConfigPath = flag.String("testconfig", "", "path to the test config file")
+	clusterConfig  api.ClusterConfig
+	testConfigPath string
 )
 
-func validateFlags() []error {
+func initClusterFlags() {
+	flag.StringVar(&clusterConfig.KubeConfigPath, "kubeconfig", "", "path to the clusterconfig file")
```
shyamjvs: clusterconfig -> kubeconfig

clusterloader2/cmd/clusterloader.go
```diff
@@ -31,38 +33,70 @@ import (
...
+	flag.StringVar(&clusterConfig.Provider, "provider", "", "kubernetes provider")
+	flag.StringVar(&clusterConfig.Host, "host", "", "host ip address")
```
shyamjvs: You can copy over these flags and their descriptions from test/e2e/framework/test_context.go.

clusterloader2/api/cluster.go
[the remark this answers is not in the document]
krzysied: I thought `api` is better place, cause the clusterconfig structure is a some kind of contract. It specifies what information (config fields) can be used by tests and  measurements implementation.

clusterloader2/pkg/test/simplecontext.go
```diff
@@ -25,15 +26,17 @@ import (
 )
 
 type simpleContext struct {
+	clusterConfig      *api.ClusterConfig
```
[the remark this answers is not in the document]
krzysied: I'll prefer not to.
SimpleContext is related to cluster (params + state) while TestConfig is a description of test steps. Those are to separate concepts and I prefer to not mix them together.

clusterloader2/api/cluster.go
krzysied: I've removed all fields except `KubeConfigPath` and `NodeCount`. The other fields will be needed when scheduler latency measurement is introduced.

clusterloader2/cmd/clusterloader.go
krzysied: done.

clusterloader2/cmd/clusterloader.go
krzysied: ok

clusterloader2/api/cluster.go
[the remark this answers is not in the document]
shyamjvs: I'm not sure if you should call that as API. They're just a set of flags (similar to your test flags) that the binary is needing for some particular things. I'm more inclined to put this into cmd/ (which is where other flags are) or if you really want to separate, put it inside some 'config/' dir.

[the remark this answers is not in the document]
shyamjvs: Some e.gs of this pattern:

- https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/options/options.go (for scheduler flags)
- https://github.com/kubernetes/kubernetes/blob/master/cmd/controller-manager/app/options/generic.go (for c-m flags)

clusterloader2/cmd/clusterloader.go
```diff
@@ -31,38 +32,58 @@ import (
...
-	if *kubeConfigPath == "" {
+	if clusterConfig.KubeConfigPath == "" {
 		errList = append(errList, fmt.Errorf("no kubeconfig path specified"))
 	}
-	if *testConfigPath == "" {
+	return errList
+}
+
+func initFlags() {
+	flag.StringVar(&testConfigPath, "testconfig", "", "path to the test config file")
```
shyamjvs: nit: path -> Path (for consistency)

clusterloader2/cmd/clusterloader.go
```diff
@@ -31,38 +32,58 @@ import (
...
 func main() {
 	defer glog.Flush()
+	initFlags()
 	flag.Parse()
 	if errList := validateFlags(); len(errList) > 0 {
 		glog.Fatalf("Parsing flags error: %v", errors.NewAggregate(errList).Error())
 	}
 
-	f, err := framework.NewFramework(*kubeConfigPath, filepath.Dir(*testConfigPath))
+	if testConfigPath == "" {
```
shyamjvs: I think this condition is already checked in validateFlags() above.

clusterloader2/run-e2e.sh
```diff
@@ -20,11 +20,5 @@ set -o pipefail
...
 
-if [ $# -lt 1 ]
-then
-	echo 'Test path not provided!'
-	exit 1
-fi
-
 cd ${CLUSTERLOADER_ROOT}/ && go build -o clusterloader './cmd/'
-./clusterloader --kubeconfig="${HOME}/.kube/config" --testconfig=$1  --alsologtostderr
+./clusterloader --kubeconfig="${HOME}/.kube/config" "$@" --alsologtostderr
```
shyamjvs: I think it's safer to move `$@` to the end (as some args in it may modify/invalidate further args).

clusterloader2/pkg/test/simplecontext.go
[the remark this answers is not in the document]
shyamjvs: Ok, that sounds reasonable.

clusterloader2/api/cluster.go
shyamjvs: Sounds good. Do we even need NodeCount right now?

krzysied: Api latency measurement uses node count. If node count is provided by flag, explicit measurement parameter can be removed from `density/config.yaml`.

shyamjvs: Ah, you're right. We're using it for varying SLO for large clusters. However, given that we're going to move to the [new SLO definition](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos) which is independent of #nodes, we can probably exclude this.

clusterloader2/cmd/clusterloader.go

clusterloader2/cmd/clusterloader.go
krzysied: Good catch. Redundant check removed.

clusterloader2/run-e2e.sh
krzysied: Done.

clusterloader2/api/cluster.go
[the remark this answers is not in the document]
krzysied: I've moved this file to `pkg/config/`.
I couldn't move this file to `cmd/`, cause ClusterConfig is used test execution, which is already referenced by `cmd/`. It would cause circular dependency, I think.

krzysied:
PTAL @shyamjvs

clusterloader2/api/cluster.go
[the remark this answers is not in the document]
shyamjvs: If you move it to an independent dir `cmd/config/` (like I was suggesting above), I don't see why there should be any circular dependency. But `pkg/config/` also is fine with me.

clusterloader2/pkg/config/cluster.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// ClusterConfig is a structure that represents cluster description.
+type ClusterConfig struct {
+	KubeConfigPath string `json: kubeConfigPath`
+	NodeCount      int    `json: nodeCount`
```
wojtek-t: Why do you need NodeCount?
@shyamjvs @krzysied

shyamjvs: Discussed earlier in https://github.com/kubernetes/perf-tests/pull/192#pullrequestreview-147560699. We should eventually get rid of it.

shyamjvs: TBH my personal take on it (if we indeed need this variable) is to get this from 'list nodes' instead of this.

wojtek-t: Yes - that's what I was going to tell.
[BTW - the link above doesn't work for me]

PR #193, opened 2018-08-13
ClusterLoader - Scheduler latency metric

krzysied:


krzysied:
@shyamjvs This metric should have the same format as the old one. Doesn't calculate through put though.
I also use some weird flags like "masterName" etc. Currently these flags are set in the test config. After  this PR merges, I will add this flags to clusterloader.

clusterloader2/pkg/measurement/common/schedulerlatency.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	return responseText, nil
+}
+
+type schedulingMetrics struct {
+	PredicateEvaluationLatency  measurementutil.LatencyMetric `json:"predicateEvaluationLatency"`
+	PriorityEvaluationLatency   measurementutil.LatencyMetric `json:"priorityEvaluationLatency"`
+	PreemptionEvaluationLatency measurementutil.LatencyMetric `json:"preemptionEvaluationLatency"`
+	BindingLatency              measurementutil.LatencyMetric `json:"bindingLatency"`
+	ThroughputAverage           float64                       `json:"throughputAverage"`
```
shyamjvs: Please remove these throughput fields if you're not computing them here.

clusterloader2/pkg/measurement/common/schedulerlatency.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		metric.SetQuantile(quantile, time.Duration(int64(float64(sample.Value)*float64(time.Second))))
+	}
+	summaries = append(summaries, &result)
+
+	return summaries, nil
+}
+
+// Sends REST request to kube scheduler metrics
+func sendRestRequestToScheduler(c clientset.Interface, op, provider, host, masterName string) (string, error) {
```
shyamjvs: Btw - even though you probably copied over this file - I think you should name this function sth better than sendRestRequest.. (as that's only one of the case, the other being SSH).

clusterloader2/pkg/measurement/common/schedulerlatency.go
```diff
@@ -0,0 +1,218 @@
...
+	"k8s.io/kubernetes/pkg/util/system"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	schedulerLatencyMetricName  = "SchedulingMetrics"
+	schedulingLatencyMetricName = model.LabelValue(schedulermetric.SchedulerSubsystem + "_" + schedulermetric.SchedulingLatencyName)
+
```
shyamjvs: This blank line can be removed.

clusterloader2/pkg/measurement/common/schedulerlatency.go
```diff
@@ -0,0 +1,218 @@
...
+	singleCallTimeout = 5 * time.Minute
```
shyamjvs: What call? I think this is not too important to need being a global variable

clusterloader2/pkg/measurement/common/schedulerlatency.go
krzysied: >  I think you should name this function sth better than sendRestRequest.. (as that's only one of the case, the other being SSH).

I will not agree with the statement that this function does something more than sending request.
In case of this ssh, ssh is here only used to execute `curl -X` which is sending the request. As you can see, main point of this function is sending request, with accessing component or sshing to master being only a way to achieve this.

shyamjvs: I'm not saying it's not a request. I meant about the 'Rest' part (which is only one of the cases - the other case SSH is not a rest request).

clusterloader2/pkg/measurement/common/schedulerlatency.go
krzysied: I will keep the global variable, only name will change.

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func createSchedulerLatencyMeasurement() measurement.Measurement {
+	return &schedulerLatencyMeasurement{}
+}
+
+type schedulerLatencyMeasurement struct{}
+
+// Execute supports two actions:
+// - reset - Resets latency data on api server side.
```
shyamjvs: -> "scheduler side"
Also fix below.

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -0,0 +1,213 @@
...
+		return summaries, err
+	}
+	masterName, err := util.GetStringOrDefault(config.Params, "masterName", "")
+	if err != nil {
+		return summaries, err
+	}
+
+	switch action {
+	case "reset":
+		glog.Infof("Resetting latency metrics in schduler...")
```
shyamjvs: -> scheduler

clusterloader2/pkg/measurement/common/schedulerlatency.go
shyamjvs: I still think the name should be more clearer, like `getRawSchedulerMetrics` or sth. The current name doesn't say what kind of request it is.

krzysied: > The current name doesn't say what kind of request it is.

It takes request kind as parameter `op`.
`getRawSchedulerMetrics` - this function doesn't always get raw scheduler metrics. If you send `DELETE`, you will reset metrics instead.

clusterloader2/pkg/measurement/common/scheduler_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/scheduler_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/schedulerlatency.go
shyamjvs: Ok, then sgtm.

PR #194, opened 2018-08-13
ClusterLoader - Adding basepath template provider

krzysied:
Object templates will use relative path (from test config directory).

### DIFF
--- a/clusterloader2/pkg/config/templateprovider.go
+++ b/clusterloader2/pkg/config/templateprovider.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"path/filepath"
 	"sync"
 	"text/template"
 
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/config/templateprovider.go
```diff
@@ -29,13 +30,15 @@ import (
 // TemplateProvider provides object templates. Templates in unstructured form
 // are served by reading file from given path or by using cache if available.
 type TemplateProvider struct {
+	basepath      string
```
shyamjvs: Please add newline above the lock's block.

clusterloader2/pkg/test/test.go
```diff
@@ -35,7 +35,7 @@ var (
 )
 
 // RunTest runs test based on provided test configuration.
-func RunTest(f *framework.Framework, conf *api.Config) []error {
+func RunTest(basepath string, f *framework.Framework, conf *api.Config) []error {
```
shyamjvs: Could you move this `basepath` parameter to framework instead of passing it as separate argument in so many functions? I think it also makes sense to have it in there.

clusterloader2/pkg/config/templateprovider.go
krzysied: done.

clusterloader2/pkg/test/test.go
krzysied: Good idea. Done.

PR #196, opened 2018-08-14
ClusterLoader - Density test

krzysied:


clusterloader2/testing/density/config.json
```diff
@@ -0,0 +1,206 @@
+{
```
wojtek-t: Can we please use yamls instead of jsons?

They are much more concise (and a bit more readable in my opinion)

shyamjvs: Can't agree more :)

shyamjvs: Also, it'll probably make this file half the size.

krzysied: Converted to yaml.
IMHO, json looked better.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,128 @@
...
+- measurements:
+  - Identifier: TimerGlobal
+    Method: Timer
+    Params:
+      action: start
+  - Identifier: TimerCreate
+    Method: Timer
+    Params:
+      action: start
+  - Identifier: SchedulerLatency
```
shyamjvs: I would suggest removing this and adding later when the PR for SchedulerLatency is merged.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,128 @@
...
+      action: reset
+      provider: gce
+      masterName: e2e-ksiedlecki2-master
+  - Identifier: APICallLatency
+    Method: APICallLatency
+    Params:
+      action: reset
+- phases:
+  - namespaceRange:
+      min: 0
```
shyamjvs: You can change in separate PR. But I think it's better user experience if the indexing for namespace range is 1-based rather than 0-based.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,128 @@
...
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        Replicas: 1
+- measurements:
+  - Identifier: WaitForRunningPods2
+    Method: WaitForRunningPods
+    Params:
+      desiredPodCount: 93
+      labelSelector: phase = phase1
+- measurements:
```
wojtek-t: IIUC we are still not measuring pod startup time, right?

krzysied: You are right. It's not implemented yet.

clusterloader2/testing/density/config.yaml
krzysied: done.

clusterloader2/testing/density/config.yaml
krzysied: done.

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,29 @@
...
+        name: {{.Basename}}
+        ports:
+        - containerPort: 80
+          protocol: TCP
+        resources:
+          requests:
+            cpu: 10m
+            memory: "26214400"
+      restartPolicy: Always
+      terminationGracePeriodSeconds: 1
```
shyamjvs: Maybe just let the default kick in instead of setting it?

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,29 @@
...
+        imagePullPolicy: IfNotPresent
+        name: {{.Basename}}
+        ports:
+        - containerPort: 80
+          protocol: TCP
+        resources:
+          requests:
+            cpu: 10m
+            memory: "26214400"
+      restartPolicy: Always
```
shyamjvs: You don't need this. It's the default - https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - image: gcr.io/kubernetes-e2e-test-images/serve-hostname-amd64:1.0
+        imagePullPolicy: IfNotPresent
+        name: {{.Basename}}
+        ports:
+        - containerPort: 80
+          protocol: TCP
+        resources:
+          requests:
+            cpu: 10m
+            memory: "26214400"
```
shyamjvs: Can you make it more understandable? Sth like `10M` should be a good value.

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,29 @@
...
+  selector:
+    name: {{.Basename}}
+  template:
+    metadata:
+      labels:
+        name: {{.Basename}}
+        phase: phase1
+    spec:
+      containers:
+      - image: gcr.io/kubernetes-e2e-test-images/serve-hostname-amd64:1.0
```
shyamjvs: Maybe change this to pause pod (like what we use in our standard density test)?  That should be `k8s.gcr.io/pause:3.1`

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  labels:
```
shyamjvs: Why do you need this label?

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: {{.Basename}}
+spec:
+  replicas: {{.Replicas}}
+  selector:
+    name: {{.Basename}}
+  template:
+    metadata:
+      labels:
+        name: {{.Basename}}
+        phase: phase1
```
shyamjvs: Can we use a better label (for e.g `app: pause`)? As I think this is not really tied to phases.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+    Method: Timer
+    Params:
+      action: stop
+      label: Pods deletion
+  - Identifier: TimerGlobal
+    Method: Timer
+    Params:
+      action: stop
+      label: Global execution
+tuningSets:
```
shyamjvs: Can you move tuningsets to above the test phases? So you know them already by the time you start reading the phases?

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+- name: TuningSet100
+  initialDelay: 0
```
shyamjvs: Do we really need to specify the 0 value? Isn't it a default?

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+    Params:
+      action: stop
+      label: Pods deletion
+  - Identifier: TimerGlobal
+    Method: Timer
+    Params:
+      action: stop
+      label: Global execution
+tuningSets:
+- name: TuningSet100
```
shyamjvs: I'd suggest a better name (like `Uniform100qps`) so that it's obvious to understand when people look at it in  phase definitions. `TuningSet` is obvious and doesn't have to be in the name.

clusterloader2/testing/density/rc.yaml
```diff
@@ -0,0 +1,25 @@
...
+  selector:
+    name: {{.Basename}}
+  template:
+    metadata:
+      labels:
+        name: {{.Basename}}
+        app: pause
+    spec:
+      containers:
+      - image: gcr.io/kubernetes-e2e-test-images/pause:3.1
```
shyamjvs: I think we're just using `k8s.gcr.io/pause:3.1` in our tests. Not sure what's the difference though.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+    Params:
+      action: reset
+- phases:
+  - namespaceRange:
+      min: 1
+      max: 3
+    replicasPerNamespace: 1
+    tuningSet: TuningSet100
+    objectBundle:
+    - basename: rc1
```
shyamjvs: Again here I suggest giving a more meaningful name to what kind of rc this is. For e.g `saturationRC` (as we create these RCs to saturate the cluster with many many pods).

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        Replicas: 30
+- measurements:
+  - Identifier: WaitForRunningPods1
```
shyamjvs: `WaitForSaturationPodsRunning`?

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+      desiredPodCount: 90
+      labelSelector: app = pause
+      timeout: 2m
+- measurements:
+  - Identifier: TimerCreate
+    Method: Timer
+    Params:
+      action: stop
+      label: Pods creation
+  - Identifier: TimerAdditionals
```
shyamjvs: Also please give more meaningful names to these timers - so that it's easy to understand what they're for.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+    Params:
+      action: start
+- phases:
+  - namespaceRange:
+      min: 1
+      max: 3
+    replicasPerNamespace: 1
+    tuningSet: TuningSet5
+    objectBundle:
+    - basename: rc1-add
```
shyamjvs: `latencyPodRC`

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        Replicas: 1
+- measurements:
+  - Identifier: APICallLatency
+    Method: APICallLatency
+    Params:
+      action: gather
+      nodeCount: 4
```
shyamjvs: Isn't it 3 nodes (based on the other values in this file)?

shyamjvs:
I've left few comments.
Besides that, one thing we should start thinking about is how to implement templating for cluster-loader config (right now we only allow it for object yamls). This is because a lot of numbers in the config are parameters depending on the no. of nodes, and ideally we shouldn't need to write different config for each different value of #nodes.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,114 @@
...
+- name: Uniform5qps
+  qpsLoad:
+    qps: 5
+steps:
+- measurements:
+  - Identifier: TimerGlobal
+    Method: Timer
+    Params:
+      action: start
+  - Identifier: TimerCreateSaturationPods
```
shyamjvs: nit: I'd suggest putting 'Timer' at the end of the name? (here and elsewhere)

krzysied: done.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,114 @@
...
+    objectBundle:
+    - basename: latency-pod-rc
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        Replicas: 1
+- measurements:
+  - Identifier: WaitForLatencyPodRunning
+    Method: WaitForRunningPods
+    Params:
+      desiredPodCount: 93
```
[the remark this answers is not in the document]
krzysied: done.

clusterloader2/testing/density/config.yaml
```diff
@@ -0,0 +1,116 @@
...
+      min: 1
+      max: 3
+    replicasPerNamespace: 1
+    tuningSet: Uniform100qps
+    objectBundle:
+    - basename: saturation-rc
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        Replicas: 30
+        PodKind: staturationPod
```
shyamjvs: saturation

krzysied: fixed

shyamjvs:
/lgtm

Nice milestone!

PR #201, opened 2018-08-17
ClusterLoader- Adding summary logging

krzysied:


clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -71,10 +71,23 @@ func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) []error
 			// index is created to make i value unchangeable during thread execution.
 			index := i
 			wg.Start(func() {
-				if err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params); err != nil {
+				summary, err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params)
+				if err != nil {
```
shyamjvs: Shouldn't we be printing summary even if measurement failed with error?

clusterloader2/pkg/measurement/common/apicalllatency.go
```diff
@@ -64,26 +64,26 @@ type apiCallLatencyMeasurement struct{}
 // Execute supports two actions:
 // - reset - Resets latency data on api server side.
 // - gather - Gathers and prints current api server latency data.
-func (*apiCallLatencyMeasurement) Execute(config *measurement.MeasurementConfig) error {
+func (*apiCallLatencyMeasurement) Execute(config *measurement.MeasurementConfig) (measurement.Summary, error) {
```
shyamjvs: Let's make the return value an array of Summaries instead of a single one - to allow generating multiple summaries within a measurement.

clusterloader2/pkg/measurement/interface.go
```diff
@@ -34,7 +34,13 @@ type MeasurementConfig struct {
...
-	Execute(config *MeasurementConfig) error
+	Execute(config *MeasurementConfig) (Summary, error)
 }
 
 type createMeasurementFunc func() Measurement
+
+// Summary represenst result of specific measurement.
+type Summary interface {
+	GetMeasurementName() string
+	PrintMeasurementResults() (string, error)
```
shyamjvs: Can we just call this `PrintSummary`? (as measurement doesn't seem too relevant here)

clusterloader2/pkg/measurement/interface.go
```diff
@@ -34,7 +34,13 @@ type MeasurementConfig struct {
...
 type Measurement interface {
-	Execute(config *MeasurementConfig) error
+	Execute(config *MeasurementConfig) (Summary, error)
 }
 
 type createMeasurementFunc func() Measurement
+
+// Summary represenst result of specific measurement.
+type Summary interface {
+	GetMeasurementName() string
```
shyamjvs: Can we call this `SummaryKind`?

clusterloader2/pkg/test/simpletestexecutor.go
krzysied: Assuming that now function returns multiple summaries, summaries will be printed even if error occurs.

clusterloader2/pkg/measurement/common/apicalllatency.go
krzysied: done.

clusterloader2/pkg/measurement/interface.go
krzysied: done.

clusterloader2/pkg/measurement/interface.go
krzysied: I've changed it to `SummaryName`.

PR #203, opened 2018-08-21
Adding echo to run-e2e script

krzysied:
For debugging purpose.

### DIFF
--- a/run-e2e.sh
+++ b/run-e2e.sh
@@ -20,6 +20,8 @@ set -o nounset
 set -o pipefail
 
 PERFTEST_ROOT=$(dirname "${BASH_SOURCE}")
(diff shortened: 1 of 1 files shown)


run-e2e.sh
```diff
@@ -29,6 +29,7 @@ case "$1" in
     ;;
   cluster-loader2 )
     #CLUSTERLOADER2
+    echo "${PERFTEST_ROOT}/clusterloader2 && ./run-e2e.sh ${@:2}"
```
shyamjvs: Maybe you should add an echo above the switch case (in case we aren't even entering here).

krzysied: Done.

PR #204, opened 2018-08-22
ClusterLoader - Adding template functions

krzysied:
Adding generating random int functions.

### DIFF
--- a/clusterloader2/pkg/config/templatefunctions.go
+++ b/clusterloader2/pkg/config/templatefunctions.go
@@ -1,0 +1,49 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"math/rand"
+	"text/template"
+	"time"
+)
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
+// GetFuncs returns map of names to functions, that are supported by template provider.
+func GetFuncs() template.FuncMap {
+	return template.FuncMap{
+		"RandInt":      randInt,
+		"RandIntRange": randIntRange,
+	}
+}
+
+// randInt returns pseudo-random int in [0, i].
+func randInt(i int) int {
+	return rand.Intn(i + 1)
+}
+
+// randIntRange returns pseudo-random int in [i, j].
+// If i >= j then i is returned.
+func randIntRange(i, j int) int {
+	if i >= j {
+		return i
+	}
+	return i + rand.Intn(j-i+1)
+}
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/config/templatefunctions.go
```diff
@@ -0,0 +1,44 @@
...
+func randInt(i int) int {
+	return rand.Intn(i)
+}
+
+func randIntRange(i, j int) int {
+	if i >= j {
+		return i
+	}
+	r := i + rand.Intn(j-i)
+	fmt.Printf("Rand: %v\n", r)
```
krzysied: My mistake - I'm going to remove this printf.

clusterloader2/pkg/config/templatefunctions.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"math/rand"
+	"text/template"
+)
+
+// GetFuncs returns map of names to functions, that are supported by template provider.
+func GetFuncs() template.FuncMap {
+	return template.FuncMap{
+		"randInt":      randInt,
```
shyamjvs: I'd suggest making these function names capitals (to show clearly they're constants).

krzysied: Done.

clusterloader2/pkg/config/templatefunctions.go
```diff
@@ -0,0 +1,41 @@
...
+
+// GetFuncs returns map of names to functions, that are supported by template provider.
+func GetFuncs() template.FuncMap {
+	return template.FuncMap{
+		"RandInt":      randInt,
+		"RandIntRange": randIntRange,
+	}
+}
+
+func randInt(i int) int {
```
shyamjvs: Could you add comment above each of these 2 functions what's the range they create randoms from? (for e.g 0-based or 1-based, open or closed interval)

krzysied: Comments added.

clusterloader2/pkg/config/templatefunctions.go
```diff
@@ -0,0 +1,44 @@
...
+		"RandIntRange": randIntRange,
+	}
+}
+
+// randInt returns pseudo-random int in [0, i).
+func randInt(i int) int {
+	return rand.Intn(i)
+}
+
+// randIntRange returns pseudo-random int in [i, j).
```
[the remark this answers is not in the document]
krzysied: Should I also change the `RandInt` function or only `RandIntRange`?
Btw, using [..) set is common programming practice - I assume that people with expect this here.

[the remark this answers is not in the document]
shyamjvs: I'd suggest changing both (for consistency).

> Btw, using [..) set is common programming practice - I assume that people with expect this here.

Yeah, that's a matter of convention (you're right that few pkgs assume open interval). But I think this'll be more convenient when we calculate and put the limits in the config.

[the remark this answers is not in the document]
krzysied: OK. Done.

clusterloader2/pkg/config/templatefunctions.go
```diff
@@ -0,0 +1,49 @@
...
+func GetFuncs() template.FuncMap {
+	return template.FuncMap{
+		"RandInt":      randInt,
+		"RandIntRange": randIntRange,
+	}
+}
+
+// randInt returns pseudo-random int in [0, i].
+func randInt(i int) int {
+	return rand.Intn(i + 1)
```
shyamjvs: 'i + 1' has spaces but similar expression in below function doesn't. I thought gofmt should be failing on this.

PR #205, opened 2018-08-22
ClusterLoader - Load test

krzysied:
Adding 3-node load test.
New measurement `WaitForRunningRCs` added.

### DIFF
--- a/clusterloader2/pkg/config/template_provider.go
+++ b/clusterloader2/pkg/config/template_provider.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"regexp"
 	"sync"
 	"text/template"
 
@@ -33,7 +34,10 @@ import (
 type TemplateProvider struct {
 	basepath string
 
-	lock          sync.RWMutex
+	binLock  sync.RWMutex
+	binCache map[string][]byte
+
+	templateLock  sync.RWMutex
 	templateCache map[string]*template.Template
 }
 
@@ -41,23 +45,62 @@ type TemplateProvider struct {
 func NewTemplateProvider(basepath string) *TemplateProvider {
 	return &TemplateProvider{
 		basepath:      basepath,
+		binCache:      make(map[string][]byte),
 		templateCache: make(map[string]*template.Template),
 	}
 }
 
-func (tp *TemplateProvider) getRawTemplate(path string) (*template.Template, error) {
-	tp.lock.RLock()
-	raw, exists := tp.templateCache[path]
-	tp.lock.RUnlock()
+func (tp *TemplateProvider) getRaw(path string) ([]byte, error) {
+	tp.binLock.RLock()
+	bin, exists := tp.binCache[path]
+	tp.binLock.RUnlock()
 	if !exists {
-		tp.lock.Lock()
-		defer tp.lock.Unlock()
+		tp.binLock.Lock()
+		defer tp.binLock.Unlock()
+		// Recheck condition.
+		bin, exists = tp.binCache[path]
+		if !exists {
+			var err error
+			bin, err = ioutil.ReadFile(filepath.Join(tp.basepath, path))
+			if err != nil {
+				return []byte{}, fmt.Errorf("reading error: %v", err)
+			}
+			tp.binCache[path] = bin
+		}
+	}
+	return bin, nil
+}
+
+// RawToObject creates object from file specified by the given path
+// or uses cached object if available.
+func (tp *TemplateProvider) RawToObject(path string) (*unstructured.Unstructured, error) {
+	bin, err := tp.getRaw(path)
+	if err != nil {
+		return nil, err
+	}
+	// Removing all placeholder from template.
+	// This needs to be done due to placeholders not being valid yaml.
+	r, err := regexp.Compile("\\{\\{.*\\}\\}")
+	if err != nil {
+		return nil, fmt.Errorf("regexp creation error: %v", err)
+	}
+	bin = r.ReplaceAll(bin, []byte{})
+	return ConvertToObject(bin)
+}
+
+func (tp *TemplateProvider) getRawTemplate(path string) (*template.Template, error) {
+	tp.templateLock.RLock()
+	raw, exists := tp.templateCache[path]
+	tp.templateLock.RUnlock()
+	if !exists {
+		tp.templateLock.Lock()
+		defer tp.templateLock.Unlock()
 		// Recheck condition.
 		raw, exists = tp.templateCache[path]
 		if !exists {
-			bin, err := ioutil.ReadFile(filepath.Join(tp.basepath, path))
+			bin, err := tp.getRaw(path)
 			if err != nil {
-				return nil, fmt.Errorf("reading error: %v", err)
+				return nil, err
 			}
 			raw = template.New("").Funcs(GetFuncs())
 			raw, err = raw.Parse(string(bin))
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -1,0 +1,236 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+TODO(krzysied): Replace ClientSet interface with dynamic client.
+This will allow more generic approach.
+*/
+
+package common
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+	batch "k8s.io/api/batch/v1"
+	"k8s.io/api/core/v1"
+	extensions "k8s.io/api/extensions/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	defaultWaitForRCPodsTimeout      = 60 * time.Second
+	defaultWaitForRCPodsInterval     = 5 * time.Second
+	waitForControlledPodsRunningName = "WaitForControlledPodsRunning"
+)
+
+func init() {
+	measurement.Register(waitForControlledPodsRunningName, createWaitForControlledPodsRunningMeasurement)
+}
+
+func createWaitForControlledPodsRunningMeasurement() measurement.Measurement {
+	return &waitForControlledPodsRunningMeasurement{}
+}
+
+type waitForControlledPodsRunningMeasurement struct{}
+
+// Execute waits until all specified RCs have all pods running or until timeout happens.
+// RCs can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	var summaries []measurement.Summary
+	kind, err := util.GetString(config.Params, "kind")
+	if err != nil {
+		return summaries, err
+	}
+	namespace, err := util.GetStringOrDefault(config.Params, "namespace", metav1.NamespaceAll)
+	if err != nil {
+		return summaries, err
+	}
+	labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")
+	if err != nil {
+		return summaries, err
+	}
+	fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")
+	if err != nil {
+		return summaries, err
+	}
+	timeout, err := util.GetDurationOrDefault(config.Params, "timeout", defaultWaitForRCPodsTimeout)
+	if err != nil {
+		return summaries, err
+	}
+
+	runtimeObjectsList, err := getRuntimeObjectsForKind(config.ClientSet, kind, namespace, labelSelector, fieldSelector)
+	if err != nil {
+		return summaries, err
+	}
+	runningRuntimeObjects := 0
+	for start := time.Now(); time.Since(start) < timeout; time.Sleep(defaultWaitForRCPodsInterval) {
+		if runningRuntimeObjects >= len(runtimeObjectsList) {
+			return summaries, nil
+		}
+
+		if err := waitForRuntimeObject(config.ClientSet, runtimeObjectsList[runningRuntimeObjects], timeout-time.Since(start)); err != nil {
+			return summaries, fmt.Errorf("waiting for %v error: %v", kind, err)
+		}
+		runningRuntimeObjects++
+		if namespace == metav1.NamespaceAll {
+			glog.Infof("%s: %d / %d %ss with all pods running", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
+		} else {
+			glog.Infof("%s: %v: %d / %d %ss with all pods running", waitForControlledPodsRunningName, namespace, runningRuntimeObjects, len(runtimeObjectsList), kind)
+		}
+	}
+
+	return summaries, fmt.Errorf("timeout while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods", kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects)
+}
+
+func waitForRuntimeObject(c clientset.Interface, obj runtime.Object, timeout time.Duration) error {
+	runtimeObjectNamespace, err := getNamespaceFromRuntimeObject(obj)
+	if err != nil {
+		return err
+	}
+	runtimeObjectSelector, err := getSelectorFromRuntimeObject(obj)
+	if err != nil {
+		return err
+	}
+	runtimeObjectReplicas, err := getReplicasFromRuntimeObject(obj)
+	if err != nil {
+		return err
+	}
+
+	err = waitForPods(c, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), timeout)
+	if err != nil {
+		return fmt.Errorf("waiting pods error: %v", err)
+	}
+	return nil
+}
+
+func getRuntimeObjectsForKind(c clientset.Interface, kind, namespace, labelSelector, fieldSelector string) ([]runtime.Object, error) {
+	listOpts := metav1.ListOptions{
+		LabelSelector: labelSelector,
+		FieldSelector: fieldSelector,
+	}
+	runtimeObjectsList := make([]runtime.Object, 0)
+	switch kind {
+	case "ReplicationController":
+		list, err := c.CoreV1().ReplicationControllers(namespace).List(listOpts)
+		if err != nil {
+			return runtimeObjectsList, err
+		}
+		for i := range list.Items {
+			runtimeObjectsList = append(runtimeObjectsList, &list.Items[i])
+		}
+		return runtimeObjectsList, nil
+	case "ReplicaSet":
+		list, err := c.ExtensionsV1beta1().ReplicaSets(namespace).List(listOpts)
+		if err != nil {
+			return runtimeObjectsList, err
+		}
+		for i := range list.Items {
+			runtimeObjectsList = append(runtimeObjectsList, &list.Items[i])
+		}
+		return runtimeObjectsList, nil
+	case "Deployment":
+		list, err := c.ExtensionsV1beta1().Deployments(namespace).List(listOpts)
+		if err != nil {
+			return runtimeObjectsList, err
+		}
+		for i := range list.Items {
+			runtimeObjectsList = append(runtimeObjectsList, &list.Items[i])
+		}
+		return runtimeObjectsList, nil
+	case "DaemonSet":
+		list, err := c.ExtensionsV1beta1().DaemonSets(namespace).List(listOpts)
+		if err != nil {
+			return runtimeObjectsList, err
+		}
+		for i := range list.Items {
+			runtimeObjectsList = append(runtimeObjectsList, &list.Items[i])
+		}
+		return runtimeObjectsList, nil
+	case "Job":
+		list, err := c.BatchV1().Jobs(namespace).List(listOpts)
+		if err != nil {
+			return runtimeObjectsList, err
+		}
+		for i := range list.Items {
+			runtimeObjectsList = append(runtimeObjectsList, &list.Items[i])
+		}
+		return runtimeObjectsList, nil
+	default:
+		return runtimeObjectsList, fmt.Errorf("unsupported kind when getting runtime object: %v", kind)
+	}
+}
+
+func getNamespaceFromRuntimeObject(obj runtime.Object) (string, error) {
+	metaObjectAccessor, ok := obj.(metav1.ObjectMetaAccessor)
+	if !ok {
+		return "", fmt.Errorf("unsupported kind when getting namespace: %v", obj)
+	}
+	return metaObjectAccessor.GetObjectMeta().GetNamespace(), nil
+}
+
+func getSelectorFromRuntimeObject(obj runtime.Object) (labels.Selector, error) {
+	switch typed := obj.(type) {
+	case *v1.ReplicationController:
+		return labels.SelectorFromSet(typed.Spec.Selector), nil
+	case *extensions.ReplicaSet:
+		return metav1.LabelSelectorAsSelector(typed.Spec.Selector)
+	case *extensions.Deployment:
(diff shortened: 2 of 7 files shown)


clusterloader2/pkg/measurement/common/waitforrcs.go
```diff
@@ -0,0 +1,92 @@
+/*
```
shyamjvs: Could you rename this file to `wait_for_rc_pods.go` to make it more clearer?
Also rename the above file to `wait_for_pods.go`.

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,147 @@
...
+      action: reset
+# Create RCs
+- phases:
+  - namespaceRange:
+      min: 1
+      max: 3
+    replicasPerNamespace: 1
+    tuningSet: Uniform100qps
+    objectBundle:
+    - basename: saturation-service
```
shyamjvs: I don't think you need to call this `saturation-service`. 'service' is enough I think.

shyamjvs: Same for basename of RC.

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,147 @@
...
+      objectTemplatePath: service.yaml
+    - basename: saturation-rc
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        ReplicasMin: 30
+        ReplicasMax: 30
+- measurements:
+  - Identifier: WaitForRunningSaturationRCs
```
shyamjvs: -> WaitForRCsRunning

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,147 @@
...
+  - Identifier: ScaleRCsTimer1
+    Method: Timer
+    Params:
+      action: stop
+      label: Scaling RCs - 1
+  - Identifier: ScaleRCsTimer2
+    Method: Timer
+    Params:
+      action: start
+# Scale RCs 1
```
shyamjvs: -> Scaling RCs first time

clusterloader2/testing/load/rc.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: {{.Basename}}
```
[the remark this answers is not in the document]
krzysied: PR with other files changed: #206

clusterloader2/pkg/measurement/common/waitforrcs.go
krzysied: This will be done in separate PR. The problem is that if I'm changing this name, I should also rename all other files to keep same convention.

krzysied:
/hold
Waiting for #206 .

clusterloader2/pkg/measurement/common/waitforrcs.go
shyamjvs: Sure. Though I don't think it's needed to rename other files (just these 2) to make the names more clearer. I'm not sure of a convention either (and have seen files named either ways).

krzysied: TBH, I will be very unhappy to have some files with `_` and some without. I think that we should stick to only convention for all the files.

shyamjvs: Sure, I'm ok with `_` everywhere then.

krzysied: Changing names PR: #207

clusterloader2/testing/load/rc.yaml
```diff
@@ -0,0 +1,24 @@
...
+    kind: loadRC
+spec:
+  replicas: {{RandIntRange .ReplicasMin .ReplicasMax}}
+  selector:
+    name: {{.Basename}}
+  template:
+    metadata:
+      labels:
+        name: {{.Basename}}
+        group: group_{{.Index}}
```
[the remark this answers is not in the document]
krzysied: I've removed `group` label. I'm using rc name label only.

clusterloader2/testing/load/service.yaml
```diff
@@ -0,0 +1,11 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: {{.Basename}}
+spec:
+  selector:
+    group: group_{{.Index}}
```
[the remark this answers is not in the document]
krzysied: Changed to rc name. Still needs to be passed by mapping, though...

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,147 @@
...
+      action: start
+  - Identifier: CreateSaturationPodsTimer
+    Method: Timer
+    Params:
+      action: start
+  - Identifier: APICallLatency
+    Method: APICallLatency
+    Params:
+      action: reset
+# Create RCs
```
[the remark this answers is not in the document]

clusterloader2/testing/load/config.yaml
krzysied: Done: `saturation-service` -> `service`.

clusterloader2/testing/load/config.yaml
krzysied: `Saturation` phrase removed from the config.

clusterloader2/testing/load/config.yaml

krzysied:
/hold cancel
All required changes are done.
@shyamjvs PTAL

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type waitForRunningRCsMeasurement struct{}
+
+// Execute waits until all specified RCs have all pods running or until timeout happens.
+// RCs can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+func (*waitForRunningRCsMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	var summaries []measurement.Summary
+	namespace, err := util.GetStringOrDefault(config.Params, "namespace", metav1.NamespaceAll)
```
shyamjvs: I don't think you need to have an option for namespace separately. It can be done through field selector on namespace.

shyamjvs: (Simiarly for wait_for_pods.go)

shyamjvs: Actually, after thinking again, maybe it's better to keep it as it offers some convenience.

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
```diff
@@ -0,0 +1,92 @@
...
+		LabelSelector: labelSelector,
+		FieldSelector: fieldSelector,
+	}
+	rcsList, err := config.ClientSet.CoreV1().ReplicationControllers(namespace).List(listOpts)
+	if err != nil {
+		return summaries, err
+	}
+	rcs := rcsList.Items
+	runningRCs := 0
+	for start := time.Now(); time.Since(start) < timeout; time.Sleep(5 * time.Second) {
```
shyamjvs: Can you make this `5 sec` value a global variable instead of hiding it within code?

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")
+	if err != nil {
+		return summaries, err
+	}
+	fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")
+	if err != nil {
+		return summaries, err
+	}
+	timeout, err := util.GetDurationOrDefault(config.Params, "timeout", 60*time.Second)
```
shyamjvs: Can you make this 60 sec default also a global variable?

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
shyamjvs: (same in wait_for_pods.go)

krzysied: I've added `globals.go` to `measurement/util` and moved this time consts there.

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
krzysied: Done.

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
shyamjvs: I would really suggest keeping these globals together with the measurement file instead of collecting them together and putting them somewhere else. This'll avoid adding extra dependency for people while adding new measurements and it makes more sense to have the globals relevant for a file in that scope.

krzysied: Ok. Done.

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
```diff
@@ -0,0 +1,97 @@
...
+const (
+	defaultWaitForRCPodsTimeout  = 60 * time.Second
+	defaultWaitForRCPodsInterval = 5 * time.Second
+)
+
+func init() {
+	measurement.Register("WaitForRunningRCs", createWaitForRunningRCsMeasurement)
+}
+
+func createWaitForRunningRCsMeasurement() measurement.Measurement {
```
shyamjvs: Can you make this file generic to work with different kinds of controllers (e.g RC, RS, deployment, job, etc)? Because we don't want to have one copy of this implementation for each of them.

Here's a reference to how we're doing it in a generic way - https://github.com/kubernetes/kubernetes/blob/7b2fa0a21f2f82dcc9b424d8f0f634e0263cb403/test/e2e/framework/util.go#L2892

shyamjvs: Also after making that change, I suggest renaming this file to `wait_for_controlled_pods.go`.

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,155 @@
...
+  - Identifier: ScaleRCsTimer1
+    Method: Timer
+    Params:
+      action: stop
+      label: Scaling RCs first time
+  - Identifier: ScaleRCsTimer2
+    Method: Timer
+    Params:
+      action: start
+# Scale RCs second time
```
shyamjvs: You can get rid of the second round of scaling which we just changed - https://github.com/kubernetes/kubernetes/pull/67769

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,155 @@
...
+  - Identifier: APICallLatency
+    Method: APICallLatency
+    Params:
+      action: gather
+      nodeCount: 3
+  - Identifier: DeleteAllPodsTimer
+    Method: Timer
+    Params:
+      action: stop
+      label: All pods deletion
```
shyamjvs: Just `Pods deletion` maybe?

clusterloader2/testing/load/rc.yaml
```diff
@@ -0,0 +1,23 @@
...
+  selector:
+    name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        name: {{.Name}}
+    spec:
+      containers:
+      - image: k8s.gcr.io/pause:3.1
+        imagePullPolicy: IfNotPresent
```
shyamjvs: This is a [default](https://kubernetes.io/docs/concepts/containers/images/#updating-images), please remove it. Also from density test's yaml.

clusterloader2/pkg/measurement/common/wait_for_rc_pods.go
krzysied: Done.

clusterloader2/testing/load/config.yaml

clusterloader2/testing/load/config.yaml

clusterloader2/testing/load/rc.yaml
krzysied: Done.

clusterloader2/testing/load/service.yaml
```diff
@@ -0,0 +1,11 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: {{.Name}}
+spec:
+  selector:
+    name: {{.RCBasename}}-{{.Index}}
+  ports:
+  - protocol: TCP
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,155 @@
...
+    Method: Timer
+    Params:
+      action: start
+# Scale RCs first time
+- phases:
+  - namespaceRange:
+      min: 1
+      max: 3
+    replicasPerNamespace: 1
+    tuningSet: Uniform5qps
```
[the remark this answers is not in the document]

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,155 @@
...
+    Method: WaitForRunningRCs
+    Params:
+      labelSelector: kind = loadRC
+      timeout: 2m
+- measurements:
+  - Identifier: ScaleRCsTimer1
+    Method: Timer
+    Params:
+      action: stop
+      label: Scaling RCs first time
```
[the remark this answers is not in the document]

krzysied:
I've reimplemented the `wait_for_rc_pods.go`. Currently it is called `wait_for_controlled_pods.go` and has support for `ReplicationController`, `ReplicaSet`, `Deployment`, `DaemonSet` and `Job`.
@shyamjvs PTAL

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -0,0 +1,233 @@
...
+
+func getRuntimeObjectForKind(c clientset.Interface, kind, namespace, labelSelector, fieldSelector string) ([]runtime.Object, error) {
+	listOpts := metav1.ListOptions{
+		LabelSelector: labelSelector,
+		FieldSelector: fieldSelector,
+	}
+	runtimeObjectsList := make([]runtime.Object, 0)
+	switch kind {
+	case "ReplicationController":
+		list, err := c.CoreV1().ReplicationControllers(namespace).List(listOpts)
```
wojtek-t: Why can't we use dynamic client as everywhere else?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -0,0 +1,233 @@
...
+		return runtimeObjectsList, nil
+	default:
+		return runtimeObjectsList, fmt.Errorf("unsupported kind when getting runtime object: %v", kind)
+	}
+}
+
+func getNamespaceRuntimeObject(obj runtime.Object) (string, error) {
+	switch typed := obj.(type) {
+	case *v1.ReplicationController:
+		return typed.ObjectMeta.Namespace, nil
```
wojtek-t: use MetaAccessor - you won't need this ugly switch,

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: I guess we can. I just need to add dynamic client to measurement config.
However, I definitely don't want to this in this PR.

krzysied:
@wojtek-t @shyamjvs PTAL

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,125 @@
...
+    tuningSet: Uniform100qps
+    objectBundle:
+    - basename: service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        RCBasename: rc
+    - basename: rc
+      objectTemplatePath: rc.yaml
+      templateFillMap:
+        ReplicasMin: 30
```
shyamjvs: Do you really need to provide this replicas info for RC when deleting them? For deleting, don't you just need the name and namespace of the object?

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,125 @@
+name: Load
+automanagedNamespaces: 3
+tuningSets:
+- name: Uniform100qps
+  qpsLoad:
+    qps: 100
+- name: Uniform1.4qps
+  qpsLoad:
+    qps: 1.4
```
shyamjvs: Have you checked this '1.4' qps value will work? From [the code](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/ticker/timer.go#L65), you seem to be doing:

```
delay = time.Duration(int(float64(time.Second) / ts.QpsLoad.Qps))
```

And when I run the following go program:

```
package main

import (
	"fmt"
	"time"
)
	
func main() {
	fmt.Println("Duration: %v", time.Duration(int(float64(time.Second) / 1.4)))
}
```

it fails with:

```
prog.go:9:47: constant truncated to integer
```

clusterloader2/testing/density/config.yaml
```diff
@@ -20,6 +20,7 @@ steps:
     Method: APICallLatency
     Params:
       action: reset
+# Creating saturation pods
```
shyamjvs: Can you use simple present tense (i.e "Create") to be consistent with your load test comments?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -0,0 +1,224 @@
...
+		for i := range list.Items {
+			runtimeObjectsList = append(runtimeObjectsList, &list.Items[i])
+		}
+		return runtimeObjectsList, nil
+	default:
+		return runtimeObjectsList, fmt.Errorf("unsupported kind when getting runtime object: %v", kind)
+	}
+}
+
+func getNamespaceRuntimeObject(obj runtime.Object) (string, error) {
```
shyamjvs: -> getNamespace"From"RuntimeObject

clusterloader2/testing/load/config.yaml
krzysied: It's true that I only need name and namespace, but...
When template is read, all template functions are evaluated. If there is no parameter provided, the function will panic. That's why parameters are always required.

clusterloader2/testing/load/config.yaml
krzysied: I tested my code and it worked :)

In your case, you are using `1.4` which is constant and `time.Second` which also is constant. I guess golang is unhappy with changing casting constant to different types ¯\_(ツ)_/¯
However, if you extract 1.4 to separate variable and you use this variable in the evaluation, it works fine.

shyamjvs: Ok, sgtm.

clusterloader2/testing/load/config.yaml
shyamjvs: Right. But in principle deletions are sth that shouldn't need any more information except probably the template.

clusterloader2/testing/load/service.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Service
```
shyamjvs: nit: Can you move the kind below the apiVersion?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -0,0 +1,224 @@
...
+func createWaitForControlledPodsRunningMeasurement() measurement.Measurement {
+	return &waitForControlledPodsRunningMeasurement{}
+}
+
+type waitForControlledPodsRunningMeasurement struct{}
+
+// Execute waits until all specified RCs have all pods running or until timeout happens.
+// RCs can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
```
shyamjvs: Can you refactor this function into at least 2 parts (instead of single long one) - one for getting list of RCs and one for waiting for them to run?

clusterloader2/testing/load/service.yaml
krzysied: Done.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/testing/density/config.yaml

clusterloader2/testing/load/config.yaml
krzysied: OK. I've change the code. It's not required to provide mapping now (for delete operation).

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -0,0 +1,224 @@
...
+	fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")
+	if err != nil {
+		return summaries, err
+	}
+	timeout, err := util.GetDurationOrDefault(config.Params, "timeout", defaultWaitForRCPodsTimeout)
+	if err != nil {
+		return summaries, err
+	}
+
+	runtimeObjectsList, err := getRuntimeObjectForKind(config.ClientSet, kind, namespace, labelSelector, fieldSelector)
```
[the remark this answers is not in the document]

clusterloader2/testing/load/config.yaml
krzysied: I've also removed mapping from deletion parts of test configs.

shyamjvs: Thanks.

clusterloader2/pkg/config/template_provider.go
```diff
@@ -34,17 +35,60 @@ type TemplateProvider struct {
...
+			defer tp.lock.Unlock()
+		}
+		var err error
+		bin, err = ioutil.ReadFile(filepath.Join(tp.basepath, path))
+		if err != nil {
+			return []byte{}, fmt.Errorf("reading error: %v", err)
+		}
+		tp.binCache[path] = bin
+	}
+
```
shyamjvs: remove extra line

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -0,0 +1,231 @@
...
+
+func getNamespaceFromRuntimeObject(obj runtime.Object) (string, error) {
+	metaObjectAccessor, ok := obj.(metav1.ObjectMetaAccessor)
+	if !ok {
+		return "", fmt.Errorf("unsupported kind when getting namespace: %v", obj)
+	}
+	return metaObjectAccessor.GetObjectMeta().GetNamespace(), nil
+}
+
+func getSelectorFromRuntimeObject(obj runtime.Object) (labels.Selector, error) {
```
shyamjvs: I think you could make this function generic by casting the obj into unstructured and accessing the selector, no?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
shyamjvs: Ping @wojtek-t. If you're ok with it.

clusterloader2/pkg/config/template_provider.go
krzysied: Done.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: TBH, I would prefer not to. Current implementation uses specific structures. Which is quite convenient due to:
- verifying types of objects
- knowing the types of fields - same fields can have different types across different object kinds.

Changing this function to use unstructured objects will make it less readable.

However, if this measurement will be reimplmented to use dynamic client, using unstructured object will be preferred.

shyamjvs: Ok, that sounds fine. Could add some high-level TODO somewhere saying that these functions should be changed after moving to dynamic client?

clusterloader2/testing/load/config.yaml
```diff
@@ -0,0 +1,120 @@
+name: Load
+automanagedNamespaces: 3
+tuningSets:
+- name: Uniform100qps
```
shyamjvs: I think you should change these QPSs to averageQPS instead of uniform to simulate load test accurately.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -210,17 +211,30 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) []err
 func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32, operation OperationType) []error {
 	var errList []error
 	objName := fmt.Sprintf("%v-%d", object.Basename, replicaIndex)
-	var mapping map[string]interface{}
-	if object.TemplateFillMap == nil {
-		mapping = make(map[string]interface{})
-	} else {
-		mapping = object.TemplateFillMap
+	if operation != CREATE_OBJECT && operation != PATCH_OBJECT && operation != DELETE_OBJECT {
```
shyamjvs: Can you please move this check to 'default' case of switch statement below?

krzysied:
All requested changes are done.
@shyamjvs @wojtek-t PTAL

clusterloader2/pkg/config/template_provider.go
```diff
@@ -34,17 +35,59 @@ type TemplateProvider struct {
...
 func NewTemplateProvider(basepath string) *TemplateProvider {
 	return &TemplateProvider{
 		basepath:      basepath,
+		binCache:      make(map[string][]byte),
 		templateCache: make(map[string]*template.Template),
 	}
 }
 
+func (tp *TemplateProvider) getRaw(path string, acquireLock bool) ([]byte, error) {
+	if acquireLock {
```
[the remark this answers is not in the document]
krzysied: Ok. Done.

shyamjvs:
/lgtm
Thanks!

PR #206, opened 2018-08-22
ClusterLoader - Changing basename placeholder to name

krzysied:


clusterloader2/pkg/test/simpletestexecutor.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	basenamePlaceholder = "Basename"
+	basenamePlaceholder = "Name"
```
shyamjvs: -> namePlaceholder

PR #210, opened 2018-08-27
ClusterLoader - Allowing multiple tests

krzysied:
Allowing `--testconfig` flag to be used multiple times.
Test will wait until resources are cleaned up.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -19,8 +19,10 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
 
 	"github.com/golang/glog"
+	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
@@ -31,10 +33,11 @@ import (
 
 var (
 	clusterLoaderConfig config.ClusterLoaderConfig
+	testConfigPaths     []string
 )
 
 func initClusterFlags() {
-	flag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 }
 
 func validateClusterFlags() []error {
@@ -46,14 +49,16 @@ func validateClusterFlags() []error {
 }
 
 func initFlags() {
-	flag.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
-	flag.StringVar(&clusterLoaderConfig.TestConfigPath, "testconfig", "", "Path to the test config file")
+	pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	pflag.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
+	pflag.StringArrayVar(&testConfigPaths, "testconfig", []string{}, "Paths to the test config files")
 	initClusterFlags()
 }
 
 func validateFlags() []error {
 	errList := make([]error, 0)
-	if clusterLoaderConfig.TestConfigPath == "" {
+	if len(testConfigPaths) < 1 {
 		errList = append(errList, fmt.Errorf("no test config path specified"))
 	}
 	errList = append(errList, validateClusterFlags()...)
@@ -63,7 +68,9 @@ func validateFlags() []error {
 func main() {
 	defer glog.Flush()
 	initFlags()
-	flag.Parse()
+	if err := pflag.CommandLine.Parse(os.Args[1:]); err != nil {
+		glog.Fatalf("Flag parse failed: %v", err)
+	}
 	if errList := validateFlags(); len(errList) > 0 {
 		glog.Fatalf("Parsing flags error: %v", errors.NewAggregate(errList).Error())
 	}
@@ -72,15 +79,16 @@ func main() {
 	if err != nil {
 		glog.Fatalf("Framework creation error: %v", err)
 	}
-	testConfig, err := config.ReadConfig(clusterLoaderConfig.TestConfigPath)
-	if err != nil {
-		glog.Fatalf("Config reading error: %v", err)
-	}
+	for _, clusterLoaderConfig.TestConfigPath = range testConfigPaths {
+		testConfig, err := config.ReadConfig(clusterLoaderConfig.TestConfigPath)
+		if err != nil {
+			glog.Fatalf("Config reading error: %v", err)
+		}
 
-	glog.Infof("Running %v test - %v", testConfig.Name, clusterLoaderConfig.TestConfigPath)
-	if errList := test.RunTest(f, &clusterLoaderConfig, testConfig); len(errList) > 0 {
-		glog.Fatalf("Test execution failed: %v", errors.NewAggregate(errList).Error())
+		glog.Infof("Running %v test - %v", testConfig.Name, clusterLoaderConfig.TestConfigPath)
+		if errList := test.RunTest(f, &clusterLoaderConfig, testConfig); len(errList) > 0 {
+			glog.Fatalf("Test execution failed: %v", errors.NewAggregate(errList).Error())
+		}
+		glog.Infof("Test %v ran successfully!", clusterLoaderConfig.TestConfigPath)
 	}
-
-	glog.Info("Test ran successfully!")
 }
--- a/clusterloader2/pkg/framework/client/objects.go
+++ b/clusterloader2/pkg/framework/client/objects.go
@@ -41,6 +41,10 @@ const (
 	retryBackoffFactor          = 3
 	retryBackoffJitter          = 0
 	retryBackoffSteps           = 6
+
+	// Parameters for namespace deletion operations.
+	defaultNamespaceDeletionTimeout  = 10 * time.Minute
+	defaultNamespaceDeletionInterval = 5 * time.Second
 )
 
 // RetryWithExponentialBackOff a utility for retrying the given function with exponential backoff.
@@ -88,7 +92,6 @@ func CreateNamespace(c clientset.Interface, namespace string) error {
 		return err
 	}
 	return RetryWithExponentialBackOff(retryFunction(createFunc, apierrs.IsAlreadyExists))
-
 }
 
 // DeleteNamespace deletes namespace with given name.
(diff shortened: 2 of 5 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -31,6 +33,7 @@ import (
 
 var (
 	clusterLoaderConfig config.ClusterLoaderConfig
+	testConfigPaths     []string
```
shyamjvs: Why are you adding new field for this? Can't you just make the one inside clusterLoaderConfig an array?

krzysied: The problem is that if you are running test, it is hard to decide which test path from testConfigPaths you should use. Best approach for me is to have only one test config path, cause only one test config path is related to this test at a time.
That's why I decided have list of test to run outside the cluster config and provide test config path one by one.

clusterloader2/cmd/clusterloader.go
```diff
@@ -47,13 +50,15 @@ func validateClusterFlags() []error {
 
 func initFlags() {
 	flag.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
-	flag.StringVar(&clusterLoaderConfig.TestConfigPath, "testconfig", "", "Path to the test config file")
+	pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
+	pflag.StringArrayVar(&testConfigPaths, "testconfig", []string{}, "Paths to the test config files")
 	initClusterFlags()
```
shyamjvs: Can you move this line to the top? (keep all pflag commands in one block)

clusterloader2/cmd/clusterloader.go
shyamjvs: Ok. Wdyt about changing `ClusterLoaderConfig.TestConfigPath` -> `ClusterLoaderConfig.TestConfig` and passing the config itself directly. Will also make `RunTest()` signature simpler.

krzysied: I still need test config path to be passed to template provider - it uses relative path for object descriptions.

clusterloader2/cmd/clusterloader.go
krzysied: I've change flag declaration to `pflag` only. However, I still need to keep `flag` due to `glog` adding it's standard flags only to `flag`.

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -99,6 +99,24 @@ func DeleteNamespace(c clientset.Interface, namespace string) error {
...
+				return true, nil
+			}
+			if IsRetryableAPIError(innerErr) {
+				return false, nil
+			}
+			return false, innerErr
+		}
+		return false, nil
+	}
+	return wait.PollImmediateInfinite(5*time.Second, retryWaitFunc)
```
shyamjvs: You shouldn't be waiting infinitely for this. There are some known bugs iirc which can lead to namespace being in "Terminating" state forever (due to some non-deletable resources).

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -99,6 +99,24 @@ func DeleteNamespace(c clientset.Interface, namespace string) error {
 	return RetryWithExponentialBackOff(retryFunction(deleteFunc, apierrs.IsNotFound))
 }
 
+// WaitForDeleteNamespace waits untils namespace is terminated.
+func WaitForDeleteNamespace(c clientset.Interface, namespace string) error {
+	retryWaitFunc := func() (bool, error) {
+		_, innerErr := c.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
```
shyamjvs: Just call it "err"?

clusterloader2/pkg/framework/framework.go
```diff
@@ -92,6 +92,12 @@ func (f *Framework) DeleteAutomanagedNamespaces() error {
 			return err
 		}
 	}
+	for i := 1; i <= f.automanagedNamespaceCount; i++ {
```
shyamjvs: I would suggest "deleting and waiting" each namespace in a separate go routine. Otherwise it is unfair for earlier namespaces which get less time to finish compared to later namespaces.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -47,11 +47,13 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) []erro
 	defer cleanupResources(ctx)
 	var errList []error
 	ctx.GetTickerFactory().Init(conf.TuningSets)
+	nsCreationStart := time.Now()
 	err := ctx.GetFramework().CreateAutomanagedNamespaces(int(conf.AutomanagedNamespaces))
 	if err != nil {
 		errList = append(errList, fmt.Errorf("automanaged namespaces creation failed: %v", err))
 		return errList
 	}
+	glog.Infof("Automanaged namespaces creation time: %v", time.Since(nsCreationStart))
```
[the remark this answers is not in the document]
krzysied: I'm not sure about this ~0, that's why I measure it.
Also, I think there is no problem to have one additional measurement.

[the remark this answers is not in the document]
shyamjvs: I still don't think this adds any value (it really should be super negligible time).

> Also, I think there is no problem to have one additional measurement.

Yes. But for e.g we're not measuring the time cluster loader takes to read config right :)
I suggest not adding extra logs which don't give valuable info.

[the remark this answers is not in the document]
krzysied: OK. I removed this measurement.

clusterloader2/pkg/framework/client/objects.go
krzysied: Timeout added.

clusterloader2/pkg/framework/client/objects.go
krzysied: Done.

clusterloader2/pkg/framework/framework.go
krzysied: Rewrited to go routines.

krzysied:
/hold cancel
There was a problem with viper (that is used for reading test config). The global instance o viper was used without resetting it, thus it stored every config directory path it used. Therefore, it sometimes read wrong config (if name matched).

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -41,6 +41,10 @@ const (
 	retryBackoffFactor          = 3
 	retryBackoffJitter          = 0
 	retryBackoffSteps           = 6
+
+	// Default time parameters
+	defaultNamespaceDeletionTimeout = 10 * time.Minute
+	defaultInterval                 = 5 * time.Second
```
shyamjvs: Please make clear - 'defaultInterval' for what?

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -41,6 +41,10 @@ const (
 	retryBackoffFactor          = 3
 	retryBackoffJitter          = 0
 	retryBackoffSteps           = 6
+
+	// Default time parameters
```
shyamjvs: Maybe sth more informative, like `Parameters for namespace deletion operations.`?

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -99,6 +102,37 @@ func DeleteNamespace(c clientset.Interface, namespace string) error {
 	return RetryWithExponentialBackOff(retryFunction(deleteFunc, apierrs.IsNotFound))
 }
 
+// ListNamespaces returns list of existing namespace names.
+func ListNamespaces(c clientset.Interface) ([]string, error) {
```
shyamjvs: It doesn't really seem worthwhile to have this extra function. Could you move the List() call into `ListAutomanagedNamespaces` (only place where it's called)?

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -99,6 +102,37 @@ func DeleteNamespace(c clientset.Interface, namespace string) error {
...
+
+// WaitForDeleteNamespace waits untils namespace is terminated.
+func WaitForDeleteNamespace(c clientset.Interface, namespace string) error {
+	retryWaitFunc := func() (bool, error) {
+		_, err := c.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
+		if err != nil {
+			if apierrs.IsNotFound(err) {
+				return true, nil
+			}
+			if IsRetryableAPIError(err) {
```
shyamjvs: Could you negate this condition and remove the extra return statement?

clusterloader2/pkg/framework/framework.go
```diff
@@ -84,16 +88,47 @@ func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
...
+	}
+	return automanagedNamespacesList, nil
+}
+
 // DeleteAutomanagedNamespaces deletes all automanged namespaces.
-func (f *Framework) DeleteAutomanagedNamespaces() error {
+func (f *Framework) DeleteAutomanagedNamespaces() []error {
+	var wg wait.Group
+	var lock sync.Mutex
+	var errList []error
```
shyamjvs: It seems like it's finally time to implement this errList class (maybe in another PR) instead of adding all this locking code everywhere.

clusterloader2/pkg/framework/framework.go
```diff
@@ -84,16 +88,47 @@ func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
...
+				defer lock.Unlock()
+				errList = append(errList, err)
+			}
+		})
 	}
-	f.automanagedNamespaceCount = 0
-	return nil
+	wg.Wait()
+	if len(errList) < 1 {
+		f.automanagedNamespaceCount = 0
```
shyamjvs: Why do you need to do this?

clusterloader2/pkg/framework/framework.go
```diff
@@ -115,3 +150,15 @@ func (f *Framework) DeleteObject(gvk schema.GroupVersionKind, namespace string,
 func (f *Framework) GetObject(gvk schema.GroupVersionKind, namespace string, name string) (*unstructured.Unstructured, error) {
 	return client.GetObject(f.dynamicClient, gvk, namespace, name)
 }
+
+func isAutomanagedNamespace(name string) bool {
+	prefix := AutomanagedNamespaceName + "-"
```
shyamjvs: Maybe just `return regexp.MatchString(AutomanagedNamespaceName + "-[0-9]+", name)`?

clusterloader2/pkg/framework/client/objects.go
krzysied: I want to keep it as it is.
`client/objects.go` is responsible for cluster operation, while `framework.go` is a facade that makes this generic operation to be clusterloader specific. Therefore, `client/objects.go` returns list of all namespaces, while `framework.go` only filters out non-autogenerated.

clusterloader2/pkg/framework/client/objects.go
krzysied: Done.

clusterloader2/pkg/framework/framework.go
krzysied: Good idea. I will make separate PR for this.

clusterloader2/pkg/framework/framework.go
krzysied: Done. I used `AutomanagedNamespaceName + "-[1-9][0-9]*"` as the patter, cause 1-based indexing.

clusterloader2/pkg/framework/framework.go
krzysied: This represents the state of the cluster. For example, if `CreateAutomanagedNamespaces` is run when not every automanaged namespace is removed, it will return error without trying to create the namespaces.

clusterloader2/pkg/framework/client/objects.go
shyamjvs: I'm not saying about how it is used. What I mean is this function is pretty much just a `c.CoreV1().Namespaces().List()` call which you can directly do inside any function where you need the namespaces.

krzysied: I understand your point. However, this change would break the convention I described in previous comment.

clusterloader2/pkg/framework/framework.go
krzysied: Lines removed.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -47,7 +48,16 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) []erro
...
 	var errList []error
 	ctx.GetTickerFactory().Init(conf.TuningSets)
-	err := ctx.GetFramework().CreateAutomanagedNamespaces(int(conf.AutomanagedNamespaces))
+	automanagedNamespacesList, err := ctx.GetFramework().ListAutomanagedNamespaces()
+	if err != nil {
+		errList = append(errList, fmt.Errorf("automanaged namespaces listing failed: %v", err))
+		return errList
+	}
+	if len(automanagedNamespacesList) > 0 {
+		errList = append(errList, fmt.Errorf("automanaged namespaces already reserved"))
```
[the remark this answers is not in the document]
krzysied: Error message updated.

clusterloader2/pkg/framework/framework.go
shyamjvs: Sure, but please remember to do it (or add TODO somewhere and fix).

krzysied: Done.

clusterloader2/pkg/framework/client/objects.go
krzysied: I've just realized that this implementation was missing retries. I've added it.

PR #211, opened 2018-08-28
ClusterLoader - Api responsiveness fix

krzysied:
Changing metric to have exactly the same name and the same layout as original one.
Needed due to backward compatibility.


krzysied:
/hold
Waiting for #205. There will be changes to load test needed.

PR #213, opened 2018-08-29
ClusterLoader - thread-safe error list

krzysied:
Adding thread-safe error list.


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -95,32 +89,27 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) []erro
 }
 
 // ExecuteStep executes single test step based on provided step configuration.
-func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) []error {
+func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) *util.ErrorList {
 	var wg wait.Group
 	// TODO(krzysied): Consider moving lock and errList to separate structure.
-	var lock sync.Mutex
-	var errList []error
+	errList := &util.ErrorList{}
```
shyamjvs: Please fix this.

clusterloader2/pkg/util/error_list.go
```diff
@@ -0,0 +1,73 @@
...
+// NewErrorList creates new error list.
+func NewErrorList(errors ...error) *ErrorList {
+	return &ErrorList{
+		errors: errors,
+	}
+}
+
+// IsEmpty returns true if there is no error in the list.
+func (e *ErrorList) IsEmpty() bool {
+	return len(e.errors) == 0
```
shyamjvs: Lock here?

clusterloader2/pkg/util/error_list.go
```diff
@@ -0,0 +1,73 @@
...
+package util
+
+import (
+	"strings"
+	"sync"
+)
+
+// ErrorList is a thread-safe error list.
+type ErrorList struct {
+	lock   sync.Mutex
```
shyamjvs: Maybe use RWMutex?

clusterloader2/pkg/test/simple_test_executor.go
krzysied: Done.

clusterloader2/pkg/util/error_list.go
krzysied: Good catch. Lock added.

clusterloader2/pkg/util/error_list.go
krzysied: I think it is not needed - this will mostly perform write operations.

shyamjvs: Yeah, that should be ok.

PR #214, opened 2018-08-30
ClusterLoader - Adding templating to test configs

krzysied:
Test configs will support templating. Nodes placeholder and template functions will available for test configs.
`nodes` must be provided!

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -38,12 +38,17 @@ var (
 
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
+	// TODO(krzysied): query actual number of nodes with client (if not set).
+	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
 }
 
 func validateClusterFlags() *util.ErrorList {
 	errList := util.NewErrorList()
 	if clusterLoaderConfig.ClusterConfig.KubeConfigPath == "" {
 		errList.Append(fmt.Errorf("no kubeconfig path specified"))
+	}
+	if clusterLoaderConfig.ClusterConfig.Nodes < 1 {
+		errList.Append(fmt.Errorf("number of nodes not specified"))
 	}
 	return errList
 }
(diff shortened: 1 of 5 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -38,13 +38,18 @@ var (
...
+	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
 }
 
 func validateClusterFlags() *util.ErrorList {
 	errList := util.NewErrorList()
 	if clusterLoaderConfig.ClusterConfig.KubeConfigPath == "" {
 		errList.Append(fmt.Errorf("no kubeconfig path specified"))
 	}
+	if clusterLoaderConfig.ClusterConfig.Nodes < 1 {
+		errList.Append(fmt.Errorf("no nodes specified"))
```
shyamjvs: -> "Number of nodes not specifiied"

clusterloader2/pkg/config/codec.go
```diff
@@ -17,45 +17,35 @@ limitations under the License.
...
+	var config api.Config
 	v := viper.New()
-	v.SetConfigName(strings.TrimSuffix(base, ext))
-	v.AddConfigPath(filepath.Dir(path))
-	if err := v.ReadInConfig(); err != nil {
-		return nil, fmt.Errorf("read config failed: %v", err)
+	if ext == "" {
+		ext = "yaml"
+	}
+	v.SetConfigType(ext)
```
shyamjvs: Instead of relying on such fragile logic, I'd suggest re-using this utility - https://github.com/kubernetes/apimachinery/blob/master/pkg/util/yaml/decoder.go#L207

clusterloader2/pkg/test/test.go
```diff
@@ -65,6 +62,14 @@ func RunTest(f *framework.Framework, clusterLoaderConfig *config.ClusterLoaderCo
 			}
 		}
 	}
+
 	ctx := CreateContext(clusterLoaderConfig, f, state.NewNamespacesState())
+	mapping := make(map[string]interface{})
```
shyamjvs: Maybe just directly do:

```
mapping := map[string]interface{"Nodes": clusterLoaderConfig.ClusterConfig.Nodes}
```

shyamjvs: Also minor nit: Move this line to after getting the test config step.

clusterloader2/pkg/config/codec.go
krzysied: Done.

clusterloader2/pkg/test/test.go
krzysied: >mapping := map[string]interface{"...

Done.

>Also minor nit: Move this line to after getting the test config step.

I've moved it one line down. Not sure if that's what you meant.

PR #217, opened 2018-09-03
ClusterLoader - test configs update

krzysied:
Changing configs to templates. Configs should be now more similar to original ones.


clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,29 @@
+#Constants
+{{$BIG_GROUP_SIZE := 250}}
+{{$MEDIUM_GROUP_SIZE := 30}}
+{{$SMALL_GROUP_SIZE := 5}}
+{{$PODS_PER_NODE := 30}}
+{{$NODES_PER_NAMESPACE := 100}}
+#Variables
+{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
```
wojtek-t: I'm against of that - I would much prefer having something simple, potentially imposing some other conditions on the setup.

So here is my thoughts:

If you assume number of nodes is divisible by 100, then:
bigRCs = num_nodes * 0.03
mediumRCs = num_nodes * 0.3
smallRCs = num_nodes * 3

Now if you assume that NODES_PER_NAMESPACE is a divisor of NUM_NODES, what you get is:
bigRCs = NODES_PER_NAMESPACE * NUM_NAMESPACES * 0.03
which means:
bigRCsPerNamespace = NODES_PER_NAMESPACE * 0.03

So with just two additional requirements:
- NUM_NODES divisable by 100
- NODES_PER_NAMESPACES being a divisor of 100

you can have it much much simple.

And our gole for these tests isn't to have a super generic config that handles all cases and noone understands it.
The purpose is to have it as simple as possible.

And I think that would exactly result in that.

@shyamjvs - thoughts?

krzysied: >bigRCs = num_nodes * 0.03
mediumRCs = num_nodes * 0.3
smallRCs = num_nodes * 3

It looks simpler, but...
- This numbers are created based on assumption on pods per nodes and RC size. If one of this variable is change, updating those equations for number of RC is also required. Based on how this equations look it's not obvious that they need to be updated.
- Having slightly more complex equations allows to show the logic behind it. People will understand this equations instead of guessing why we used those constants.

wojtek-t: It would be true if we would be changing those numbers. But we aren't and tbh people won't really be doing that.
So introducing complexity just to make it more generic, knowing that probably noone will take advantage of that is overkill.

If needed we can always change that later - fo now, it's overkill that:
- makes it hard to parse - seriously, if I didn't know what you're going to achieve I wouldn't easily parse that
- noone will take advantage of

shyamjvs: Krzysiek has a point with making it more understandable, but I agree more with @wojtek-t that we should try to keep it as simple as possible - optimizing for generic scenario is probably an overkill for us atm.

wojtek-t: I'm fine with some comment with a link to a doc where how it's computed and maybe what you currently have if someone would like to use that in the future.

But for now, I would like to keep it as simple as possible.

krzysied: I've changed it to simpler version. However, I've also added comments to these variables.

krzysied: @wojtek-t  Are you ok with this change?

clusterloader2/testing/density/config.yaml
```diff
@@ -25,22 +33,23 @@ steps:
...
       templateFillMap:
-        Replicas: 30
+        Replicas: {{$replicas}}
         PodKind: saturationPod
 - measurements:
   - Identifier: WaitForSaturationRunning
     Method: WaitForRunningPods
     Params:
-      desiredPodCount: 90
+      desiredPodCount: {{MultiplyInt $namespaces $replicas}}
```
shyamjvs: Isn't this just `$totalpods`?

shyamjvs: Or you're trying to account for rounding off?

shyamjvs:
Btw - you could've broken the 2 commits into 2 different PRs (so you're not blocked on fixing the CI job).

clusterloader2/testing/density/config.yaml
```diff
@@ -1,5 +1,13 @@
+#Constants
+{{$NODES_PER_NAMESPACE := 100}}
+{{$PODS_PER_NODE := 30}}
+#Variables
+{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
+{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
```
shyamjvs: (IIUC) Why are you subtracting 1 from #nodes?

clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,33 @@
...
+{{$BIG_GROUP_SIZE := 250}}
+{{$MEDIUM_GROUP_SIZE := 30}}
+{{$SMALL_GROUP_SIZE := 5}}
+{{$PODS_PER_NODE := 30}}
+{{$NODES_PER_NAMESPACE := 100}}
+#Variables
+# bigRCs - 1/4 of all pods should be in big RCs.
+{{$bigRCs := MultiplyInt .Nodes 0.03}}
+# mediumRCs - 1/4 of all pods should be in medium RCs.
+{{$mediumRCs := MultiplyInt .Nodes 0.3}}
```
shyamjvs: Shouldn't this be 0.25? (Nodes*30 / 4*30)

clusterloader2/testing/density/config.yaml
krzysied: To not create additional namespace when we have `nodes` == `nodes_per_namespace` (* `n`). This is a handling for this one specific case.

clusterloader2/testing/load/config.yaml
krzysied: It should. Good catch :)

krzysied: Fixed.

krzysied:
> Btw - you could've broken the 2 commits into 2 different PRs (so you're not blocked on fix for CI job).

You are right. I created separate PR #220 .

clusterloader2/testing/density/config.yaml
shyamjvs: SG

clusterloader2/testing/density/config.yaml
shyamjvs: Then instead of doing this multiplication at places, could you instead just compute a `$totalPodsRounded` at the top and use it?

clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,33 @@
...
+{{$mediumRCs := MultiplyInt .Nodes 0.25}}
+# smallRCs - 1/2 of all pods should be in small RCs.
+{{$smallRCs := MultiplyInt .Nodes 3}}
+{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
+{{$bigRCsPerNamespace := DivideInt $bigRCs $namespaces}}
+{{$mediumRCsPerNamespace := DivideInt $mediumRCs $namespaces}}
+{{$smallRCsPerNamespace := DivideInt $smallRCs $namespaces}}
+{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
+# saturationQPS - qps for RC operations should be an equivalent to 10 [pods/s].
+{{$saturationQPS := DivideFloat (AddInt $bigRCs $mediumRCs | AddInt $smallRCs) (DivideFloat $totalPods 20)}}
```
shyamjvs: Is it possible to change AddInt to take arbitrary no. of arguments? (to make this cleaner)

clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,33 @@
...
+# mediumRCs - 1/4 of all pods should be in medium RCs.
+{{$mediumRCs := MultiplyInt .Nodes 0.25}}
+# smallRCs - 1/2 of all pods should be in small RCs.
+{{$smallRCs := MultiplyInt .Nodes 3}}
+{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
+{{$bigRCsPerNamespace := DivideInt $bigRCs $namespaces}}
+{{$mediumRCsPerNamespace := DivideInt $mediumRCs $namespaces}}
+{{$smallRCsPerNamespace := DivideInt $smallRCs $namespaces}}
+{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
+# saturationQPS - qps for RC operations should be an equivalent to 10 [pods/s].
```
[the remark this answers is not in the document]
krzysied: The problem is that this qps is calculated with assumption that we want to achieve 10pods/s. However, there is also service created for every RC. I applied x2, because there are always two objects that should be created.

clusterloader2/testing/density/config.yaml

clusterloader2/testing/load/config.yaml
krzysied: At first I thought that template functions will required fixed number of parameters. However it turned out that template functions support variadic args.
Change implemented.

clusterloader2/testing/load/config.yaml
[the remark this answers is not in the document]
krzysied: Maybe I should separate service creation from rc creation?

clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,33 @@
...
+{{$saturationQPS := DivideFloat (AddInt $bigRCs $mediumRCs | AddInt $smallRCs) (DivideFloat $totalPods 20)}}
+#Test
 name: load
-automanagedNamespaces: 3
+automanagedNamespaces: {{$namespaces}}
 tuningSets:
-- name: Average0.33qps
+- name: AverageSaturate
```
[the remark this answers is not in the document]

clusterloader2/testing/load/config.yaml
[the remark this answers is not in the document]
shyamjvs: I'm not sure I understand the relation b/w services and RC creations here. Throughput is calculated from the rate at which we want pods to be created (which is decided by the rate at which we create RCs).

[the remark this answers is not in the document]
krzysied: For every RC, one service is created. So I assumed that I should use bundle = [service, RC].
However, in the original load test, all services are created and then all the RC. The `AvgSaturationQPS` was originally only used for RC creation, so I needed to recalculated it for svc + RC bundle. That's why I used x2 multiplier.

Maybe it was my mistake to use bundle and I should use original test logic.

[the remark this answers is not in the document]
wojtek-t: Bundling service and RC together is the right decision.

But I also don't understand why the throughput was changed - the throughput as Shyam mentioned above is for pods and whether we create a service or not is orthogonal.

clusterloader2/testing/density/config.yaml
```diff
@@ -25,22 +34,23 @@ steps:
...
         PodKind: saturationPod
 - measurements:
   - Identifier: WaitForSaturationRunning
     Method: WaitForRunningPods
     Params:
-      desiredPodCount: 90
+      desiredPodCount: {{$totalPodsRounded}}
       labelSelector: kind = saturationPod
-      timeout: 2m
+      # timeout = (nodes * 30 /20)s + 3m
```
shyamjvs: Update this comment in terms of #totalPods.

clusterloader2/testing/load/config.yaml
```diff
@@ -25,26 +46,57 @@ steps:
...
+        ReplicasMin: {{$SMALL_GROUP_SIZE}}
+        ReplicasMax: {{$SMALL_GROUP_SIZE}}
 - measurements:
   - Identifier: WaitForRunningRCs
     Method: WaitForControlledPodsRunning
     Params:
       kind: ReplicationController
       labelSelector: kind = loadRC
-      timeout: 2m
+      # timeout = (nodes * 30 /20)s + 3m
```
shyamjvs: Also here.

clusterloader2/testing/load/config.yaml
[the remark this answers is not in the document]
krzysied: Cause after this change (making bundle), total number of pods remained the same and number of operations (for object creation) that need tick from tuning set increased two times.
To keep the same expected time for pod creation, I had to increase qps two times - half of the ticks are consumed by the services.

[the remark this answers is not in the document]
shyamjvs: Why isn't one tick used per object bundle?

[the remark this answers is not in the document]
wojtek-t: yes - that sounds liek a bug.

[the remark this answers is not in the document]
krzysied: Tuning set is all about operation per seconds. Inside the object bundle you can have any number of objects (operation). The real operations per seconds would be tuning set qps + all operations from any started bundles that haven't finished yet.

I think that tick per object operation approach is easier to understand and better implements the idea behind tuning sets.

clusterloader2/testing/density/config.yaml
krzysied: Comment updated.

clusterloader2/testing/load/config.yaml
krzysied: Comment updated.

clusterloader2/testing/load/config.yaml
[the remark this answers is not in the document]
shyamjvs: > Tuning set is all about operation per seconds.

I'm not sure I agree with that because TuningSet is given per object bundle as opposed to per object (which kind of makes sense given that the bundle forms a logical unit of user workload). @wojtek-t - wdyt?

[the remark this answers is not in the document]
wojtek-t: +1 to Shyam - your "unit" in ClusterLoader is object bundle and that is what we are executing.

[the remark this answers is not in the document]
krzysied: I stand corrected then. I always thought that object bundle was introduced only to ensure order of object operations :/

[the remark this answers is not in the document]
krzysied: I created #223 to fix the ticker.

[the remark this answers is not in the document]
krzysied: I've changed this number from `20` to `10`.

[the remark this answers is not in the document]
krzysied: #223 merged.
@shyamjvs @wojtek-t Friendly ping to keep PR alive.

clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,33 @@
...
-- name: Average0.33qps
+- name: AvgSaturationQPS
   randomizedLoad:
-    averageQps: 0.33
-- name: Average1.4qps
+    averageQps: {{$saturationQPS}}
+- name: AvgScalingQPS
   randomizedLoad:
-    averageQps: 1.4
+    averageQps: {{MultiplyFloat $saturationQPS 4}}
```
shyamjvs: Could you add a comment here explaining this factor of 4? (you should be able to ~copy it from our load test)

clusterloader2/testing/load/config.yaml
```diff
@@ -1,12 +1,33 @@
...
+{{$bigRCs := MultiplyInt .Nodes 0.03}}
+# mediumRCs - 1/4 of all pods should be in medium RCs.
+{{$mediumRCs := MultiplyInt .Nodes 0.25}}
+# smallRCs - 1/2 of all pods should be in small RCs.
+{{$smallRCs := MultiplyInt .Nodes 3}}
+{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
+{{$bigRCsPerNamespace := DivideInt $bigRCs $namespaces}}
+{{$mediumRCsPerNamespace := DivideInt $mediumRCs $namespaces}}
+{{$smallRCsPerNamespace := DivideInt $smallRCs $namespaces}}
+{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
```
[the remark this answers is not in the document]
krzysied: I've changed it to totalPodsRounded.
It would still work with totalPods though. Only timeouts and qps are calculated using this variable - in some cases tests would take few minutes longer to be executed.

clusterloader2/testing/load/config.yaml
krzysied: Comment added.

krzysied:
@shyamjvs Pinging to keep this PR alive.

I think all of the comments are resolved.

krzysied:
/assign @wojtek-t 
PTAL

PR #218, opened 2018-09-03
ClusterLoader - Adding parallelism

krzysied:
Executing object operations (namespace x replicaIndex) in parallel.

### DIFF
--- a/clusterloader2/pkg/test/simple_test_executor.go
+++ b/clusterloader2/pkg/test/simple_test_executor.go
@@ -128,83 +128,88 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) *util
 		return util.NewErrorList(fmt.Errorf("ticker creation error: %v", err))
 	}
 	defer ticker.Stop()
-	for _, nsName := range nsList {
-		instancesStates := make([]*state.InstancesState, 0)
-		// Updating state (DesiredReplicaCount) of every object in object bundle.
-		for j := range phase.ObjectBundle {
-			id, err := getIdentifier(ctx, &phase.ObjectBundle[j])
-			if err != nil {
-				errList.Append(err)
-				return errList
-			}
-			instances, exists := ctx.GetState().Get(nsName, id)
-			if !exists {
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -128,83 +128,88 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) *util
...
-				} else if replicaIndex >= instancesStates[j].CurrentReplicaCount {
-					<-ticker.C
-					if objectErrList := ste.ExecuteObject(ctx, &phase.ObjectBundle[j], nsName, replicaIndex, CREATE_OBJECT); !objectErrList.IsEmpty() {
-						errList.Concat(objectErrList)
-						if isErrsCritical(objectErrList) {
-							return errList
+				})
+			}
+			// Handling for update/create objects.
+			for replicaCounter := int32(0); replicaCounter < phase.ReplicasPerNamespace; replicaCounter++ {
```
shyamjvs: It seems like you are introducing goroutines proportional to #namespaces x #replicas. This can be a really huge number of them and I'm not sure if you should have such uncontrolled parallelism. Instead, bound parallelism / fixed no. of workers is better IMO.

@wojtek-t - for thoughts

wojtek-t: go-routines is fairly cheap mechanism - it's completely fine to have thousands of them running at the same time.
Didn't look into the code, but assuming most of them are sleeping at any given point in time, that sounds fine to me.

shyamjvs: You're right. Since most of them would be blocked on timer channel at all times, that should be ok.

Couple of points though:

- I'm not sure if it makes sense to have more goroutines than our qps (which will be a bottleneck anyway)
- There is some small constant [memory overhead associated](https://stackoverflow.com/questions/8509152/max-number-of-goroutines) with each goroutine - about 4MB per 1000 goroutines - but we may be able to live with that

PR #219, opened 2018-09-03
ClusterLoader - Adding parallelism to wait_for_controlled_pods measurement

krzysied:
Making  wait_for_controlled_pods measurement be executed in parallel.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -23,6 +23,7 @@ package common
 
 import (
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"github.com/golang/glog"
@@ -32,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
@@ -83,24 +85,39 @@ func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.Meas
 	if err != nil {
 		return summaries, err
 	}
-	runningRuntimeObjects := 0
-	for start := time.Now(); time.Since(start) < timeout; time.Sleep(defaultWaitForRCPodsInterval) {
-		if runningRuntimeObjects >= len(runtimeObjectsList) {
-			return summaries, nil
-		}
+	var wg wait.Group
+	errList := util.NewErrorList()
+	var runningRuntimeObjects int32
+	for objectCounter := 0; objectCounter < len(runtimeObjectsList); objectCounter++ {
+		objectIndex := objectCounter
+		wg.Start(func() {
+			if err := waitForRuntimeObject(config.ClientSet, runtimeObjectsList[objectIndex], timeout); err != nil {
+				errList.Append(fmt.Errorf("waiting for %v error: %v", kind, err))
+				return
+			}
 
-		if err := waitForRuntimeObject(config.ClientSet, runtimeObjectsList[runningRuntimeObjects], timeout-time.Since(start)); err != nil {
-			return summaries, fmt.Errorf("waiting for %v error: %v", kind, err)
-		}
-		runningRuntimeObjects++
-		if namespace == metav1.NamespaceAll {
-			glog.Infof("%s: %d / %d %ss with all pods running", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		} else {
-			glog.Infof("%s: %v: %d / %d %ss with all pods running", waitForControlledPodsRunningName, namespace, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		}
+			atomic.AddInt32(&runningRuntimeObjects, 1)
+			objName, err := getNameFromRuntimeObject(runtimeObjectsList[objectIndex])
+			if err != nil {
+				errList.Append(fmt.Errorf("reading object name error: %v", err))
+				return
+			}
+			objNamespace, err := getNamespaceFromRuntimeObject(runtimeObjectsList[objectIndex])
+			if err != nil {
+				errList.Append(fmt.Errorf("reading object namespace error: %v", err))
+				return
+			}
+			glog.Infof("%s: %s (%s) has all pods running", waitForControlledPodsRunningName, objName, objNamespace)
+		})
 	}
+	wg.Wait()
 
-	return summaries, fmt.Errorf("timeout while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods", kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects)
+	if int(runningRuntimeObjects) == len(runtimeObjectsList) {
+		glog.Infof("%s: %d / %d %ss are running with all pods", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
+		return summaries, nil
+	}
+	return summaries, fmt.Errorf("error while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods: %v",
+		kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects, errList.String())
 }
 
 func waitForRuntimeObject(c clientset.Interface, obj runtime.Object, timeout time.Duration) error {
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -83,24 +85,35 @@ func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.Meas
...
-			glog.Infof("%s: %d / %d %ss with all pods running", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		} else {
-			glog.Infof("%s: %v: %d / %d %ss with all pods running", waitForControlledPodsRunningName, namespace, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		}
-	}
-
-	return summaries, fmt.Errorf("timeout while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods", kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects)
+	for objectCouter := 0; objectCouter < len(runtimeObjectsList); objectCouter++ {
+		objectIndex := objectCouter
+		wg.Start(func() {
```
krzysied: @shyamjvs Assuming that we agree on approach for parallelism in #218, I will add the same solution here.

clusterloader2/pkg/measurement/common/wait_for_pods.go
```diff
@@ -91,10 +91,12 @@ func waitForPods(clientSet clientset.Interface, namespace, labelSelector, fieldS
 				runningPodsCount++
 			}
 		}
-		if namespace == metav1.NamespaceAll {
-			glog.Infof("WaitForRunningPods: running %d / %d", runningPodsCount, desiredPodCount)
-		} else {
-			glog.Infof("WaitForRunningPods: %s: running %d / %d", namespace, runningPodsCount, desiredPodCount)
+		if log {
```
shyamjvs: In general, we're using glog [verbosity](https://godoc.org/github.com/golang/glog#V) to control amount of logs to print - that precisely seems to be the use-case here. So I suggest removing this bool checks and just log those statements at appropriate verbosity (maybe V(4)?).

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -83,24 +85,35 @@ func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.Meas
...
-		runningRuntimeObjects++
-		if namespace == metav1.NamespaceAll {
-			glog.Infof("%s: %d / %d %ss with all pods running", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		} else {
-			glog.Infof("%s: %v: %d / %d %ss with all pods running", waitForControlledPodsRunningName, namespace, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		}
-	}
-
-	return summaries, fmt.Errorf("timeout while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods", kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects)
+	for objectCouter := 0; objectCouter < len(runtimeObjectsList); objectCouter++ {
```
shyamjvs: -> objectCounter (typo)

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -83,24 +85,35 @@ func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.Meas
...
+		objectIndex := objectCouter
+		wg.Start(func() {
+			if err := waitForRuntimeObject(config.ClientSet, runtimeObjectsList[objectIndex], timeout); err != nil {
+				errList.Append(fmt.Errorf("waiting for %v error: %v", kind, err))
+				return
+			}
+
+			lock.Lock()
```
shyamjvs: Please use [atomic counters](https://gobyexample.com/atomic-counters) instead of manually managing all this locking.

krzysied: This lock is not only for runningRuntimeObjects variable, but also for logging.
I agree that if I remove the lock and use atomic counter, the runningRuntimeObjects will be calculated correctly. However, if I remove lock from logging part, it is possible that variable will change in a meantime. Potential can log same number of running objects multiple times.
WYT?

clusterloader2/pkg/measurement/common/wait_for_pods.go
krzysied: TBH, I wanted to disable log for `waitForPods` if it called by other measurement.
For example this function is called by `waitForControlledPodsRunningMeasurement` in parallel, so it will create many log entries and I'm not sure it those logs will be helpful.
I agree that verbosity can be applied here, but I'm not sure whether I want to remove the switch.
WYT?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: Typo fixed.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
shyamjvs: You're right, but I don't think that's a big problem. You can for e.g fix the log by just printing the current RC instead of the count till then.

clusterloader2/pkg/measurement/common/wait_for_pods.go
shyamjvs: If you think those logs aren't really helpful, then you can remove it. If you're think they're helpful and there's no measurement wanting to turn it off, then there's no need to add such extra switch atm.

krzysied: The thing is that this log are or are not helpful depending on use case. For example:
- If we use `wait_for_pods` measurement, you have one group of pods you're waiting for. In this case, this logs are very helpful, cause it shows how this specific group becomes ready.
- If we use `wait_for_controlled_pod`, `waiting_for_pod` is just a subfunction. In this case we have many pod groups, so seeing the changes in every group (especially if it's in parallel) is not so helpful. IMHO  it's better to observer only number of controlling object being ready.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: According to what we discussed offline, replaced lock with atomic counter.
Only name of a ready object is printed to log.

clusterloader2/pkg/measurement/common/wait_for_pods.go
```diff
@@ -91,10 +91,12 @@ func waitForPods(clientSet clientset.Interface, namespace, labelSelector, fieldS
...
+			if namespace == metav1.NamespaceAll {
+				glog.Infof("WaitForRunningPods: running %d / %d", runningPodsCount, desiredPodCount)
```
shyamjvs: IMO you should also include label and field selectors to the log. Otherwise they won't be much informative I think, when there are many calls to this function.

krzysied: I agree with you that we should be able to identify measurement.
IMHO I'm not sure that putting every parameter that function have to the log is the best option. Maybe we want to have logs like "<metric_type>(<metric_identifier>): <log text>"? We already provide identifiers in test definitions. WYT?

krzysied: Label selector and field selector added to the log.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -83,24 +85,39 @@ func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.Meas
...
+				return
+			}
+			glog.Infof("%s: %s (%s) has all pods running", waitForControlledPodsRunningName, objName, objNamespace)
+		})
 	}
+	wg.Wait()
 
-	return summaries, fmt.Errorf("timeout while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods", kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects)
+	if int(runningRuntimeObjects) == len(runtimeObjectsList) {
+		glog.Infof("%s: %d / %d %ss are running with all pods", waitForControlledPodsRunningName, runningRuntimeObjects, runningRuntimeObjects, kind)
```
shyamjvs: Something seems wrong with this log. Why are you printing `runningRuntimeObjects` twice?

clusterloader2/pkg/measurement/common/wait_for_pods.go
shyamjvs: I'm ok with keeping it this way.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: I've changed the second one to len(runtimeObjectsList).

PR #226, opened 2018-09-14
ClusterLoader -  Pod startup latency measurement

krzysied:


clusterloader2/pkg/measurement/util/pod_informer.go
```diff
@@ -0,0 +1,109 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+// PodInformer observes pods spcecified by given namespace, label selector and field selector.
+// There is an option to register callback for add and update pod event.
+type PodInformer struct {
```
shyamjvs: I think a big part of this file is just re-implementing [shared informers](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L43). For e.g registering event handlers, starting/stopping the informer. You should try to use that instead.. It'll also easily let you do things like checking if the informer has synced.

shyamjvs: The only new part I see is setting the listwatcher based on namespace and label/field selectors.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	podStartupLatencyMetricName = "PodStartupLatency"
+	syncTimeout                 = time.Minute
```
shyamjvs: Please use better name (like "informerSyncTimeout" or sth). This doesn't give much useful information.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+	resyncPeriod                = time.Second
```
shyamjvs: 1 second seems too less to me. I expect the resync period to be much higher (like 1 hour or sth).

@wojtek-t - Wdyt?

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+	return &podStartupLatencyMeasurement{}
+}
+
+type podStartupLatencyMeasurement struct {
+	namespace     string
+	labelSelector string
+	fieldSelector string
+	informer      cache.SharedInformer
+	stopCh        chan struct{}
+	createTimes   map[string]map[string]metav1.Time
```
wojtek-t: IIUC this is map from namespace -> createTimes.

Do you want to allow for the same names across namespaces?
Then much easier would be to make the key of the form "namespace/name" [it's done in many places already, e.g. it's what informer cache is also doing].

MetaNamespaceKey (or sth like) that in tools/cache in client-go is what you're looking for

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+	for _, event := range schedEvents.Items {
+		namespaceDict := getNamespaceDict(p.createTimes, event.InvolvedObject.Namespace)
+		if _, ok := namespaceDict[event.InvolvedObject.Name]; ok {
+			getNamespaceDict(p.scheduleTimes, event.InvolvedObject.Namespace)[event.InvolvedObject.Name] = event.FirstTimestamp
+		}
+	}
+	return nil
+}
+
+func checkPodFunc(p *podStartupLatencyMeasurement) func(interface{}) {
```
wojtek-t: this syntax is pretty strange - it should be method in podStartupLatencymeasurement.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+	p.createTimes = make(map[string]map[string]metav1.Time)
```
wojtek-t: initialize all of these in constructor instead.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+	perc50 := latencies[int(math.Ceil(float64(length*50)/100))-1]
+	perc90 := latencies[int(math.Ceil(float64(length*90)/100))-1]
+	perc99 := latencies[int(math.Ceil(float64(length*99)/100))-1]
+	return measurementutil.LatencyMetric{Perc50: perc50, Perc90: perc90, Perc99: perc99}
+}
+
+func latencyToPerfData(l measurementutil.LatencyMetric, name string) measurementutil.DataItem {
+	return measurementutil.DataItem{
+		Data: map[string]float64{
+			"Perc50": float64(l.Perc50) / 1000000, // us -> ms
```
wojtek-t: us -> ms - would be 1000

it's probably ns -> ms, right?

krzysied: You are right. Fixed.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: Changed to "namespace/name".

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,302 @@
...
+			checkPod(newObj)
+		},
+	})
+
+	go p.informer.Run(p.stopCh)
+	timeoutCh := make(chan struct{})
+	timeoutTimer := time.AfterFunc(syncTimeout, func() {
+		close(timeoutCh)
+	})
+	defer timeoutTimer.Stop()
```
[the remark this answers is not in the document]
krzysied: It just a timer to create the timeout signal. The defer part is to stop the timer, if informer synced before the timeout.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: I thought that resync is needed to update the newly detected objects. It turned out that this resync is full resync, so it is not needed.
Resync removed.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,296 @@
...
+	if !cache.WaitForCacheSync(timeoutCh, p.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+	return nil
+}
+
+func (p *podStartupLatencyMeasurement) gather(c clientset.Interface) ([]measurement.Summary, error) {
+	glog.Infof("Gathering pod startup latency measurement...")
+	if p.informer == nil || p.stopCh == nil {
```
wojtek-t: This assumes that there is no concurrency.

Is that true, that we don't support that (if so, we should be pretty clear in some comemnts/docs).
If not, we should probably add a lock.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,296 @@
...
+		"involvedObject.kind": "Pod",
+		"source":              corev1.DefaultSchedulerName,
+	}.AsSelector().String()
+	options := metav1.ListOptions{FieldSelector: selector}
+	schedEvents, err := c.CoreV1().Events(p.namespace).List(options)
+	if err != nil {
+		return err
+	}
+	for _, event := range schedEvents.Items {
+		key := createMetaNamespaceKey(event.InvolvedObject.Namespace, event.InvolvedObject.Name)
```
wojtek-t: Instead of creating a dedicated function, use:
cache.MetaNamespaceKeyFunc

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,296 @@
...
+		if _, ok := p.createTimes[key]; ok {
+			p.scheduleTimes[key] = event.FirstTimestamp
+		}
+	}
+	return nil
+}
+
+func (p *podStartupLatencyMeasurement) checkPodFunc() func(interface{}) {
+	var mutex sync.Mutex
```
wojtek-t: Why do you put this mutex here - it seems strange.

In general, you're trying to protect maps of p, so it should be mutex at the p level.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: The problem is that event.InvolvedObject doesn't have meta, so the MetaNamespaceKeyFunc will not work.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: I pretty sure that we don't want to have concurrency for the measurements. I will mention it in README.

krzysied:
@wojtek-t PTAL

clusterloader2/pkg/measurement/common/pod_startup_latency.go
wojtek-t: This should be mentioned also in this file (next to struct definition or its initialization or sth like that)

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,293 @@
...
+
+	close(p.stopCh)
+	if err := p.gatherScheduleTimes(c); err != nil {
+		return []measurement.Summary{}, err
+	}
+	for key, create := range p.createTimes {
+		sched, ok := p.scheduleTimes[key]
+		if !ok {
+			glog.Infof("Failed to find schedule time for %v", key)
+			continue
```
wojtek-t: This is tricky.
If there is pod that stuck in scheduling, starting, being reported etc., this will simply ignore it.
This is wrong - we should ignore them.

Maybe it's fine to not have it as part of histogram (though it's also not obvious to me), but at the very least, we should fail the test (well not immediately, but ensure that the test will fail), if there are more than X% of missed observations.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,293 @@
...
+	case "gather":
+		return p.gather(config.ClientSet)
+	default:
+		return summaries, fmt.Errorf("unknown action %v", action)
+	}
+
+}
+
+func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
+	glog.Infof("Starting pod startup latency measurement...")
```
wojtek-t: You should check if it wasn't already started before.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: Done.
Measurement will verify if >= 99% pods were successfully scheduled.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
krzysied: Done.

krzysied:
@wojtek-t PTAL

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,305 @@
...
+	createTimes   map[string]metav1.Time
+	scheduleTimes map[string]metav1.Time
+	runTimes      map[string]metav1.Time
+	watchTimes    map[string]metav1.Time
+}
+
+// Execute supports two actions:
+// - start - Starts to observe pods and pods events.
+// - gather - Gathers and prints current pod latency data.
+// This measurment doesn't support concurrency. Multiple calls to this measurement
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/measurement/common/pod_startup_latency.go
```diff
@@ -0,0 +1,305 @@
...
+	defer timeoutTimer.Stop()
+	if !cache.WaitForCacheSync(timeoutCh, p.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+	return nil
+}
+
+func (p *podStartupLatencyMeasurement) gather(c clientset.Interface) ([]measurement.Summary, error) {
+	glog.Infof("%v: gathering pod startup latency measurement...", podStartupLatencyMetricName)
+	if p.informer == nil || p.stopCh == nil {
```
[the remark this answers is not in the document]
krzysied: Done.
I've also added check for closed channel (line 150) - I don't want this measurement to panic.

clusterloader2/pkg/util/util.go
```diff
@@ -217,3 +217,14 @@ func CopyMap(src, dest map[string]interface{}) {
 		dest[k] = v
 	}
 }
+
+// IsChannelClosed returns true if given channel is closed.
+func IsChannelClosed(ch <-chan struct{}) bool {
+	select {
+	case <-ch:
```
wojtek-t: This isn't correct - if the channel is open but something is writing to it (or something is already buffered) it will also return true. But channel is not closed.
It's not really possible to check it efficiently.

Let's just remove it and assume channel is not closed - this class is not thread-safe anyway.

krzysied: Removed.

PR #229, opened 2018-09-19
ClusterLoader - Adding additional cluster config flags

krzysied:
Added flags:
- provider

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/pflag"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	"k8s.io/perf-tests/clusterloader2/pkg/test"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 
@@ -39,6 +40,7 @@ var (
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
 }
 
 func validateClusterFlags() *util.ErrorList {
@@ -88,6 +90,7 @@ func main() {
 		glog.Fatalf("Parsing flags error: %v", errList.String())
 	}
 
+	measurement.ClusterConfig = &clusterLoaderConfig.ClusterConfig
 	f, err := framework.NewFramework(clusterLoaderConfig.ClusterConfig.KubeConfigPath)
 	if err != nil {
 		glog.Fatalf("Framework creation error: %v", err)
--- a/clusterloader2/pkg/measurement/common/api_responsiveness.go
+++ b/clusterloader2/pkg/measurement/common/api_responsiveness.go
@@ -78,7 +78,7 @@ func (*apiResponsivenessMeasurement) Execute(config *measurement.MeasurementConf
 	case "gather":
 		// TODO(krzysied): Implement new method of collecting latency metrics.
 		// New method is defined here: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos.
-		nodeCount, err := util.GetInt(config.Params, "nodeCount")
+		nodeCount, err := util.GetIntOrDefault(config.Params, "nodeCount", measurement.ClusterConfig.Nodes)
 		if err != nil {
 			return summaries, err
 		}
--- a/clusterloader2/pkg/measurement/manager.go
+++ b/clusterloader2/pkg/measurement/manager.go
@@ -20,6 +20,13 @@ import (
 	"sync"
 
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+)
+
+var (
+	// ClusterConfig represents cluster configuration.
+	// TODO(issue 249): Move this global variable to MeasurementManager structure.
+	ClusterConfig *config.ClusterConfig
 )
 
 // MeasurementManager manages all measurement executions.
--- a/clusterloader2/pkg/test/simple_test_executor.go
+++ b/clusterloader2/pkg/test/simple_test_executor.go
@@ -99,7 +99,9 @@ func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) *util.Er
 			// index is created to make i value unchangeable during thread execution.
 			index := i
 			wg.Start(func() {
-				err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params)
+				err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method,
+					step.Measurements[index].Identifier,
+					step.Measurements[index].Params)
 				if err != nil {
 					errList.Append(fmt.Errorf("measurement call %s - %s error: %v", step.Measurements[index].Method, step.Measurements[index].Identifier, err))
 				}
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -27,4 +27,8 @@ type ClusterLoaderConfig struct {
 type ClusterConfig struct {
 	KubeConfigPath string `json: kubeConfigPath`
 	Nodes          int    `json: nodes`
+	Provider       string `json: provider`
+	// TODO(krzysied): Add support for HA cluster with more than one master.
+	MasterIP   string `json: masterIP`
+	MasterName string `json: masterName`
 }
--- a/clusterloader2/pkg/measurement/common/etcd_metrics.go
+++ b/clusterloader2/pkg/measurement/common/etcd_metrics.go
@@ -60,11 +60,11 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig) 
 	if err != nil {
 		return summaries, err
 	}
-	provider, err := util.GetString(config.Params, "provider")
+	provider, err := util.GetStringOrDefault(config.Params, "provider", measurement.ClusterConfig.Provider)
 	if err != nil {
 		return summaries, err
 	}
-	host, err := util.GetString(config.Params, "host")
+	host, err := util.GetStringOrDefault(config.Params, "host", measurement.ClusterConfig.MasterIP)
 	if err != nil {
 		return summaries, err
 	}
--- a/clusterloader2/pkg/measurement/common/metrics_for_e2e.go
+++ b/clusterloader2/pkg/measurement/common/metrics_for_e2e.go
@@ -86,7 +86,7 @@ type metricsForE2EMeasurement struct{}
 func (m *metricsForE2EMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
 	var summaries []measurement.Summary
 
-	provider, err := util.GetString(config.Params, "provider")
+	provider, err := util.GetStringOrDefault(config.Params, "provider", measurement.ClusterConfig.Provider)
 	if err != nil {
 		return summaries, err
 	}
--- a/clusterloader2/pkg/measurement/common/profile.go
+++ b/clusterloader2/pkg/measurement/common/profile.go
@@ -63,11 +63,11 @@ func createMeasurement(config *measurement.MeasurementConfig, profileKind string
 	if err != nil {
 		return summaries, err
 	}
-	provider, err := util.GetString(config.Params, "provider")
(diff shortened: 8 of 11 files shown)


clusterloader2/pkg/measurement/common/api_responsiveness.go
```diff
@@ -78,7 +78,7 @@ func (*apiResponsivenessMeasurement) Execute(config *measurement.MeasurementConf
 	case "gather":
 		// TODO(krzysied): Implement new method of collecting latency metrics.
 		// New method is defined here: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos.
-		nodeCount, err := util.GetInt(config.Params, "nodeCount")
+		nodeCount, err := util.GetIntOrDefault(config.Params, "nodeCount", config.ClusterConfig.Nodes)
```
shyamjvs: Weren't we gonna compute this by api call instead of taking it as parameter?

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -99,7 +99,7 @@ func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) *util.Er
 			// index is created to make i value unchangeable during thread execution.
 			index := i
 			wg.Start(func() {
-				err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params)
+				err := ctx.GetMeasurementManager().Execute(step.Measurements[index].Method, step.Measurements[index].Identifier, step.Measurements[index].Params, &ctx.GetClusterLoaderConfig().ClusterConfig)
```
shyamjvs: Can you split the arguments across multiple lines?

clusterloader2/pkg/measurement/manager.go
```diff
@@ -42,14 +43,15 @@ func CreateMeasurementManager(clientSet clientset.Interface) *MeasurementManager
...
 	measurementInstance, err := mm.getMeasurementInstance(methodName, identifier)
 	if err != nil {
 		return err
 	}
 	config := &MeasurementConfig{
-		ClientSet: mm.clientSet,
-		Params:    params,
+		ClientSet:     mm.clientSet,
+		Params:        params,
+		ClusterConfig: clusterConfig,
```
shyamjvs: Do you really need to pass and copy the cluster config each time for every measurement? It seems to me like the whole thing is just a global constant.

clusterloader2/pkg/measurement/common/api_responsiveness.go
krzysied: Computing parameters will be next step. I will create PR after this one is merged.

clusterloader2/pkg/measurement/manager.go
krzysied: I'm not convinced that global constant is the best solution. Here are the reasons why I think we shouldn't try this approach:
- Creates unnecessary dependency
It make metrics only be usable with this specific global variable. If anyone wants to use it in different project, he will need to assure that this global variable is initialized. Setting some external variable to change function result is not very intuitive. It also makes testing more complex.
- Scatters information across the whole project.
I assume that metrics is the part of project that at some point will be developed by other contributors from different projects. If we provide all available information in the interface signature, it's easier to understand what parameters are available and which not. IMHO going through the code to find the global variable you can use is more troublesome then just looking at the signature.

I think that we can make client config to be part of measurement manager (the same why as the clientSet is). This will remove cluster config from the signature of `Execute` of measurement manager.

clusterloader2/pkg/measurement/common/api_responsiveness.go
shyamjvs: Then please at least add a TODO.

clusterloader2/pkg/measurement/manager.go
shyamjvs: > It make metrics only be usable with this specific global variable. If anyone wants to use it in different project, he will need to assure that this global variable is initialized. 

If that's your problem, then just add another file in measurement pkg where you can initialize one single ClusterConfig object by calling some method at the start and then use the same object as a global constant for all measurements. Otherwise the problems with current style is:

- You're creating multiple copies of same global constant
- Storing constants as fields inside the struct
- Adding extra code complexity (due to new field that needs to be maintained, making function signatures more complex, asking measurement implementers to pass an extra arg in the Execute() method that they may not even need or have no clue about)

> Setting some external variable to change function result is not very intuitive.

I think this is the concept of global variables, and we do use them :)

> It also makes testing more complex.

It's not possible to write unit tests for these measurements even now. You're using methods like `SSH` which can't be mocked in the tests as the function is not passed as an argument to the measurement method - which should be the case in ideal unit-testing world where everything used in function should be passed as args to the function.

> If we provide all available information in the interface signature, it's easier to understand what parameters are available and which not.

I don't really agree with that we should pass all configurable stuff as params and not as constants. I believe in practice most people would just look at other measurement functions before implementing and it shouldn't be very non-trivial.

clusterloader2/cmd/clusterloader.go
```diff
@@ -39,7 +39,10 @@ var (
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 	// TODO(krzysied): query actual number of nodes with client (if not set).
```
wojtek-t: There is a separate PR doing that. Can you remove this comment in that PR.

clusterloader2/cmd/clusterloader.go
```diff
@@ -39,7 +39,10 @@ var (
...
-	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
+	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "Number of nodes")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "master-ip", "", "IP of the master node")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterName, "master-name", "", "Name of the master node")
```
wojtek-t: Why do you want this? This is implementation detail. What about HA clusters?

clusterloader2/cmd/clusterloader.go
```diff
@@ -39,7 +39,10 @@ var (
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 	// TODO(krzysied): query actual number of nodes with client (if not set).
-	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
+	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "Number of nodes")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "master-ip", "", "IP of the master node")
```
wojtek-t: Why do you need this? Again - what about HA clusters?

clusterloader2/cmd/clusterloader.go
```diff
@@ -39,7 +39,10 @@ var (
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 	// TODO(krzysied): query actual number of nodes with client (if not set).
-	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
+	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "Number of nodes")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
```
wojtek-t: What is the usecase for provider?

I know that there are features that may be provider specific (e.g. storage), but does it really matter?

clusterloader2/pkg/measurement/interface.go
```diff
@@ -28,6 +29,8 @@ type MeasurementConfig struct {
 	// into the Execute method.
 	Params map[string]interface{}
 	// TODO(krzysied): add CloudProvider.
```
wojtek-t: Why do you need CloudProvider?

clusterloader2/pkg/measurement/manager.go
wojtek-t: I would say, that ClusterConfig should be constant.
And if we agree on that, this should be a constant field for MeasurementManager 9and then you don't need to pass it here).

clusterloader2/cmd/clusterloader.go
krzysied: I assume that most of the variables will be automatically set based on data form cluster client. However, if someone wants to set specific value, there should be a way to do this.

clusterloader2/pkg/measurement/interface.go
krzysied: TBH, it was suggested by @shyamjvs . At this moment I'm not sure whether we will need cloud provider or not, so I would like to keep the TODO.

clusterloader2/pkg/measurement/manager.go
krzysied: I added the ClusterConfig variable to ```measurement/manger.go```.

clusterloader2/cmd/clusterloader.go
```diff
@@ -39,6 +40,10 @@ var (
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
+	// TODO(krzysied): Consider support for HA cluster with more than one master.
```
wojtek-t: It's not "consider" - we should do that.
Also - please open an issue for that.

clusterloader2/cmd/clusterloader.go
wojtek-t: Sure - I didn't mean to do that in this PR. I just want this to not be forgot.

clusterloader2/cmd/clusterloader.go
wojtek-t: So what's the point in exposing them now? Do we really need it now?
If not - let's not overgeneralize until it's really needed.

wojtek-t: So let me clarify my point: - I'm fine with making those part of ClusterLoader config.
But let's not expose them via flags until it's really needed.

krzysied: For example `master-ip` is used for sshing into the master node. Etcd metrics are using it to access the master. If master is not register AFAIK there is no way to calculate master ip - it needs to be provided.

wojtek-t: Even though in theory it's possible, we don't have such usecase for now.
So until it's really needed, let's avoid adding flags for such things.

krzysied: OK. Flags removed.

clusterloader2/cmd/clusterloader.go
```diff
@@ -39,6 +40,8 @@ var (
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
 	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
+	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
+	// TODO(krzysied): Add support for HA cluster with more than one master.
```
wojtek-t: This TODO should be moved somewhere else.
I thing that probably to the ClusterConfig definition.

clusterloader2/pkg/measurement/manager.go
```diff
@@ -20,6 +20,12 @@ import (
 	"sync"
 
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+)
+
+var (
+	// ClusterConfig represents cluster configuration.
+	ClusterConfig *config.ClusterConfig
```
wojtek-t: I'm not a fan of this global variable.

Can you make add a TODO to make this a class variable or parameter somewhere or something like that?

clusterloader2/cmd/clusterloader.go
krzysied: TODO moved to the ClusterConfig definition.

clusterloader2/pkg/measurement/manager.go
krzysied: TODO added.

wojtek-t: Instead of TODO(krzysied) we should move towards TODO(issue id).
Please open an issue for that and replace.

krzysied: Done.
I thought that putting TODO creator login is a best practice is k8s repo.

PR #230, opened 2018-09-19
ClusterLoader - Adding bool parsing from config

krzysied:
Adding support for bool values.
\+ Fixing typos.

### DIFF
--- a/clusterloader2/pkg/util/util.go
+++ b/clusterloader2/pkg/util/util.go
@@ -24,50 +24,96 @@ import (
 	"time"
 )
 
+// ErrKeyNotFound is returned when key doesn't exists in a map.
+type ErrKeyNotFound struct {
+	key string
+}
+
+// Erros is an error interface implementation.
+func (e *ErrKeyNotFound) Error() string {
+	return fmt.Sprintf("key %s not found", e.key)
+}
+
+// IsErrKeyNotFound returns true only if error type is ErrKeyNotFound.
+func IsErrKeyNotFound(err error) bool {
+	_, isErrKeyNotFound := err.(*ErrKeyNotFound)
+	return isErrKeyNotFound
+}
+
 // GetString tries to return value from map cast to string type. If value doesn't exist, error is returned.
 func GetString(dict map[string]interface{}, key string) (string, error) {
-	return getString(dict, key, "", fmt.Errorf("key %s not found", key))
+	return getString(dict, key)
 }
 
 // GetInt tries to return value from map cast to int type. If value doesn't exist, error is returned.
 func GetInt(dict map[string]interface{}, key string) (int, error) {
-	return getInt(dict, key, 0, fmt.Errorf("key %s not found", key))
+	return getInt(dict, key)
 }
 
 // GetFloat64 tries to return value from map cast to float64 type. If value doesn't exist, error is returned.
 func GetFloat64(dict map[string]interface{}, key string) (float64, error) {
-	return getFloat64(dict, key, 0, fmt.Errorf("key %s not found", key))
+	return getFloat64(dict, key)
 }
 
 // GetDuration tries to return value from map cast to duration type. If value doesn't exist, error is returned.
 func GetDuration(dict map[string]interface{}, key string) (time.Duration, error) {
-	return getDuration(dict, key, 0, fmt.Errorf("key %s not found", key))
+	return getDuration(dict, key)
+}
+
+// GetBool tries to return value from map cast to bool type. If value doesn't exist, error is returned.
+func GetBool(dict map[string]interface{}, key string) (bool, error) {
+	return getBool(dict, key)
 }
 
 // GetStringOrDefault tries to return value from map cast to string type. If value doesn't exist default value is used.
 func GetStringOrDefault(dict map[string]interface{}, key string, defaultValue string) (string, error) {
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/util/util.go
```diff
@@ -137,6 +147,25 @@ func getDuration(dict map[string]interface{}, key string, defaultValue time.Dura
 	return duration, nil
 }
 
+func getBool(dict map[string]interface{}, key string, defaultValue bool, defaultError error) (bool, error) {
+	value, exists := dict[key]
+	if !exists {
+		return defaultValue, defaultError
```
wojtek-t: Returning "defaultValue, defaultError" here (and everywhere else in this file) looks really strange here.

Instead you should have a dedicated "notFoundError" and return
"zeroValueForThatType", errNotFound

krzysied: OK. I've changed the functions co they return zero-value every time.

clusterloader2/pkg/util/util.go
```diff
@@ -24,50 +24,102 @@ import (
 	"time"
 )
 
+var errKeyNotFound = fmt.Errorf("key not found")
```
wojtek-t: You want to define a type for it, e.g. like here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/scheduler_assume_cache.go#L58

clusterloader2/pkg/util/util.go
```diff
@@ -24,50 +24,102 @@ import (
...
+
 // GetString tries to return value from map cast to string type. If value doesn't exist, error is returned.
 func GetString(dict map[string]interface{}, key string) (string, error) {
-	return getString(dict, key, "", fmt.Errorf("key %s not found", key))
+	value, err := getString(dict, key)
+	if err == errKeyNotFound {
```
wojtek-t: If you create the type I suggested, you won't need all of those strange iffs.

clusterloader2/pkg/util/util.go
```diff
@@ -24,50 +24,102 @@ import (
...
+		return value, fmt.Errorf("key %s not found", key)
+	}
+	return value, err
 }
 
 // GetStringOrDefault tries to return value from map cast to string type. If value doesn't exist default value is used.
 func GetStringOrDefault(dict map[string]interface{}, key string, defaultValue string) (string, error) {
-	return getString(dict, key, defaultValue, nil)
+	value, err := getString(dict, key)
```
wojtek-t: What are the usecases for "Get*OrDefault" - do we even use non-zero value for defaults?

krzysied: > // GetStringOrDefault tries to return value from map cast to string type. If value doesn't exist default value is used.

The assumption is that value can be optional - will not exists in the map. In this case you want to use default value (not necessary a zero value) specified for certain use case.

Use case:
In metric we can enable option to gather data only from master, so we use:
```util.GetBoolOrDefault(config.Params, "masterOnly", false)```
which assumes that user can specify value or default (false) will be used.

clusterloader2/pkg/util/util.go
krzysied: OK. Done.

clusterloader2/pkg/util/util.go
krzysied: I got rid of half of the ifs.
I think ifs for handling the default values are still needed. WYT?

clusterloader2/pkg/util/util.go
krzysied: There are additional two in `pkg/measurement/common/scheduler_latency.go`:
```host, err := util.GetStringOrDefault(config.Params, "host", "")```
```masterName, err := util.GetStringOrDefault(config.Params, "masterName", "")```

Four more in `pkg/measurement/common/wait_for_controlled_pods` (and also in `wait_for_pods`):
```namespace, err := util.GetStringOrDefault(config.Params, "namespace", metav1.NamespaceAll)```
```labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")```
```fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")```
```timeout, err := util.GetDurationOrDefault(config.Params, "timeout", defaultWaitForRCPodsTimeout)```

There are more use cases in metric I'm currently trying to add to clusterloader.
Assuming that measurements are the main part that will be developed, number of use cases will grow.

wojtek-t: Most of these aren't good examples, because they are using zero-value ("" is string zero-value).
NamespaceAll also turns out to be empty string :)

So timeout is the only case where you need it.

krzysied: > Most of these aren't good examples, because they are using zero-value ("" is string zero-value).
NamespaceAll also turns out to be empty string :)

But still I think that those are the places when you want to use already defined function instead of implementing error handling on your own. In these cases, you still want to use zero value **only if** value is not provided. Plus, having a way to define default value is convenient.

krzysied: After offline discussion we agreed to make following changes:
- Making the ErrKeyNotFound globally accessible.
- Added function to check whether a given error is ErrKeyNotFound type.

PR #236, opened 2018-10-01
ClusterLoader - Adding node number computing

krzysied:
Adding computing number of nodes if flag is not provided.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -38,7 +38,6 @@ var (
 
 func initClusterFlags() {
 	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
-	// TODO(krzysied): query actual number of nodes with client (if not set).
 	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
 }
 
@@ -46,9 +45,6 @@ func validateClusterFlags() *util.ErrorList {
 	errList := util.NewErrorList()
 	if clusterLoaderConfig.ClusterConfig.KubeConfigPath == "" {
 		errList.Append(fmt.Errorf("no kubeconfig path specified"))
-	}
-	if clusterLoaderConfig.ClusterConfig.Nodes < 1 {
-		errList.Append(fmt.Errorf("number of nodes not specified"))
 	}
 	return errList
 }
@@ -70,6 +66,18 @@ func validateFlags() *util.ErrorList {
 	return errList
 }
 
+func completeConfig(f *framework.Framework) error {
+	if clusterLoaderConfig.ClusterConfig.Nodes == 0 {
+		nodes, err := util.GetSchedulableUntainedNodesNumber(f.GetClientSet())
+		if err != nil {
+			return fmt.Errorf("Getting number of nodes error: %v", err)
+		}
+		clusterLoaderConfig.ClusterConfig.Nodes = nodes
+		glog.Infof("ClusterConfig.Nodes set to %v", nodes)
+	}
+	return nil
+}
+
 func main() {
 	defer glog.Flush()
 	initFlags()
@@ -84,6 +92,10 @@ func main() {
 	if err != nil {
 		glog.Fatalf("Framework creation error: %v", err)
 	}
+	if err = completeConfig(f); err != nil {
+		glog.Fatalf("Config completing error: %v", err)
+	}
+
 	for _, clusterLoaderConfig.TestConfigPath = range testConfigPaths {
 		glog.Infof("Running %v", clusterLoaderConfig.TestConfigPath)
 		if errList := test.RunTest(f, &clusterLoaderConfig); !errList.IsEmpty() {
--- a/clusterloader2/pkg/util/cluster.go
+++ b/clusterloader2/pkg/util/cluster.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"github.com/golang/glog"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// GetSchedulableUntainedNodesNumber returns number of nodes in the cluster.
+func GetSchedulableUntainedNodesNumber(c clientset.Interface) (int, error) {
+	nodeList, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return 0, err
+	}
+	numNodes := 0
+	for i := range nodeList.Items {
+		if isNodeSchedulable(&nodeList.Items[i]) && isNodeUntainted(&nodeList.Items[i]) {
+			numNodes++
+		}
+	}
+	return numNodes, err
+}
+
+// Node is schedulable if:
+// 1) doesn't have "unschedulable" field set
+// 2) it's Ready condition is set to true
+// 3) doesn't have NetworkUnavailable condition set to true
+func isNodeSchedulable(node *corev1.Node) bool {
+	nodeReady := isNodeConditionSetAsExpected(node, corev1.NodeReady, true, false)
+	networkReady := isNodeConditionUnset(node, corev1.NodeNetworkUnavailable) ||
+		isNodeConditionSetAsExpected(node, corev1.NodeNetworkUnavailable, false, true)
+	return !node.Spec.Unschedulable && nodeReady && networkReady
+}
+
+// Tests whether node doesn't have any taint with "NoSchedule" or "NoExecute" effect.
+func isNodeUntainted(node *corev1.Node) bool {
+	for i := range node.Spec.Taints {
+		if node.Spec.Taints[i].Effect == corev1.TaintEffectNoSchedule || node.Spec.Taints[i].Effect == corev1.TaintEffectNoExecute {
+			return false
+		}
+	}
+	return true
+}
+
+func isNodeConditionSetAsExpected(node *corev1.Node, conditionType corev1.NodeConditionType, wantTrue, silent bool) bool {
+	// Check the node readiness condition (logging all).
+	for _, cond := range node.Status.Conditions {
+		// Ensure that the condition type and the status matches as desired.
+		if cond.Type == conditionType {
+			if wantTrue == (cond.Status == corev1.ConditionTrue) {
+				return true
+			}
+			if !silent {
+				glog.Infof("Condition %s of node %s is %v instead of %t. Reason: %v, message: %v",
+					conditionType, node.Name, cond.Status == corev1.ConditionTrue, wantTrue, cond.Reason, cond.Message)
+			}
+			return false
+		}
+	}
+	if !silent {
+		glog.Infof("Couldn't find condition %v on node %v", conditionType, node.Name)
+	}
+	return false
+}
+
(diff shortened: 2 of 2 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -70,6 +67,19 @@ func validateFlags() *util.ErrorList {
 	return errList
 }
 
+func completeConfig(f *framework.Framework) error {
+	if clusterLoaderConfig.ClusterConfig.Nodes == 0 {
+		nodes, err := util.GetNodesNumber(f.GetClientSet())
+		if err == nil {
```
wojtek-t: The convention is:

if err != nil {
  return ...
}
// the rest

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,37 @@
...
+
+// GetNodesNumber returns number of nodes in the cluster.
+func GetNodesNumber(c clientset.Interface) (int, error) {
+	nodeList, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return 0, err
+	}
+	numNodes := 0
+	for i := range nodeList.Items {
+		if nodeList.Items[i].Spec.Unschedulable == false {
```
wojtek-t: There is question what about not-ready nodes (or in general, nodes that has taints preventing scheduling on that nodes).

In load tests (consciously) we decided for:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/load.go#L148

I would say - let's do the same.

clusterloader2/cmd/clusterloader.go

clusterloader2/pkg/util/cluster.go
krzysied: Done.
I copied the logic from test/framework. It created a lot of dependencies, though.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,157 @@
...
+// 3) doesn't have NetworkUnavailable condition set to true
+func isNodeSchedulable(node *corev1.Node) bool {
+	nodeReady := isNodeConditionSetAsExpected(node, corev1.NodeReady, true, false)
+	networkReady := isNodeConditionUnset(node, corev1.NodeNetworkUnavailable) ||
+		isNodeConditionSetAsExpected(node, corev1.NodeNetworkUnavailable, false, true)
+	return !node.Spec.Unschedulable && nodeReady && networkReady
+}
+
+// Test whether a fake pod can be scheduled on "node", given its current taints.
+func isNodeUntainted(node *corev1.Node) bool {
```
wojtek-t: I don't like this amount of dependencies.

But in general, this is a bit of overkill. You can pretty much copy this one instead:
https://github.com/kubernetes/kubernetes/blob/807eb6b24b953ff3f59c8049eff7e0edbc279e18/pkg/scheduler/algorithm/predicates/predicates.go#L1506

and you won't even need those dependencies at all.

krzysied: Done.
I've also copied variables from `pkg/controller/nodelifecycle`, so dependency to the package could be removed.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,185 @@
...
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "fake-not-scheduled",
+					Image: "fake-not-scheduled",
+				},
+			},
+		},
+	}
+	nodeInfo := schedulercache.NewNodeInfo()
```
wojtek-t: This is overkill. All you need are taints and it's easier to extract them from the node itself.

[And you won't need scheduler dependency]

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,185 @@
...
+		return false, err
+	}
+
+	if v1helper.TolerationsTolerateTaintsWithFilter(pod.Spec.Tolerations, taints, filter) {
+		return true, nil
+	}
+	return false, nil
+}
+
+func isNodeConditionSetAsExpected(node *corev1.Node, conditionType corev1.NodeConditionType, wantTrue, silent bool) bool {
```
wojtek-t: Copying this function is unfortunate - it's pretty complicated...

krzysied: The problem is that original function is located at `kubernete/test/e2e/framework`. If I vendor this package, I also need to vendor the whole kubernetes.
I will try to simplify to the code.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,177 @@
...
+		return t.Effect == corev1.TaintEffectNoSchedule || t.Effect == corev1.TaintEffectNoExecute
+	})
+	if err != nil {
+		glog.Errorf("Can't test predicates for node %s: %v", node.Name, err)
+		return false
+	}
+	return fit
+}
+
+func podToleratesNodeTaints(pod *corev1.Pod, node *corev1.Node, filter func(t *corev1.Taint) bool) (bool, error) {
```
wojtek-t: this never returns error - so let's remove it from signature :)

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,177 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+	"k8s.io/kubernetes/pkg/scheduler/algorithm"
+)
+
+var (
+	// UnreachableTaintTemplate is the taint for when a node becomes unreachable.
+	unreachableTaintTemplate = &corev1.Taint{
+		Key:    algorithm.TaintNodeUnreachable,
```
wojtek-t: It's so freaking ridiculous that you need to important half of Kubernetes to get those constants.
Those constants should be moved to a package that doesn't depend on anything else.

So let's do the following:
- instead of importing those constants, let's hardcode it here
- add a TODO to important those constants once they can be imported without importating 50 other packages that are useless...

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,179 @@
...
+	return false
+}
+
+func isNodeConditionSetAsExpected(node *corev1.Node, conditionType corev1.NodeConditionType, wantTrue, silent bool) bool {
+	// Check the node readiness condition (logging all).
+	for _, cond := range node.Status.Conditions {
+		// Ensure that the condition type and the status matches as desired.
+		if cond.Type == conditionType {
+			// For NodeReady condition we need to check Taints as well
+			if cond.Type == corev1.NodeReady {
```
wojtek-t: I would really like to simplify this function.
In particular:

1. do we really need to check taints? we are checking those separately in isNodeUntainted anyway.
Or do I miss something?

2. Probably more, I think. But let's ensure that taints aren't really needed and that would simplify stuff..

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,179 @@
...
+
+	fit := podToleratesNodeTaints(fakePod, node, func(t *corev1.Taint) bool {
+		// PodToleratesNodeTaints is only interested in NoSchedule and NoExecute taints.
+		return t.Effect == corev1.TaintEffectNoSchedule || t.Effect == corev1.TaintEffectNoExecute
+	})
+	return fit
+}
+
+func podToleratesNodeTaints(pod *corev1.Pod, node *corev1.Node, filter func(t *corev1.Taint) bool) bool {
+	if v1helper.TolerationsTolerateTaintsWithFilter(pod.Spec.Tolerations, node.Spec.Taints, filter) {
```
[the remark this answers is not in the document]
krzysied: TBH, I look into TolerationsTolerateTaintsWithFilter implementation and it looks like for every taint (if it satisfy the filer functions) and for every toleration check `toleration.ToleratesTaint(taint)`. Assuming that there is no toleration passed, this whole function does nothing...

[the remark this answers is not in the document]
wojtek-t: No - that's not true.

It iterates over all taints (which comes from node) and checks if there is a corresponding tolleration for each of them.
If there isn't (and taint will cause pods to be not scheduled there) it will return false (and that is exactly the case we want to catch - if there are any taints than).

This is exactly what we want to do.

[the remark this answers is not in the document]
krzysied: You are right.
I misunderstood one case.

clusterloader2/pkg/util/cluster.go
krzysied: If I understand isNodeUntainted correctly, it will return false if any of taints has TaintEffectNoExecute effect. Assuming that unreachableTaintTemplate and notReadyTaintTemplate also have the TaintEffectNoExecute effect, this case will be verified by isNodeUntainted.

I agree that taints check can be removed.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,179 @@
...
+			Containers: []corev1.Container{
+				{
+					Name:  "fake-not-scheduled",
+					Image: "fake-not-scheduled",
+				},
+			},
+		},
+	}
+
+	fit := podToleratesNodeTaints(fakePod, node, func(t *corev1.Taint) bool {
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,179 @@
...
+		// PodToleratesNodeTaints is only interested in NoSchedule and NoExecute taints.
+		return t.Effect == corev1.TaintEffectNoSchedule || t.Effect == corev1.TaintEffectNoExecute
+	})
+	return fit
+}
+
+func podToleratesNodeTaints(pod *corev1.Pod, node *corev1.Node, filter func(t *corev1.Taint) bool) bool {
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/util/cluster.go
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/util/cluster.go
krzysied: After removing taint part, it turned out that the whole `NodeReady` case check is not needed.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"github.com/golang/glog"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+)
+
+// GetNodesNumber returns number of nodes in the cluster.
+func GetNodesNumber(c clientset.Interface) (int, error) {
```
wojtek-t: GetSchedulableUntainedNodesNumber

- we need to be pretty precise here.

clusterloader2/pkg/util/cluster.go
```diff
@@ -0,0 +1,90 @@
...
+		return t.Effect == corev1.TaintEffectNoSchedule || t.Effect == corev1.TaintEffectNoExecute
+	})
+}
+
+func isNodeConditionSetAsExpected(node *corev1.Node, conditionType corev1.NodeConditionType, wantTrue, silent bool) bool {
+	// Check the node readiness condition (logging all).
+	for _, cond := range node.Status.Conditions {
+		// Ensure that the condition type and the status matches as desired.
+		if cond.Type == conditionType {
+			if (wantTrue && (cond.Status == corev1.ConditionTrue)) || (!wantTrue && (cond.Status != corev1.ConditionTrue)) {
```
wojtek-t: Why not:

```wantTrue == (cond.Status == corev1.ConditionTrue) ```

clusterloader2/pkg/util/cluster.go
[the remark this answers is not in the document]
krzysied: So, if I understand correctly, it just loops over the taints and return false if any of them satisfy the filter function. Maybe I should just implement the loop here, so no vendor will be needed?

clusterloader2/pkg/util/cluster.go
krzysied: Done.

clusterloader2/pkg/util/cluster.go
krzysied: Done.

clusterloader2/pkg/util/cluster.go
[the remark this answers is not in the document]
krzysied: I've removed the helper call.

PR #242, opened 2018-10-12
ClusterLoader - density test update

krzysied:
Adding multiple iterations of latency pods creation.

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -1,18 +1,25 @@
+# ASSUMPTIONS:
+# - Underlying cluster should have 100+ nodes.
+# - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
+
 #Constants
 {{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
+{{$PODS_PER_NAMESPACE := 3000}}
+{{$MIN_LATENCY_PODS := 500}}
+{{$MIN_SATURATION_PODS_TIMEOUT := 180}}
+{{$TEST_THROUGHPUT := 20}}
+
 #Variables
-{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
-{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
-{{$replicas := DivideInt $totalPods $namespaces}}
-{{$totalPodsRounded := MultiplyInt $namespaces $replicas}}
-#Test
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$totalPods := MultiplyInt $PODS_PER_NAMESPACE $namespaces}}
+{{$latencyReplicas := DivideInt (MaxInt $MIN_LATENCY_PODS .Nodes) $namespaces}}
+{{$totalLatencyPods := MultiplyInt $namespaces $latencyReplicas}}
+{{$saturationRCTimeout := DivideFloat $totalPods $TEST_THROUGHPUT | AddInt $MIN_SATURATION_PODS_TIMEOUT}}
+
+
 name: density
 automanagedNamespaces: {{$namespaces}}
 tuningSets:
-- name: Uniform100qps
-  qpsLoad:
-    qps: 100
 - name: Uniform5qps
   qpsLoad:
     qps: 5
@@ -27,80 +34,103 @@ steps:
     Params:
       action: reset
 # Create saturation pods
+- measurements:
+  - Identifier: WaitForRunningSaturationRCs
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      kind: ReplicationController
+      labelSelector: group = saturation
+      operationTimeout: {{$saturationRCTimeout}}s
 - phases:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
     replicasPerNamespace: 1
-    tuningSet: Uniform100qps
+    tuningSet: Uniform5qps
     objectBundle:
     - basename: saturation-rc
       objectTemplatePath: rc.yaml
       templateFillMap:
-        Replicas: {{$replicas}}
-        PodKind: saturationPod
+        Replicas: {{$PODS_PER_NAMESPACE}}
+        Group: saturation
 - measurements:
-  - Identifier: WaitForSaturationRunning
-    Method: WaitForRunningPods
+  - Identifier: WaitForRunningSaturationRCs
+    Method: WaitForControlledPodsRunning
     Params:
-      desiredPodCount: {{$totalPodsRounded}}
-      labelSelector: kind = saturationPod
-      # timeout = (totalPodsRounded /20)s + 3m
-      timeout: {{DivideInt $totalPodsRounded 20 | AddInt 180}}s
+      action: gather
+- name: Creating saturation pods
+# Create latency pods
 - measurements:
   - Identifier: PodStartupLatency
     Method: PodStartupLatency
     Params:
       action: start
-      labelSelector: kind = latencyPod
-- name: Creating saturation pods
-# Create latency pods
+      labelSelector: group = latency
+- measurements:
+  - Identifier: WaitForRunningLatencyRCs
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      kind: ReplicationController
+      labelSelector: group = latency
+      operationTimeout: 10m
 - phases:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{DivideInt .Nodes $namespaces}}
+    replicasPerNamespace: {{$latencyReplicas}}
     tuningSet: Uniform5qps
     objectBundle:
     - basename: latency-pod-rc
       objectTemplatePath: rc.yaml
       templateFillMap:
         Replicas: 1
-        PodKind: latencyPod
+        Group: latency
 - measurements:
-  - Identifier: WaitForLatencyPodRunning
-    Method: WaitForRunningPods
+  - Identifier: WaitForRunningLatencyRCs
+    Method: WaitForControlledPodsRunning
     Params:
-      desiredPodCount: {{.Nodes}}
-      labelSelector: kind = latencyPod
-      timeout: 10m
+      action: gather
+- name: Creating latency pods
+# Remove latency pods
+- phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: Uniform5qps
+    objectBundle:
+    - basename: latency-pod-rc
+      objectTemplatePath: rc.yaml
+- measurements:
+  - Identifier: WaitForRunningLatencyRCs
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
 - measurements:
   - Identifier: PodStartupLatency
     Method: PodStartupLatency
     Params:
       action: gather
-- name: Creating latency pods
+- name: Deleting latancy pods
 # Delete pods
 - phases:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
     replicasPerNamespace: 0
-    tuningSet: Uniform100qps
-    objectBundle:
-    - basename: latency-pod-rc
-      objectTemplatePath: rc.yaml
-- name: Deleting latancy pods
-- phases:
-  - namespaceRange:
-      min: 1
-      max: {{$namespaces}}
-    replicasPerNamespace: 0
-    tuningSet: Uniform100qps
+    tuningSet: Uniform5qps
     objectBundle:
     - basename: saturation-rc
       objectTemplatePath: rc.yaml
+- measurements:
+  - Identifier: WaitForRunningSaturationRCs
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
 - name: Deleting saturation pods
+# Collect measurements
 - measurements:
   - Identifier: APIResponsiveness
     Method: APIResponsiveness
--- a/clusterloader2/testing/density/rc.yaml
+++ b/clusterloader2/testing/density/rc.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: ReplicationController
 metadata:
   name: {{.Name}}
+  labels:
+    group: {{.Group}}
 spec:
   replicas: {{.Replicas}}
   selector:
@@ -10,7 +12,7 @@ spec:
     metadata:
       labels:
         name: {{.Name}}
-        kind: {{.PodKind}}
+        group: {{.Group}}
     spec:
       containers:
       - image: k8s.gcr.io/pause:3.1


clusterloader2/pkg/framework/client/objects.go
```diff
@@ -172,7 +172,9 @@ func PatchObject(dynamicClient dynamic.Interface, namespace string, name string,
 func DeleteObject(dynamicClient dynamic.Interface, gvk schema.GroupVersionKind, namespace string, name string) error {
 	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
 	deleteFunc := func() error {
-		return dynamicClient.Resource(gvr).Namespace(namespace).Delete(name, nil)
+		falseVar := false
+		deleteOption := &metav1.DeleteOptions{OrphanDependents: &falseVar}
```
wojtek-t: This means that deletion is no longer a default k8s deletion.

I'm fine with that semantic, but we need to be very clear about it:
- in comments in this file
- in design doc (it should be updated together with this PR)
- additionally in readme too

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,14 @@
 #Constants
 {{$NODES_PER_NAMESPACE := 100}}
 {{$PODS_PER_NODE := 30}}
+{{$MIN_LATENCY_PODS := 500}}
```
wojtek-t: Why do we make it constant?

clusterloader2/testing/density/config.yaml
```diff
@@ -69,12 +72,39 @@ steps:
     Params:
       action: start
       labelSelector: kind = latencyPod
+{{range $iteration := (RangeInt 1 $latencyPodsIterations)}}
```
wojtek-t: I hate it - it starts looking like coding in yaml.

Why do we need iterations?
Why can't we simply created all latency pods in a single iteration?
I don't see any reason why we would need iterations.

krzysied: > I hate it - it starts looking like coding in yaml.

I agree. However I'm not sure what's the alternative.

> Why do we need iterations?

I would like to implement (more or less) the same logic as in original test - this way it's easier to compare the results.

> Why can't we simply created all latency pods in a single iteration?

That's tricky. Assuming that we already have a lot of saturation pods, adding 1 pod is not a problem. Adding 5 can exceed the quota. (Needs to be verified.)
It also depends on what we want to achieve and what we want to measure. Creating 5 times 20 pods and deleting them is different than creating and deleting 100 pods.

wojtek-t: So I can tell you what the motivation for that was.
In general, if there were two few pods, a single pods may affect 99th percentile. So we want to schedule bit enough number of those to avoid such situation.

TBH, I'm not convinced that what exactly was done is the best way. And while in general I agree replicating the test existing is better, we should do it carefully.

In this particular case, I really think that we don't want to do iteration - this precedent is too dangerous.
If you remember, I already told you in the past that I'm against such configs, and I would prefer hardcoding stuff. But I gave up.
But this is waaay to much for me.

re quota issues - it's about maxPodsPerNode, not quota.
And for all tests we run, we would be fine even with 5 pods.

Let's just change it to schedule max(500, #nodes) latency pods and not introduce iterations.

In fact, I still believe this test is too complicated due to all those fancy parameters that noone will never change...

clusterloader2/pkg/framework/client/objects.go
krzysied: Added in 3rd commit.

clusterloader2/testing/density/config.yaml
krzysied: Iteration removed.
The only problem is that max(500, #nodes) works really bad if someone uses it with 5-nodes cluster.

wojtek-t: This test has other problems for small clusters...

Let's add a comment that this test is supposed to run on clusters larger than X (50?, 100?) and that's fine.

krzysied: Comment added.

> This test has other problems for small clusters...

The results was meaningless in terms of scalability - the sample was to small, right?
However, it was good as an example - everyone could run it on any cluster.

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,17 @@
...
+
 #Constants
 {{$NODES_PER_NAMESPACE := 100}}
 {{$PODS_PER_NODE := 30}}
+{{$MIN_LATENCY_PODS := 500}}
 #Variables
 {{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
 {{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
 {{$replicas := DivideInt $totalPods $namespaces}}
 {{$totalPodsRounded := MultiplyInt $namespaces $replicas}}
```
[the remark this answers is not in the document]
krzysied: > The first thing that I would do is to understand why we really want all those Timers?

The main reasons are:
- it creates json file with all timers measurements. It can be easily parsed.
- allows to measure execution time between any two points.
- allows to connect execution time with a label. Even if we add timers in background to every phase, we still want to name these phases.

> One quick observation is that if you want to measure all phases, maybe that should just happen in the background.

I would be OK with this solution. However it's only because I know how cluster loader works.
IMHO if we want cluster loader to user friendly, we should have more explicit declaration approach instead of things running in the background.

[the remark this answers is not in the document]
wojtek-t: What I'm saying is that I don't see real benefits from having those (at least not enough benefits to justify 25% bigger config).
1. We are not tracking those times anywhere now
2. Just having a log" Phase %s started", "phase %s ended" is more than we need now.

I agree that if we want to track latency of some phase, it's useful. But we really don't care about it now.

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -172,7 +172,10 @@ func PatchObject(dynamicClient dynamic.Interface, namespace string, name string,
 func DeleteObject(dynamicClient dynamic.Interface, gvk schema.GroupVersionKind, namespace string, name string) error {
 	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
 	deleteFunc := func() error {
-		return dynamicClient.Resource(gvr).Namespace(namespace).Delete(name, nil)
+		// Delete operation removes object with all of the dependants.
+		falseVar := false
+		deleteOption := &metav1.DeleteOptions{OrphanDependents: &falseVar}
+		return dynamicClient.Resource(gvr).Namespace(namespace).Delete(name, deleteOption)
```
wojtek-t: Can you split all of the above to a separate PR?

I think this is much less controversial.
Wrt to changes to the test itself, I think it requires significantly more work.

krzysied: Code changes moved to #255.

krzysied:
/hold
Waiting for #255 to be merged.

clusterloader2/testing/density/config.yaml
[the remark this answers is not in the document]
krzysied: > We are not tracking those times anywhere now

There is a problem with Timer name - I need to revert it's name to `TestPhaseTimer` (backward compatibility). If I do this, the results will appear on perfdash.

> Just having a log" Phase %s started", "phase %s ended" is more than we need now

Just to clarify, what is an example of a text that %s will be replaced with?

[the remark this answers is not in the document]
wojtek-t: Discussed offline.
To summarize:
- let's introduce an optional "name" field to Step
- if it is defined, fire a timer for that step automatically

clusterloader2/testing/density/config.yaml
```diff
@@ -90,7 +96,27 @@ steps:
...
+# Remove latency pods
+- phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: Uniform5qps
+    objectBundle:
+    - basename: latency-pod-rc
+      objectTemplatePath: rc.yaml
```
wojtek-t: If replicasPerNamespace is 0 (or in general smaller than before so that it results in deletion of objects),
do we really need to pass objectTemplatePath and templateFillMap?
Conceptually this shouldn't be needed.

krzysied: > do we really need to pass objectTemplatePath (...)?

Yes. Previously there was `ObjectKind` structure that described kind and api group. Currently this data has to be taken from yaml file. It required due to possible names collision.

> do we really need to pass (...) templateFillMap?

No. It's not required. I've removed it from test.

wojtek-t: OK thanks. That makes sense.

clusterloader2/testing/density/config.yaml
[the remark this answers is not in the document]
krzysied: Adding "name" field - #257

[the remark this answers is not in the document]
krzysied: Adding timer - #258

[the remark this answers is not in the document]
wojtek-t: Well - it's not timer per se, just some logs. But yes - it's definitely enough (at least for now).

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,17 @@
+# This test should be run on cluster with 100+ nodes!
+
 #Constants
 {{$NODES_PER_NAMESPACE := 100}}
 {{$PODS_PER_NODE := 30}}
+{{$MIN_LATENCY_PODS := 500}}
 #Variables
 {{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
 {{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
```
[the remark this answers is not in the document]
krzysied: Assumption added.
It solved the problem with saturation pods, but latency pods still require rounding (due to the `MaxInt` being used).

krzysied:
@wojtek-t 
Code is rebased. All changes from other PRs are applied.

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,15 @@
+# This test should be run on cluster with 100+ nodes!
```
wojtek-t:
```
# ASSUMPTIONS:
- Underlying cluster should have 100+ nodes.
- Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
```

clusterloader2/testing/density/config.yaml
```diff
@@ -37,16 +41,16 @@ steps:
...
       templateFillMap:
-        Replicas: {{$replicas}}
+        Replicas: {{$PODS_PER_NAMESPACE}}
         PodKind: saturationPod
 - measurements:
   - Identifier: WaitForSaturationRunning
     Method: WaitForRunningPods
     Params:
-      desiredPodCount: {{$totalPodsRounded}}
+      desiredPodCount: {{$totalPods}}
```
wojtek-t: I'm not convinced we should do this - I think that more standard our tests are as a whole, the better.

As a result, I would prefer using the same measurements in both tests when possible.
So I think we should use:
WaitForControlledPodsRunning (given that we have RCs here anyway).

And then you won't need totalPods'
The same observation is for latencyPods.

krzysied: Ok. I see your point.
I assume that this PR is blocked until #264 is merged. Or should I use the old version of WaitForControlledPodsRunning?

wojtek-t: Yes - it's blocked.

krzysied:
#264 merged. PR rebased.

krzysied:
I applied changes from #264 . Currently we are waiting for RCs being ready instead of waiting just for pods. There are two separate measurement instances. One for saturation RCs, the other for latency RCs.  I also changed the operationTimeout for saturation RCs, due to 20pods/s assumption.
@wojtek-t PTAL

clusterloader2/testing/density/config.yaml
wojtek-t: please fix

clusterloader2/testing/density/config.yaml
```diff
@@ -55,42 +67,60 @@ steps:
...
+- measurements:
+  - Identifier: PodStartupLatency
+    Method: PodStartupLatency
+    Params:
+      action: gather
+- measurements:
+  - Identifier: WaitForRunningLatencyRCs
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: stop
```
wojtek-t: Do we need to stop measurements explicitly?
This is additional 5 lines of the code (2 times) that are boilerplate.

If we need, can we introduce an action "gather-and-stop" (or sth like that)?

clusterloader2/testing/density/rc.yaml
```diff
@@ -10,7 +12,9 @@ spec:
     metadata:
       labels:
         name: {{.Name}}
+{{if .PodKind}}
```
wojtek-t: No ifs please. Seriously - this is making this unreadable.

wojtek-t: We should reuse the same label. Maybe sth like:
group: {{.Group}}

[where group is saturation or latency]?

clusterloader2/testing/density/config.yaml
krzysied: I think in this case should be ok to not stop the measurement, although I'm not a fan of this approach.

I would prefer to add `Finalize()` function to measurement interface. This function would be called for every measurement instance during cleanup phase. Wdyt?

clusterloader2/testing/density/config.yaml

clusterloader2/testing/density/config.yaml
wojtek-t: > I would prefer to add Finalize() function to measurement interface. This function would be called for every measurement instance during cleanup phase. Wd

This also sounds fine.

clusterloader2/testing/density/rc.yaml
krzysied: `if` removed.

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,19 @@
...
 #Variables
-{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
-{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
-{{$replicas := DivideInt $totalPods $namespaces}}
-{{$totalPodsRounded := MultiplyInt $namespaces $replicas}}
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$totalPods := MultiplyInt $PODS_PER_NAMESPACE $namespaces}}
+{{$latencyReplicas := DivideInt (MaxInt $MIN_LATENCY_PODS .Nodes) $namespaces}}
+{{$totalLatencyPods := MultiplyInt $namespaces $latencyReplicas}}
+{{$saturationRCTimeout := MaxInt $MIN_SATURATION_PODS_TIMEOUT_S (DivideFloat $totalPods $POD_THROUGHPUT_PER_S)}}
```
wojtek-t: That is too short:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/density.go#L637

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,19 @@
+# ASSUMPTIONS:
+# - Underlying cluster should have 100+ nodes.
+# - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
+
 #Constants
 {{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
+{{$PODS_PER_NAMESPACE := 3000}}
+{{$MIN_LATENCY_PODS := 500}}
+{{$MIN_SATURATION_PODS_TIMEOUT_S := 120}}
```
wojtek-t: I don't think that's needed.

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,19 @@
...
+{{$POD_THROUGHPUT_PER_S := 20}}
```
wojtek-t: TEST_THROUGHPUT ?

clusterloader2/testing/density/config.yaml
krzysied: I'm still using it to calculate saturationRCTimeout. Value changed to 180.

clusterloader2/testing/density/config.yaml

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,19 @@
+# ASSUMPTIONS:
+# - Underlying cluster should have 100+ nodes.
+# - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
+
 #Constants
 {{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
+{{$PODS_PER_NAMESPACE := 3000}}
+{{$MIN_LATENCY_PODS := 500}}
+{{$MIN_SATURATION_PODS_TIMEOUT := 180}}
```
wojtek-t: I wouldn't call that min. because we add it every time.
OTOH, I don't have suggestion for better name...

clusterloader2/testing/density/config.yaml
```diff
@@ -37,16 +53,14 @@ steps:
     - basename: saturation-rc
       objectTemplatePath: rc.yaml
       templateFillMap:
-        Replicas: {{$replicas}}
+        Replicas: {{$PODS_PER_NAMESPACE}}
+        RCKind: saturationRC
```
wojtek-t: I think I wasn't clear enough. I don't want to have two separate labels - this should be the same label.
It's very common pattern that you propagate a bunch of labels of grouping object to the pod template.

We should reuse the same label (let's call it "group") and set it to "saturation" or "latency".

clusterloader2/testing/density/config.yaml
```diff
@@ -55,42 +69,55 @@ steps:
       labelSelector: kind = latencyPod
 - name: Creating saturation pods
 # Create latency pods
+- measurements:
```
wojtek-t: For consistency - the comment should be before the measurement.

clusterloader2/testing/density/config.yaml
```diff
@@ -37,16 +53,14 @@ steps:
     - basename: saturation-rc
```
wojtek-t: I don't see a good reason for introducing "Uniform100qps" tuningset.
To simplify the config, let's remove this and use Uniform5qps here too.

clusterloader2/testing/density/config.yaml
```diff
@@ -1,11 +1,19 @@
...
-{{$totalPods := MultiplyInt $PODS_PER_NODE .Nodes}}
-{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
-{{$replicas := DivideInt $totalPods $namespaces}}
-{{$totalPodsRounded := MultiplyInt $namespaces $replicas}}
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$totalPods := MultiplyInt $PODS_PER_NAMESPACE $namespaces}}
+{{$latencyReplicas := DivideInt (MaxInt $MIN_LATENCY_PODS .Nodes) $namespaces}}
+{{$totalLatencyPods := MultiplyInt $namespaces $latencyReplicas}}
+{{$saturationRCTimeout := DivideFloat $totalPods $TEST_THROUGHPUT | AddInt $MIN_SATURATION_PODS_TIMEOUT}}
 #Test
```
wojtek-t: s/#Test/empty line/
[or maybe two empty lines?]

clusterloader2/testing/density/config.yaml

clusterloader2/testing/density/config.yaml
krzysied: OK. Done.

clusterloader2/testing/density/config.yaml

PR #253, opened 2018-10-17
ClusterLoader - Updating load test

krzysied:
- Creating all services and then all RCs.
- Fixing tuning sets.

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -1,36 +1,39 @@
+# ASSUMPTIONS:
+# - Underlying cluster should have 100+ nodes.
+# - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
+
 #Constants
 {{$BIG_GROUP_SIZE := 250}}
 {{$MEDIUM_GROUP_SIZE := 30}}
 {{$SMALL_GROUP_SIZE := 5}}
-{{$PODS_PER_NODE := 30}}
 {{$NODES_PER_NAMESPACE := 100}}
-#Variables
+{{$PODS_PER_NODE := 30}}
+{{$TEST_THROUGHPUT := 10}}
 # bigRCs - 1/4 of all pods should be in big RCs.
-{{$bigRCs := MultiplyInt .Nodes 0.03}}
+{{$BIG_RCS_PER_NAMESPACE := 3}}
 # mediumRCs - 1/4 of all pods should be in medium RCs.
-{{$mediumRCs := MultiplyInt .Nodes 0.25}}
+{{$MEDIUM_RCS_PER_NAMESPACE := 25}}
 # smallRCs - 1/2 of all pods should be in small RCs.
-{{$smallRCs := MultiplyInt .Nodes 3}}
-{{$namespaces := DivideInt (SubtractInt .Nodes 1) $NODES_PER_NAMESPACE | AddInt 1}}
-{{$bigRCsPerNamespace := DivideInt $bigRCs $namespaces}}
-{{$mediumRCsPerNamespace := DivideInt $mediumRCs $namespaces}}
-{{$smallRCsPerNamespace := DivideInt $smallRCs $namespaces}}
-{{$totalPodsRounded := AddInt (MultiplyInt $bigRCs $BIG_GROUP_SIZE) (MultiplyInt $mediumRCs $MEDIUM_GROUP_SIZE) (MultiplyInt $smallRCs $SMALL_GROUP_SIZE)}}
-# saturationQPS - qps for RC operations should be an equivalent to 10 [pods/s].
-{{$saturationQPS := DivideFloat (AddInt $bigRCs $mediumRCs $smallRCs) (DivideFloat $totalPodsRounded 10)}}
-#Test
+{{$SMALL_RCS_PER_NAMESPACE := 300}}
+
+#Variables
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
+{{$saturationTime := DivideInt $totalPods $TEST_THROUGHPUT}}
+
+
 name: load
 automanagedNamespaces: {{$namespaces}}
 tuningSets:
-- name: AvgSaturationQPS
-  randomizedLoad:
-    averageQps: {{$saturationQPS}}
-- name: AvgScalingQPS
-  randomizedLoad:
+- name: RandomizedSaturationTimeLimited
+  RandomizedTimeLimitedLoad:
+    timeLimit: {{$saturationTime}}s
+- name: RandomizedScalingTimeLimited
+  RandomizedTimeLimitedLoad:
     # The expected number of created/deleted pods is totalPods/4 when scaling,
     # as each RC changes its size from X to a uniform random value in [X/2, 3X/2].
-    # To match 10 [pods/s] requirement, we need to multiply saturationQPS by 4.
-    averageQps: {{MultiplyFloat $saturationQPS 4}}
+    # To match 10 [pods/s] requirement, we need to divide saturationTime by 4.
+    timeLimit: {{DivideInt $saturationTime 4}}s
 steps:
 - measurements:
   - Identifier: APIResponsiveness
@@ -44,14 +47,14 @@ steps:
     Params:
       action: start
       kind: ReplicationController
-      labelSelector: kind = loadRC
+      labelSelector: group = load
       operationTimeout: 10m
 - phases:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{$bigRCsPerNamespace}}
-    tuningSet: AvgSaturationQPS
+    replicasPerNamespace: {{$BIG_RCS_PER_NAMESPACE}}
+    tuningSet: RandomizedSaturationTimeLimited
     objectBundle:
     - basename: big-service
       objectTemplatePath: service.yaml
@@ -65,8 +68,8 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{$mediumRCsPerNamespace}}
-    tuningSet: AvgSaturationQPS
+    replicasPerNamespace: {{$MEDIUM_RCS_PER_NAMESPACE}}
+    tuningSet: RandomizedSaturationTimeLimited
     objectBundle:
     - basename: medium-service
       objectTemplatePath: service.yaml
@@ -80,8 +83,8 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{$smallRCsPerNamespace}}
-    tuningSet: AvgSaturationQPS
+    replicasPerNamespace: {{$SMALL_RCS_PER_NAMESPACE}}
+    tuningSet: RandomizedSaturationTimeLimited
     objectBundle:
     - basename: small-service
       objectTemplatePath: service.yaml
@@ -103,8 +106,8 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{$bigRCsPerNamespace}}
-    tuningSet: AvgScalingQPS
+    replicasPerNamespace: {{$BIG_RCS_PER_NAMESPACE}}
+    tuningSet: RandomizedScalingTimeLimited
     objectBundle:
     - basename: big-rc
       objectTemplatePath: rc.yaml
@@ -114,8 +117,8 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{$mediumRCsPerNamespace}}
-    tuningSet: AvgScalingQPS
+    replicasPerNamespace: {{$MEDIUM_RCS_PER_NAMESPACE}}
+    tuningSet: RandomizedScalingTimeLimited
     objectBundle:
     - basename: medium-rc
       objectTemplatePath: rc.yaml
@@ -125,8 +128,8 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: {{$smallRCsPerNamespace}}
-    tuningSet: AvgScalingQPS
+    replicasPerNamespace: {{$SMALL_RCS_PER_NAMESPACE}}
+    tuningSet: RandomizedScalingTimeLimited
     objectBundle:
     - basename: small-rc
       objectTemplatePath: rc.yaml
@@ -145,7 +148,7 @@ steps:
       min: 1
       max: {{$namespaces}}
     replicasPerNamespace: 0
-    tuningSet: AvgSaturationQPS
+    tuningSet: RandomizedSaturationTimeLimited
     objectBundle:
     - basename: big-service
       objectTemplatePath: service.yaml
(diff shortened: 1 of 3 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -17,20 +17,37 @@
 {{$smallRCsPerNamespace := DivideInt $smallRCs $namespaces}}
 {{$totalPodsRounded := AddInt (MultiplyInt $bigRCs $BIG_GROUP_SIZE) (MultiplyInt $mediumRCs $MEDIUM_GROUP_SIZE) (MultiplyInt $smallRCs $SMALL_GROUP_SIZE)}}
 # saturationQPS - qps for RC operations should be an equivalent to 10 [pods/s].
-{{$saturationQPS := DivideFloat (AddInt $bigRCs $mediumRCs $smallRCs) (DivideFloat $totalPodsRounded 10)}}
+{{$bigSaturationQPS := DivideFloat $bigRCs (DivideFloat $totalPodsRounded 10)}}
+{{$mediumSaturationQPS := DivideFloat $mediumRCs (DivideFloat $totalPodsRounded 10)}}
```
wojtek-t: TBH, I don't think it's exaclty what we want.
I think we should go a different direction to make the test much simpler.

What we know is that we would like to generate average 10pods/s throughput.
So I would suggest introducing a different TuningSet type like:
RandomizedLoad
but where you pass total time over which you spread operations, not the QPS.

That would make it waaaay simpler, because all you will need is one tuningset, not one tunningset per type.

krzysied: > That would make it waaaay simpler, because all you will need is one tuningset, not one tunningset per type.

I guess I have to disagree. If you want to spreed something randomly, you also have to know the number of action to be spread.
Here I see three options:
 - Tuning set takes time and number of operations as a parameters. Still separate tuning sets will be required.
- Add options to have tuning set on step level. It propagates to phases. This one tuning set will be shared. (Still requires previous point)
- Reimplement tuning sets so it takes list of operation to run. (I'm not a fan of this idea though.)

wojtek-t: > Reimplement tuning sets so it takes list of operation to run. (I'm not a fan of this idea though.)

This is what I had on my mind.
Why do you say it's a bad idea?

krzysied: > Why do you say it's a bad idea?

It's not a bad idea. It will just require reimplementing all tuning sets and how the phases are handled and this will take time.

clusterloader2/testing/load/config.yaml
```diff
@@ -15,22 +16,39 @@
 {{$bigRCsPerNamespace := DivideInt $bigRCs $namespaces}}
 {{$mediumRCsPerNamespace := DivideInt $mediumRCs $namespaces}}
 {{$smallRCsPerNamespace := DivideInt $smallRCs $namespaces}}
-{{$totalPodsRounded := AddInt (MultiplyInt $bigRCs $BIG_GROUP_SIZE) (MultiplyInt $mediumRCs $MEDIUM_GROUP_SIZE) (MultiplyInt $smallRCs $SMALL_GROUP_SIZE)}}
+{{$totalPods := AddInt (MultiplyInt $bigRCs $BIG_GROUP_SIZE) (MultiplyInt $mediumRCs $MEDIUM_GROUP_SIZE) (MultiplyInt $smallRCs $SMALL_GROUP_SIZE)}}
 # saturationQPS - qps for RC operations should be an equivalent to 10 [pods/s].
-{{$saturationQPS := DivideFloat (AddInt $bigRCs $mediumRCs $smallRCs) (DivideFloat $totalPodsRounded 10)}}
+{{$bigSaturationQPS := DivideFloat $bigRCs (DivideFloat $totalPods 10)}}
+{{$mediumSaturationQPS := DivideFloat $mediumRCs (DivideFloat $totalPods 10)}}
+{{$smallSaturationQPS := DivideFloat $smallRCs (DivideFloat $totalPods 10)}}
```
wojtek-t: >> Why do you say it's a bad idea?

> It's not a bad idea. It will just require reimplementing all tuning sets and how the phases are handled and this will take time.

I just looked into it and I didn't realize that tuning sets are implemented based on tickers.
I think it's really limiting implementation.

In fact that they should have access to the operations that they are supposed to run.
[So I think that at the very high-level the api for tuning-set implementatio should be:
func (ts *TuningSet) Execute(actions) {}
rather than implemented based on ticker.

That is much more powerful and gives more freedom in defining their types (as an example what I suggested with randomized load).

I understand that it will take time to implement it, but I really think we should do this (I wish I was around when this was implemented to point that at that time).

clusterloader2/testing/load/config.yaml
```diff
@@ -100,8 +138,8 @@ steps:
     Params:
       kind: ReplicationController
       labelSelector: kind = loadRC
-      # timeout = (totalPodsRounded /20)s + 3m
-      timeout: {{DivideInt $totalPodsRounded 20 | AddInt 180}}s
+      # timeout = (totalPods /20)s + 3m
```
wojtek-t: This is also a thing that we need to rethink how to do.
The way how it is done now is:
- you create everything (in the previous step), spreaded over extended period of time, potentially hours
- only then you actually verify if pods were starting (so for some RC, the timeout may be hours, not really what you have below).

That is clearly not what we want. What we probably want is:
- initialize WaitForControlledPodsRunning
- it will be watching RCs and for each of them will be spawning goroutine watching its pods
- these go-routines may finish (with success or failure or timeout)
- the at the end you gather results of this measurement
[This gathering is synchronous and make take time]

But that seems to be the only reasonable way to do it.

clusterloader2/testing/load/config.yaml
krzysied: #259 - Rework of TuningSets.

wojtek-t: So to copy my initial comment on that:

> TBH, I don't think it's exaclty what we want.
I think we should go a different direction to make the test much simpler.

> What we know is that we would like to generate average 10pods/s throughput.
So I would suggest introducing a different TuningSet type like:
RandomizedLoad
but where you pass total time over which you spread operations, not the QPS.

>That would make it waaaay simpler, because all you will need is one tuningset, not one tunningset per type.

that should be relatively simple to do now

krzysied: #262 - Time limited tuning sets.

clusterloader2/testing/load/config.yaml
```diff
@@ -15,22 +16,39 @@
...
+    averageQps: {{MultiplyFloat $bigSaturationQPS 4}}
+- name: MediumAvgScalingQPS
+  randomizedLoad:
+    averageQps: {{MultiplyFloat $mediumSaturationQPS 4}}
+- name: SmallAvgScalingQPS
+  randomizedLoad:
+    averageQps: {{MultiplyFloat $smallSaturationQPS 4}}
 steps:
 - measurements:
   - Identifier: PhaseTimer
```
wojtek-t: Can you now get rid of those timers and instead provide names for steps (if we believe those times are useful).
Maybe even a separate PR, to already simplify configs asap.

krzysied: #263 - Removing timers from test configs.

clusterloader2/testing/load/config.yaml
wojtek-t: OK - this is merged. So we can use it here.

clusterloader2/testing/load/config.yaml
krzysied: #264 - Reimplementing WaitForControlledPodsRunning

clusterloader2/testing/load/config.yaml
krzysied: Changes applied.

clusterloader2/testing/load/config.yaml
krzysied: Changes applied.

krzysied:
#264 Merged. PR rebased.

clusterloader2/testing/load/config.yaml
```diff
@@ -15,28 +18,63 @@
...
-  randomizedLoad:
-    averageQps: {{$saturationQPS}}
-- name: AvgScalingQPS
-  randomizedLoad:
+- name: SvcSaturationQPS
+  qpsLoad:
+    qps: {{$SVC_SATURATION_QPS}}
+- name: RandomizedSaturationTimeLimited
+  RandomizedTimeLimitedLoad:
+    timeLimit: {{MultiplyInt $saturationTime $SECOND_TO_NANO}}
```
wojtek-t: Please don't.
 API should be reasonable. Passing timeouts in nanoseconds is not reasonable.

Seconds is what we should pass.

clusterloader2/testing/load/config.yaml
```diff
@@ -4,6 +4,9 @@
 {{$SMALL_GROUP_SIZE := 5}}
 {{$PODS_PER_NODE := 30}}
 {{$NODES_PER_NAMESPACE := 100}}
+{{$SVC_SATURATION_QPS := 10}}
+{{$PODS_SATURATION_QPS := 10}}
```
wojtek-t: Can we collapse these to TEST_THROUGHPUT ?

clusterloader2/testing/load/config.yaml
```diff
@@ -15,28 +18,63 @@
 {{$bigRCsPerNamespace := DivideInt $bigRCs $namespaces}}
```
wojtek-t: Same as in density test - let's assume that number of nodes is divisible by nodes per-namespace.

clusterloader2/testing/load/config.yaml
```diff
@@ -15,28 +18,63 @@
...
   - Identifier: APIResponsiveness
     Method: APIResponsiveness
     Params:
       action: reset
+# Create SVCs
+- phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: {{$bigRCsPerNamespace}}
```
wojtek-t: We don't create service per RC currently:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/load.go#L553

That said, apparently each pod is part of some service now.
So we should try:
- changing the current test to have service per RC
- if that works (and it should), then bundle service with RC creation together

clusterloader2/testing/load/config.yaml
krzysied: I know, it ugly. But that's how goalng  parses time.Duration using json deserializer.
I haven't found any clean solution for this problem yet:
- Creating additional type that implements Marshal and Unmarshal interfaces. Requires to be cast to time.Duration when used.
- Implementing new json encoder. Sounds like a bad idea.
- Having one api for serializing and other for internal use. Also bad idea.

clusterloader2/testing/load/config.yaml
krzysied: Done. Test simplified.

clusterloader2/testing/load/config.yaml
krzysied: Services moved to the bundles.
Looks like fixing namespace problem (#256) and spreading creation randomly through the time, has fixed the performance issue.

clusterloader2/testing/load/config.yaml
wojtek-t: Maybe we should use string representation of duration?
I think that then you san do something like "5s" for example.

wojtek-t: I guess what i would like to have is something like this:
https://github.com/kubernetes/perf-tests/pull/242/files#diff-8b375d89a1280308d37301996c699680R45

krzysied: Done. Changes from #277 applied.

clusterloader2/testing/load/config.yaml
```diff
@@ -1,36 +1,40 @@
...
+# - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
+
 #Constants
 {{$BIG_GROUP_SIZE := 250}}
 {{$MEDIUM_GROUP_SIZE := 30}}
 {{$SMALL_GROUP_SIZE := 5}}
-{{$PODS_PER_NODE := 30}}
 {{$NODES_PER_NAMESPACE := 100}}
-#Variables
+{{$PODS_PER_NAMESPACE := 3000}}
```
wojtek-t: The constant to use should be "PODS_PER_NODE" (also in density test).

This is the thing that is better understood.

clusterloader2/testing/load/config.yaml
```diff
@@ -1,36 +1,40 @@
...
+{{$TEST_THROUGHPUT := 10}}
```
wojtek-t: Just for reusability of the test - is it possible to somehow overwrite it when calling a test?
i.e. If I would like to run test 5x faster, then the whole config would be exactly the same modulo this line.

clusterloader2/testing/load/config.yaml
krzysied: Changed to `PODS_PER_NODE`.

clusterloader2/testing/load/config.yaml
krzysied: Depends what you mean by that:
- In load test `TEST_THROUGHPUT` is a constant that determines the execution speed. However, there is no way to set it through the test call.
- There is no easy way to propagate value from call to the test. The one we support is `Nodes`. The question is whether is necessary to support overrides for test variables? (This is tempalting, not shell scripting)
- If time is the only concern here, we can introduce flag to define speed multiplier. This flag would be used by pass to tuning sets. Will it be useful?

clusterloader2/testing/load/config.yaml
krzysied: "PODS_PER_NODE" in density test - #280

clusterloader2/testing/load/config.yaml
wojtek-t: So what we do now in our tests is:
- we *have* a way to overwrite speed of test (we're using env-var which is hacky, but it's possible)
- I would like to preserve that ability, without touching test config

[e.g. I would like to be able to e.g. run the same test with different speed on different suites]

wojtek-t: Time is not the only thing. E.g. pod per node is the same thing (we have suite with high-density that again is running the same suite, but with different pods per node).

krzysied: OK. I understand.
I created issue #281 regarding this feature. IMHO we should continue with this PR as it is and I will try to figure out the solution for #281 in the meantime.

wojtek-t: That's fine

clusterloader2/testing/load/config.yaml
```diff
@@ -1,36 +1,40 @@
...
-{{$totalPodsRounded := AddInt (MultiplyInt $bigRCs $BIG_GROUP_SIZE) (MultiplyInt $mediumRCs $MEDIUM_GROUP_SIZE) (MultiplyInt $smallRCs $SMALL_GROUP_SIZE)}}
-# saturationQPS - qps for RC operations should be an equivalent to 10 [pods/s].
-{{$saturationQPS := DivideFloat (AddInt $bigRCs $mediumRCs $smallRCs) (DivideFloat $totalPodsRounded 10)}}
-#Test
+{{$SMALL_RCS_PER_NAMESPACE := 300}}
+
+#Variables
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
+# saturationTime - time for RC saturation should be an equivalent to 10 [pods/s] qps.
```
wojtek-t: This comment doesn't provide any value imho

clusterloader2/testing/load/config.yaml
```diff
@@ -50,8 +54,8 @@ steps:
   - namespaceRange:
```
wojtek-t: for consistency with density, maybe it should be "group = load" (I mean the label above)?

clusterloader2/testing/load/config.yaml
krzysied: Comment removed.

clusterloader2/testing/load/config.yaml

PR #256, opened 2018-10-18
ClusterLoader - Random automanaged namespace name

krzysied:
This is the fix for requests latency problem. Each test will use separate namespaces.

### DIFF
--- a/clusterloader2/pkg/framework/framework.go
+++ b/clusterloader2/pkg/framework/framework.go
@@ -33,17 +33,13 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
-const (
-	// AutomanagedNamespaceName is a basename for automanaged namespaces.
-	AutomanagedNamespaceName = "namespace"
-)
-
 // Framework allows for interacting with Kubernetes cluster via
 // official Kubernetes client.
 type Framework struct {
-	automanagedNamespaceCount int
-	clientSet                 clientset.Interface
-	dynamicClient             dynamic.Interface
+	automanagedNamespacePrefix string
+	automanagedNamespaceCount  int
+	clientSet                  clientset.Interface
+	dynamicClient              dynamic.Interface
 }
 
 // NewFramework creates new framework based on given kubeconfig.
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -45,6 +46,8 @@ func createSimpleTestExecutor() TestExecutor {
 
 // ExecuteTest executes test based on provided configuration.
 func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *util.ErrorList {
+	framework.AutomanagedNamespaceName = fmt.Sprintf("namespace-%v", rand.Int())
```
wojtek-t: Why do we need randomization?
Why can't we do sth like "test-<test_name from config>" or sth like that?

Names should generally be different anyway :)

krzysied: > sth like "test-<test_name from config>" 

I thought about this. In most cases it is a proper solution.
The only case it doesn't solve is when somebody runs the same twice. This is allowed, so IMHO this also be handled correctly.

wojtek-t: OK - that's a bit artificial, but I agree that probably worth supporting.

Then let's do: "test-<test name>-<hash>" (but let make hash smaller than int32 and with constant amount of characters)

krzysied: Done. Turns out this hash needs to be in [a-z0-9].

clusterloader2/pkg/framework/framework.go
```diff
@@ -33,7 +33,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
-const (
+var (
```
wojtek-t: This shouldn't be global variable.

It should be a field of framework.Framework.

krzysied: Done.

clusterloader2/pkg/framework/framework.go
```diff
@@ -33,14 +33,10 @@ import (
...
 
-const (
-	// AutomanagedNamespaceName is a basename for automanaged namespaces.
-	AutomanagedNamespaceName = "namespace"
-)
-
 // Framework allows for interacting with Kubernetes cluster via
 // official Kubernetes client.
 type Framework struct {
+	automanagedNamespaceName  string
```
wojtek-t: automanagedNamespacePrefix ?

in the end, names have "-number"
https://github.com/kubernetes/perf-tests/blob/40503f822c514584dad1306da3494234221d13c4/clusterloader2/pkg/framework/framework.go#L81

krzysied: Done.

PR #258, opened 2018-10-18
ClusterLoader - Adding step timer

krzysied:


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -92,7 +92,13 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *util.
...
 func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) *util.ErrorList {
 	var wg wait.Group
-	// TODO(krzysied): Consider moving lock and errList to separate structure.
+	if step.Name != "" {
+		startTime := time.Now()
+		glog.Infof("Step \"%s\": started", step.Name)
+		defer func() {
+			glog.Infof("Step \"%s\": ended - %v", step.Name, time.Since(startTime))
+		}()
+	}
```
wojtek-t: That part looks fine.

However, this function is racy - the errList should be protected by muxes, since you're appending to it from different go-routines.

wojtek-t: s/muxes/mutex/

wojtek-t: Aaah ok - I misread it. Thanks.

PR #259, opened 2018-10-18
ClusterLoader - Tuning set rework

krzysied:
`pkg/tuningset/` - tuning set executors
`pkg/ticker/` - removed
`pkg/test` - updated with new tuning sets. All wait groups replaces with appending func(){...} to actions.


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -126,105 +126,102 @@ func (ste *simpleTestExecutor) ExecutePhase(ctx Context, phase *api.Phase) *util
...
 				}
+			})
+		}
+
+		// Calculating minimal replica count of objects from object bundle.
+		// If there is update operation to be executed, minimal replica count is set to zero.
+		minCurrentReplicaCount := int32(math.MaxInt32)
+		for j := range instancesStates {
+			if instancesStates[j].CurrentReplicaCount == phase.ReplicasPerNamespace {
+				minCurrentReplicaCount = 0
```
wojtek-t: This makes me think that the logic here is not what we want.
I think that we should assume that "bundle is bundle".

If say bundle consists of objects A, B and C, there should be the same amount of those before and after the phase.
Otherwise, what does it mean that:
- you want to get to 5 replicas of bundle
- but there are 3 As, 5 Bs, and 7 Cs

For me - that's a bad input.

You're trying to solve it hackily by computing min and max, and iterating from desired to them, but that will cause more confusion to users than saying "bad input".

We should change that so that : if number of instances of particular objects within a bundle are different, then the input is bad, we return error and skip the phase.

wojtek-t: Probably should be done in a separate PR though

clusterloader2/pkg/tuningset/qps_load.go
```diff
@@ -14,32 +14,27 @@ See the License for the specific language governing permissions and
...
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/perf-tests/clusterloader2/api"
 )
 
-type ticker struct {
-	C     <-chan struct{}
-	timer *time.Timer
-}
+type qpsLoad api.QpsLoad
```
wojtek-t: This is technically correct, but I'm not a fan of such construct.
Instead, can you do simply:

type qpsLoad struct {
   config *api.QpsLoad   (maybe config is not the best name)
}

clusterloader2/pkg/tuningset/qps_load.go
```diff
@@ -14,32 +14,27 @@ See the License for the specific language governing permissions and
...
-	ch := make(chan struct{})
-	timer, err := newTimer(ch, ts)
-	if err != nil {
-		return nil, err
-	}
-	return &ticker{
-		C:     ch,
-		timer: timer,
-	}, nil
+func newQpsLoad(ts *api.TuningSet) TuningSet {
```
wojtek-t: this should take api.QpsLoad as as a parameter.

clusterloader2/pkg/tuningset/randimized_load.go
```diff
@@ -0,0 +1,46 @@
+/*
```
wojtek-t: s/randomized/randomized/ in file name :)

clusterloader2/pkg/tuningset/randimized_load.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"math/rand"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+type randomizedLoad api.RandomizedLoad
```
wojtek-t: Same comments as in above file

clusterloader2/pkg/tuningset/simple_tuning_set_factory.go
```diff
@@ -14,38 +14,47 @@ See the License for the specific language governing permissions and
...
-// CreateTicker creates new ticker based on provided tuning set name.
-func (tf *simpleTickerFactory) CreateTicker(name string) (*ticker, error) {
+// CreateTuningSet creates new tuning set based on provided tuning set name.
+func (tf *simpleTuningSetFactory) CreateTuningSet(name string) (TuningSet, error) {
 	tuningSet, exists := tf.tuningSetMap[name]
 	if !exists {
 		return nil, fmt.Errorf("tuningset %s not found", name)
 	}
-	return newTicker(tuningSet)
+	if tuningSet.QpsLoad != nil {
```
wojtek-t: switch {
case tuningSet.QpsLoad != nil:
  return newQpsLoad(tuningSet.QpsLoad), nil
...
}

clusterloader2/pkg/tuningset/stepped_load.go
```diff
@@ -0,0 +1,42 @@
...
+package tuningset
+
+import (
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+type steppedLoad api.SteppedLoad
```
wojtek-t: Again - the same comments.

clusterloader2/pkg/tuningset/qps_load.go
krzysied: Done.

clusterloader2/pkg/tuningset/qps_load.go
krzysied: Done.

clusterloader2/pkg/tuningset/randimized_load.go
krzysied: Done.

clusterloader2/pkg/tuningset/simple_tuning_set_factory.go
krzysied: Done.

clusterloader2/pkg/test/simple_test_executor.go
krzysied: This is a very strong assumption. Definitely this should be also mention in docs.

> Probably should be done in a separate PR though

I agree. I will create one after this one is merged.

wojtek-t: Yes - but if I'm saying that "this group of objects is a bundle", I don't want to assume that for some bundles there exist some subset of it, for some other superset and so on - this doesn't make any sense.

Filed https://github.com/kubernetes/perf-tests/issues/261 for it.

PR #262, opened 2018-10-19
ClusterLoader - Adding time limited tuning sets

krzysied:
Two new tuning sets added:
- TimeLimitedLoad
Spreads action evenly over given time.
- RandomizedTimeLimitedLoad
Runs each action at random point of time (within given time limit).

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -102,6 +102,10 @@ type TuningSet struct {
 	RandomizedLoad *RandomizedLoad `json: randomizedLoad`
 	// SteppedLoad is a definition for SteppedLoad tuning set.
 	SteppedLoad *SteppedLoad `json: steppedLoad`
+	// TimeLimitedLoad is a definition for TimeLimitedLoad tuning set.
+	TimeLimitedLoad *TimeLimitedLoad `json: timeLimitedLoad`
+	// RandomizedTimeLimitedLoad is a definition for RandomizedTimeLimitedLoad tuning set.
+	RandomizedTimeLimitedLoad *RandomizedTimeLimitedLoad `json: randomizedTimeLimitedLoad`
 }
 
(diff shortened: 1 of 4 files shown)


clusterloader2/api/types.go
```diff
@@ -102,6 +102,10 @@ type TuningSet struct {
 	RandomizedLoad *RandomizedLoad `json: randomizedLoad`
 	// SteppedLoad is a definition for SteppedLoad tuning set.
 	SteppedLoad *SteppedLoad `json: steppedLoad`
+	// RandomizedLoad is a definition for TimeLimitedLoad tuning set.
```
wojtek-t: TimeLimitedLoad

clusterloader2/pkg/tuningset/randomized_time_limited.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func (r *randomizedTimeLimitedLoad) Execute(actions []func()) {
+	var wg wait.Group
+	for i := range actions {
+		index := i
+		wg.Start(func() {
+			// Sleeps for random duration in [0, TimeLimit].
+			time.Sleep(time.Duration(rand.Intn(int(r.params.TimeLimit))))
```
wojtek-t: This may not work - time.Duration is converted to nanoseconds, and int can be 4 bytes...
Should be sth like:

time.Sleep(time.Duration(rand.Int63n(r.params.TimeLimit.Nanoseconds()))

clusterloader2/api/types.go
krzysied: Fixed.

clusterloader2/pkg/tuningset/randomized_time_limited.go
krzysied: Done.

PR #263, opened 2018-10-19
ClusterLoader - Removing unnecessary timers from tests

krzysied:


clusterloader2/testing/density/config.yaml
```diff
@@ -58,23 +50,14 @@ steps:
...
-      label: CreateLatencyPods
-      action: start
   - Identifier: PodStartupLatency
     Method: PodStartupLatency
     Params:
       action: start
       labelSelector: kind = latencyPod
 # Create latency pods
-- phases:
+- name: Create latency pods
```
wojtek-t: This one is also tricky - creation will take exactly: #pods / 5 - we perform 5qps :)

clusterloader2/testing/density/config.yaml
```diff
@@ -86,30 +69,22 @@ steps:
       templateFillMap:
         Replicas: 1
         PodKind: latencyPod
-- measurements:
+- name: Wait for latency pods
```
wojtek-t: And this should be fairly fast - it should be pretty much <5s according to our SLO :-)

wojtek-t: So in other words - I don't think these two provide any value.

clusterloader2/testing/density/config.yaml
```diff
@@ -37,7 +27,8 @@ steps:
     Params:
       action: reset
 # Create saturation pods
-- phases:
+- name: Create saturation pods
```
wojtek-t: I started thinking about it more, and I don't think that this PR is exactly what we want:
https://github.com/kubernetes/perf-tests/pull/258

Many steps are not self-contained (e.g. here - we create stuff, but not everything is created when the step is finished).

So instead of that, I would say that:
- let's log only when step finishes [step X done]
- don't modify the existing steps (let's leave them without name)
- when you want to get a timestamp, create a dummy empty step just to print the name

For example:
```
# Create saturation pods
- phases:
...
- measurements:
...
- name: Saturation pods created
```

krzysied: TBH, I'm not convinced that it's the best solution.

>when you want to get a timestamp, create a dummy empty step just to print the name

I really don't like that we have name just to use `Log` function in a hacky way. It's not intuitive how to use it. Also we will lose the ability to measure time.

IMHO better solution would be:
1. Leaving timers for named step (implementation).
2. Removing name from current steps from configs.
3. Changing `Step` implementation. Adding `Substeps []Step` field to `Step`.
4. Changing config to have substeps and named global steps.

Reason behind it:
When people describe any algorithm, they mostly use multi-level list to describe steps that need to be follow. Point 3 will allow us to do exactly the same thing with test configs. It'll be easy to understand and config will look better due to additional indents.

Example:
```
- name: Create saturation pods
   substeps:
   - phases:
      ...
   - measurements
     ...
- name: Create latency pods
  substeps:
   - phases:
      ...
```

wojtek-t: > Also we will lose the ability to measure time.

If you really need it, you can use timers - you're always free to do that.
I just don't want to have 30% of configs being timers in our current tests.

> Changing Step implementation. Adding Substeps []Step field to Step.

No. That's complicating the API for no real reason.

wojtek-t: Another strong argument is that it would make it possible to express the same thing in multiple ways in the api. That's always a bad thing.

clusterloader2/testing/density/config.yaml
```diff
@@ -94,20 +75,11 @@ steps:
...
     Method: PodStartupLatency
     Params:
       action: gather
-  - Identifier: PhaseTimer
-    Method: Timer
-    Params:
-      label: DeleteAllPods
-      action: start
+- name: Creating latency pods
 # Delete pods
```
wojtek-t: Please split the deletion of latency pods and saturation pods into two.

Also please add the timestamp after latency pods deletion.

clusterloader2/testing/load/config.yaml
```diff
@@ -155,17 +135,7 @@ steps:
...
-    Method: Timer
-    Params:
-      label: ScaleRCs
-      action: stop
-  - Identifier: PhaseTimer
-    Method: Timer
-    Params:
-      label: DeleteAllPods
-      action: start
+- name: Scaling RCs
```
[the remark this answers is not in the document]

clusterloader2/testing/density/config.yaml

PR #264, opened 2018-10-23
ClusterLoader - Reimplementing wait for controlled pods measurement

krzysied:
Reimplementing wait for controlled pods measurement. Now it support two actions:
- start - starts to observe the controlling objects.
- gather - waits for objects to have all pods running.
- stop - stops the observation.

Minor changes:
- parallel group that limits number of routines running in parallel added.
- `waitForPods` has stop channel. Timeouts should be handled by the caller.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -24,7 +24,6 @@ package common
 import (
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/golang/glog"
@@ -32,15 +31,19 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
 
 const (
-	defaultWaitForRCPodsTimeout      = 60 * time.Second
-	defaultWaitForRCPodsInterval     = 5 * time.Second
+	defaultSyncTimeout               = 60 * time.Second
+	defaultOperationTimeout          = 10 * time.Minute
+	checkControlledPodsInterval      = 5 * time.Second
 	waitForControlledPodsRunningName = "WaitForControlledPodsRunning"
+	waitForControlledPodsWorkers     = 10
 )
 
 func init() {
@@ -48,99 +51,365 @@ func init() {
 }
 
 func createWaitForControlledPodsRunningMeasurement() measurement.Measurement {
-	return &waitForControlledPodsRunningMeasurement{}
+	return &waitForControlledPodsRunningMeasurement{
+		queue:      workqueue.New(),
+		checkerMap: make(map[string]*objectChecker),
+	}
 }
 
-type waitForControlledPodsRunningMeasurement struct{}
+type waitForControlledPodsRunningMeasurement struct {
+	informer          cache.SharedInformer
+	kind              string
+	namespace         string
+	labelSelector     string
+	fieldSelector     string
+	operationTimeout  time.Duration
+	stopCh            chan struct{}
+	isRunning         bool
+	queue             workqueue.Interface
+	workerGroup       wait.Group
+	handlingGroup     wait.Group
+	lock              sync.Mutex
+	opResourceVersion uint64
+	checkerMap        map[string]*objectChecker
+}
 
-// Execute waits until all specified RCs have all pods running or until timeout happens.
-// RCs can be specified by field and/or label selectors.
+// Execute waits until all specified controlling objects have all pods running or until timeout happens.
+// Controlling objects can be specified by field and/or label selectors.
 // If namespace is not passed by parameter, all-namespace scope is assumed.
-func (*waitForControlledPodsRunningMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+// "Start" action starts observation of the controlling objects, while "gather" waits for until
+// specified number of controlling objects have all pods running.
+func (w *waitForControlledPodsRunningMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
 	var summaries []measurement.Summary
-	kind, err := util.GetString(config.Params, "kind")
-	if err != nil {
-		return summaries, err
-	}
-	namespace, err := util.GetStringOrDefault(config.Params, "namespace", metav1.NamespaceAll)
-	if err != nil {
-		return summaries, err
-	}
-	labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")
-	if err != nil {
-		return summaries, err
-	}
-	fieldSelector, err := util.GetStringOrDefault(config.Params, "fieldSelector", "")
-	if err != nil {
-		return summaries, err
-	}
-	timeout, err := util.GetDurationOrDefault(config.Params, "timeout", defaultWaitForRCPodsTimeout)
+	action, err := util.GetString(config.Params, "action")
 	if err != nil {
 		return summaries, err
 	}
 
-	runtimeObjectsList, err := runtimeobjects.ListRuntimeObjectsForKind(config.ClientSet, kind, namespace, labelSelector, fieldSelector)
-	if err != nil {
-		return summaries, err
-	}
-	var wg wait.Group
-	errList := util.NewErrorList()
-	var runningRuntimeObjects int32
-	for objectCounter := 0; objectCounter < len(runtimeObjectsList); objectCounter++ {
-		objectIndex := objectCounter
-		wg.Start(func() {
-			if err := waitForRuntimeObject(config.ClientSet, runtimeObjectsList[objectIndex], timeout); err != nil {
-				errList.Append(fmt.Errorf("waiting for %v error: %v", kind, err))
-				return
-			}
-
-			atomic.AddInt32(&runningRuntimeObjects, 1)
-			objName, err := runtimeobjects.GetNameFromRuntimeObject(runtimeObjectsList[objectIndex])
-			if err != nil {
-				errList.Append(fmt.Errorf("reading object name error: %v", err))
-				return
-			}
-			objNamespace, err := runtimeobjects.GetNamespaceFromRuntimeObject(runtimeObjectsList[objectIndex])
-			if err != nil {
-				errList.Append(fmt.Errorf("reading object namespace error: %v", err))
-				return
-			}
-			glog.Infof("%s: %s (%s) has all pods running", waitForControlledPodsRunningName, objName, objNamespace)
-		})
-	}
-	wg.Wait()
-
-	if int(runningRuntimeObjects) == len(runtimeObjectsList) {
-		glog.Infof("%s: %d / %d %ss are running with all pods", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
+	switch action {
+	case "start":
+		w.kind, err = util.GetString(config.Params, "kind")
+		if err != nil {
+			return summaries, err
+		}
+		w.namespace, err = util.GetStringOrDefault(config.Params, "namespace", metav1.NamespaceAll)
+		if err != nil {
+			return summaries, err
+		}
+		w.labelSelector, err = util.GetStringOrDefault(config.Params, "labelSelector", "")
+		if err != nil {
+			return summaries, err
+		}
+		w.fieldSelector, err = util.GetStringOrDefault(config.Params, "fieldSelector", "")
+		if err != nil {
+			return summaries, err
+		}
+		w.operationTimeout, err = util.GetDurationOrDefault(config.Params, "operationTimeout", defaultOperationTimeout)
+		if err != nil {
+			return summaries, err
+		}
+		return summaries, w.start(config.ClientSet)
+	case "gather":
+		syncTimeout, err := util.GetDurationOrDefault(config.Params, "syncTimeout", defaultSyncTimeout)
+		if err != nil {
+			return summaries, err
+		}
+		return summaries, w.gather(config.ClientSet, syncTimeout)
+	case "stop":
+		w.stop()
 		return summaries, nil
+	default:
+		return summaries, fmt.Errorf("unknown action %v", action)
 	}
-	return summaries, fmt.Errorf("error while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods: %v",
-		kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects, errList.String())
 }
 
-func waitForRuntimeObject(c clientset.Interface, obj runtime.Object, timeout time.Duration) error {
-	runtimeObjectNamespace, err := runtimeobjects.GetNamespaceFromRuntimeObject(obj)
+// String returns a string representation of the metric.
+func (*waitForControlledPodsRunningMeasurement) String() string {
+	return waitForControlledPodsRunningName
+}
+
+func (w *waitForControlledPodsRunningMeasurement) start(c clientset.Interface) error {
+	if w.informer != nil {
+		glog.Infof("%v: wait for controlled pods measurement already running", w)
+		return nil
+	}
+	glog.Infof("%v: starting wait for controlled pods measurement...", w)
+	optionsModifier := func(options *metav1.ListOptions) {
+		options.FieldSelector = w.fieldSelector
+		options.LabelSelector = w.labelSelector
+	}
+	listerWatcher := cache.NewFilteredListWatchFromClient(c.CoreV1().RESTClient(), w.kind+"s", w.namespace, optionsModifier)
+	w.isRunning = true
+	w.stopCh = make(chan struct{})
+	w.informer = cache.NewSharedInformer(listerWatcher, nil, 0)
+	w.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			addF := func() {
+				w.handleObject(c, nil, obj)
+			}
+			w.queue.Add(&addF)
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			updateF := func() {
+				w.handleObject(c, oldObj, newObj)
+			}
+			w.queue.Add(&updateF)
+		},
+		DeleteFunc: func(obj interface{}) {
+			deleteF := func() {
+				w.handleObject(c, obj, nil)
+			}
+			w.queue.Add(&deleteF)
+		},
+	})
+
+	for i := 0; i < waitForControlledPodsWorkers; i++ {
+		w.workerGroup.Start(w.worker)
+	}
+	go w.informer.Run(w.stopCh)
+	timeoutCh := make(chan struct{})
+	timeoutTimer := time.AfterFunc(informerSyncTimeout, func() {
+		close(timeoutCh)
+	})
+	defer timeoutTimer.Stop()
+	if !cache.WaitForCacheSync(timeoutCh, w.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+
+	return nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) worker() {
+	for {
+		f, stop := w.queue.Get()
+		if stop {
+			return
+		}
+		(*f.(*func()))()
+		w.queue.Done(f)
+	}
+}
+
+func (w *waitForControlledPodsRunningMeasurement) gather(c clientset.Interface, syncTimeout time.Duration) error {
+	glog.Infof("%v: waiting for controlled pods measurement...", w)
+	if w.informer == nil {
+		return fmt.Errorf("metric %s has not been started", w)
+	}
+	desiredCount, maxResourceVersion, err := w.getObjectCountAndMaxVersion(c)
 	if err != nil {
 		return err
+	}
+
+	cond := func() (bool, error) {
+		return w.opResourceVersion >= maxResourceVersion, nil
+	}
+	if err := wait.Poll(checkControlledPodsInterval, syncTimeout, cond); err != nil {
+		return fmt.Errorf("timed out while waiting for controlled pods")
+	}
+
+	w.handlingGroup.Wait()
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	numberRunning := 0
+	for _, checker := range w.checkerMap {
+		if status := checker.getStatus(); status {
+			numberRunning++
+		}
+	}
+	if desiredCount != numberRunning {
+		glog.Errorf("%s: incorrect objects number: %d/%d %ss are running with all pods", w, numberRunning, desiredCount, w.kind)
+		return fmt.Errorf("incorrect objects number")
+	}
+
+	glog.Infof("%s: %d/%d %ss are running with all pods", w, numberRunning, desiredCount, w.kind)
+	return nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) stop() {
+	close(w.stopCh)
+	w.queue.ShutDown()
+	w.workerGroup.Wait()
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	w.isRunning = false
+	for _, checker := range w.checkerMap {
+		checker.terminate()
+	}
+}
+
+func (w *waitForControlledPodsRunningMeasurement) handleObject(c clientset.Interface, oldObj, newObj interface{}) {
+	var oldRuntimeObj runtime.Object
+	var newRuntimeObj runtime.Object
+	var ok bool
+	oldRuntimeObj, ok = oldObj.(runtime.Object)
+	if oldObj != nil && !ok {
+		glog.Errorf("%s: uncastable old object: %v", w, oldObj)
+		return
+	}
+	newRuntimeObj, ok = newObj.(runtime.Object)
+	if newObj != nil && !ok {
+		glog.Errorf("%s: uncastable new object: %v", w, newObj)
+		return
+	}
+	defer func() {
+		// We want to update version after (potentially) creating goroutine.
+		if err := w.updateOpResourceVersion(oldRuntimeObj); err != nil {
+			glog.Errorf("%s: updating resource version error: %v", w, err)
+		}
+		if err := w.updateOpResourceVersion(newRuntimeObj); err != nil {
+			glog.Errorf("%s: updating resource version error: %v", w, err)
+		}
+	}()
+	isEqual, err := runtimeobjects.IsEqualRuntimeObjectsSpec(oldRuntimeObj, newRuntimeObj)
+	if err != nil {
+		glog.Errorf("%s: comparing specs error: %v", w, err)
+		return
+	}
+	if isEqual {
+		// Skip updates without changes in the spec.
+		return
+	}
+
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if !w.isRunning {
+		return
+	}
+
+	if err := w.deleteObjectLocked(c, oldRuntimeObj); err != nil {
+		glog.Errorf("%s: delete checker error: %v", w, err)
+	} else if newRuntimeObj == nil {
+		key, err := runtimeobjects.CreateMetaNamespaceKey(oldRuntimeObj)
+		if err != nil {
+			glog.Errorf("%s: meta key creation error: %v", w, err)
+			return
+		}
+		glog.Infof("%s: %v has been deleted", w, key)
+	}
+	if err := w.createObjectLocked(c, newRuntimeObj); err != nil {
+		glog.Errorf("%s: create checker error: %v", w, err)
+	}
+}
+
+func (w *waitForControlledPodsRunningMeasurement) createObjectLocked(c clientset.Interface, obj runtime.Object) error {
+	if obj == nil {
+		return nil
+	}
+	key, err := runtimeobjects.CreateMetaNamespaceKey(obj)
+	if err != nil {
+		return fmt.Errorf("meta key creation error: %v", err)
+	}
+	checker, err := w.waitForRuntimeObject(c, obj)
+	if err != nil {
+		return fmt.Errorf("waiting for %v error: %v", key, err)
+	}
+	time.AfterFunc(w.operationTimeout, func() {
+		checker.terminate()
+	})
+	w.checkerMap[key] = checker
+	return nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) deleteObjectLocked(c clientset.Interface, obj runtime.Object) error {
+	if obj == nil {
+		return nil
+	}
+	key, err := runtimeobjects.CreateMetaNamespaceKey(obj)
+	if err != nil {
+		return fmt.Errorf("meta key creation error: %v", err)
+	}
+	if checker, exists := w.checkerMap[key]; exists {
+		checker.terminate()
+		delete(w.checkerMap, key)
+	}
+	return nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(runtimeObj runtime.Object) error {
+	if runtimeObj == nil {
+		return nil
+	}
+	version, err := runtimeobjects.GetResourceVersionFromRuntimeObject(runtimeObj)
+	if err != nil {
+		return fmt.Errorf("retriving resource version error: %v", err)
+	}
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if version > w.opResourceVersion {
+		w.opResourceVersion = version
+	}
+	return nil
+}
+
+// getObjectCountAndMaxVersion returns number of objects that satisfy measurements parameters
+// and maximal resource version of these objects.
+// These two values allow to properly handle all object operations:
+// - When create/delete operation are called we expect the exact number of objects.
+// - When objects is updated we expect to receive event referencing this specific version.
+//   Using maximum from objects resource versions assures that all updates will be processed.
+func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion(c clientset.Interface) (int, uint64, error) {
+	var desiredCount int
+	var maxResourceVersion uint64
+	objects, err := runtimeobjects.ListRuntimeObjectsForKind(c, w.kind, w.namespace, w.labelSelector, w.fieldSelector)
+	if err != nil {
+		return desiredCount, maxResourceVersion, fmt.Errorf("listing objects error: %v", err)
+	}
+
+	for i := range objects {
+		runtimeObj, ok := objects[i].(runtime.Object)
+		if !ok {
+			glog.Errorf("%s: cannot cast to runtime.Object: %v", objects[i])
+			continue
+		}
+		version, err := runtimeobjects.GetResourceVersionFromRuntimeObject(runtimeObj)
+		if err != nil {
+			glog.Errorf("%s: retriving resource version error: %v", w, err)
+			continue
+		}
+		desiredCount++
+		if version > maxResourceVersion {
+			maxResourceVersion = version
+		}
+	}
+	return desiredCount, maxResourceVersion, nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) waitForRuntimeObject(clientSet clientset.Interface, obj runtime.Object) (*objectChecker, error) {
+	runtimeObjectNamespace, err := runtimeobjects.GetNamespaceFromRuntimeObject(obj)
+	if err != nil {
+		return nil, err
 	}
 	runtimeObjectSelector, err := runtimeobjects.GetSelectorFromRuntimeObject(obj)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	runtimeObjectReplicas, err := runtimeobjects.GetReplicasFromRuntimeObject(obj)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	key, err := runtimeobjects.CreateMetaNamespaceKey(obj)
+	if err != nil {
+		return nil, fmt.Errorf("meta key creation error: %v", err)
 	}
 
-	stopCh := make(chan struct{})
-	time.AfterFunc(timeout, func() {
-		close(stopCh)
+	o := newObjectCheker()
+	o.lock.Lock()
+	defer o.lock.Unlock()
+	w.handlingGroup.Start(func() {
+		err = waitForPods(clientSet, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), o.stopCh, false)
+		o.lock.Lock()
+		defer o.lock.Unlock()
+		if err != nil {
+			if o.isRunning {
+				// Log error only if checker wasn't terminated.
+				glog.Errorf("%s: error for %v: %v", w, key, err)
+			}
+			return
+		}
+		glog.Infof("%s: %v has all pods (%d) running", w, key, runtimeObjectReplicas)
+		o.status = true
 	})
-	err = waitForPods(c, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), stopCh, false)
-	if err != nil {
-		return fmt.Errorf("waiting pods error: %v", err)
-	}
-	return nil
+	return o, nil
 }
 
 type objectChecker struct {
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
@@ -32,24 +32,6 @@ import (
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
 )
 
-// GetObjectForKind returns object instance of given kind.
-func GetObjectForKind(kind string) (runtime.Object, error) {
-	switch kind {
-	case "ReplicationController":
-		return &corev1.ReplicationController{}, nil
-	case "ReplicaSet":
-		return &extensions.ReplicaSet{}, nil
-	case "Deployment":
-		return &extensions.Deployment{}, nil
-	case "DaemonSet":
-		return &extensions.DaemonSet{}, nil
-	case "Job":
-		return &batch.Job{}, nil
-	default:
-		return nil, fmt.Errorf("unsupported kind when getting runtime object: %v", kind)
-	}
-}
-
 // ListRuntimeObjectsForKind returns objects of given kind that satisfy given namespace, labelSelector and fieldSelector.
 func ListRuntimeObjectsForKind(c clientset.Interface, kind, namespace, labelSelector, fieldSelector string) ([]runtime.Object, error) {
 	var runtimeObjectsList []runtime.Object
@@ -180,6 +162,9 @@ func GetSelectorFromRuntimeObject(obj runtime.Object) (labels.Selector, error) {
 
 // GetSpecFromRuntimeObject returns spec of given runtime object.
 func GetSpecFromRuntimeObject(obj runtime.Object) (interface{}, error) {
+	if obj == nil {
+		return nil, nil
+	}
 	switch typed := obj.(type) {
 	case *corev1.ReplicationController:
 		return typed.Spec, nil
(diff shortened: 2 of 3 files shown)


krzysied:
@wojtek-t 
I created `pkg/util/parallel.go` cause I couldn't find any existing implementation for limiting parallelism.
If you agree that this is needed, I can move this code to separate PR.

clusterloader2/pkg/util/parallel.go
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+package util
+
+// GoroutineGroup executes routines.
+type GoroutineGroup interface {
+	Execute(routine func())
+}
+
+type limitedParallelGroup struct {
```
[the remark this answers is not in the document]
krzysied: This was introduced due to problem with pod-stores. When `WaitForControlledPodsRunning` is used during scaling RCs, it creates pod-stores for each existing RC and some of the pod-store fail. I though that putting limit on number of running goroutines, will stagger the pod-stores creations.

In original tests this problem didn't exist, cause after scaling one RC single pod-store was created. Scaling RCs was spread over time.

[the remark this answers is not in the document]
wojtek-t: I'm not sure why do you say it's a problem now?

Since we are saying that we're only care about changes of a Spec, then change in a Spec of some RC should trigger starting a PodStore.
So spreading those scales over time makes it equivalent to what we have currently.

Is there anything I'm missing?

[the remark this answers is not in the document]
krzysied: The measurement I implemented works like this:
1. Lists all RCs that satisfy the requirements
2. Starts to verify statuses of each listed RC
3. Gets events for RCs that satisfy the requirements.
- Create runs verifying for given RC
- Update stops verifying for older version (if any is running) and starts new verification
- Delete removes RC info from the list.

When "gather" called, it waits until there is specific number of RCs with all pods running.

The problem is with point 2. This synchronize inner RC list with existing RCs. E.g. if there is 10000 RC running, 10000 pod-stors will be created (at the same time) to verify RC statuses.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,178 +23,419 @@ package common
...
-
-			atomic.AddInt32(&runningRuntimeObjects, 1)
-			objName, err := getNameFromRuntimeObject(runtimeObjectsList[objectIndex])
+			resourceVersion, err := getResourceVersionFromRuntimeObject(runtimeObj)
 			if err != nil {
-				errList.Append(fmt.Errorf("reading object name error: %v", err))
 				return
 			}
-			objNamespace, err := getNamespaceFromRuntimeObject(runtimeObjectsList[objectIndex])
+			w.updateOpResourceVersion(resourceVersion)
```
wojtek-t: this should take object (to avoid parsing RV in 3 places separately)

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,178 +23,419 @@ package common
...
 	}
-	wg.Wait()
+	timeoutTimer.Stop()
 
-	if int(runningRuntimeObjects) == len(runtimeObjectsList) {
-		glog.Infof("%s: %d / %d %ss are running with all pods", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		return summaries, nil
+	store := w.informer.GetStore()
+	objects := store.List()
+	for i := range objects {
```
wojtek-t: To we need it?
Won't informer send all Add events at the beginning anyway?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,178 +23,419 @@ package common
...
+}
+
+func (w *waitForControlledPodsRunningMeasurement) gather(timeout time.Duration) error {
+	glog.Infof("%v: waiting for controlled pods measurement...", waitForControlledPodsRunningName)
+	if w.informer == nil {
+		return fmt.Errorf("metric %s has not been started", waitForControlledPodsRunningName)
+	}
+	startTime := time.Now()
+	store := w.informer.GetStore()
+	objects := store.List()
```
wojtek-t: This is incorrect. The list should be done from apiserver directly - the cache may be lagging.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,178 +23,419 @@ package common
...
 	}
-	return summaries, fmt.Errorf("error while waiting for %ss to be running in namespace '%v' with labels '%v' and fields '%v' - only %d found running with all pods: %v",
-		kind, namespace, labelSelector, fieldSelector, runningRuntimeObjects, errList.String())
 }
 
-func waitForRuntimeObject(c clientset.Interface, obj runtime.Object, timeout time.Duration) error {
+func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(resourceVersion string) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	version, err := strconv.ParseUint(resourceVersion, 10, 64)
```
wojtek-t: Please opaque this into function.

Also this is the function you want to copy:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd/api_object_versioner.go#L73

BTW - you should also copy the last one for comparing from that file too.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,178 +23,419 @@ package common
...
+		atomic.AddInt32(&w.runningRoutines, 1)
+		defer func() { atomic.AddInt32(&w.runningRoutines, -1) }()
+		err = waitForPods(clientSet, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), o.stopCh, false)
+		o.lock.Lock()
+		defer o.lock.Unlock()
+		objName, _ := getNameFromRuntimeObject(obj)
+		if err != nil {
+			if o.isRunning {
+				// Log error only if checker wasn't terminated.
+				glog.Infof("%s: error for %v %v: %v", waitForControlledPodsRunningName, runtimeObjectNamespace, objName, err)
```
wojtek-t: Shouldn't this report an error?

clusterloader2/pkg/util/parallel.go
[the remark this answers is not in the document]
wojtek-t: Having this parallel group defeats the whole purpose of timeout.

In the current implementation of our tests, it won't hurt us, because we start the measurement when we don't have any RC.
If we want to do that correctly, the solution is to share a single pod store (the TODo that I suggested couple days ago: https://github.com/kubernetes/perf-tests/pull/264/files#r227685929), not introducing a limitted parallel group.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: Fixed.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: I copied `ObjectResourceVersion`. I think comparing function will not be needed.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: TBH, I want to keep the sync part.
I agree that in our case it may be not needed. However, I want this function to be more generic, able to handle many cases. IMHO if somebody uses this measurement when RCs already exist it should behave correctly.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,178 +23,348 @@ package common
...
+	if err != nil {
+		glog.Errorf("%s: reading object name error: %v", waitForControlledPodsRunningName, err)
+		return
+	}
+	objNamespace, err := getNamespaceFromRuntimeObject(obj)
+	if err != nil {
+		glog.Errorf("%s: reading object namespace error: %v", waitForControlledPodsRunningName, err)
+		return
+	}
+	w.lock.Lock()
```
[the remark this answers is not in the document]
krzysied: Changed to goroutines.

clusterloader2/pkg/measurement/common/wait_for_pods.go
```diff
@@ -70,31 +70,38 @@ func (*waitForRunningPodsMeasurement) Execute(config *measurement.MeasurementCon
...
 
-	return summaries, waitForPods(config.ClientSet, namespace, labelSelector, fieldSelector, desiredPodCount, timeout, true)
+	stopCh := make(chan struct{})
+	time.AfterFunc(timeout, func() { close(stopCh) })
+	return summaries, waitForPods(config.ClientSet, namespace, labelSelector, fieldSelector, desiredPodCount, stopCh, true)
 }
 
-func waitForPods(clientSet clientset.Interface, namespace, labelSelector, fieldSelector string, desiredPodCount int, timeout time.Duration, log bool) error {
+func waitForPods(clientSet clientset.Interface, namespace, labelSelector, fieldSelector string, desiredPodCount int, stopCh <-chan struct{}, log bool) error {
 	ps, err := measurementutil.NewPodStore(clientSet, namespace, labelSelector, fieldSelector)
```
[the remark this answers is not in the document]
krzysied: TODO added.

clusterloader2/pkg/measurement/common/wait_for_pods.go
```diff
@@ -70,31 +70,38 @@ func (*waitForRunningPodsMeasurement) Execute(config *measurement.MeasurementCon
...
+				}
 			}
-		}
-		if log {
-			glog.Infof("WaitForRunningPods: namespace(%s), labelSelector(%s), fieldSelector(%s): running %d / %d", namespace, labelSelector, fieldSelector, runningPodsCount, desiredPodCount)
-		}
-		if runningPodsCount == desiredPodCount {
-			return nil
+			if log {
+				glog.Infof("WaitForRunningPods: namespace(%s), labelSelector(%s), fieldSelector(%s): running %d / %d", namespace, labelSelector, fieldSelector, runningPodsCount, desiredPodCount)
```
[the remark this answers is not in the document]
krzysied: I had discussion with @shyamjvs about it. The reason behind the decision was to have a easy way to identify the measurement the call is coming from and understand what condition this pod had to satisfy.

clusterloader2/pkg/util/parallel.go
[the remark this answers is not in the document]
krzysied: After offline discussion, we came to agreement  to remove parallel group.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: You are right - I removed this check.
I ran a test and it behaved as you decried. However, I haven't found the code responsible for this yet.

krzysied:
@wojtek-t I think I responded to all comments. Also #269 was created to fix pod store problem. PTAL.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,122 +23,397 @@ package common
...
+	labelSelector     string
+	fieldSelector     string
+	operationTimeout  time.Duration
+	stopCh            chan struct{}
+	isRunning         bool
+	runningRoutines   int32
+	lock              sync.Mutex
+	opResourceVersion uint64
+	checkerMap        map[string]*objectChecker
+}
```
wojtek-t: The problem with current implementation is that you can start only a single measurement.

If I would like to track different sets of objects at the same time, that's currently impossible.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,122 +23,397 @@ package common
...
+	w.informer = cache.NewSharedInformer(listerWatcher, obj, 0)
+	w.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			go func() {
+				runtimeObj, ok := obj.(runtime.Object)
+				if !ok {
+					return
+				}
+
+				w.updateOpResourceVersion(runtimeObj)
```
wojtek-t: handle error

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,122 +23,397 @@ package common
...
 	if err != nil {
-		return summaries, err
+		return err
+	}
+	w.isRunning = true
+	w.stopCh = make(chan struct{})
+	w.informer = cache.NewSharedInformer(listerWatcher, obj, 0)
+	w.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			go func() {
```
wojtek-t: I'm not a fan of starting go-routine in handler.
Generally, the pattern we're using is:
- in handler add object to the queue
- have a worker (set of workers) that take items from queue and process them

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,122 +23,397 @@ package common
...
+		version, err := getResourceVersionFromRuntimeObject(runtimeObj)
+		if err != nil {
+			glog.Errorf("%s: retriving resource version error: %v", waitForControlledPodsRunningName, err)
+			continue
+		}
+		desiredCount++
+		if version > maxResourceVersion {
+			maxResourceVersion = version
+		}
+	}
```
wojtek-t: Please split that into separate function i.e. function that return maxRV from current object (list + compute)

clusterloader2/pkg/measurement/common/wait_for_pods.go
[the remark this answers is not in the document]
wojtek-t: I don't really agree with that, but let's discuss offline.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -231,9 +530,26 @@ func getSelectorFromRuntimeObject(obj runtime.Object) (labels.Selector, error) {
...
+		return typed.Spec, nil
+	case *extensions.DaemonSet:
+		return typed.Spec, nil
+	case *batch.Job:
+		return typed.Spec, nil
+	default:
+		return nil, fmt.Errorf("unsupported kind when getting selector: %v", obj)
+	}
+}
+
```
wojtek-t: Would it be possible to split this PR a bit so that it's smaller.
Maybe those helper functions and objectChecker struct with its methods?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: It is possible to have more then one instance of given measurement.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: Runtime object related functions moved to #270 .

krzysied:
/hold
Waiting for #270 being merged.

krzysied:
@wojtek-t vendor update moved to separate PR - #271

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,239 +23,410 @@ package common
...
+				if err := w.updateOpResourceVersion(runtimeObj); err != nil {
+					return
+				}
+				w.handleObject(c, runtimeObj)
+			}
+			w.queue.Add(&addF)
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			updateF := func() {
+				oldRuntimeObj, ok := oldObj.(runtime.Object)
```
wojtek-t: Can you opaque this into function (and share that between add/update/delete).

```
func (w *...) handleUpdateObject(oldObj, newObj interface{}) {
   // what you have here (maybe you need to handle nil differently, but I think it will work)
}
```

And then here:

```
AddFunc: func(obj interface{}) {
  w.queue.Add(w.handleUpdateObject(nil, obj)
}
...
```

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,239 +23,410 @@ package common
...
+				}
+				w.deleteObject(runtimeObj)
+			}
+			w.queue.Add(&deleteF)
+		},
+	})
+
+	for i := 0; i < waitForControlledPodsWorkers; i++ {
+		w.wg.Start(func() {
+			for {
```
wojtek-t: Opaque this into a separate worker function.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -23,239 +23,410 @@ package common
...
+		glog.Errorf("%s: reading object namespace error: %v", waitForControlledPodsRunningName, err)
+		return
 	}
-	runtimeObjectReplicas, err := getReplicasFromRuntimeObject(obj)
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if !w.isRunning {
+		return
+	}
+	key := createMetaNamespaceKey(objNamespace, objName)
```
wojtek-t: Can you make this function to take runtime.Object instead?

The longer function you write, the harder it is to follow.

krzysied:
/hold
Waiting for #272 and #273 to be merged.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: > Can you opaque this into function (and share that between add/update/delete).

Done.

```
AddFunc: func(obj interface{}) {
  w.queue.Add(w.handleUpdateObject(nil, obj)
}
...
```
Not sure how to do this:
- Function needs to be called after retrieving it from queue, not before. I can change `handleObject` so it returns `func()` but I'm not sure if this will improve readability.
- Due to queue using map internally, function are not support. Although, there is no problem with function pointers.

krzysied:
@wojtek-t PR rebased. PTAL.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+	if version > w.opResourceVersion {
+		w.opResourceVersion = version
+	}
+	return nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion(c clientset.Interface) (int, uint64, error) {
+	var desiredCount int
+	var maxResourceVersion uint64
+	objects, err := runtimeobjects.ListRuntimeObjectsForKind(c, w.kind, w.namespace, w.labelSelector, w.fieldSelector)
```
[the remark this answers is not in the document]
krzysied: Change moved to #274.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+		glog.Infof("%v: wait for controlled pods measurement already running", waitForControlledPodsRunningName)
+		return nil
+	}
+	glog.Infof("%v: starting wait for controlled pods measurement...", waitForControlledPodsRunningName)
+	optionsModifier := func(options *metav1.ListOptions) {
+		options.FieldSelector = w.fieldSelector
+		options.LabelSelector = w.labelSelector
+	}
+	listerWatcher := cache.NewFilteredListWatchFromClient(c.CoreV1().RESTClient(), w.kind+"s", w.namespace, optionsModifier)
+	obj, err := runtimeobjects.GetObjectForKind(w.kind)
```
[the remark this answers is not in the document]
krzysied: Code removed.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+	}
+	go w.informer.Run(w.stopCh)
+	timeoutCh := make(chan struct{})
+	timeoutTimer := time.AfterFunc(informerSyncTimeout, func() {
+		close(timeoutCh)
+	})
+	if !cache.WaitForCacheSync(timeoutCh, w.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+	timeoutTimer.Stop()
```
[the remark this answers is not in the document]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+	return nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) worker() {
+	for {
+		f, stop := w.queue.Get()
+		if stop {
+			return
+		}
+		funcF, ok := f.(*func())
```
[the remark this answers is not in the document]
krzysied: Error shouldn't happen here. I changed it to cast without check.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+		if ok {
+			(*funcF)()
+		}
+		w.queue.Done(f)
+	}
+}
+
+func (w *waitForControlledPodsRunningMeasurement) gather(c clientset.Interface, timeout time.Duration) error {
```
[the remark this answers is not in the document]
krzysied: `timeout` replaced with `syncTimeout`.
Currently `gather` waits until event version >= max (objects versions), then until handling gouroutines finish and finally it verifies if desired object count is acquired. `synTimeout` can happen while waiting for event version being synced.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+		checker.terminate()
+	}
+}
+
+func (w *waitForControlledPodsRunningMeasurement) handleObject(c clientset.Interface, oldObj, newObj interface{}) {
+	var oldRuntimeObj runtime.Object
+	var newRuntimeObj runtime.Object
+	var ok bool
+	oldRuntimeObj, ok = oldObj.(runtime.Object)
+	if oldObj != nil && !ok {
```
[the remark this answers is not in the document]
krzysied: The problem is that we also want to handle `nil`. When `nil` is passed from the event, it doesn't implement `runtime.Object` interface, so it needs to be handled separately.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
-	err = waitForPods(c, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), stopCh, false)
-	if err != nil {
-		return fmt.Errorf("waiting pods error: %v", err)
-	}
-	return nil
+	o := newObjectCheker()
+	o.lock.Lock()
+	defer o.lock.Unlock()
+	go func() {
+		atomic.AddInt32(&w.runningRoutines, 1)
```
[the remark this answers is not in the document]
krzysied: Done. `runningRoutines` replaced with `handlingGroup`.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -32,115 +32,367 @@ import (
...
+		defer func() { atomic.AddInt32(&w.runningRoutines, -1) }()
+		err = waitForPods(clientSet, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), o.stopCh, false)
+		o.lock.Lock()
+		defer o.lock.Unlock()
+		key, _ := runtimeobjects.CreateMetaNamespaceKey(obj)
```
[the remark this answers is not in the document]
krzysied: Handling added.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -210,6 +210,12 @@ func GetReplicasFromRuntimeObject(obj runtime.Object) (int32, error) {
 
 // IsEqualRuntimeObjectsSpec returns true if given runtime objects have identical specs.
 func IsEqualRuntimeObjectsSpec(runtimeObj1, runtimeObj2 runtime.Object) (bool, error) {
+	if runtimeObj1 == nil && runtimeObj2 == nil {
+		return true, nil
+	}
+	if runtimeObj1 == nil || runtimeObj2 == nil {
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+		if err != nil {
+			return summaries, err
+		}
+		w.operationTimeout, err = util.GetDurationOrDefault(config.Params, "operationTimeout", defaultOperationTimeout)
+		if err != nil {
+			return summaries, err
+		}
+		return summaries, w.start(config.ClientSet)
+	case "gather":
+		syncTimeout, err := util.GetDurationOrDefault(config.Params, "syncTimeout", defaultWaitForRCPodsTimeout)
```
wojtek-t: defaultWaitForRCPodsTimeout name is misleading

defaultSyncTimeout maybe?

[BTW - I like the syncTimeout name]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+		w.stop()
+		return summaries, nil
+	default:
+		return summaries, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (w *waitForControlledPodsRunningMeasurement) start(c clientset.Interface) error {
+	if w.informer != nil {
+		glog.Infof("%v: wait for controlled pods measurement already running", waitForControlledPodsRunningName)
```
wojtek-t: printing waitForControlledPodsRunningName (which btw is a pointer) doesn't seem to be extremely helpful, I think.

Maybe just print kind, namespace, field and label selector?

[Same below]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+}
+
+func (w *waitForControlledPodsRunningMeasurement) worker() {
+	for {
+		f, stop := w.queue.Get()
+		if stop {
+			return
+		}
+		funcF := f.(*func())
+		(*funcF)()
```
wojtek-t: (*f.(func())()

[if you're not handling casting errors anyway]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
wojtek-t: Or actually, let's introduce `string()` method on a measurement and just use this in every place in logs.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+	}
+	startTime := time.Now()
+	desiredCount, maxResourceVersion, err := w.getObjectCountAndMaxVersion(c)
 	if err != nil {
-		return summaries, err
+		return err
+	}
+
+	isUpToDate := false
+	for ; time.Since(startTime) < syncTimeout; time.Sleep(checkControlledPodsInterval) {
```
wojtek-t:
```
if err := wait.Poll(checkControllerPodsInterval, syncTimeout, func() (bool, error) { return w.opResourceVersion >= maxResourceVersion, nil}); err != nil {
  rteurn fmt.Errorf(...)
}
```

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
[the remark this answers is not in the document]
wojtek-t: given that it's you who is calling that function, you can also pass runtime.Object(nil) [then it implements runtime.Object]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+
+func (w *waitForControlledPodsRunningMeasurement) handleObject(c clientset.Interface, oldObj, newObj interface{}) {
+	var oldRuntimeObj runtime.Object
+	var newRuntimeObj runtime.Object
+	var ok bool
+	oldRuntimeObj, ok = oldObj.(runtime.Object)
+	if oldObj != nil && !ok {
+		return
+	}
+	defer func() {
```
wojtek-t: Add a comment why you call it in defer (what we discussed offline)

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+		if err := w.updateOpResourceVersion(oldRuntimeObj); err != nil {
+			glog.Errorf("%s: updating resource version error: %v", waitForControlledPodsRunningName, err)
+		}
+	}()
+	newRuntimeObj, ok = newObj.(runtime.Object)
+	if newRuntimeObj != nil && !ok {
+		return
+	}
+	defer func() {
```
wojtek-t: This is incorrect :-) - deferred functions are called in the reversed order.

But you don't need to call that twice - we know that newObj (if exists) has higher RV than the old one.
So if the former is non-nil, use the former, otherwise use the old one [and just one defer]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+		if err := w.updateOpResourceVersion(newRuntimeObj); err != nil {
+			glog.Errorf("%s: updating resource version error: %v", waitForControlledPodsRunningName, err)
+		}
+	}()
+	if isEqual, err := runtimeobjects.IsEqualRuntimeObjectsSpec(oldRuntimeObj, newRuntimeObj); err != nil || isEqual {
+		return
```
wojtek-t: please handle error

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if !w.isRunning {
+		return
+	}
+
+	if oldRuntimeObj != nil {
+		if err := w.deleteObjectLocked(c, oldRuntimeObj); err != nil {
+			glog.Errorf("%s: delete checker error: %v", waitForControlledPodsRunningName, err)
+		} else {
```
wojtek-t: } else if newRuntimeObj == nil {

[If possible, we want to avoid nesting]

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
-			objNamespace, err := runtimeobjects.GetNamespaceFromRuntimeObject(runtimeObjectsList[objectIndex])
-			if err != nil {
-				errList.Append(fmt.Errorf("reading object namespace error: %v", err))
-				return
-			}
-			glog.Infof("%s: %s (%s) has all pods running", waitForControlledPodsRunningName, objName, objNamespace)
-		})
+func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion(c clientset.Interface) (int, uint64, error) {
+	var desiredCount int
+	var maxResourceVersion uint64
```
wojtek-t: Please add a comment why the approach with count + resourceVersion works (something that we discussed offline).

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
 	}
-	wg.Wait()
 
-	if int(runningRuntimeObjects) == len(runtimeObjectsList) {
-		glog.Infof("%s: %d / %d %ss are running with all pods", waitForControlledPodsRunningName, runningRuntimeObjects, len(runtimeObjectsList), kind)
-		return summaries, nil
+	for i := range objects {
+		runtimeObj, ok := objects[i].(runtime.Object)
+		if !ok {
+			glog.Infof("%s: cannot cast to runtime.Object: %v", objects[i])
```
wojtek-t: error?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
-	time.AfterFunc(timeout, func() {
-		close(stopCh)
+	o := newObjectCheker()
+	o.lock.Lock()
+	defer o.lock.Unlock()
+	w.handlingGroup.Start(func() {
+		err = waitForPods(clientSet, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), o.stopCh, false)
+		o.lock.Lock()
+		defer o.lock.Unlock()
+		key, keyErr := runtimeobjects.CreateMetaNamespaceKey(obj)
```
wojtek-t: compute it outside lock (or probably even in the outer function)

clusterloader2/testing/load/config.yaml
```diff
@@ -38,6 +38,15 @@ steps:
...
       action: reset
 # Create RCs
+- measurements:
+  - Identifier: WaitForRunningRCs
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      kind: ReplicationController
+      labelSelector: kind = loadRC
+      # operationTimeout = (totalPodsRounded /20)s + 3m
```
wojtek-t: Operation timeout is for a single RC - it now should depend on the size of the individual RC.

Currently we set it to 10m: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/load.go#L544
but ideally (not in this PR, but would be good to have some TODO for it somewhere) we should make it depend on how much work should be done.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: > Maybe just print kind, namespace, field and label selector

I think that we discussed this offline (in context of wait_for_pods.go) and decided to change measurement so it uses identifier in logs.

> Or actually, let's introduce string() method on a measurement and just use this in every place in logs.

Let's discuss this offline.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: > This is incorrect :-) - deferred functions are called in the reversed order.

`updateOpResourceVersion` updates only to higher version. Order shouldn't matter here.

> But you don't need to call that twice - we know that newObj (if exists) has higher RV than the old one.
So if the former is non-nil, use the former, otherwise use the old one [and just one defer]

Will performance gain justify adding additional `if`? I mean this code already looks complicated :/ Nils are already handled within `updateOpResourceVersion`.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/testing/load/config.yaml

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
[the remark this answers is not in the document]
krzysied: `runtime.Object(nil)` doesn't help. `x.(type)` cast doesn't work for `nil`.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,380 @@ package common
...
+	if w.informer == nil {
+		return fmt.Errorf("metric %s has not been started", waitForControlledPodsRunningName)
+	}
+	desiredCount, maxResourceVersion, err := w.getObjectCountAndMaxVersion(c)
 	if err != nil {
-		return summaries, err
+		return err
+	}
+
+	if err := wait.Poll(checkControlledPodsInterval, syncTimeout, func() (bool, error) { return w.opResourceVersion >= maxResourceVersion, nil }); err != nil {
```
wojtek-t: nit: can you please split this line?
or maybe:
```
cond := func() (bool error) { return w.oResourceVersion >= maxResourceVersion, nil }
if err := wait.Poll(checkControllerPodInterval, cond); err != nil {
   ...
```

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
wojtek-t: The order doesn't matter now, but if we change the function to do something else, it may result in error.
That said, I agree that maybe additional if would be an overkill.
But I think the following would be clearer:
```
if oldRuntimeObj, ok = oldObj.(runtime.Object); !ok && oldObj != nil {
  glog.Errorf(...) 
  return
}
if newRuntimeObj, ok := newObj.(runtime.Object); !ok && newObj != nil {
  glog.Errorf(...)
  return
}
defer func() {
  // We want ...
  if err := w.updateOpResourceVersion(oldRuntimeObj); err != nil  { ...
  if err := w.updateOpResourceVersion(newRuntimeObj); err != nil { ..
}
```

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,385 @@ package common
...
+		return
+	}
+	defer func() {
+		// We want to update version after (potentially) creating goroutine.
+		if err := w.updateOpResourceVersion(newRuntimeObj); err != nil {
+			glog.Errorf("%s: updating resource version error: %v", w, err)
+		}
+	}()
+	if isEqual, err := runtimeobjects.IsEqualRuntimeObjectsSpec(oldRuntimeObj, newRuntimeObj); err != nil || isEqual {
+		if err != nil {
```
wojtek-t: We generally should try to avoid nesting whenever it's possible. So this is better.
```
isEqual, err := ...
if err != nil {
  glog.Errorf("...")
  return
}
if isEqual {
  // Add comment
  return
}
```

Even though it's a bit longer, the flow is much more obvious now.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,385 @@ package common
...
+		return
 	}
-	labelSelector, err := util.GetStringOrDefault(config.Params, "labelSelector", "")
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if !w.isRunning {
+		return
+	}
+
+	if oldRuntimeObj != nil {
```
wojtek-t: Can you move handling errors to the createObject and deleteObject methods?
[i.e. if obj == nil {
  return
}]

This one would be cleaner then.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: After offline discussion, we agreed to add `String()` method to waitForControlledPods measurement. Additionally #275 issue has been created to update all metrics.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
wojtek-t: Why not compute it in the outer function?
if the object is incorrect, we have a way to return an error there.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,385 @@ package common
...
+		if err := w.deleteObjectLocked(c, oldRuntimeObj); err != nil {
+			glog.Errorf("%s: delete checker error: %v", w, err)
```
wojtek-t: I just realized that a bunch of those glog.Errors shoudl actually fail the test (in the end, not immediately).
glog.Error doesn't have that property.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: That's true. However, I expect that all small errors will have impact on final measurement check (e.g. failing to check object deletion will result on failing object count check), so all in all whole measurement will fail.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: Good idea. Done.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,390 @@ package common
...
+func (w *waitForControlledPodsRunningMeasurement) handleObject(c clientset.Interface, oldObj, newObj interface{}) {
+	var oldRuntimeObj runtime.Object
+	var newRuntimeObj runtime.Object
+	var ok bool
+	oldRuntimeObj, ok = oldObj.(runtime.Object)
+	if oldObj != nil && !ok {
+		return
+	}
+	newRuntimeObj, ok = newObj.(runtime.Object)
+	if newRuntimeObj != nil && !ok {
```
wojtek-t: newObj (or oldRuntimeObj above)

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -24,123 +24,390 @@ package common
...
+		return
```
wojtek-t: don't fail silently

> That's true. However, I expect that all small errors will have impact on final measurement check (e.g. failing to check object deletion will result on failing object count check), so all in all whole measurement will fail.

Not necessary. For this particular case, imagine that RC was successfully created and thus it has a checker with final status.
But then if you want to update it and you fail e.g. here, it simply won't have any effect. - we won't do nothing, but the measurement itself will succeed.

As i said, a bunch of errors here, should cause test to fail (though not immediately).

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: Good catch. Fixed.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
krzysied: I've added `glog.Errorf`.

> But then if you want to update it and you fail e.g. here, it simply won't have any effect. - we won't do nothing, but the measurement itself will succeed.

I see two possible scenarios:
- There is unrecognizable object - the cast fails or some of necessary fields are not accessible. I think it should be log - it something unexpected, but that's definitely not a RC (or other controlling pods object), so it shouldn't affect wait for controlled pods measurement.
- Object is recognized, therefore checker is created. If checker fails, it will be visible because of he status. The measurement will fail.

IMHO, this is handled properly.

Anyway, if you're suggesting that errors should be handle in the different way, maybe we want to create issue? TBH I think we want to moved this topic to the different PR to iron out the proper solution.

wojtek-t: Being unable to cast means some logical error in the code - we are doing something really wrong then. This should fail the test.

I'm fine with opening a separate issue for that and solving it separately (it may be the case fo different measurements and other parts of code). But this needs to be fixed before we switch to cluster-loader for running our tests.

krzysied: > But this needs to be fixed before we switch to cluster-loader for running our tests.

I agree. I created #276 - this will allow us to track error fixing status for all measurements.

PR #266, opened 2018-10-24
ClusterLoader - Adding resources versions state

krzysied:
- Adding resources versions state
This is a map {apiVersion, Kind} -> {resourceVersion}
File `pkg/state/resource_state.go`.
- Adding state that contains both namespaces state (previous state) and resources versions state
Files: `pkg/state/state.go`, `pkg/state/namespaces_state.go`.
- Applying changes
Files: all not listed above.

### DIFF
--- a/clusterloader2/pkg/state/resource_state.go
+++ b/clusterloader2/pkg/state/resource_state.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state
+
+import (
+	"fmt"
+	"strconv"
+	"sync"
+)
+
+// ResourceTypeIdentifier is a unique identifier for a resource type.
+type ResourceTypeIdentifier struct {
+	ObjectKind string
+	ApiGroup   string
+}
+
+// resourcesVersionsState represents most recent resources versions for a given object types.
+// These versions are available only for resource types of objects that were create/updated
+// by the clusterloader.
+type resourcesVersionsState struct {
+	lock              sync.RWMutex
+	resourcesVersions map[ResourceTypeIdentifier]uint64
+}
+
+// newResourcesVersionState creates new resources versions state.
+func newResourcesVersionsState() *resourcesVersionsState {
+	return &resourcesVersionsState{
+		resourcesVersions: make(map[ResourceTypeIdentifier]uint64),
+	}
+}
+
+// Get returns state of current resource version.
+func (rs *resourcesVersionsState) Get(identifier ResourceTypeIdentifier) (string, bool) {
+	rs.lock.RLock()
+	defer rs.lock.RUnlock()
+	version, exists := rs.resourcesVersions[identifier]
+	if !exists {
+		return "0", false
+	}
+	return strconv.FormatUint(version, 10), true
+}
+
+// Set stores information about current resource version.
+func (rs *resourcesVersionsState) Set(identifier ResourceTypeIdentifier, resourceVersion string) error {
+	rs.lock.Lock()
+	defer rs.lock.Unlock()
+	version, err := strconv.ParseUint(resourceVersion, 10, 64)
+	if err != nil {
+		return fmt.Errorf("incorrect version format")
+	}
+	if current, exists := rs.resourcesVersions[identifier]; exists && current > version {
+		version = current
+	}
+	rs.resourcesVersions[identifier] = version
+	return nil
+}
--- a/clusterloader2/pkg/state/namespaces_state.go
+++ b/clusterloader2/pkg/state/namespaces_state.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state
+
+import (
+	"fmt"
+	"sync"
+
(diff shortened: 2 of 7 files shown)


clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,68 @@
...
+	"sync"
+)
+
+// ResourceIdentifier is a unique identifier for resource
+type ResourceIdentifier struct {
+	ObjectKind string
+	ApiGroup   string
+}
+
+// resourcesVersionsState represents current version of all used resources.
```
wojtek-t: This comment is misleading.

It represents the most recent RV from objects of a given type that we created/updated/deleted ourselves.

clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Set stores information about current resource version.
+func (rs *resourcesVersionsState) Set(identifier ResourceIdentifier, version string) {
+	rs.lock.Lock()
+	defer rs.lock.Unlock()
+	rs.resourcesVersions[identifier] = version
+}
+
+// Delete removes information about given resource type.
+func (rs *resourcesVersionsState) Delete(identifier ResourceIdentifier) error {
```
wojtek-t: I don't think that's useful - I don't see a way how this could be used :-)

krzysied: Function removed.

clusterloader2/pkg/state/resource_state.go
krzysied: Comment updated.

krzysied:
@wojtek-t  I've changed version type from string to uint64.

clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,63 @@
...
+
+// newResourcesVersionState creates new resources versions state.
+func newResourcesVersionsState() *resourcesVersionsState {
+	return &resourcesVersionsState{
+		resourcesVersions: make(map[ResourceIdentifier]uint64),
+	}
+}
+
+// Get returns state of current resource version.
+func (rs *resourcesVersionsState) Get(identifier ResourceIdentifier) (uint64, bool) {
```
wojtek-t: The API should be string-based.
We may hold uint64 internally, but the API should be string-based.

wojtek-t: i.e. the RV is string and we should do convertions here, rather than everywhere where we will use it.

The same for Set.

krzysied: Done.

clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+// Get returns state of current resource version.
+func (rs *resourcesVersionsState) Get(identifier ResourceIdentifier) (string, bool) {
+	rs.lock.RLock()
+	defer rs.lock.RUnlock()
+	versionUInt64, exists := rs.resourcesVersions[identifier]
+	if !exists {
+		return "", false
```
wojtek-t: Maybe that's fine, but I'm thinking about two other possible apis:
- return just string (and in this case, default to "0") [though it has disadvantages]
- returning (string, error) and returning notFound here (that makes it more extendable in the future if needed).

clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Set stores information about current resource version.
+func (rs *resourcesVersionsState) Set(identifier ResourceIdentifier, version string) error {
+	rs.lock.Lock()
+	defer rs.lock.Unlock()
+	versionUInt64, err := strconv.ParseUint(version, 10, 64)
+	if err != nil {
+		return fmt.Errorf("incorrect version format")
+	}
+	if _, exists := rs.resourcesVersions[identifier]; !exists {
```
wojtek-t: I think that the following is easier to follow:
```
if currrent, exists := rs.resourceVersions[identified]; exists && current > versionUint64 {
  versionUint64 = current
}
rs.resourceVersions[identifier] = versionUint64
```

clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,73 @@
...
+		return "", false
+	}
+	return strconv.FormatUint(versionUInt64, 10), true
+}
+
+// Set stores information about current resource version.
+func (rs *resourcesVersionsState) Set(identifier ResourceIdentifier, version string) error {
+	rs.lock.Lock()
+	defer rs.lock.Unlock()
+	versionUInt64, err := strconv.ParseUint(version, 10, 64)
```
wojtek-t: versionUint64 is a strange name :-)
Maybe make this version and function arg resourceVersion ?

clusterloader2/pkg/state/resource_state.go
```diff
@@ -0,0 +1,73 @@
...
+package state
+
+import (
+	"fmt"
+	"strconv"
+	"sync"
+)
+
+// ResourceIdentifier is a unique identifier for resource
+type ResourceIdentifier struct {
```
wojtek-t: nit: ResourceIdentifier may be a misleading name.
Maybe simply ResourceType ?

clusterloader2/pkg/state/resource_state.go
krzysied: The reasons I would prefer to keep current api:
- It similar to what golang map has.
- `namespaceState` has the same api. IMHO having two similar structs with two different apis sounds bad.

We can change both apis. The question here is whether implementing `notFound` error brings anything new.

clusterloader2/pkg/state/resource_state.go
krzysied: You are right. Done.

clusterloader2/pkg/state/resource_state.go
krzysied: Changed to `version` and `resourceVersion`.

clusterloader2/pkg/state/resource_state.go
krzysied: TBH I prefer `ResourceIdentifier`, cause it an identifier. It just happened to be the same thing as resource type at this point.
I can see a case when we decide to add (for example) namespace to this struct.  It would be still identifier, but not a resource type.

clusterloader2/pkg/state/resource_state.go
wojtek-t: It's more clean in my opinion, but I'm not going to insist on that. My arguments are comparably weak to yours, to let's not change anything :-)

clusterloader2/pkg/state/resource_state.go
wojtek-t: But Resource is something different - it's an instance of an object for example.

I'm fine with leaving "Identifier" suffix, but it's not a resource.
Since I don't think we will add anything to it anytime soon, I would go with ResourceTypeIdentifier, and if at some point we will have to add something, we will change the name.

krzysied: I get your point. Renamed to `ResourceTypeIdentifier`.

PR #267, opened 2018-10-25
ClusterLoader - Storing resources versions

krzysied:
- `CreateObject` and `PatchObject` return server objects
- test executor stores resource types versions in State.

### DIFF
--- a/clusterloader2/pkg/framework/client/objects.go
+++ b/clusterloader2/pkg/framework/client/objects.go
@@ -137,19 +137,25 @@ func WaitForDeleteNamespace(c clientset.Interface, namespace string) error {
 }
 
 // CreateObject creates object based on given object description.
-func CreateObject(dynamicClient dynamic.Interface, namespace string, name string, obj *unstructured.Unstructured) error {
+func CreateObject(dynamicClient dynamic.Interface, namespace string, name string, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	var newObj *unstructured.Unstructured
 	gvk := obj.GroupVersionKind()
 	gvr, _ := meta.UnsafeGuessKindToResource(gvk)
 	obj.SetName(name)
 	createFunc := func() error {
-		_, err := dynamicClient.Resource(gvr).Namespace(namespace).Create(obj)
+		var err error
+		newObj, err = dynamicClient.Resource(gvr).Namespace(namespace).Create(obj)
 		return err
 	}
-	return RetryWithExponentialBackOff(retryFunction(createFunc, apierrs.IsAlreadyExists))
+	if err := RetryWithExponentialBackOff(retryFunction(createFunc, apierrs.IsAlreadyExists)); err != nil {
+		return nil, err
+	}
+	return newObj, nil
 }
 
 // PatchObject updates (using patch) object with given name, group, version and kind based on given object description.
-func PatchObject(dynamicClient dynamic.Interface, namespace string, name string, obj *unstructured.Unstructured) error {
+func PatchObject(dynamicClient dynamic.Interface, namespace string, name string, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	var newObj *unstructured.Unstructured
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -258,12 +258,22 @@ func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, na
 	errList := util.NewErrorList()
 	switch operation {
 	case CREATE_OBJECT:
-		if err := ctx.GetFramework().CreateObject(namespace, objName, obj); err != nil {
+		if newObj, err := ctx.GetFramework().CreateObject(namespace, objName, obj); err != nil {
 			errList.Append(fmt.Errorf("namespace %v object %v creation error: %v", namespace, objName, err))
+		} else {
+			err = updateResourceVersion(ctx, newObj)
```
wojtek-t: if err := updateResourceVersion(ctx, newObj); err != nil {
}

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -258,12 +258,22 @@ func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, na
...
+			if err != nil {
+				errList.Append(fmt.Errorf("namespace %v object %v resource version paring error: %v", namespace, objName, err))
+			}
 		}
 	case PATCH_OBJECT:
-		if err := ctx.GetFramework().PatchObject(namespace, objName, obj); err != nil {
+		if newObj, err := ctx.GetFramework().PatchObject(namespace, objName, obj); err != nil {
 			errList.Append(fmt.Errorf("namespace %v object %v updating error: %v", namespace, objName, err))
+		} else {
```
wojtek-t: same here

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -311,6 +321,14 @@ func createNamespacesList(ctx Context, namespaceRange *api.NamespaceRange) []str
 	return nsList
 }
 
+func updateResourceVersion(ctx Context, obj *unstructured.Unstructured) error {
+	identifier := state.ResourceTypeIdentifier{
+		ObjectKind: obj.GetKind(),
+		ApiGroup:   obj.GetAPIVersion(),
```
wojtek-t: GetAPIVersion is not ApiGroup.

GetAPIVersion returns "group/version", which is not exactly what we want.

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -162,10 +168,13 @@ func PatchObject(dynamicClient dynamic.Interface, namespace string, name string,
...
 		}
-		_, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.StrategicMergePatchType, patch)
+		newObj, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.StrategicMergePatchType, patch)
 		return err
 	}
-	return RetryWithExponentialBackOff(retryFunction(updateFunc, nil))
+	if err := RetryWithExponentialBackOff(retryFunction(updateFunc, nil)); err != nil {
+		return nil, err
+	}
+	return newObj, nil
```
wojtek-t: This approach has a visible problem, that we don't have a way to detect RV of Delete operation (which is pretty crucial) - we want to track that too.

And I don't see a good solution for that now.
And it seems that we may have over-engineered that.

After thinking about that more, I think we may significantly simplify that by:
- not tracking RVs at state level at all
- when you call "gather" in the measurement you:
 (a) first call "LIST" on that kind (say RCs) and store the RV from the ListMeta
 (b) wait in the watch until it's synced with that RV
 (c) do the real waiting for all of the started go-routines

Sorry for introducing unnecessary complications - the above seems to be much easier and doesn't have the problem with deletions (and doesn't require changes all over the code).

wojtek-t:
As discussed offline - this won't be needed. So closing.

PR #273, opened 2018-10-29
ClusterLoader - Wait for pods update

krzysied:


krzysied:
/hold 
Requires #271 
Possible conflict with #272

clusterloader2/pkg/measurement/common/wait_for_pods.go
```diff
@@ -70,31 +70,39 @@ func (*waitForRunningPodsMeasurement) Execute(config *measurement.MeasurementCon
...
-		}
-		if runningPodsCount == desiredPodCount {
-			return nil
+			if log {
+				glog.Infof("WaitForRunningPods: namespace(%s), labelSelector(%s), fieldSelector(%s): running %d / %d", namespace, labelSelector, fieldSelector, runningPodsCount, desiredPodCount)
+			}
+			if runningPodsCount == desiredPodCount {
+				return nil
+			}
+			time.Sleep(defaultWaitForPodsInterval)
```
wojtek-t: select {
case <- stopCh
case <- time.After()
}

[we want to break as soon as we can]

Issue #269, opened 2018-10-26
ClusterLoader: Using shared podStrore


krzysied:
Creating multiple podStores at the same time result in some podStores not being created. Measurements should use shared podStore instead of creating multiple podStores to avoid this problem.

mm4tt:
Looks like there was an attempt to fix that, but it was decided to "not do that now" - https://github.com/kubernetes/perf-tests/pull/279#issuecomment-460151213
@wojtek-t could you elaborate more on the reasoning behind that decision? 

It seems that the issue still happens and causes flakiness, see https://github.com/kubernetes/kubernetes/issues/73461#issuecomment-460964397

wojtek-t:
I didn't look into the failure itself, but I'm not convinced this would fix it.

There are couple reasons why I decided to drop it now:
- shared pod store would mean we are keeping all pods in test memory, which itself is pretty expensive
- creating pod store for every RC is a bit better simulating what users may be doing (i.e. list+watch what is happening with my pods)

krzysied:
As for now, we decided to not implement shared pod store.
I'm closing the issue.

PR #279, opened 2018-11-05
ClusterLoader - Implementing shared pod store

krzysied:
- Implementing shared pod store. Shared pod store collects information about all pods and allows to filter pods on client side.
- Removing field selector.

ref #269


wojtek-t:
Let's close - we decided not to do that now.

Issue #281, opened 2018-11-06
ClusterLoader: adding overrides to test configs


krzysied:
Cluster loader test configs should support overrides:
- There should be a way to pass value to variable through the test call.
- If value is not provided, default (specified in config) should be used.

PR #282, opened 2018-11-06
ClusterLoader - Adding mapping overrides

krzysied:
ref #281


clusterloader2/pkg/config/codec.go
```diff
@@ -44,3 +44,10 @@ func convertToObject(raw []byte) (*unstructured.Unstructured, error) {
 	}
 	return obj, nil
 }
+
+func convertToGivenStruct(raw []byte, v interface{}) error {
```
wojtek-t: decodeInto

krzysied: Done.

PR #284, opened 2018-11-07
ClusterLoader - Applying default params to test configs

### LINKED ISSUE #281: ClusterLoader: adding overrides to test configs

krzysied:
fixes #281

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -3,11 +3,11 @@
 # - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
 
 #Constants
-{{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+{{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
+{{$TEST_THROUGHPUT := DefaultParam .TEST_THROUGHPUT 20}}
 {{$MIN_LATENCY_PODS := 500}}
 {{$MIN_SATURATION_PODS_TIMEOUT := 180}}
-{{$TEST_THROUGHPUT := 20}}
 
 #Variables
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -3,23 +3,23 @@
 # - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
 
 #Constants
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+{{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
+{{$TEST_THROUGHPUT := DefaultParam .TEST_THROUGHPUT 10}}
 {{$BIG_GROUP_SIZE := 250}}
 {{$MEDIUM_GROUP_SIZE := 30}}
 {{$SMALL_GROUP_SIZE := 5}}
-{{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
-{{$TEST_THROUGHPUT := 10}}
-# bigRCs - 1/4 of all pods should be in big RCs.
-{{$BIG_RCS_PER_NAMESPACE := 3}}
-# mediumRCs - 1/4 of all pods should be in medium RCs.
-{{$MEDIUM_RCS_PER_NAMESPACE := 25}}
-# smallRCs - 1/2 of all pods should be in small RCs.
-{{$SMALL_RCS_PER_NAMESPACE := 300}}
 
 #Variables
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
 {{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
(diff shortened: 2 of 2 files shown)


krzysied:
@wojtek-t `TEST_THROUGHPUT` is an example of variable that has different values in different tests. IMHO at some point we should make overrides per tests instead of having global one.

clusterloader2/testing/density/config.yaml
```diff
@@ -3,11 +3,11 @@
...
 
 #Constants
-{{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
-{{$MIN_LATENCY_PODS := 500}}
-{{$MIN_SATURATION_PODS_TIMEOUT := 180}}
-{{$TEST_THROUGHPUT := 20}}
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+{{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
+{{$MIN_LATENCY_PODS := DefaultParam .MIN_LATENCY_PODS 500}}
```
wojtek-t: This shouldn't be modify-able.

clusterloader2/testing/density/config.yaml
```diff
@@ -3,11 +3,11 @@
...
+{{$MIN_SATURATION_PODS_TIMEOUT := DefaultParam .MIN_SATURATION_PODS_TIMEOUT 180}}
```
wojtek-t: Sa,e here.

clusterloader2/testing/load/config.yaml
```diff
@@ -3,18 +3,18 @@
...
 #Constants
-{{$BIG_GROUP_SIZE := 250}}
-{{$MEDIUM_GROUP_SIZE := 30}}
-{{$SMALL_GROUP_SIZE := 5}}
-{{$NODES_PER_NAMESPACE := 100}}
-{{$PODS_PER_NODE := 30}}
-{{$TEST_THROUGHPUT := 10}}
+{{$BIG_GROUP_SIZE := DefaultParam .BIG_GROUP_SIZE 250}}
+{{$MEDIUM_GROUP_SIZE := DefaultParam .MEDIUM_GROUP_SIZE 30}}
+{{$SMALL_GROUP_SIZE := DefaultParam .SMALL_GROUP_SIZE 5}}
```
wojtek-t: Neither of those.

clusterloader2/testing/load/config.yaml
```diff
@@ -3,18 +3,18 @@
...
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+{{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
+{{$TEST_THROUGHPUT := DefaultParam .TEST_THROUGHPUT 10}}
 # bigRCs - 1/4 of all pods should be in big RCs.
-{{$BIG_RCS_PER_NAMESPACE := 3}}
+{{$BIG_RCS_PER_NAMESPACE := DefaultParam .BIG_RCS_PER_NAMESPACE 3}}
```
wojtek-t: Also neither of those.

clusterloader2/testing/density/config.yaml

clusterloader2/testing/density/config.yaml

clusterloader2/testing/load/config.yaml

clusterloader2/testing/load/config.yaml
krzysied: I'm not convinced.
`BIG_RCS_PER_NAMESPACE` is a value that depends on `BIG_GROUP_SIZE`, `NODES_PER_NAMESPACE` and `PODS_PER_NODE`. Assuming that we allow to modify any of those, `BIG_RCS_PER_NAMESPACE` should also be modifiable.

wojtek-t: OK - I see your point.
So maybe this should actually not be hardcoded by switched to: "POD_PER_NODE * Nodes / 1000" (and similarly for the below ones).

However, I changed assumption from "all pods / 4" to "namespace / 4", due to the fact, that we want to have identical namespaces.

clusterloader2/testing/load/config.yaml
```diff
@@ -3,23 +3,23 @@
...
-{{$MEDIUM_RCS_PER_NAMESPACE := 25}}
-# smallRCs - 1/2 of all pods should be in small RCs.
-{{$SMALL_RCS_PER_NAMESPACE := 300}}
 
 #Variables
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
 {{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
 {{$saturationTime := DivideInt $totalPods $TEST_THROUGHPUT}}
+# bigRCs - 1/4 of namespace pods should be in big RCs.
+{{$bigRcsPerNamespace := DivideInt (MultiplyInt $NODES_PER_NAMESPACE $PODS_PER_NODE) (MultiplyInt 4 $BIG_GROUP_SIZE)}}
```
wojtek-t: I don't think that's correct. This is 

"pods per namespace / 1000" - it doesn't depend on cluster size. And clearly it should.

krzysied: >  it doesn't depend on cluster size. And clearly it should.

I don't agree. At least as long as we keep the assumption that each namespace is identical.
Currently we need to define the content of namespace. Cluster size has only impact on namespace count.

We can drop the assumption that each namespace is identical, however it will make the test config much more complex.

wojtek-t: aah ok - beause it's just for single namespace.

PR #286, opened 2018-11-07
ClusterLoader - Test metrics measurements bundle

krzysied:


clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+func init() {
+	measurement.Register(testMetricsMeasurementName, createTestMetricsMeasurment)
+}
+
+func createTestMetricsMeasurment() measurement.Measurement {
+	var metrics testMetrics
+	var err error
+	if metrics.apiResponsiveness, err = measurement.CreateMeasurement("APIResponsiveness"); err != nil {
```
wojtek-t: This can fail test, right? Then it shouldn't be here.

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -0,0 +1,184 @@
...
+		summary, err = execute(t.resourceUsageSummary, config, map[string]interface{}{
+			"action": "start",
+		})
+		appendResults(&summaries, &errs, summary, err)
+		summary, err = execute(t.schedulingMetrics, config, map[string]interface{}{
+			"action": "reset",
+		})
+		appendResults(&summaries, &errs, summary, err)
+	case "gather":
+		summary, err := execute(t.apiResponsiveness, config, map[string]interface{}{
```
wojtek-t: Define options once, instead of defining it for every call separately.

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -0,0 +1,184 @@
...
+		appendResults(&summaries, &errs, summary, err)
+	default:
+		return summaries, fmt.Errorf("unknown action %v", action)
+	}
+
+	if len(errs) > 0 {
+		for i := range errs {
+			glog.Errorf("%s: %v", testMetricsMeasurementName, errs[i])
+		}
+		return summaries, fmt.Errorf("test error")
```
wojtek-t: return errors aggregation of errs

clusterloader2/pkg/measurement/common/simple/api_responsiveness.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package common
+package simple
```
wojtek-t: This isn't simple - this is regular SLO measurement.
It shouldn't be part of that bundle.

clusterloader2/testing/density/config.yaml
```diff
@@ -26,14 +26,10 @@ tuningSets:
     qps: 5
 steps:
 - measurements:
-  - Identifier: APIResponsiveness
-    Method: APIResponsiveness
+  - Identifier: TestMetrics
```
wojtek-t: APIResponsiveness should be defined explicitly - (at least for now).

Medium term, we may create a bundle for "slo validation" or sth like that.

clusterloader2/testing/load/config.yaml
```diff
@@ -36,10 +36,10 @@ tuningSets:
     timeLimit: {{DivideInt $saturationTime 4}}s
 steps:
 - measurements:
-  - Identifier: APIResponsiveness
-    Method: APIResponsiveness
+  - Identifier: TestMetrics
```
wojtek-t: Same here.

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
krzysied: APIResponsiveness removed from the bundle.

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
krzysied: The problem is that there are metrics that use the same key but with different values ("action": "start" | "action": "reset"), so it can't be defined once.

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
krzysied: Done.

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
wojtek-t: Then define two or three or sth sets of those, e.g.

startParams, resetParams, and so on.

clusterloader2/pkg/measurement/common/simple/api_responsiveness.go
krzysied: I've assumed that there are two types of metrics:
- simple - which is a standalone metric.
- bundle - aggregation of other metrics.

Even though the `ApiResponsiveness` is not a part of a bundle, IMHO it should be classified as simple.

clusterloader2/testing/density/config.yaml

clusterloader2/testing/load/config.yaml

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/api_responsiveness.go
wojtek-t: it should be classified as slos (or something like that) together with pod startup time

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		summary, err = execute(t.resourceUsageSummary, actionStartConfig)
+		appendResults(&summaries, errList, summary, err)
+		summary, err = execute(t.schedulingMetrics, actionResetConfig)
+		appendResults(&summaries, errList, summary, err)
+	case "gather":
+		summary, err := execute(t.etcdMetrics, actionGatherConfig)
+		appendResults(&summaries, errList, summary, err)
+		summary, err = execute(t.metricsForE2E, config)
+		appendResults(&summaries, errList, summary, err)
+		summary, err = execute(t.apiserverCPUProfile, kubeApiserverConfig)
```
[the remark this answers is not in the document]
krzysied: Issue created #296

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -0,0 +1,170 @@
...
+	if metrics.apiserverMemoryProfile, err = measurement.CreateMeasurement("MemoryProfile"); err != nil {
+		glog.Errorf("TestMetrics: apiserverMemoryProfile creation error: %v", err)
+	}
+	if metrics.schedulerMemoryProfile, err = measurement.CreateMeasurement("MemoryProfile"); err != nil {
+		glog.Errorf("TestMetrics: schedulerMemoryProfile creation error: %v", err)
+	}
+	if metrics.resourceUsageSummary, err = measurement.CreateMeasurement("ResourceUsageSummary"); err != nil {
+		glog.Errorf("TestMetrics: resourceUsageSummary creation error: %v", err)
+	}
+	if metrics.schedulingMetrics, err = measurement.CreateMeasurement("SchedulingMetrics"); err != nil {
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/api_responsiveness.go
krzysied: `ApiResponsiveness` and `PodStartupLatency` moved to slos subdirectory.

PR #300, opened 2018-11-16
ClusterLoader - Logging nodes

krzysied:
Logging available nodes list (name, internal ip, external ip) before running tests.


clusterloader2/cmd/clusterloader.go
```diff
@@ -100,6 +101,27 @@ func completeConfig(f *framework.Framework) error {
 	return nil
 }
 
+func printClusterNodes(f *framework.Framework) error {
```
wojtek-t: Can we move it out of main file? Probably to some utility file somewhere (in framework?)

clusterloader2/cmd/clusterloader.go
```diff
@@ -100,6 +101,27 @@ func completeConfig(f *framework.Framework) error {
...
+	nodes, err := util.GetSchedulableUntainedNodes(f.GetClientSet())
+	if err != nil {
+		return fmt.Errorf("getting nodes error: %v", err)
+	}
+	glog.Infof("Listing clsuter nodes:")
```
wojtek-t: cluster

clusterloader2/cmd/clusterloader.go
```diff
@@ -100,6 +101,27 @@ func completeConfig(f *framework.Framework) error {
 	return nil
 }
 
+func printClusterNodes(f *framework.Framework) error {
+	nodes, err := util.GetSchedulableUntainedNodes(f.GetClientSet())
```
wojtek-t: Let's list all the nodes (not just schedulable and untainted) - something that may initially be tainted may be untainted later e.g.

However I also added information about node being schedulable/unschedulable to log. IMHO this information will be useful, due to the nodes parameter from cluster config being initialize based on this property.

clusterloader2/pkg/util/cluster.go
```diff
@@ -41,6 +41,29 @@ func GetSchedulableUntainedNodesNumber(c clientset.Interface) (int, error) {
 	return numNodes, err
 }
 
+// LogClusterNodes prints nodes infromation (name, internal ip, external ip) to log.
```
wojtek-t: information

clusterloader2/cmd/clusterloader.go
```diff
@@ -119,6 +119,10 @@ func main() {
 		glog.Fatalf("Config completing error: %v", err)
 	}
 
+	if err = util.LogClusterNodes(f.GetClientSet()); err != nil {
+		glog.Fatalf("Nodes info logging error: %v", err)
```
wojtek-t: This probably shouldn't fail the test immediately.

clusterloader2/pkg/util/cluster.go
```diff
@@ -41,6 +41,29 @@ func GetSchedulableUntainedNodesNumber(c clientset.Interface) (int, error) {
...
+func LogClusterNodes(c clientset.Interface) error {
+	nodeList, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
```
wojtek-t: Can we retry this?

clusterloader2/pkg/util/cluster.go
krzysied: Done.

clusterloader2/cmd/clusterloader.go
krzysied: Changed to `Errorf`. However, I we are not able to list nodes, should it mean that something serious happened?

clusterloader2/pkg/util/cluster.go
krzysied: Retry added to this function (and all other list functions in this file).

clusterloader2/cmd/clusterloader.go
wojtek-t: Not necessary - maybe your request just failed because of network issue?

PR #303, opened 2018-11-19
Add support to test node-local-cache dns server

prameshj:
Added new test to query multiple service names
Also a flag to run large number of external hostname queries


krzysied:
@prameshj
Is there a way to split this PR into smaller ones (or at least divide this into commits)?
IIUC there are three separate changes here:
- adding new query file
- running top command and gathering cpu and memory usage
- adding flags and additional logic

dns/py/runner.py
```diff
@@ -197,6 +262,10 @@ def _run_perf(self, test_case, inputs):
 
         for key, value in parser.results.items():
           results['data'][key] = value
+        results['data']['max_perfserver_cpu'] = res_usage.get()
```
krzysied: IIUC `_run_time` calculates max resource usage  and puts it into the queue every 2 second. Here we use `get`, so it returns first element of mentioned queue. So will it use max resource usage of 2 first second of runtime?

prameshj: Sorry for the very late response, I somehow thought i already replied to this. Great catch, I see I have some indenting issue. We want to add to the queue at the end of the while loop. Fixed it.

dns/py/params.py
```diff
@@ -276,6 +276,15 @@ def iterate(remaining, pv):
 
     return cases
 
+  def add_param(self, param_name, values):
```
bowei: semantics for this seem weird -- it won't add unless it already exists?

dns/py/params.py
```diff
@@ -276,6 +276,15 @@ def iterate(remaining, pv):
...
+    if param_name not in self.values:
+      return
+    self.values[param_name].append(values)
+
+  def get_param(self, param_name):
+    if param_name not in self.values:
+      return
```
bowei: return None?

dns/py/params.py
prameshj: good point, renamed it to set_param

dns/py/params.py
prameshj: done

bowei:
I kind of regret implementing this in Python :-P

PR #307, opened 2018-11-22
ClusterLoader - handling SLOs fails

krzysied:
Tests will continue if there are only metric violation errors.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -127,7 +127,8 @@ func main() {
 	for _, clusterLoaderConfig.TestConfigPath = range testConfigPaths {
 		glog.Infof("Running %v", clusterLoaderConfig.TestConfigPath)
 		if errList := test.RunTest(f, &clusterLoaderConfig); !errList.IsEmpty() {
-			glog.Fatalf("Test execution failed: %v", errList.String())
+			glog.Errorf("Test execution failed: %v", errList.String())
+			continue
 		}
 		glog.Infof("Test %v ran successfully!", clusterLoaderConfig.TestConfigPath)
 	}


clusterloader2/pkg/errors/error_list.go
```diff
@@ -58,6 +58,22 @@ func (e *ErrorList) Concat(e2 *ErrorList) {
...
 
+// Filter returns list of errors for which a filter function returns true.
+func (e *ErrorList) Filter(f func(error) bool) *ErrorList {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	errList := NewErrorList()
+	if f == nil {
+		return errList
+	}
+	for i := range e.errors {
```
wojtek-t: lock here - locking earlier isn't needed

clusterloader2/cmd/clusterloader.go
```diff
@@ -127,7 +127,13 @@ func main() {
 	for _, clusterLoaderConfig.TestConfigPath = range testConfigPaths {
 		glog.Infof("Running %v", clusterLoaderConfig.TestConfigPath)
 		if errList := test.RunTest(f, &clusterLoaderConfig); !errList.IsEmpty() {
-			glog.Fatalf("Test execution failed: %v", errList.String())
+			if metricErrList := errList.Filter(errors.IsMetricViolationError); !metricErrList.IsEmpty() {
+				glog.Errorf("Test slos failed: %v", errList.String())
```
wojtek-t: I'm not entirely sure this is the right approach.
What are the other errors that could be returned from here?

[In general, there aren't many errors that should block the next test, so I would like to understand cases where we don't allow next test]

clusterloader2/pkg/errors/error_list.go
krzysied: Done.

clusterloader2/cmd/clusterloader.go
krzysied: I've changed `glog.Fatal` to `glog.Error` so it will not block next tests.

TBH I don't thing that this is final version of error handling. Previous implementation blocked tests if any error was risen. I'm positive that metric violation errors shouldn't block test execution and this main change in this PR.
I added TODO, so execution errors handling will be fixed in the future.

wojtek-t: Does this glog error provide any value? We should have logged that already earlier anyway.

clusterloader2/cmd/clusterloader.go
```diff
@@ -127,7 +127,14 @@ func main() {
...
+			}
+			// TODO: Reimplement execution errors handling.
+			if executionErrList := errList.Filter(func(e error) bool { return !errors.IsMetricViolationError(e) }); !executionErrList.IsEmpty() {
```
wojtek-t: Can we introduce IsNotMetricViolationError helper?

clusterloader2/cmd/clusterloader.go
krzysied: IMHO it is like a summary of what happened. Instead of jumping through whole log, you can see type of metric violations in one place.

clusterloader2/cmd/clusterloader.go

clusterloader2/cmd/clusterloader.go
wojtek-t: So given though we are logging all errors anyway now, why not simplify all this stuff (for now) to:
```
glog.Errorf("Test execution failed: %v", errList.String())
continue
```

krzysied: That's true. I will change it.

Anyway I would like to log metric violation errors and execution errors separately - those are IMHO two different thing. I will create separate PR in future regarding this matter.

PR #309, opened 2018-11-23
ClusterLoader - 5x100 density test update

krzysied:
Temporally changing test to make 5x  create latency pods iterations. Required to compare different latency pods creation approaches.

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -13,7 +13,7 @@
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
 {{$podsPerNamespace := MultiplyInt $PODS_PER_NODE $NODES_PER_NAMESPACE}}
 {{$totalPods := MultiplyInt $podsPerNamespace $namespaces}}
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/density/config.yaml
```diff
@@ -61,7 +61,8 @@ steps:
     Params:
       action: gather
 - name: Creating saturation pods
-# Create latency pods
+# TODO: Change this explicite iterations when pod startup problem is fixed!
```
wojtek-t: Ohh man - how ugly that is...

Can you at least open an issue to trigger reverting that?

wojtek-t: I meant "to not forget about reverting" :-)

wojtek-t:
BTW - I really doubt this is related. But you can try...

clusterloader2/testing/density/config.yaml
krzysied: Issue #310

krzysied:
@wojtek-t After running some local test I observed ~0.5s difference. I'm curious if there will be difference while using test environment.

wojtek-t:
Well - there are differences between runs (especially if you're running from your desktop). So I still doubt it's that. But let's see.

/lgtm

PR #322, opened 2018-11-30
ClusterLoader - Adding resource constraints to 100 nodes density test

krzysied:


clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,21 @@
...
+  cpuConstraint: 0.01
+  memoryConstraint: 26214400 #25 * (1024 * 1024)
+l7-lb-controller:
+  cpuConstraint: 0.17
+  memoryConstraint: 69206016 #66 * (1024 * 1024)
+coredns:
+  cpuConstraint: 0.13
+  memoryConstraint: 14680064 #14 * (1024 * 1024)
+kube-apiserver:
+  cpuConstraint: 2.4
```
wojtek-t: Where are those numbers coming from?
These are different than the current ones:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/density.go#L176

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,21 @@
+heapster:
```
wojtek-t: fluentd is missing

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,21 @@
...
+  nemoryConstraint: 387973120 #370 * (1024 * 1024)
+kube-proxy:
+  cpuConstraint: 0.01
+  memoryConstraint: 26214400 #25 * (1024 * 1024)
+l7-lb-controller:
+  cpuConstraint: 0.17
+  memoryConstraint: 69206016 #66 * (1024 * 1024)
+coredns:
+  cpuConstraint: 0.13
+  memoryConstraint: 14680064 #14 * (1024 * 1024)
```
wojtek-t: I have no idea where are those numbers coming from...

clusterloader2/testing/density/constraints_100.yaml
wojtek-t: I think that explains why some numbers are smaller (i.e. the values need to work for larger clusters).
But why some values are higher (e.g. apiserver cpu)?

BTW - if we want to adjust them later, please file an issue for that.

krzysied: > But why some values are higher (e.g. apiserver cpu)?

I investigated this topic some time ago, so I also took original tests into account - during that time clusterloader density test was not stable.

Since two days ago density test is stable. Maybe better option would be just wait till Monday and if everything goes well with the tests, it will better to readjust the value based only on clusterloader results?

wojtek-t: sgtm

krzysied:
/hold
This PR will be updated with resource constraints based on clusterloader density test.

clusterloader2/testing/density/constraints_100.yaml
krzysied: In original density test I found constraints only for "fluentd-elasticsearch" which doesn't appear in clusterloader nor original density test results. I've replaced it with "fluentd-gcp".

clusterloader2/testing/density/constraints_100.yaml
krzysied: I updated the values. All of them are set based on clusterloader density test only.
What I find weird is "kube-controller-manager" cpu usage being much higher than in original test.

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,24 @@
+coredns:
+  cpuConstraint: 0.01
+  memoryConstraint: 9437184 #9 * (1024 * 1024)
```
wojtek-t: Some constrantings seem to be pretty aggresive here - e.g. this 9MB.
I think with such constraints we will see flake sooner or later.

Setting memory to less than 20-30MB I think is too aggressive.

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,24 @@
...
+fluentd-gcp:
+  cpuConstraint: 1
+  memoryConstraint: 220200960 #210 * (1024 * 1024)
+heapster:
+  cpuConstraint: 0.1
+  nemoryConstraint: 340787200 #325 * (1024 * 1024)
+kube-apiserver:
+  cpuConstraint: 1.7
+  memoryConstraint: 975175680 #930 * (1024 * 1024)
```
wojtek-t: This is really small comparing to this one: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/density.go#L177
Are we sure it's fine?

wojtek-t: Ohh - I think I may understand it now - in ClusterLoader we are *always* running density before load. So that may explain some differences, I think.

If so, it's probably worth adding a comment somewhere (that the constraints assume that nothing was really running before this test).

clusterloader2/testing/density/config.yaml
```diff
@@ -3,6 +3,7 @@
 # - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
 
 #Constants
+{{$RESOURCE_CONSTRAINTS_FILE := DefaultParam .RESOURCE_CONSTRAINTS_FILE "constraints_100.yaml"}}
```
wojtek-t: I don't think it should be default.
Default should be empty, and we should overwrite it for 100-node test only.

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,24 @@
...
+kube-controller-manager:
+  cpuConstraint: 0.9
+  memoryConstraint: 241172480 #230 * (1024 * 1024)
+kube-proxy:
+  cpuConstraint: 0.01
+  memoryConstraint: 10485760 #10 * (1024 * 1024)
```
wojtek-t: Same here - 10MB seems too aggresive to me.

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,24 @@
...
+  cpuConstraint: 1
+  memoryConstraint: 220200960 #210 * (1024 * 1024)
+heapster:
+  cpuConstraint: 0.1
+  nemoryConstraint: 340787200 #325 * (1024 * 1024)
+kube-apiserver:
+  cpuConstraint: 1.7
+  memoryConstraint: 975175680 #930 * (1024 * 1024)
+kube-controller-manager:
+  cpuConstraint: 0.9
```
wojtek-t: Why this is higher than orginally:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/scalability/density.go#L178

I don't think we observed issues with that...

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,24 @@
...
+  memoryConstraint: 241172480 #230 * (1024 * 1024)
+kube-proxy:
+  cpuConstraint: 0.01
+  memoryConstraint: 10485760 #10 * (1024 * 1024)
+kube-scheduler:
+  cpuConstraint: 0.22
```
wojtek-t: Let's make 0.25.

clusterloader2/testing/density/constraints_100.yaml
```diff
@@ -0,0 +1,24 @@
...
+  memoryConstraint: 94371840 #90 * (1024 * 1024)
+l7-lb-controller:
+  cpuConstraint: 0.08
+  memoryConstraint: 65011712 #62 * (1024 * 1024)
```
wojtek-t: 70 maybe? Let's not make them too artificial (62 seem too artificial to me).

clusterloader2/testing/density/constraints_100.yaml
krzysied: Do we want to find constraints for every density test run or just for our ci job density test?

TBH I prefer the solution when we set default constraints to empty and set constraints in ci job definition, so this will be job specific.

clusterloader2/testing/density/constraints_100.yaml
krzysied: Changed to 80.

clusterloader2/testing/density/constraints_100.yaml
krzysied: Done.

clusterloader2/testing/density/100_nodes/constraints.yaml
```diff
@@ -0,0 +1,24 @@
...
+  cpuConstraint: 1
+  memoryConstraint: 241172480 #230 * (1024 * 1024)
+heapster:
+  cpuConstraint: 0.15
+  nemoryConstraint: 367001600 #350 * (1024 * 1024)
+kube-apiserver:
+  cpuConstraint: 1.7
+  memoryConstraint: 1048576000 #1000 * (1024 * 1024)
+kube-controller-manager:
+  cpuConstraint: 0.9
```
wojtek-t: I think this one got lost. Why this is higher than it used to be before?

clusterloader2/testing/density/config.yaml
```diff
@@ -3,6 +3,7 @@
 # - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
 
 #Constants
+{{$DENSITY_RESOURCE_CONSTRAINTS_FILE := DefaultParam .DENSITY_RESOURCE_CONSTRAINTS_FILE ""}}
```
wojtek-t: I'm not sure I understand this.
So the goal is to set two different files, the second file setting the override?

How this will look like on test-infra side?

clusterloader2/pkg/measurement/common/simple/resource_usage.go
```diff
@@ -80,7 +80,10 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Measurement
 		if constraintsPath != "" {
 			mapping := make(map[string]interface{})
 			mapping["Nodes"] = config.ClusterConfig.Nodes
-			config.TemplateProvider.TemplateInto(constraintsPath, mapping, &e.resourceConstraints)
+			err = config.TemplateProvider.TemplateInto(constraintsPath, mapping, &e.resourceConstraints)
```
wojtek-t: if err := ...; err != nil {

clusterloader2/testing/density/constraints_100.yaml
krzysied: It rarely happens that cpu usage is above 0.8, but it happens. I'm aware that this value is greater than in original test, but I don't know the reason why yet.

clusterloader2/testing/density/constraints_100.yaml
krzysied: Changed to 30.

clusterloader2/testing/density/config.yaml
krzysied: I removed default value. This value will be provided with overrides.

clusterloader2/testing/density/config.yaml
krzysied: All overrides are provided via yaml file. On test-infrs side it will require additional test argument
```
--testoverrides=./testing/density/100_nodes/override.yaml
```

clusterloader2/pkg/measurement/common/simple/resource_usage.go
krzysied: Done.

clusterloader2/testing/density/100_nodes/constraints.yaml
wojtek-t: Probably we should understand that... it seems strange that it is higher, given that we didn't observe any issues with that in current tests where this is set to 0.8.

krzysied: I agree. I added issue #326.
Nevertheless I think this should not be a blocker for this PR.

clusterloader2/testing/density/config.yaml
wojtek-t: got it - thanks

clusterloader2/testing/density/100_nodes/constraints.yaml
wojtek-t: debatable, but ok.

PR #323, opened 2018-11-30
Updating dns benchmark run script

krzysied:
- dns/run takes care of running benchmark. It runs tests, creates required directories and creates metrics.
- run-e2e.sh only calls dns/run


prameshj:
/lgtm

I do not know of any other script using dns/run. Adding @bowei to comment.

PR #325, opened 2018-12-03
Allow to set masternode / masterip through CLI

mrueg:
When trying to run the benchmark locally this was helpful


krzysied:
Thanks @mrueg !

PR #331, opened 2018-12-10
Dns benchmark - Fixing dns benchmark path

krzysied:


wojtek-t:
@krzysied - please fix, tests are failing

krzysied:
@wojtek-t The problem is it verifies that labels used by perfdash. I think we don't want to disable this check, so I created #332 to add dns config to perfdash.

PR #333, opened 2018-12-10
Dns benchmark - Adding numpy installation to run script

krzysied:
Adding numpy installation to run script, due to numpy being required by `py/run_perf.py`. This command installs numpy or does nothing is numpy s already available.


krzysied:
@wojtek-t For some reason lgmt and approve were not triggered.

PR #340, opened 2018-12-17
ClusterLoader- Fixing measurements for kubemark

krzysied:


clusterloader2/pkg/measurement/common/simple/resource_usage.go
```diff
@@ -103,7 +103,7 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Measurement
 		}
 
 		glog.Infof("%s: starting resource usage collecting...", e)
-		e.gatherer, err = gatherers.NewResourceUsageGatherer(config.ClientSet, provider, host, gatherers.ResourceGathererOptions{
+		e.gatherer, err = gatherers.NewResourceUsageGatherer(config.ClientSet, host, provider, gatherers.ResourceGathererOptions{
```
wojtek-t: This is a bit misleading (if you look into below fine, where the order is "provider, host - line 149 scheduler_latency.go).
Can we agree on one order of those and use everywhere?

krzysied: Changed to "host, provider" in every file.

PR #341, opened 2018-12-18
ClusterLoader - Fixing RC deletion handling

krzysied:
- Waiting for pods.
Measurement will wait until desired pods are running and there are no others pods that match given labels and namespace (regardless their statuses).
- Waiting for controlled pod.
Handling of object deletion changed to waiting for pods with desired replicas equaling 0.

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -33,6 +33,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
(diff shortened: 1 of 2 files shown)


wojtek-t:
I would also like to look into it tomorrow...

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -221,16 +221,26 @@ func (w *waitForControlledPodsRunningMeasurement) gather(c clientset.Interface,
...
 	defer w.lock.Unlock()
-	numberRunning := 0
+	var numberRunning, numberTerminated, numberUnknown int
 	for _, checker := range w.checkerMap {
-		if status := checker.getStatus(); status {
+		switch checker.getStatus() {
+		case running:
 			numberRunning++
+		case terminated:
+			numberTerminated++
```
wojtek-t: This is used only in log.Infof - is that really expected?

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -221,16 +221,26 @@ func (w *waitForControlledPodsRunningMeasurement) gather(c clientset.Interface,
...
 		}
 	}
+	glog.Infof("%s: running %d, terminated %d, unknown: %d", w, numberRunning, numberTerminated, numberUnknown)
 	if desiredCount != numberRunning {
 		glog.Errorf("%s: incorrect objects number: %d/%d %ss are running with all pods", w, numberRunning, desiredCount, w.kind)
 		return fmt.Errorf("incorrect objects number")
 	}
+	if numberUnknown > 0 {
+		glog.Errorf("%s: unknown status for %d %ss", w, numberUnknown, w.kind)
+		return fmt.Errorf("unknown objects statuses")
```
wojtek-t: I'm not sure it's really unknown - usuallly, it would mean that simply we didn't yet finish reconciling. Is that right?

clusterloader2/pkg/measurement/common/simple/wait_for_pods.go
```diff
@@ -139,7 +139,7 @@ func waitForPods(clientSet clientset.Interface, namespace, labelSelector, fieldS
 			if log {
 				glog.Infof("%s: %s: %s", callerName, selectorsString, podsStatus.String())
 			}
-			if podsStatus.Running == desiredPodCount {
+			if podsStatus.Running == desiredPodCount && len(pods) == desiredPodCount {
```
wojtek-t: nit: can we change to:

len(pods) == podsStatus.Running && podsStatus.Running == desiredPodCount

This is more intuitive (all pods are running and there is correct number of running pods).

[I know it's technically the same, but it's easier to reason about.]

krzysied: Done.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
krzysied: I treat `numberTerminated` as sanity check. It's only used for logging objects status.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
krzysied: It's true. However the reason of operation not being finished is unknown. That's why I though "status unknown" is a good name.
If this error is returned, there definitely will be some logs explaining the situation.

krzysied: Do we want to have more information in the error message or in error log? What kind of information should be added?
For me this error is only a signal that something went wrong and I expect user to look into log file to see more details - that's why I tried to keep error message simple.

wojtek-t: That's a bad way - error should tell you what happened. Users may look into logs to get information "why", but not to look "what really happened".

Things that are part of log (e.g. what object that was) should be part of error.

In fact (as I mentioned sometime in the past), I'm not convinced our logs are what we really need. We should probably do the audit of them soon.

krzysied: Ok.
I added the list of errors for the objects with unknown status. This errors contain object names with error descriptions.

wojtek-t:
OK - let's see and tune if needed later.

/lgtm

PR #346, opened 2018-12-19
Add a way to optionally enable chaos monkey in tests.

mborsz:
Chaos monkey is disabled by default, but can be enabled by adding --testoverrides=testing/chaosmonkey/override.yaml.

/assign @wojtek-t

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -69,7 +69,7 @@ func initFlags() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
 	pflag.StringArrayVar(&testConfigPaths, "testconfig", []string{}, "Paths to the test config files")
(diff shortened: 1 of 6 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -69,7 +69,7 @@ func initFlags() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
 	pflag.StringArrayVar(&testConfigPaths, "testconfig", []string{}, "Paths to the test config files")
-	pflag.StringVar(&clusterLoaderConfig.TestOverridesPath, "testoverrides", "", "Path to the config overrides file")
+	pflag.StringArrayVar(&clusterLoaderConfig.TestOverridesPath, "testoverrides", []string{}, "Paths to the config overrides file. The latter overrides take precedence over changes in former files.")
```
wojtek-t: @krzysied - i think it's ok, but can you just confirm that it will work fine?

krzysied: lgtm

PR #350, opened 2018-12-19
ClusterLoader - Adding env flags

krzysied:
Adding env flags that can read env variable if flag is not provided.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"path"
@@ -27,9 +26,9 @@ import (
 	ginkgoconfig "github.com/onsi/ginkgo/config"
 	ginkgoreporters "github.com/onsi/ginkgo/reporters"
 	ginkgotypes "github.com/onsi/ginkgo/types"
-	"github.com/spf13/pflag"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/flags"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/test"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
@@ -49,11 +48,11 @@ var (
 )
 
 func initClusterFlags() {
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
-	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "", "Name of the masternode")
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "masterip", "", "Hostname/IP of the masternode")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "masterip", "MASTER_IP", "", "Hostname/IP of the masternode")
 }
 
 func validateClusterFlags() *errors.ErrorList {
@@ -65,11 +64,9 @@ func validateClusterFlags() *errors.ErrorList {
 }
 
 func initFlags() {
-	pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
-	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	pflag.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
-	pflag.StringArrayVar(&testConfigPaths, "testconfig", []string{}, "Paths to the test config files")
-	pflag.StringArrayVar(&clusterLoaderConfig.TestOverridesPath, "testoverrides", []string{}, "Paths to the config overrides file. The latter overrides take precedence over changes in former files.")
+	flags.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
+	flags.StringArrayVar(&testConfigPaths, "testconfig", []string{}, "Paths to the test config files")
+	flags.StringArrayVar(&clusterLoaderConfig.TestOverridesPath, "testoverrides", []string{}, "Paths to the config overrides file. The latter overrides take precedence over changes in former files.")
 	initClusterFlags()
 }
 
@@ -150,7 +147,7 @@ func printTestResult(name, status, errors string) {
 func main() {
 	defer glog.Flush()
 	initFlags()
-	if err := pflag.CommandLine.Parse(os.Args[1:]); err != nil {
+	if err := flags.Parse(); err != nil {
 		glog.Fatalf("Flag parse failed: %v", err)
 	}
 	if errList := validateFlags(); !errList.IsEmpty() {
--- a/clusterloader2/pkg/flags/flags.go
+++ b/clusterloader2/pkg/flags/flags.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package flags
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"strconv"
+
+	"github.com/spf13/pflag"
+)
+
+func init() {
+	pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+}
+
+var flags []flagFunc
+
+// StringVar creates string flag with given parameters.
+func StringVar(s *string, flagName, defaultValue, description string) {
+	pflag.StringVar(s, flagName, defaultValue, description)
+}
+
+// IntVar creates int flag with given parameters.
+func IntVar(i *int, flagName string, defaultValue int, description string) {
+	pflag.IntVar(i, flagName, defaultValue, description)
+}
+
+// StringEnvVar creates string flag with given parameters.
+// If flag is not provided, it will try to get env variable.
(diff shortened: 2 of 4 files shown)


clusterloader2/run-e2e.sh
```diff
@@ -19,6 +19,7 @@ set -o nounset
 set -o pipefail
 
 CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
+export KUBECONFIG=${KUBECONFIG:-"${HOME}/.kube/config"}
```
wojtek-t: "${KUBECONFIG:-${HOME}/.kube/config}"

clusterloader2/cmd/clusterloader.go
```diff
@@ -49,11 +48,11 @@ var (
 )
 
 func initClusterFlags() {
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file")
-	pflag.IntVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", 0, "number of nodes")
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "", "Cluster provider")
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "", "Name of the masternode")
-	pflag.StringVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "masterip", "", "Hostname/IP of the masternode")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NODES", 0, "number of nodes")
```
wojtek-t: Do we really need this env?

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,132 @@
...
+	"os"
+	"strconv"
+
+	"github.com/golang/glog"
+	"github.com/spf13/pflag"
+)
+
+const (
+	noInt    = -12345
+	noString = "xxxNoStringxxx"
```
wojtek-t: I'm not a fan of those strange defaults. Defaults are part of documentation (in fact they are kind of part of api also). So I don't want to have stuff like that.
I would really prefer using standard defaults (empty values) - which always has to be overwritten anyway.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,132 @@
...
+type stringFlagEnv struct {
+	v            *string
+	envVariable  string
+	defaultValue string
+}
+
+// ParseEnv assigns env variable value if needed.
+func (s *stringFlagEnv) ParseEnv() {
+	if *s.v == noString {
+		if s.envVariable != "" {
```
wojtek-t: reduce nesting:

if ... && ... {
}

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,132 @@
...
+func (i *intFlagEnv) ParseEnv() {
+	if *i.v == noInt {
+		if i.envVariable != "" {
+			if val, ok := os.LookupEnv(i.envVariable); ok {
+				iVal, err := strconv.Atoi(val)
+				if err != nil {
+					*i.v = iVal
+					return
+				}
+				glog.Errorf("Parsing env variable %s failed. Using default %d", i.envVariable, i.defaultValue)
```
wojtek-t: I would say that we should fail the test if we are not able to parse the flag.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,132 @@
...
+type intFlagEnv struct {
+	v            *int
+	envVariable  string
+	defaultValue int
+}
+
+// ParseEnv assigns env variable value if needed.
+func (i *intFlagEnv) ParseEnv() {
+	if *i.v == noInt {
+		if i.envVariable != "" {
```
wojtek-t: Same here

clusterloader2/cmd/clusterloader.go
wojtek-t: If so, isn't that NUM_NODES ?

clusterloader2/run-e2e.sh
krzysied: Done.

clusterloader2/cmd/clusterloader.go
krzysied: I assumed that there possible should be env variable for each cluster flag.
Changed to "NUM_NODES".

clusterloader2/pkg/flags/flags.go
krzysied: I changed consts to 0 and "" respectively.
I though about having some magic variables that user will definitely not use. It was useful to verifying whether flag was provided or not.

clusterloader2/pkg/flags/flags.go
krzysied: Not sure how to && those two if's. Inverted if used instead.

clusterloader2/pkg/flags/flags.go
krzysied: Done.

clusterloader2/pkg/flags/flags.go
krzysied: Done.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,138 @@
...
+	if *s.v != noString {
+		return nil
+	}
+	if s.envVariable != "" {
+		if val, ok := os.LookupEnv(s.envVariable); ok {
+			*s.v = val
+			return nil
+		}
+	}
+	*s.v = s.defaultValue
```
wojtek-t: Actually, I just realized that it doesn't work.
What if we say default is 5, but I set "--my-flag=0". It will then be overwritten to 5.

I think that this can be solved differently.
- you first parse EnvVars (with defaulting to default value)
- then you parse flags (to overwrite stuff)

That would solve your problem, right?

krzysied: The problem is we don't know whether the flag was provided or not - parse flag always returns value. If someone set "--my-flag=0" it will return the same value as default. In this case we don't know if we should override the env var.

krzysied: I saw that flag module provides interface `flag.Value` that handles setting. I need to verify is pflag also supports this interface.

wojtek-t: Did you see the flag.Var() as described here:
https://godoc.org/github.com/spf13/pflag

It seems that it's pretty much what you need.

krzysied: Ok. I changed flag declaration to flag.Var().
I also moved inner structures definitions to types.go.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,83 @@
...
+func IntVar(i *int, flagName string, defaultValue int, description string) {
+	pflag.IntVar(i, flagName, defaultValue, description)
+}
+
+// StringEnvVar creates string flag with given parameters.
+// If flag is not provided, it will try to get env variable.
+func StringEnvVar(s *string, flagName, envVariable, defaultValue, description string) {
+	stringFlag := &stringFlagEnv{
+		v:            s,
+		envVariable:  envVariable,
```
wojtek-t: Given that you are introducing Var() anyway, this is too complicated..
It should work as following:
- read env-var (if it doesn't exist, use default)
- create a struct implementing Var interface that keeps just a single value equal to the one from above
- calling Set() is modifying that value

krzysied: I'm not sure why env value should be used first? IMHO, explicit flag sounds more like final setting provided by the user.

krzysied: Implementation changed to match what @wojtek-t described above. Priority for parameters: flag > env var > default.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/spf13/pflag"
+)
+
+func init() {
+	pflag.CommandLine = pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+}
+
+var flags []flagEnv
```
wojtek-t: I'm generally not a fan of having global variable, and I actually think it's not really needed here.
See my comments below.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,83 @@
...
+	pflag.IntVar(i, flagName, defaultValue, description)
+}
+
+// StringEnvVar creates string flag with given parameters.
+// If flag is not provided, it will try to get env variable.
+func StringEnvVar(s *string, flagName, envVariable, defaultValue, description string) {
+	*s = defaultValue
+	stringFlag := &stringFlagEnv{
+		v:           s,
+		envVariable: envVariable,
```
wojtek-t: Why do we need to pass envVariable to the bottom?
I would say it's cleaner to say:

*s = ParseStringEnv(envVariable, defaultValue)
stringFlag := &stringFlagEnv{v: s}
pflag.Var(stringFlag, flagName, description)

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,83 @@
...
+		v:           i,
+		envVariable: envVariable,
+	}
+	pflag.Var(intFlag, flagName, description)
+	flags = append(flags, intFlag)
+}
+
+// Parse parses provided flags and env variables.
+func Parse() error {
+	for i := range flags {
```
wojtek-t: Then this one isn't really needed.

clusterloader2/pkg/flags/types.go
```diff
@@ -0,0 +1,106 @@
...
+	ParseEnv() error
+}
+
+type stringFlagEnv struct {
+	v           *string
+	envVariable string
+}
+
+// ParseEnv assigns env variable value if needed.
+func (s *stringFlagEnv) ParseEnv() error {
```
wojtek-t: This shouldn't be public method.

krzysied:
@wojtek-t According to our offline discussion:
- types.go contains only generic flags that can evaluate additional function.
- flags.go have all environment parsing logic.

PTAL

clusterloader2/pkg/flags/types.go
```diff
@@ -0,0 +1,88 @@
...
+var _ pflag.Value = (*intFlagFunc)(nil)
+var _ flagFunc = (*intFlagFunc)(nil)
+
+type flagFunc interface {
+	parseFunc() error
+}
+
+type stringFlagFunc struct {
+	v *string
+	f func() error
```
wojtek-t: Can we have a better name for it?
Maybe initialize? Or sth like that?

clusterloader2/pkg/flags/types.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// parseFunc runs additional parsing function.
+func (s *stringFlagFunc) parseFunc() error {
```
wojtek-t: Shouldn't contain "Func" - so if any "parse()".
But maybe initialize() ?

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,109 @@
...
+	pflag.IntVar(i, flagName, defaultValue, description)
+}
+
+// StringEnvVar creates string flag with given parameters.
+// If flag is not provided, it will try to get env variable.
+func StringEnvVar(s *string, flagName, envVariable, defaultValue, description string) {
+	*s = defaultValue
+	stringFlag := &stringFlagFunc{
+		v: s,
+		f: func() error { return parseEnvString(s, envVariable) },
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,109 @@
...
+	pflag.StringArrayVar(s, flagName, defaultValue, description)
+}
+
+// IntEnvVar creates int flag with given parameters.
+// If flag is not provided, it will try to get env variable.
+func IntEnvVar(i *int, flagName, envVariable string, defaultValue int, description string) {
+	*i = defaultValue
+	intFlag := &intFlagFunc{
+		v: i,
+		f: func() error { return parseEnvInt(i, envVariable) },
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/flags/flags.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}
+	return nil
+}
+
+func parseEnvInt(i *int, envVariable string) error {
+	if envVariable != "" {
+		if val, ok := os.LookupEnv(envVariable); ok {
+			iVal, err := strconv.Atoi(val)
+			if err == nil {
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/flags/types.go
krzysied: Renamed:
- v -> valPtr
- f -> initializeFunc

clusterloader2/pkg/flags/types.go
krzysied: Changed to "initiazlize".

PR #353, opened 2018-12-20
Change operationTimeout to 15m in scalability tests

mborsz:
As operationTimeout for WaitForRunningSaturationRCs was dynamically computed, I changed hard deadline to 15m and moved logic to compare perc99 in SaturationPodStartupLatency.

/assign @wojtek-t 
/assign @krzysied

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -36,10 +36,10 @@ import (
 )
 
 const (
-	podStartupLatencyThreshold       = 5 * time.Second
-	podStartupLatencyMeasurementName = "PodStartupLatency"
-	informerSyncTimeout              = time.Minute
-	successfulStartupRatioTreshold   = 0.99
+	defaultPodStartupLatencyThreshold = 5 * time.Second
+	podStartupLatencyMeasurementName  = "PodStartupLatency"
+	informerSyncTimeout               = time.Minute
+	successfulStartupRatioThreshold   = 0.99
 )
 
 func init() {
@@ -69,6 +69,7 @@ type podStartupLatencyMeasurement struct {
 	runTimes      map[string]metav1.Time
 	watchTimes    map[string]metav1.Time
 	nodeNames     map[string]string
+	threshold     time.Duration
 }
 
 // Execute supports two actions:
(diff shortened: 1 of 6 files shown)


clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -320,7 +327,7 @@ type podStartupLatency struct {
 
 // SummaryName returns name of the summary.
 func (p *podStartupLatency) SummaryName() string {
-	return podStartupLatencyMeasurementName
+	return p.summaryName
```
wojtek-t: Maybe sth like "podStartupLatencyMeasurementName + p.summaryName"

clusterloader2/testing/density/config.yaml
```diff
@@ -5,7 +5,7 @@
 #Constants
 {{$DENSITY_RESOURCE_CONSTRAINTS_FILE := DefaultParam .DENSITY_RESOURCE_CONSTRAINTS_FILE ""}}
 {{$NODE_MODE := DefaultParam .NODE_MODE "allnodes"}}
-{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 1}}
```
wojtek-t: Revert? I guess it's coming from your individual testing...

clusterloader2/testing/density/config.yaml
```diff
@@ -53,14 +53,21 @@ steps:
...
+      threshold: {{$saturationRCTimeout}}s
 - measurements:
   - Identifier: WaitForRunningSaturationRCs
     Method: WaitForControlledPodsRunning
     Params:
       action: start
       kind: ReplicationController
       labelSelector: group = saturation
-      operationTimeout: {{$saturationRCTimeout}}s
+      operationTimeout: 15m
```
wojtek-t: This will be too low in large clusters (where we create all RCs at the very beginning, but creating all pods take 1h+)

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
krzysied: If we change this we should remember to update perfdash (iirc it should be able to handle multiple parser for one metric). There was an assumption to not change summaries names, though.

wojtek-t: If we want to have two measurements of the same type (as here), they can't have the same name, right?

krzysied: You are right. I missed the metric adding part when I look at the PR.

mborsz:
PTAL

wojtek-t:
/lgtm

/hold
Feel free to unhold, if we don't want to fix perf-dash first.

mborsz:
#354 is merged.

/unhold

mborsz:
#354 is merged.

/hold cancel

PR #354, opened 2018-12-20
Change perfdash to look for PodStartupLatency metrics in new place.

mborsz:
The path changes in #353 from PodStartupLatency to PodStartupLatency_PodStartupLatency and a new metric PodStartupLatency_SaturationPodStartupLatency is added.

/assign @krzysied

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 1.15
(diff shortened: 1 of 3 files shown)


perfdash/Makefile
```diff
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 1.15
```
krzysied: Can you also update deployment.yaml file?

PR #366, opened 2019-01-04
ClusterLoader - Executing load test service operation in a sequence.

krzysied:
Adding and deleting services 1 by 1.

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -27,9 +27,9 @@
 name: load
 automanagedNamespaces: {{$namespaces}}
 tuningSets:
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -27,9 +27,9 @@
 name: load
 automanagedNamespaces: {{$namespaces}}
 tuningSets:
-- name: Uniform100qps
-  qpsLoad:
-    qps: 100
+- name: Sequence
+  parallelismLimitedLoad:
+    parallelismLimit: 1
```
wojtek-t: We can probably increase it to more than one, but I'm fine with doing that later.

PR #367, opened 2019-01-08
ClusterLoader - Fixing multiclient

krzysied:
Forcing clients to no share underling transport.


krzysied:
@mborsz https://github.com/kubernetes/kubernetes/blob/cc67ccfd7f4f0bc96d7f1c8e5fe8577821757d03/test/e2e/scalability/load.go#L354 is the config creation form original test. I added the missing transport part to clusterloader.

Issue #371, opened 2019-01-11
CLusterLoader: DNS is using more memory than it used to


krzysied:
After changing tests to clusterloader's, coreDNS (and also kubeDNS) is using more memory than it used to.
![coredns_memory](https://user-images.githubusercontent.com/25119199/51028390-8e1aaf00-1593-11e9-987b-e9d73d27aaea.png)

This issue causes cordns on  5k nodes to OOM.

krzysied:
The two biggest difference in api calls:
![endpotins_put](https://user-images.githubusercontent.com/25119199/51028757-ae973900-1594-11e9-8b44-2bfff8e2d3d9.png)
![pods_post](https://user-images.githubusercontent.com/25119199/51028762-b1922980-1594-11e9-8849-dfb50b0f419b.png)

krzysied:
During few last days I verified, that:
- using patch instead scale doesn't have impact on metrics mentioned above.
- Reducing number of created pods should reduce the number of endpoints calls, however I don't expect this change to be big.

krzysied:
I think I found the cause: https://github.com/kubernetes/kubernetes/blob/33a9c6e892f69e20be9527ba00bd33dfa5de221b/test/e2e/scalability/load.go#L602
In original load test, the service has additional label that is used as a selector. This label points to one specific RC. In theory every two RC should use one SVC, however it looks like every second RC uses SVC.

In clusterloader implementation, every RC is connected to some SVC, therefore it created twice as many connection as in original test.

krzysied:
I've run some tests svc implemented as it is on original tests (svc connected to every second rc). The memory used by coredns dropped to ~11468800, which is the same amount as original tests used. Without this change coredns consumes ~14831616.

krzysied:
/cc @wojtek-t @mborsz 
I created #374 to see if it reducing number of endpoints will fix the issue.

krzysied:
Here's impact of #374:
![coredns_memory_change](https://user-images.githubusercontent.com/25119199/51105535-82282a80-17e9-11e9-9a9d-16c053bfb889.png)
![2k_coredns_memory_change](https://user-images.githubusercontent.com/25119199/51105634-d29f8800-17e9-11e9-965f-cdc013746646.png)
It looks like increased number of endpoints was the main issue.

wojtek-t:
Fixed now.

PR #372, opened 2019-01-11
ClusterLoader - Changing to original load RC randimized range

krzysied:
Changing replicas random range to match original tests: https://github.com/kubernetes/kubernetes/blob/49891cc270019245a3d4796e84b33bf36d0bae08/test/e2e/scalability/load.go#L674

ref #371 

Should reduce pod post calls.

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -149,7 +149,7 @@ steps:
       objectTemplatePath: rc.yaml
       templateFillMap:
         ReplicasMin: {{MultiplyInt $BIG_GROUP_SIZE 0.5}}
-        ReplicasMax: {{MultiplyInt $BIG_GROUP_SIZE 1.5}}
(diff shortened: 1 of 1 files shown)


wojtek-t:
@krzysied - can we revert this one? This doesn't introduce viusible difference and is complicating the config :-)

krzysied:
I agree that it should be reverted. However, first I want to have at least one correct-ish run of https://testgrid.k8s.io/sig-scalability-perf-tests#ci-kubernetes-e2e-gce-scale-performance-cl.
Then I can revert this change. Thanks to this:
- 5k clusterlaoder job will run with similar setup to original 5k test
- Allows to see impact of this change in big scale.

The first run of kubernetes-e2e-gce-scale-performance-cl that contains all changes related to #371 is happening right now. If it goes well, I will create commit revert tomorrow morning.
@wojtek-t Wdyt?

wojtek-t:
> The first run of kubernetes-e2e-gce-scale-performance-cl that contains all changes related to #371 is happening right now. If it goes well, I will create commit revert tomorrow morning.
@wojtek-t Wdyt?

makes sense

PR #376, opened 2019-01-15
Clusterloader - Bumping up scheduler memory constraint

krzysied:
Scheduler avg memory usage (gce 100) has increased ~5MB. Bumping up constraint to prevent flakes.

Memory has increased between run 21637 and 21640, possible due to https://github.com/kubernetes/kubernetes/pull/71731 PR.

### DIFF
--- a/clusterloader2/testing/density/100_nodes/constraints.yaml
+++ b/clusterloader2/testing/density/100_nodes/constraints.yaml
@@ -18,7 +18,7 @@ kube-proxy:
   memoryConstraint: 31457280 #30 * (1024 * 1024)
 kube-scheduler:
   cpuConstraint: 0.35
-  memoryConstraint: 115343360 #110 * (1024 * 1024)
+  memoryConstraint: 125829120 #120 * (1024 * 1024)
 l7-lb-controller:
   cpuConstraint: 0.15
   memoryConstraint: 83886080 #80 * (1024 * 1024)


clusterloader2/testing/density/100_nodes/constraints.yaml
```diff
@@ -18,7 +18,7 @@ kube-proxy:
   memoryConstraint: 31457280 #30 * (1024 * 1024)
 kube-scheduler:
   cpuConstraint: 0.35
-  memoryConstraint: 115343360 #110 * (1024 * 1024)
+  memoryConstraint: 120586240 #115 * (1024 * 1024)
```
wojtek-t: Can we bup it a bit more aggresively - bumping by 5MB doesn't make much sense.
I told that couple weeks ago already and i tseems we're still suffering from that.

krzysied: Ok. I changed it to 120MB. It is a ~10% change.

> I told that couple weeks ago already and i tseems we're still suffering from that.

I will not agree with word "suffering". The average usage of memory increased, the test responded with flakes. It drew our attention, so we are aware of the changes made to the kubernetes. From my point of view, it's a good thing.

wojtek-t: Did it result in any action item? If not, then it's "suffering".

krzysied:
@wojtek-t I think all comments are resolved. PTAL

wojtek-t:
I asked this question above:
"Did it result in any action item? If not, then it's "suffering"."

krzysied:
> I asked this question above:
"Did it result in any action item? If not, then it's "suffering"."

I didn't realized that I should respond to it. I thought it was rhetorical question.
Anyway, it hasn't resulted in action item and after thinking about it I agree with what you written.

wojtek-t:
> Anyway, it hasn't resulted in action item and after thinking about it I agree with what you written.

Thanks :-)
That's why I said that we shouldn't be so aggresive in those limits :-)

/lgtm
/approve

PR #382, opened 2019-01-17
ClusterLoader - Logging profile error for gke

krzysied:
Profiles will only log master sshing error instead of propagating it to caller.
Gathering profiles from apiserver is not a standard thing in gke.

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/profile.go
+++ b/clusterloader2/pkg/measurement/common/simple/profile.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/golang/glog"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/simple/profile.go
```diff
@@ -102,6 +103,11 @@ func gatherProfile(componentName, profileKind, host, provider string) ([]measure
 	getCommand := fmt.Sprintf("curl -s localhost:%v/debug/pprof/%s", profilePort, profileKind)
 	sshResult, err := measurementutil.SSH(getCommand, host+":22", provider)
 	if err != nil {
+		if provider == "gke" {
+			// Only logging error for gke. SHHing to gke master is not a standard case.
```
wojtek-t: s/is not a standard case/is not supported/

krzysied: Done.

PR #388, opened 2019-01-25
ClusterLoader - Fixing wait for controlling objects timeout

krzysied:
Fixing bug introduced in #386 . It caused setting timeout status even if there was no timeout.

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -421,6 +421,8 @@ func (w *waitForControlledPodsRunningMeasurement) waitForRuntimeObject(clientSet
 	o.lock.Lock()
 	defer o.lock.Unlock()
 	w.handlingGroup.Start(func() {
+		// This function sets the status (and error message) for the object checker.
+		// The handling of bad statuses and errors is done by gather() function of the measurement.
 		err = waitForPods(clientSet, runtimeObjectNamespace, runtimeObjectSelector.String(), "", int(runtimeObjectReplicas), o.stopCh, true, w.String())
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -430,8 +430,12 @@ func (w *waitForControlledPodsRunningMeasurement) waitForRuntimeObject(clientSet
 				glog.Errorf("%s: error for %v: %v", w, key, err)
 				o.err = fmt.Errorf("%s: %v", key, err)
 			}
+			if o.status == timeout {
+				glog.Errorf("%s: %s timed out", w, key)
```
wojtek-t: glog error won't fail the test and we don't return error - I don't think it's expected.

krzysied: That's true. However the test failing is done by https://github.com/kubernetes/perf-tests/blob/61c121b65366dcb94efcfae54813190b3a78f1dc/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go#L253

This file starts to look to complicated. I will think about how to rewrite it (in different PR).

wojtek-t: I don't think it's time to rewrite it now. I actually believe it shouldn't be a measurement.
So let's don't rewrite it now.

wojtek-t: Please add a comment why we don't need to return error here.

krzysied: Done.

PR #392, opened 2019-01-28
ClusterLoader - Increasing timeout for waiting for deletion operations

krzysied:
Increasing timeout for operation related to the pod deletion.

This change is required due to the garbage collector requiring twice as much time to remove the pods as it was needed for pods creation.

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -303,29 +303,52 @@ func (w *waitForControlledPodsRunningMeasurement) handleObject(clients *framewor
 
 	if err := w.deleteObjectLocked(clients.GetClient(), oldRuntimeObj); err != nil {
 		glog.Errorf("%s: delete checker error: %v", w, err)
-	} else if newRuntimeObj == nil {
-		if err := w.handleObjectLocked(clients.GetClient(), oldRuntimeObj, true); err != nil {
-			glog.Errorf("%s: create checker error: %v", w, err)
-		}
 	}
-	if err := w.handleObjectLocked(clients.GetClient(), newRuntimeObj, false); err != nil {
+	if err := w.handleObjectLocked(clients.GetClient(), oldRuntimeObj, newRuntimeObj); err != nil {
 		glog.Errorf("%s: create checker error: %v", w, err)
 	}
 }
 
-func (w *waitForControlledPodsRunningMeasurement) handleObjectLocked(c clientset.Interface, obj runtime.Object, isDeleted bool) error {
-	if obj == nil {
-		return nil
+func checkScaledown(oldObj, newObj runtime.Object) (bool, error) {
+	oldReplicas, err := runtimeobjects.GetReplicasFromRuntimeObject(oldObj)
+	if err != nil {
+		return false, err
 	}
-	key, err := runtimeobjects.CreateMetaNamespaceKey(obj)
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -304,16 +304,41 @@ func (w *waitForControlledPodsRunningMeasurement) handleObject(clients *framewor
 	if err := w.deleteObjectLocked(clients.GetClient(), oldRuntimeObj); err != nil {
 		glog.Errorf("%s: delete checker error: %v", w, err)
 	} else if newRuntimeObj == nil {
-		if err := w.handleObjectLocked(clients.GetClient(), oldRuntimeObj, true); err != nil {
+		if err := w.handleObjectLocked(clients.GetClient(), oldRuntimeObj, true, true); err != nil {
```
wojtek-t: I'm not a fan of passing a list of bools, as this is hard to follow.

I think it would be clear if handleObjectLocked would take both old and new version and encapsulate that logic in the handleObjectLock function?

krzysied: Done.

krzysied: As you suggested, I moved most of the logic to the handleObjectLocked function.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -303,29 +303,63 @@ func (w *waitForControlledPodsRunningMeasurement) handleObject(clients *framewor
...
+	newReplicas, err := runtimeobjects.GetReplicasFromRuntimeObject(newObj)
+	if err != nil {
+		return false, err
+	}
+
+	return newReplicas < oldReplicas, nil
+}
+
+func (w *waitForControlledPodsRunningMeasurement) handleObjectLocked(c clientset.Interface, oldObj, newObj runtime.Object) error {
+	if oldObj == nil && newObj == nil {
```
wojtek-t: This should never be the case.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -303,29 +303,63 @@ func (w *waitForControlledPodsRunningMeasurement) handleObject(clients *framewor
...
 
-func (w *waitForControlledPodsRunningMeasurement) handleObjectLocked(c clientset.Interface, obj runtime.Object, isDeleted bool) error {
-	if obj == nil {
+func checkScaledown(oldObj, newObj runtime.Object) (bool, error) {
+	if newObj == nil {
+		return true, nil
+	}
+	if oldObj == nil {
+		return false, nil
+	}
```
wojtek-t: Instead of having those if`s, can you please:
- add "if obj == nil return 0" statement to GetReplicasFromRuntimeObject function (similarly as it is handled in GetSpecFrom..
- remove those if`s here

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
krzysied: Done.

PR #393, opened 2019-01-29
ClusterLoader - log which ReplicationControllers have timed-out.

mm4tt:
This will help us debug issues like https://github.com/kubernetes/kubernetes/issues/73461

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -23,6 +23,7 @@ package simple
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -463,13 +468,16 @@ type objectChecker struct {
 	stopCh    chan struct{}
 	status    objectStatus
 	err       error
+	// namespacedName of the object being checked.
+	namespacedName string
```
wojtek-t: I would call it a key.
While this is namespacedName in all current usecase, this is an implementation detail (e.g. in the future we may want to identify objects by UID, or we will have CRDs that aren't namespaced, or sth).

mm4tt: Good point, done.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -463,13 +468,17 @@ type objectChecker struct {
...
-func newObjectCheker() *objectChecker {
+func newObjectChecker(key string) *objectChecker {
 	return &objectChecker{
-		stopCh:    make(chan struct{}),
-		isRunning: true,
-		status:    unknown,
+		stopCh:         make(chan struct{}),
+		isRunning:      true,
+		status:         unknown,
+		key: key,
```
wojtek-t: this will fail gofmt

mm4tt: Doh, always forgetting about gofmt. 
Done.

PR #395, opened 2019-01-31
ClusterLoader - Updating README

krzysied:
Adding info about:
- overrides
- metrics: APIResponsiveness, PodStartupLatency, ResourceUsageSummary, SchedulingThroughput

### DIFF
--- a/clusterloader2/README.md
+++ b/clusterloader2/README.md
@@ -6,6 +6,10 @@ To run ClusterLoader type:
 ```
 go run cmd/clusterloader.go --kubeconfig=kubeConfig.yaml --testconfig=config.yaml
 ```
+OR
+```
+./run-e2e.sh --testconfig=config.yaml
+```
 Flags kubeconfig and testconfig are necessary.
 
 ### Flags
@@ -26,6 +30,7 @@ If not specified, summaries are printed to standard log.
  - provider - Cluster provider, options are: gce, gke, kubemark, aws, local, vsphere, skeleton
  - mastername - Name of the master node
  - masterip - DNS Name / IP of the master node
+ - testoverrides - path to file with overrides.
 
 ## Tests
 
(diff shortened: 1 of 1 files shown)


clusterloader2/README.md
```diff
@@ -48,12 +53,23 @@ Two always available parameters are ```{{.Name}}``` and ```{{.Index}}```
 which specifies object name and object replica index respectively. \
 Example of a template can be found here: [load rc template].
 
+### Overrides
+
+Overrides allow to inject new variables values to the template. \
+Many tests define input parameters. Input parameter is a variable
+that potentially will be provided by the test framework. Cause input parameters are optional,
+each reference should be opaqued with ```DefaultParam``` function that will
```
wojtek-t: Each reference has to be opaqued with ..., because input parameters are option and we need to handle cases when env variable is unset.

clusterloader2/README.md
```diff
@@ -48,12 +53,23 @@ Two always available parameters are ```{{.Name}}``` and ```{{.Index}}```
...
+Example of overrides can be found here: [overrides]
+
 ## Measurement
 
 Currently available measurements are:
 - **APIResponsiveness** \
 This measurement creates summary for latency and number for server api calls.
- Api calls are divided by resource, subresource, verb and scope.
+Api calls are divided by resource, subresource, verb and scope. \
+This measurement is a SLO which means if [api call latency restrictions] are violated,
```
wojtek-t: This measurement verifies if API call latencies SLO is satisfied.

clusterloader2/README.md
```diff
@@ -63,8 +79,21 @@ This measurement gathers the memory profile provided by pprof for a given compon
 - **MetricsForE2E** \
 The measurement gathers metrics from kube-apiserver, controller manager,
 scheduler and optionally all kubelets.
+- **PodStartupLatency** \
+This measurement observes pods being created and collects is startup times. \
+This measurement is a SLO. If given startup time threshold is violated, the error is returned.
```
wojtek-t: This measurement verifies if pod startup SLO is satisfied.

clusterloader2/README.md
```diff
@@ -63,8 +79,21 @@ This measurement gathers the memory profile provided by pprof for a given compon
 - **MetricsForE2E** \
 The measurement gathers metrics from kube-apiserver, controller manager,
 scheduler and optionally all kubelets.
+- **PodStartupLatency** \
+This measurement observes pods being created and collects is startup times. \
```
wojtek-t: This is implementation detail that people shouldn't care about.

clusterloader2/README.md
krzysied: OK. Fixed.

clusterloader2/README.md
krzysied: Done.

clusterloader2/README.md
krzysied: Done.

clusterloader2/README.md
krzysied: Removed.

Issue #396, opened 2019-02-01
ClusterLoader2 assumes ssh service should be available


Betula-L:
**What happened?**

In our cluster, ssh is not available. However, `clusterloader2/pkg/measurement/util/ssh.go:88` assumes every master should own an available ssh service. 

Errors occurred as follow, 

```
[root@Betula-L kubernetes]# ./clusterloader --kubeconfig="${HOME}/.kube/config" --testconfig=testing/density/config.yaml --provider=kubemark --masterip=10.142.43.51 --mastername=localhost
E0201 11:30:17.948014   11090 test_metrics.go:127] TestMetrics: [unexpected error (code: 0) in ssh connection to master: &errors.errorString{s:"error getting signer for provider kubemark: 'error reading SSH key /root/.ssh/google_compute_engine: 'open /root/.ssh/google_compute_engine: no such file or directory''"}]
E0201 11:30:17.948183   11090 kubemark.go:50] error when trying to SSH to master machine. Skipping probe. error getting signer for provider kubemark: 'error reading SSH key /root/.ssh/google_compute_engine: 'open /root/.ssh/google_compute_engine: no such file or directory''

```
**What you expected to happen**:

We may run clusterloader on master without ssh. If parameter `masterip==localhost`, clusterloader could execute `func execLocal()`  rather than `func SSH()`

wojtek-t:
The requirement for ssh is limiting (I agree), but it should break your test.
If anything is failing due to that reason, then we should fix that.

Otherwise (if these are only error lines in the log file), you can probably ignore them.

@krzysied

wojtek-t:
> I think it's hard to be ignore. i can't acquire most of results if errors occurred.

The most crucial one (e.g. those validating SLOs, etc.) should be available.
Those "auxiliary" ones (like profiles, etc.) won't be, which is kind of expected.

I guess I'm trying to set expectations right.
Also, running tests on your master machine is really bad idea due to resource consumption of the tests itself.

Betula-L:
@wojtek-t , could provider `kubemark` acquire SSH_KEY_FILE from env `KUBEMARK_SSH_KEY` rather than constant string `google_compute_engine`? 
```
case "kubemark":
  keyfile = os.Getenv("KUBEMARK_SSH_KEY")
  if len(keyfile) == 0 {
  	keyfile = "google_compute_engine"
  }
```
[...]

PR #397, opened 2019-02-01
fix can't run clusterloader without ssh

Betula-L:
Related to Issue: https://github.com/kubernetes/perf-tests/issues/396

### DIFF
--- a/clusterloader2/pkg/measurement/util/ssh.go
+++ b/clusterloader2/pkg/measurement/util/ssh.go
@@ -17,10 +17,13 @@ limitations under the License.
 package util
 
 import (
+	"bytes"
 	"fmt"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"syscall"
 
 	"golang.org/x/crypto/ssh"
 	sshutil "k8s.io/kubernetes/pkg/ssh"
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/util/ssh.go
```diff
@@ -45,8 +48,47 @@ type SSHResult struct {
...
 }
 
+func execLocal(cmd string) (SSHResult, error) {
+	result := SSHResult{Cmd: cmd}
+	result.User = "localhost"
+
+	c := exec.Command("sh", "-c", cmd)
+	var outBuf, errBuf bytes.Buffer
+	c.Stdout = &outBuf
+	c.Stdout = &errBuf
```
krzysied: c.Stderr = &errBuf

clusterloader2/pkg/measurement/util/ssh.go
```diff
@@ -45,8 +48,47 @@ type SSHResult struct {
...
+			// The program has exited with an exit code != 0
+
+			// This works on both Unix and Windows. Although package
+			// syscall is generally platform dependent, WaitStatus is
+			// defined for both Unix and Windows and in both cases has
+			// an ExitStatus() method with the same signature.
+			if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+				code = status.ExitStatus()
+			}
+		} else {
```
krzysied: Maybe we can invert the `if` to reduce the nesting?
```
v, ok := ...
if !ok {
    return err
}
do_stuff_with_v()
````

Betula-L: Done. @kyzysied

PR #401, opened 2019-02-06
ClusterLoader - Initializing klog

krzysied:
Adding klog initialization.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -159,6 +161,20 @@ func main() {
 		glog.Fatalf("Parsing flags error: %v", errList.String())
 	}
 
+	// TODO: Replce glog with klog.
```
mm4tt: NIT: A good TODO should be  associated with some issue or at least assigned to someone. 
An issue is preferable as usually it provides all the details for someone else to take over in case you won't be able to finish / do it yourself.

Could you create an issue and associate it with this TODO?

clusterloader2/cmd/clusterloader.go
```diff
@@ -159,6 +161,20 @@ func main() {
...
+	// Init klog.
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
+
+	// Sync the glog and klog flags.
```
mm4tt: IIUC, we can get rid of this once we replace glog with klog in cluster loader? If so, could you document that?

clusterloader2/cmd/clusterloader.go

clusterloader2/cmd/clusterloader.go
krzysied: Comment added.

krzysied:
#403 is merged. Already contains changes of #401.

PR #405, opened 2019-02-08
Add more descriptive error messages.

mborsz:
The goal is to have errors that doesn't require to check build-log to see what exact object/metric is incorrect.

/assign @krzysied

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness.go
@@ -108,7 +108,7 @@ func (a *apiResponsivenessMeasurement) apiserverMetricsGather(c clientset.Interf
 		return nil, err
 	}
 	sort.Sort(sort.Reverse(metrics))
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness.go
```diff
@@ -124,19 +124,19 @@ func (a *apiResponsivenessMeasurement) apiserverMetricsGather(c clientset.Interf
 		}
 		if latency > latencyThreshold {
 			isBad = true
-			badMetrics++
+			badMetrics = append(badMetrics, fmt.Sprintf("got: %v; expected perc99 <= %v", metrics.ApiCalls[i], latencyThreshold))
```
krzysied: nit: `%v` prints only values, while `%+v` prints field names with values.

mborsz: Done.

PR #406, opened 2019-02-11
Reimplementing randomized_time_limited tuning set

krzysied:
Reimplementing randomized_time_limited tuning set, so it uses fewer goroutines.

### DIFF
--- a/clusterloader2/pkg/tuningset/randomized_time_limited.go
+++ b/clusterloader2/pkg/tuningset/randomized_time_limited.go
@@ -18,6 +18,7 @@ package tuningset
 
 import (
 	"math/rand"
+	"sort"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/tuningset/randomized_time_limited.go
```diff
@@ -34,15 +35,34 @@ func newRandomizedTimeLimitedLoad(params *api.RandomizedTimeLimitedLoad) TuningS
...
 }
 
+type scheduleEntry struct {
+	action    func()
+	startTime time.Time
+}
+
 func (r *randomizedTimeLimitedLoad) Execute(actions []func()) {
-	var wg wait.Group
+	var schedule []scheduleEntry
```
wojtek-t: Why? What's the point of changing it?

krzysied: Reducing number of gorouitnes (and memory) used.
TBH, I wanted to introduced this changed due to pod store creation flake. In theory goroutines shouldn't be the bottleneck. However if we can reduce something that potentially can be a limit, I think that's something we want to do.

krzysied:
After offline discussion, we decided to drop this PR.

Issue #411, opened 2019-02-13
ClusterLoader2 doesn't handle tombstones


wojtek-t:
Example failure:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/73959/pull-kubernetes-kubemark-e2e-gce-big/37500/build-log.txt

The interested line is this one;
```
[...]

PR #412, opened 2019-02-13
Handle tombstones in clusterloader2

wojtek-t:
Fix https://github.com/kubernetes/perf-tests/issues/411

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -171,7 +171,11 @@ func (w *waitForControlledPodsRunningMeasurement) start(clients *framework.Multi
 		},
 		DeleteFunc: func(obj interface{}) {
 			deleteF := func() {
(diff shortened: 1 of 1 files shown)


wojtek-t:
Fixed type

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -171,6 +171,9 @@ func (w *waitForControlledPodsRunningMeasurement) start(clients *framework.Multi
 		},
 		DeleteFunc: func(obj interface{}) {
 			deleteF := func() {
+				if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+					w.handleObject(clients, tombstone.Obj, nil)
```
krzysied: Should it handle `tombstone.Obj` and then `obj`, or is here a `return` missing?

wojtek-t: should return - will fix

wojtek-t:
PTAL

Issue #414, opened 2019-02-14
ClusterLoader2 should support controllers beside ReplicationController


Betula-L:
**What would you like to be added**:
1. Add parameter `apiVersion` for `type waitForControlledPodsRunningMeasurement struct `
2. Support measurement`WaitForControlledPodsRunning` for kinds in different apiVersions (different controllers)

**Why is this needed**:

**Kubernetes e2e test:** Kubernetes supports multiple controllers such as Deployment, Job, ReplicationController, etc. For kubernetes/kubernetes density test, all of them are supported.

https://github.com/kubernetes/kubernetes/blob/2b9c9154b52bdac914bf3ab37c8d349d926e6e16/test/e2e/scalability/density.go#L557

**ClusterLoad2:**  Clusterload2 only support controllers in CoreV1. Controllers `Deployment`, `Job`, etc. can't be measured by clusterload2.  

 Clusterload2 does not provides parameter `apiVersion` for measurement `WaitForControlledPodsRunning`, due to codes as followed, it can't create listwatch expect in CoreV1().

https://github.com/kubernetes/perf-tests/blob/74aa0691d943234621454aab72c180d4648f9ee6/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go#L155

wojtek-t:
Great point.

@krzysied - please fix

krzysied:
- [ ] Hot fix #415 
Fixing ReplicaSet, Deployment, DaemonSet and Job handling.
- [ ] Replacing clientset with dynamic clent.
This will allow more generic handling.

wojtek-t:
I actually claim that it's not that urgent to make a hack to fix it.
We should wait for a proper fix.

Betula-L:
This is fixed in https://github.com/kubernetes/perf-tests/pull/438

PR #415, opened 2019-02-14
ClusterLoader - Fixing different kinds handling in wait_for_controlled_pods

krzysied:
Adding fix for ReplicaSet, Deployment, DaemonSet and Job.

ref #414

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 /*
-TODO(krzysied): Replace ClientSet interface with dynamic client.
+TODO(#414): Replace ClientSet interface with dynamic client.
 This will allow more generic approach.
 */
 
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -152,7 +152,17 @@ func (w *waitForControlledPodsRunningMeasurement) start(clients *framework.Multi
 		options.FieldSelector = w.fieldSelector
 		options.LabelSelector = w.labelSelector
 	}
-	listerWatcher := cache.NewFilteredListWatchFromClient(clients.GetClient().CoreV1().RESTClient(), w.kind+"s", w.namespace, optionsModifier)
+	var listerWatcher *cache.ListWatch
+	switch w.kind {
+	case "ReplicationController":
+		listerWatcher = cache.NewFilteredListWatchFromClient(clients.GetClient().CoreV1().RESTClient(), w.kind+"s", w.namespace, optionsModifier)
```
wojtek-t: This isn't very generic. I would like to avoid introducing switches that you need to expand whenever you want to use a new type...

krzysied: I agree. 
However this PR is just a hot fix for object kinds CL should support. To implement generic approach, I will need definitely more time than this. Thus I thought it will better to create hot fix and then spend time on reimplementation. There is TODO list in #414.

wojtek-t: I don't think we need this fix that urgently. I would prefer waiting for a proper fix.

Betula-L: > I agree.
> However this PR is just a hot fix for object kinds CL should support. To implement generic approach, I will need definitely more time than this. Thus I thought it will better to create hot fix and then spend time on reimplementation. There is TODO list in #414.

I try to implement a dynamic client for informer by package `k8s.io/client-go/dynamic/dynamicinformer`. It works well. 

However, in my private branch, it applies parameter `resource` rather `kind`, since i can't find any way covert type `GroupVersionKind` to type `GroupVersinResource`. Actually, i do not know what difference between `kind` and `resource`.

krzysied: > i can't find any way covert type `GroupVersionKind` to type `GroupVersinResource`

The simplest way: https://github.com/kubernetes/perf-tests/blob/3fb03e234ca8d50a931cbe94800df4265a9df211/clusterloader2/pkg/framework/client/objects.go#L160

Assuming that clusterlaoder uses in all object operations, I think you can also use in this case.

krzysied:
After offline discussion, it was decided to close this PR.

PR #419, opened 2019-02-18
Allow starting prometheus stack in clusterloader2

mm4tt:
See https://github.com/kubernetes/kubernetes/issues/74213

The feature is flag-gated behind the --set-up-prometheus flag.
I will be gradually launching it, starting from the smallest tests/clusters.
It's probable that we will have to tune the cpu/memory limits/requests in order to make it work for the big clusters, but as the yaml as read via the template provider it shouldn't be hard to have memory/cpu as a function of # of nodes in the cluster.


clusterloader2/cmd/clusterloader.go
```diff
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/prometheus"
```
krzysied: nit: Please move this import one block bellow.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,101 @@
...
+	klog.Infof("Namespace %s created", namespace)
+	if err := applyManifests(framework); err != nil {
+		return err
+	}
+	klog.Info("Prometheus stack set up successfully")
+	return nil
+}
+
+func isPrometheusStackSetUp(client clientset.Interface) (bool, error) {
+	_, err := client.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
```
krzysied: Retry missing. Example: https://github.com/kubernetes/perf-tests/blob/6cde98fdcc96aaff7e0c12f6c4b6bfb767362497/clusterloader2/pkg/framework/client/objects.go#L100
This namespace getter can be added to "k8s.io/perf-tests/clusterloader2/pkg/framework/client".

clusterloader2/cmd/clusterloader.go
mm4tt: Done.

clusterloader2/pkg/prometheus/prometheus.go

clusterloader2/cmd/clusterloader.go
```diff
@@ -30,6 +30,7 @@ import (
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/flags"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/prometheus"
```
wojtek-t: Please remove the binary file from this PR:  clusterloader2/clusterloader

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+	namespace = "monitoring"
+)
+
+// SetUpPrometheusStack sets up prometheus stack in the cluster.
+// It does nothing if the prometheus stack is already set up in the cluster.
+func SetUpPrometheusStack(framework *framework.Framework) error {
+	k8sClient := framework.GetClientSets().GetClient()
+	if isPrometheusStackSetUp, err := isPrometheusStackSetUp(k8sClient); err != nil {
+		return err
+	} else if isPrometheusStackSetUp {
```
wojtek-t: I'm not huge fan of "else if" constructions. What about simply:
isPrometheusStackSetUp, err := isProme...
if err != nil {
  return err
}
if isPromethe... {
}

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+		klog.Info("Prometheus stack already set up.")
+		return nil
+	}
+
+	klog.Info("Setting up prometheus stack")
+	if err := client.CreateNamespace(k8sClient, namespace); err != nil {
+		return err
+	}
+	klog.Infof("Namespace %s created", namespace)
```
wojtek-t: This log isn't particularly useful?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := applyManifests(framework); err != nil {
+		return err
+	}
+	klog.Info("Prometheus stack set up successfully")
+	return nil
+}
+
+func isPrometheusStackSetUp(c clientset.Interface) (bool, error) {
+	if ns, err := client.GetNamespace(c, namespace); err != nil {
```
wojtek-t: I'm not a fan of this - the fact that namespace exists doesn't mean that prometheus stack is actually set up properly.

Instead of handling that internally, I would prefer to move that responsibility on the user. In general, applying a set of manifests (if they didn't change) should be no-op. 
Also, if a code structre initializing it may call the setup function more than once, this can be opaqued in sync.Once.

So I would personally remove this one.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+		return false, err
+	} else {
+		return ns != nil, nil
+	}
+}
+
+func applyManifests(framework *framework.Framework) error {
```
wojtek-t: Can you add a TODO for the future to consider using out-of-the-box "kubectl create -f <directory>"?

clusterloader2/cmd/clusterloader.go
```diff
@@ -66,6 +67,7 @@ func validateClusterFlags() *errors.ErrorList {
 
 func initFlags() {
 	flags.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
+	flags.BoolVar(&clusterLoaderConfig.SetUpPrometheus, "set-up-prometheus", false, "Whether to set-up the prometheus monitoring in the cluster.")
```
wojtek-t: Maybe we should call it "enable-prometheus-server" ?

[I prefer enable than set-up, and I think we should be clear that it's about the server]

mm4tt: Done.

PR #423, opened 2019-02-25
ClusterLoader - Add dynmaic client for WaitForControlledPodsRunning

Betula-L:
**Why we need it**:
Fix https://github.com/kubernetes/perf-tests/issues/414

**What this PR does**:
1. Add vendor k8s.io/client-go/dynamic/dynamicinformer
 (There is still a bug fix not merged https://github.com/kubernetes/kubernetes/pull/74344, but in my PR, it's the fixed version)
2. Add parameter apiVersion for `WaitForControlledPodsRunning`
3. Support dynamic informer for `WaitForControlledPodsRunning`

**Something not sure**:
1. I think my PR handle tombstone in a wrong way, but i do not what's right.
https://github.com/Betula-L/perf-tests/blob/60100ff37cf2ad366366dd7b298602da14fe602c/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go#L179
2. I prefer to remove the useless `switch-case` for followed code.
https://github.com/kubernetes/perf-tests/blob/3ea5f47bceb00d54f6edb3cf36bb1ce98646577d/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go#L169

**Something need to be improved**:
It still has `switch-case` in the code. I do not modify the followed code due to the complex reference for it. 
https://github.com/Betula-L/perf-tests/blob/60100ff37cf2ad366366dd7b298602da14fe602c/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go#L34

### DIFF
--- a/clusterloader2/pkg/framework/framework.go
+++ b/clusterloader2/pkg/framework/framework.go
@@ -69,6 +69,11 @@ func (f *Framework) GetClientSets() *MultiClientSet {
 	return f.clientSets
 }
 
+// GetDynamicClients returns dynamic clients.
+func (f *Framework) GetDynamicClients() *MultiDynamicClient {
+	return f.dynamicClients
+}
+
 // CreateAutomanagedNamespaces creates automanged namespaces.
 func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
 	if f.automanagedNamespaceCount != 0 {
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -30,7 +30,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
-	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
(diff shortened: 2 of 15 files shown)


krzysied:
Please split changes into separate commits. For example:
- vendor update
- adding dynamic clients set getter to Framework + changing MeasurmentManager to use framework
- reimplementation of WaitForControlledPodsRunning

It will make it easier to review.

And also:
- remove clusterlaoder binary file
- there is desnity test is update, however load test and node-throughput test should also be updated

clusterloader2/pkg/measurement/util/pod_store.go
```diff
@@ -40,14 +45,45 @@ type PodStore struct {
 	Reflector *cache.Reflector
 }
 
+var podGroupVersion, _ = schema.ParseGroupVersion("v1")
+var podGropuVersionResource = podGroupVersion.WithResource("pods")
+
+// NewPodStore creates PodStore based on given namespace, label selector and field selector.
+func NewPodStoreForDynamicClient(c dynamic.Interface, namespace string, labelSelector string, fieldSelector string) (*PodStore, error) {
```
wojtek-t: I don't see any reason for introducing pod store for dynamic client - what was our motivation? It doesn't give us anything.

Betula-L: It's because, in my current commit, the caller for func `NewPodStore` does not have clients `clienset.interface`. I will improve it.

Betula-L: Done.

Betula-L:
Hi guys, i commit my code again. @krzysied @wojtek-t 

**The following is a checklist for your suggestion.**
- [x] split changes into separate commits.
- [x] remove clusterlaoder binary file
- [x] update all of tests' config
- [x] do not introduce pod store for dynamic client

**The following is a checklist which is still in question.**
- [ ] how to handle over tombstone problem for dynamic informer? 
https://github.com/kubernetes/perf-tests/blob/3ea5f47bceb00d54f6edb3cf36bb1ce98646577d/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go#L174
- [ ] should i remove the redundant `switch-case` in runtimeobjects.go?

wojtek-t:
@Betula-L - thanks a lot for working on that.
Would you be able to split that PR into multiple ones? Most of the commits are non-controversial, so probably we would be able to merge them immediately (it also significantly simplifies and speeds up review).

Betula-L:
OK. Got it.

Betula-L:
**Please review those double PR first.**
https://github.com/kubernetes/perf-tests/pull/425
https://github.com/kubernetes/perf-tests/pull/426

The following commits are relied on those.
@wojtek-t

PR #424, opened 2019-02-25
ClusterLoader - Limiting pod startup logs to worst 100 pods

krzysied:
Limiting pod startup latencies listing to only 100 worst instead of 10% of all.

Ref discussion: https://github.com/kubernetes/kubernetes/issues/74088#issuecomment-464680872

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -303,7 +303,11 @@ func (p *podStartupLatencyMeasurement) checkPod(obj interface{}) {
 
 func (p *podStartupLatencyMeasurement) printLatencies(latencies []podLatencyData, header string) {
 	metrics := extractLatencyMetrics(latencies)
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -303,7 +303,11 @@ func (p *podStartupLatencyMeasurement) checkPod(obj interface{}) {
 
 func (p *podStartupLatencyMeasurement) printLatencies(latencies []podLatencyData, header string) {
 	metrics := extractLatencyMetrics(latencies)
-	klog.Infof("%s: 10%% %s: %v", p, header, latencies[(len(latencies)*9)/10:])
+	index := len(latencies) - 100
+	if index < 0 {
+		index = 0
+	}
+	klog.Infof("%s: 100 %s: %v", p, header, latencies[index:])
```
wojtek-t: Not necessary 100.

s/100/len(latencies) - index/

krzysied: Done.

PR #426, opened 2019-02-26
ClusterLoader - Add dynamic clients for measurement manager

Betula-L:
**Why we need it**:
https://github.com/kubernetes/perf-tests/issues/414

**What this PR does**:
Related to https://github.com/kubernetes/perf-tests/pull/423

### DIFF
--- a/clusterloader2/pkg/measurement/interface.go
+++ b/clusterloader2/pkg/measurement/interface.go
@@ -25,6 +25,8 @@ import (
 type MeasurementConfig struct {
 	// Clientsets are kubernetes clients.
 	ClientSets *framework.MultiClientSet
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/measurement/interface.go
```diff
@@ -25,6 +25,8 @@ import (
 type MeasurementConfig struct {
 	// Clientsets are kubernetes clients.
 	ClientSets *framework.MultiClientSet
+	// DynamicClients are kuberentes dynamic clients.
```
wojtek-t: s/kuberentes/kubernetes/

Betula-L: Done. @wojtek-t

PR #429, opened 2019-02-26
Tune prometheus stack to make it work for 5k node clusters

mm4tt:
https://github.com/kubernetes/kubernetes/issues/74213


wojtek-t:
thanks!

PR #430, opened 2019-02-26
Recording rule for Network Programming Latency.

mm4tt:
The rule will be used in the Network Programming Latency dashboard.
It's recommended by Prometheus documentation to use recording rules for
dashboards, see https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/

I've also modified the Network Programming Latency dashboard to align it with the SLI/SLO definition.

I also modified the kube-proxy scrape interval, as now thanks to the recording rule we should be able to scrape the metrics more often.

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/grafana-dashboardDefinitions.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/grafana-dashboardDefinitions.yaml
@@ -25,22 +25,9 @@ items:
           "editable": true,
           "gnetId": null,
           "graphTooltip": 0,
-          "id": 10,
+          "id": 2,
           "links": [],
           "panels": [
-            {
-              "collapsed": false,
-              "gridPos": {
-                "h": 1,
-                "w": 24,
-                "x": 0,
-                "y": 0
-              },
-              "id": 4,
-              "panels": [],
-              "title": "Network Programming Latency",
-              "type": "row"
-            },
             {
               "aliasColors": {},
               "bars": false,
@@ -49,10 +36,10 @@ items:
               "datasource": "prometheus",
               "fill": 1,
               "gridPos": {
-                "h": 11,
-                "w": 13,
+                "h": 18,
+                "w": 12,
                 "x": 0,
-                "y": 1
+                "y": 0
               },
               "id": 2,
               "legend": {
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
```diff
@@ -0,0 +1,27 @@
...
+    prometheus: k8s
+    role: alert-rules
+  name: prometheus-k8s-rules
+  namespace: monitoring
+spec:
+  groups:
+  - name: kube-proxy.rules
+    rules:
+    - expr: |
+        histogram_quantile(0.99, sum(rate(kubeproxy_network_programming_latency_seconds_bucket[10m])) by (le))
```
wojtek-t: 5m: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/network_programming_latency.md

(here and below)

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
```diff
@@ -0,0 +1,27 @@
...
+      record: kubeproxy:kubeproxy_network_programming_latency:histogram_quantile
```
wojtek-t: The other thing that is different than in the definition is that you make aggregation across all kube-proxies, whereas in definition we say:
```In default Kubernetes installation, 99th percentile of (99th percentiles across all programmers (e.g. iptables)) per cluster-day <= X```

So we take 99% of fastest kube-proxies and check those.

Maybe we should consider changing the definition - I wrote it that way because I wasn't sure if aggregation percentiles across all nodes is feasible, but it seems it is. Thoughts?

Comment (mm4tt) [reply clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml]: Discussed personally -  https://github.com/kubernetes/community/pull/3307

**Comment by mm4tt** [comment]
Here is how the updated dashboard will look like
0n5brsad2pf

**Comment by mm4tt** [comment]

File: clusterloader2/pkg/prometheus/manifests/prometheus-serviceMonitorKubeProxy.yaml [hunk 3]
```diff
@@ -12,9 +12,9 @@ spec:
       # collect them as often as in smaller clusters. For example if in 100 node cluster the test
       # takes 10min it's useful to gather metrics every 30s. On the other hand if the test takes
       # 10h collecting metrics every 5min is still good enough. The expression below should give us
-      # 30s interval for small clusters (up to 500 nodes), 2.5min interval for 2K node clusters and
-      # 5.5min interval for 5K node clusters.
-    - interval: {{MultiplyInt 30 (AddInt 1 (DivideInt .Nodes 500))}}s
+      # 30s interval for small clusters (up to 1000 nodes), 1.5min interval for 2K node clusters and
```
wojtek-t: Would it be possible to have:
30s if N<=1000
1m for bigger ones?

Or would it kill us?
I think I would prefer to have smaller intervals if possible.

mm4tt: Hard to tell, the only way of finding out is trying it. 

Changed as you proposed, if it causes any issue I will change it back to what it was.

clusterloader2/pkg/prometheus/manifests/grafana-dashboardDefinitions.yaml
```diff
@@ -77,32 +148,102 @@ items:
...
+              "percentage": false,
+              "pointradius": 2,
+              "points": false,
+              "renderer": "flot",
+              "seriesOverrides": [],
+              "stack": false,
+              "steppedLine": false,
+              "targets": [
+                {
+                  "expr": "quantile_over_time(0.99, kubeproxy:kubeproxy_network_programming_latency:histogram_quantile{quantile='0.99'}[1h])",
```
wojtek-t: Why this is 1h here?

wojtek-t: Discussed offline - given it's supposed to be visualization of SLO (and SLo will be computed independetly anyway), let's leave it for now as is.

mm4tt:
/hold
Let's merge https://github.com/kubernetes/perf-tests/pull/447 first, this one is less important and can potentially break something (e.g. we're lowering the scrape interval here).

mm4tt:
I'm OOO next week, will get back to this PR once I'm back.

wojtek-t:
@mm4tt - what's with this one? Do we still need it or can we close it?

mm4tt:
This was taken over by @oxddr and in fact got merged in https://github.com/kubernetes/perf-tests/pull/557
/close

PR #432, opened 2019-02-26
Add ability to tear down prometheus stack.

mm4tt:
I cofirmed that deleting the namespace deletes the PV which in turns deletes disks in GCE.
This should help us fix the "leaked resources" errors, e.g. https://gubernator.k8s.io/build/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-killer/1964

Putting this behind a flag, so one can disable it and leave prometheus stack for further analysis when running tests manually.


mm4tt:

wojtek-t:
> I cofirmed that deleting the namespace deletes the PV which in turns deletes disks in GCE.

My main question is whether this is synchronous - if not there can be a race. I can't remember from the top of my head, (I think it is - so let's try).

/approve

PR #439, opened 2019-02-28
Wait for prometheus to become healthy.

mm4tt:
We shouldn't run tests until the prometheus server is healthy and collects metrics from all targets.

https://github.com/kubernetes/kubernetes/issues/74213

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -17,17 +17,24 @@ limitations under the License.
 package prometheus
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
-	"os"
-	"path/filepath"
 )
 
 const (
-	namespace = "monitoring"
+	namespace                    = "monitoring"
+	checkPrometheusReadyInterval = 30 * time.Second
+	checkPrometheusReadyTimeout  = 10 * time.Minute
 )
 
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -99,3 +110,51 @@ func applyManifests(
...
+
+	var response targetsResponse
+	if err := json.Unmarshal(raw, &response); err != nil {
+		// This shouldn't happen, return error.
+		return false, err
+	}
+
+	for _, t := range response.Data.ActiveTargets {
+		if t.Health != "up" {
+			klog.Infof("Target {job=%s, pod=%s} not ready...", t.Labels["job"], t.Labels["pod"])
```
wojtek-t: How does registering targets work?
Is it possible that we call this function immediately and there won't be any targets yet (and thus we satisfy it), and then the new (no yet up) will reigster?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -99,3 +110,51 @@ func applyManifests(
...
+	klog.Info("Waiting for Prometheus stack to become healthy...")
+	return wait.Poll(
+		checkPrometheusReadyInterval,
+		checkPrometheusReadyTimeout,
+		func() (bool, error) { return isPrometheusReady(client) })
+}
+
+func isPrometheusReady(client clientset.Interface) (bool, error) {
+	raw, err := client.CoreV1().
+		Services("monitoring").
```
krzysied: nit: You added const for namespace name.

clusterloader2/pkg/prometheus/prometheus.go
mm4tt: IIUC, they're created from the ServiceMonitor (CRD) objects and it's very unlikely that the prometheus server will see no targets - even though they might be created after the prometheus server deployment, it's quite unlikely that starting up prometheus server pod will outrun creating these objects. 

Nevertheless I added a condition to expect at least # nodes ready target. That should guarantee that we won't run into the situation you described.

clusterloader2/pkg/prometheus/prometheus.go
mm4tt: Thanks, done!

PR #444, opened 2019-03-07
Add deep links to PerfDash. 

misterikkit:
This change adds two-way binding between the drop-down menus of PerfDash
and the set of URL query parameters. The result is that changing the
drop-downs will change the URL, and changing the URL will update the UI.

More importantly, these URLs can be shared, causing PerfDash to load a
specific graph.

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 1.19
+TAG = 1.20
 
 REPO = gcr.io/k8s-testimages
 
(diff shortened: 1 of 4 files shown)


krzysied:
@misterikkit  Good job!
TBH I was thinking about this feature for some time, but I didn't know how to implement it in an efficient way.

Could you also add the commit that bumps up the perfdash version, please? The version needs to be incremented in Makefile and deployment.yaml files.
Apart from that it looks good to me :+1:

misterikkit:
> The version needs to be incremented in Makefile and deployment.yaml files.

Done.

Is the build and deploy step something that you take care of?

krzysied:
I bumped up the version to 1.19 already, please change to 1.20.

Pushing new version requires using `make push` command - this will push the image and then updating deployment in `k8s-mungegithub` project.
I can take care of this after this PR is merged.

misterikkit:
Whoops, rebased.

I'm pretty sure I don't have permission to push the image or modify the running deployment. Thanks for taking care of that.

PR #445, opened 2019-03-07
Bump l7-lb-controller cpu constraint to 0.20

mm4tt:
This is to address the following test failure (flakiness): https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-killer/2153


mm4tt:
Ref https://github.com/kubernetes/kubernetes/issues/73461

PR #446, opened 2019-03-07
Make Prometheus work in kubemark clusters.

mm4tt:


mm4tt:
Something is wrong with author of my commit and the cla check fails, will resend

krzysied:
> Something is wrong with author of my commit and the cla check fails

@mm4tt It could happen when you are pushing changes from another device and you haven't properly configured github account.
Sometimes it happens without any reason ¯\_(ツ)_/¯ - push with -f flag helps.

PR #434, opened 2019-02-27
ClusterLoader - ApiResponsiveness measurement

krzysied:
- Adding new method for parsing method samples.
- Adding new api responsiveness measurement based on prometheus server data.

PR #447, opened 2019-03-07
Make Prometheus work in kubemark clusters.

mm4tt:
Ref kubernetes/kubernetes#74213


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -99,29 +156,67 @@ func applyManifests(
...
 	}
 	return nil
 }
 
-func waitForPrometheusToBeHealthy(client clientset.Interface, nodeCount int) error {
+// exposeKubemarkApiServerMetrics configures anonymous access to the apiserver metrics in the
+// kubemark cluster.
+func (pc *PrometheusController) exposeKubemarkApiServerMetrics() error {
+	klog.Info("Exposing kube-apiserver metrics in kubemark cluster")
+	// This has to be done in the kubemark cluster, thus we need to client a new client.
```
krzysied: s/client a new client/create a new client/

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -99,29 +156,67 @@ func applyManifests(
...
+			ObjectMeta: metav1.ObjectMeta{Name: "system:anonymous"},
+			RoleRef:    rbacv1.RoleRef{Kind: "ClusterRole", Name: "apiserver-metrics-viewer"},
+			Subjects: []rbacv1.Subject{
+				{Kind: "User", Name: "system:anonymous"},
+			},
+		})
+		return err
+	}
+	if err := retryCreateFunction(createClusterRole); err != nil {
+		return nil
```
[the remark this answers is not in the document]
mm4tt: :) Done

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -99,29 +156,67 @@ func applyManifests(
...
+	}
+	if err := retryCreateFunction(createClusterRoleBinding); err != nil {
+		return nil
```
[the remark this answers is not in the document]
mm4tt: Done

clusterloader2/pkg/prometheus/prometheus.go
mm4tt: Done, thanks!

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -172,3 +272,14 @@ type target struct {
 	Labels map[string]string `json:"labels"`
 	Health string            `json:"health"`
 }
+
+func getKubemarkMasterClusterIp() (ip string, err error) {
```
krzysied: Why do use named return values if there are not used in the code?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -33,37 +36,97 @@ import (
...
+}
+
+// NewPrometheusController creates a new instance of PrometheusController for the given config.
+func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *PrometheusController, err error) {
+	pc = &PrometheusController{
+		clusterLoaderConfig: clusterLoaderConfig,
+		isKubemark:          clusterLoaderConfig.ClusterConfig.Provider == "kubemark",
+	}
+
+	if pc.isKubemark {
```
krzysied: Just suggestion:
```
kubeConfigPath := clusterLoaderConfig.ClusterConfig.KubeConfigPath
if pc.isKubemark {
	// For kubemark we will be setting up Prometheus stack in the root cluster.
	kubeConfigPath = clusterLoaderConfig.ClusterConfig.KubemarkRootKubeConfigPath
}
if pc.framework, err = framework.NewFramework(kubeConfigPath, numK8sClients); err != nil {
	return nil, err
}
```

clusterloader2/pkg/prometheus/prometheus.go
mm4tt: This was a leftover from a previous version. Removed, thanks!

clusterloader2/cmd/clusterloader.go
```diff
@@ -55,13 +55,20 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "masterip", "MASTER_IP", "", "Hostname/IP of the masternode")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubemarkRootKubeConfigPath, "kubemark-root-kubeconfig", "DEFAULT_KUBECONFIG", "",
```
wojtek-t: s/DEFAULT_KUBECONFIG/KUBEMARK_ROOT_KUBECONFIG/

default is misleading and inconsistent with all other names in this line

clusterloader2/pkg/prometheus/manifests/kubemark/prometheus-roleBinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: prometheus-k8s
+  namespace: kubemark
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: monitoring
```
wojtek-t: missing new line?

clusterloader2/pkg/prometheus/manifests/kubemark/prometheus-role.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
```
wojtek-t: missing new line?

clusterloader2/pkg/prometheus/manifests/kubemark/kube-apiserver-endpoints.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
wojtek-t: Out of curiosity - why do we need this?

clusterloader2/pkg/prometheus/manifests/kubemark/kube-apiserver-service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
wojtek-t: Apiserver itself creates a default k8s service. Can't we use that one?

clusterloader2/cmd/clusterloader.go
mm4tt: I totally agree, but the DEFAULT_KUBECONFIG already exists and does exactly what we need (see e.g. https://github.com/kubernetes/kubernetes/blob/c9e14c85f5b826493a2376ab4743499ac834b8ee/test/kubemark/run-e2e-tests.sh#L42)

In that case I decided to use it instead of introducing another env variable. 

But I'm happy to change it to KUBEMARK_ROOT_KUBECONFIG (+ setting it in the cluster set-up scripts).
LMK WDYT

clusterloader2/pkg/prometheus/manifests/kubemark/prometheus-role.yaml
mm4tt: Done.

clusterloader2/pkg/prometheus/manifests/kubemark/prometheus-roleBinding.yaml
mm4tt: Done.

clusterloader2/pkg/prometheus/manifests/kubemark/kube-apiserver-service.yaml
mm4tt: As discussed, this service points to the Kubemark ApiServer, which runs outside of the main cluster.
Added a comment to explain that.

clusterloader2/pkg/prometheus/manifests/kubemark/kube-apiserver-endpoints.yaml
mm4tt: Added comment. PTAL.

mm4tt:
@wojtek-t, this PR has the highest priority, would be great if you can review it first.  It will unblock https://github.com/kubernetes/perf-tests/pull/434

clusterloader2/cmd/clusterloader.go
wojtek-t: OK - if it already exists, I can live with it.

Please just add a TODO to fix that later so that we don't forget why this is like this.

mm4tt: Done.

spiffxp:
FYI https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-kubemark-e2e-gce-big has become quite flaky in the last 2 days, and this PR was suggested as a likely or possible culprit

wojtek-t:
@spiffxp - reverting in https://github.com/kubernetes/test-infra/pull/11811

PR #452, opened 2019-03-12
ClusterLoader2 should retry retriable net errors.

mm4tt:
Example error is:
```
Post https://34.73.197.67/apis/rbac.authorization.k8s.io/v1/clusterrolebindings: net/http: TLS handshake timeout
```


wojtek-t:
/lgtm
/approve

Thanks!

PR #453, opened 2019-03-12
Prometheus - disable creating Kublet Endpoints

mm4tt:
We don't monitor kubelets via Prometheus in scale tests, thus we can disable creation of the Endpoints object for kubelets.

An alternative would be to wait for the Kubelet Endpoints object to be created before running tests. 
But, for now, disabling the Kubelet Endpoints object creation is simply a better solution.


mm4tt:
/assign @krzysied 
/assign @wojtek-t 

Looks like Prow missed Wojtek's approval, could you re-approve?

PR #454, opened 2019-03-12
CL2 should fail on Prometheus set-up errors.

mm4tt:
As the Prometheus Server is going to become a standard way of measuring SLIs in scale-tests we should make the ClusterLoader2 fail if it's not able to set up the Prometheus stack. 
It doesn't make sense to run tests if we're not able to measure SLIs.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -188,7 +188,7 @@ func main() {
 	}
 	if clusterLoaderConfig.EnablePrometheusServer {
 		if err := prometheus.SetUpPrometheusStack(f, &clusterLoaderConfig); err != nil {
-			klog.Errorf("Error while setting up prometheus stack: %v", err)
+			klog.Fatalf("Error while setting up prometheus stack: %v", err)
 		}
 	}
 


clusterloader2/cmd/clusterloader.go
```diff
@@ -188,7 +188,7 @@ func main() {
 	}
 	if clusterLoaderConfig.EnablePrometheusServer {
 		if err := prometheus.SetUpPrometheusStack(f, &clusterLoaderConfig); err != nil {
-			klog.Errorf("Error while setting up prometheus stack: %v", err)
+			klog.Fatalf("Error while setting up prometheus stack: %v", err)
```
wojtek-t: Can we retry instead (e.g. retry 5 times and fail only if never succeeded).

What I'm afraid is that if we simply fail, then with some transient error, we will loose the whole day at least for larger tests.

mm4tt: We're already retrying everything that can fail transiently, e.g. applying manifests or probing whether Prometheus stack has been set up.
Now we've fixed the retry-able net errors, I don't expect it to ever fail due to errors that can be fixed with top-level retry.

What about having an agreement that we will submit it as is, but if it ever fails then we will add the top level retry?

wojtek-t: sgtm

Issue #457, opened 2019-03-19
ClusterLoader2: pull-kubernetes-e2e-gce-100-performance test fails when etcd metrics port is changed


wenjiaswe:
etcd_metrics.go is always checking "curl http://127.0.0.1:2379/metrics",
https://github.com/kubernetes/perf-tests/blob/38cd4ef9bff2646f63690b6137a70a3705e5148b/clusterloader2/pkg/measurement/common/simple/etcd_metrics.go#L157 however in https://github.com/kubernetes/kubernetes/pull/74690, port number 2382 is used in GCE provider, so pull-kubernetes-e2e-gce-100-performance [failed](https://gubernator.k8s.io/build/kubernetes-jenkins/pr-logs/pull/74690/pull-kubernetes-e2e-gce-100-performance/38332?log#log)

PR #458, opened 2019-03-19
Fix etcd metrics port for GCE

wenjiaswe:
When provider is GCE, etcd metrics port will be set to 2382 in https://github.com/kubernetes/kubernetes/pull/74690. Now etcd_metrics.go is always checking on 2379 for metrics, pull-kubernetes-e2e-gce-100-performance fails. 

It's the first I make any change to perf-tests repo, I have some questions:
1. This PR only makes sense when https://github.com/kubernetes/kubernetes/pull/74690 is merged. But if this pull-kubernetes-e2e-gce-100-performance fails, I cannot check in https://github.com/kubernetes/kubernetes/pull/74690. In this case, what shall I do? Should I make change to the perf-tests locally to test it first, get #74690 merged, and then push this PR?

2. Even after https://github.com/kubernetes/kubernetes/pull/74690 is merged, this change to perf-tests PR should only work with k8s source code after the merge, is there any way for me to do the version control in this PR?

Fix: https://github.com/kubernetes/perf-tests/issues/457

Thank you very much! 
/cc @wojtek-t

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/etcd_metrics.go
+++ b/clusterloader2/pkg/measurement/common/simple/etcd_metrics.go
@@ -150,11 +150,22 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string) error {
 func (e *etcdMetricsMeasurement) getEtcdMetrics(host, provider string) ([]*model.Sample, error) {
 	// Etcd is only exposed on localhost level. We are using ssh method
 	if provider == "gke" {
-		klog.Infof("%s: not grabbing scheduler metrics through master SSH: unsupported for gke", e)
+		klog.Infof("%s: not grabbing etcd metrics through master SSH: unsupported for gke", e)
 		return nil, nil
 	}
 
-	cmd := "curl http://localhost:2379/metrics"
+	// In https://github.com/kubernetes/kubernetes/pull/74690, mTLS is enabled for etcd server
+	// http://localhost:2382 is specified to bypass TLS credential requirement when checking
+	// etcd /metrics and /health.
+	if samples, err := e.sshEtcdMetrics("curl http://localhost:2382/metrics", host, provider); err == nil {
+		return samples, nil
+	}
+
+	// Use old endpoint if new one fails.
+	return e.sshEtcdMetrics("curl http://localhost:2379/metrics", host, provider)
+}
+
+func (e *etcdMetricsMeasurement) sshEtcdMetrics(cmd, host, provider string) ([]*model.Sample, error) {
 	sshResult, err := measurementutil.SSH(cmd, host+":22", provider)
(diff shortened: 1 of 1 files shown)


krzysied:
After offline discussion with @wojtek-t, I would suggest this solution:
```
func (e *etcdMetricsMeasurement) getEtcdMetrics(host, provider string) ([]*model.Sample, error) {
	// Etcd is only exposed on localhost level. We are using ssh method
	if provider == "gke" {
		klog.Infof("%s: not grabbing etcd metrics through master SSH: unsupported for gke", e)
		return nil, nil
	}

	// In https://github.com/kubernetes/kubernetes/pull/74690, mTLS is enabled for etcd server
	// http://localhost:2382 is specified to bypass TLS credential requirement when checking
	// etcd /metrics and /health.
	if samples, err := e.sshEtcdMetrics("curl http://localhost:2382/metrics", host, provider); err == nil {
		return samples, nil
	}

	// Use old endpoint if new one fails.
	return e.sshEtcdMetrics("curl http://localhost:2379/metrics", host, provider)
}

func (e *etcdMetricsMeasurement) sshEtcdMetrics(cmd, host, provider string) ([]*model.Sample, error) {
	sshResult, err := measurementutil.SSH(cmd, host+":22", provider)
	if err != nil || sshResult.Code != 0 {
		return nil, fmt.Errorf("unexpected error (code: %d) in ssh connection to master: %#v", sshResult.Code, err)
	}
	data := sshResult.Stdout

	return measurementutil.ExtractMetricSamples(data)
}
```
Cluster loader is used for testing older version of kubernetes - even if provider is gce it should query old endpoint. If curling 2382 fails it should check 2379.

I'm not sure if it only affects gce.
@wojtek-t This will also change the etcd port on kubemark master, right?

wenjiaswe:
@krzysied thank you very much! I have made the change as you suggested, PTAL. 

Do I need to wait till this PR merged in order to get pull-kubernetes-e2e-gce-100-performance test passed for https://github.com/kubernetes/kubernetes/pull/74690?

wojtek-t:
> Do I need to wait till this PR merged in order to get pull-kubernetes-e2e-gce-100-performance test passed for kubernetes/kubernetes#74690?

Yes

wenjiaswe:
Thanks @wojtek-t , so that means I cannot submit  kubernetes/kubernetes#74690 until this PR is merged right? 

I will wait for @krzysied to lgtm and approve. Thanks!

krzysied:
@wenjiaswe I've re-run pull-kubernetes-e2e-gce-100-performance on kubernetes/kubernetes#74690 and it passed. The PR is not blocked anymore :)

wenjiaswe:
Thank you very much @krzysied !

PR #460, opened 2019-03-20
ClusterLoader - Increasing apiserver cpu limit to 2.5

krzysied:
ref https://github.com/kubernetes/kubernetes/issues/75509


wojtek-t:
Can we first check with PR author that the increase is expected? If so, I'm fine with bumping, if not, we should understand it.

krzysied:
/hold
There is a PR that should fix the issue - https://github.com/kubernetes/kubernetes/pull/75570

krzysied:
/close
kubernetes/kubernetes#75570 fixed the issue

Issue #459, opened 2019-03-19
ClusterLoader: Unsafe prometheus getKubemarkMasterClusterIp


krzysied:
There is a problem with flaking kubemark tests.

IIUC the problem is caused by https://github.com/kubernetes/perf-tests/blob/38cd4ef9bff2646f63690b6137a70a3705e5148b/clusterloader2/pkg/prometheus/prometheus.go#L274
The requirement for this bug to happen is for example having two kubemark clusters within one project. The instance list will return two ip addresses, which generates bad manifest yaml.
I'm also guessing that naming cluster "...-kubemark" will result in having instances "...-kubemark-master" and "...-kubemark-kubemark-master". Both instances will satisfy the command filter. 

getKubemarkMasterClusterIp should be implemented in a different way.

ref https://github.com/kubernetes/kubernetes/issues/73105

mm4tt:
/assign

Thanks @krzysied, will take a look.

mm4tt:
/close
Fixed by https://github.com/kubernetes/perf-tests/pull/465

PR #465, opened 2019-03-25
Use $MASTER_IP to scrape kubemark's apiserver metrics

mm4tt:
Previously we've used kubemark master's cluster ip, but I don't think there is any gain in doing that.
What's more the way we implemented getting the master's cluster ip turned out to be not safe when there are multiple clusters running in the same gcp project, see https://github.com/kubernetes/perf-tests/issues/459

Below a proof that the kubemark master can be successfully scraped over public ip
5iggcC0AvTq


mm4tt:
Ref. https://github.com/kubernetes/kubernetes/issues/74213

PR #466, opened 2019-03-25
Retry 404 errors when creating prometheus objects

mm4tt:
The 404 errors during object creations may be caused by races.
E.g. when a ServiceAccount is created the associated token is created asynchronously, when creating an object that needs the token we may get a retryable 404 error if the token has not been created yet.

This should fix the issue described in https://github.com/kubernetes/kubernetes/issues/74213#issuecomment-476247400

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -149,12 +149,12 @@ func (pc *PrometheusController) applyManifests(manifestGlob string) error {
 				return err
 			}
 			for _, item := range objList.Items {
-				if err := pc.framework.CreateObject(item.GetNamespace(), item.GetName(), &item); err != nil {
+				if err := pc.framework.CreateObject(item.GetNamespace(), item.GetName(), &item, client.Retry(apierrs.IsNotFound)); err != nil {
 					return fmt.Errorf("error while applying (%s): %v", manifest, err)
 				}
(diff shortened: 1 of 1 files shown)


wojtek-t:
/approve cancel

I don't agree this is a correct PR in general - see comment there.

clusterloader2/pkg/framework/client/objects.go
```diff
@@ -172,7 +180,14 @@ func CreateObject(dynamicClient dynamic.Interface, namespace string, name string
...
 	}
-	return RetryWithExponentialBackOff(RetryFunction(createFunc, apierrs.IsAlreadyExists))
+	return RetryWithExponentialBackOff(createRetryFunction(
+		createFunc,
+		apierrs.IsAlreadyExists /* isAllowedError */,
+		// We retry the 404 errors for the create object operations as they may be caused by races,
+		// e.g. when a ServiceAccount is created the associated token is created asynchronously, when
+		// creating an object that needs the token we may get retryable 404 error if the token has not
+		// been created yet.
+		logError(apierrs.IsNotFound) /* isRetryableError */))
```
wojtek-t: I don't agree with that fix - while it may make sense in some context, it's not correct in generic case.

If we really need to tolerate some other errors in some contexts, we should make the interface to also accept allowed errors (and default to just Already Exists or sth like that).

PR #467, opened 2019-03-26
Change objects API to allow specifying options

mm4tt:
Options will allow stating how particular errors should be treated, e.g. which additional errors should be allowed (ignored), which should be retried.

This change is almost transparent to the existing code, e.g. the old code like

```
framework.CreateObject(item.GetNamespace(), item.GetName(), &item)
```
will be still valid and will behave as it used to, but it will be also possible to do something like

```
framework.CreateObject(item.GetNamespace(), item.GetName(), &item, client.Allow(apierrs.IsUnauthorized), client.Retry(apierrs.IsNotFound))
```

which in addition will allow IsUnauthorized errors and will retry IsNotFound.

This PR is to address https://github.com/kubernetes/perf-tests/pull/466#discussion_r268795519

### DIFF
--- a/clusterloader2/pkg/framework/client/objects.go
+++ b/clusterloader2/pkg/framework/client/objects.go
@@ -81,15 +81,51 @@ func IsRetryableNetError(err error) bool {
 	return false
 }
 
+// ApiCallOptions describes how api call errors should be treated, i.e. which errors should be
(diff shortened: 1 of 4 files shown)


wojtek-t:
I like it - thanks!

/lgtm
/approve

PR #474, opened 2019-04-02
Probes - boilerplate and first probe

mm4tt:
Probes are simple jobs running inside kubernetes cluster measuring SLIs that could not be measured through other means.

This PR introduces probes boilerplate and implements two probes for measuring the in-cluster network latency SLI.


probes/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.12 AS build-env
```
wojtek-t: 1.12.1 maybe?

I would like to avoid this being silently changed :-)

probes/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+ARG gopkg=github.com/mm4tt/k8s-util/experimental/probes
```
wojtek-t: mm4tt/k8s-util/experimental?

probes/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ADD ["cmd", "/go/src/$gopkg/cmd"]
+ADD ["pkg", "/go/src/$gopkg/pkg"]
+ADD ["go.mod", "/go/src/$gopkg"]
+ADD ["go.sum", "/go/src/$gopkg"]
+
+ENV GO111MODULE on
+WORKDIR /go/src/$gopkg
+RUN CGO_ENABLED=0 go build -o /workspace/probes ./cmd
+
+FROM golang:1.12-alpine
```
wojtek-t: Why de we need alpine?

probes/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+WORKDIR /workspace
+COPY --from=build-env /workspace/probes .
+ENTRYPOINT ["/workspace/probes"]
```
wojtek-t: I'm not huge fan of entrypoint - I personally prefer explicitly running command when starting a container rather than relying on some default for this. Can we remove this?

probes/README.md
```diff
@@ -0,0 +1,41 @@
+# Probes
+
+Probes are simple jobs running inside kubernetes cluster measuring SLIs that could not be measured through other means.
+## Probe Types
+  
+### Ping Client
+
+This probe exports the `probes_in_cluster_network_latency_seconds` metric.
+It implements the [Network Latency Scalability SLI]
```
wojtek-t: This link doesn't work (I think it requires additiona [] at the end).

probes/pkg/ping/client/client.go
```diff
@@ -0,0 +1,58 @@
...
+	"k8s.io/perf-tests/probes/pkg/metrics"
+)
+
+var (
+	pingServerAddress = flag.String("ping-server-address", "", "The address of the ping server")
+	pingSleepDuration = flag.Duration("ping-sleep-duration", 1*time.Second, "Duration of the sleep between pings")
+)
+
+// Run runs the ping client probe that periodically pings the ping server and exports latency metric.
+func Run() {
```
wojtek-t: How about:
- create a pingClientConfig struct
- add a defaultPingClientConfig() method (that will initialize config from flags)
- let Run() take the config as a parameter

I think it would be a bit cleaner.
And same for server below.

probes/pkg/ping/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+		pingServerAddress: *pingServerAddress,
+		pingSleepDuration: *pingSleepDuration,
+	}
+}
+
+// Run runs the ping client probe that periodically pings the ping server and exports latency metric.
+func Run(config *PingClientConfig) {
+	for {
+		time.Sleep(config.pingSleepDuration)
+		klog.Infof("ping -> %s...\n", config.pingServerAddress)
```
wojtek-t: This will be too spammy in my opinion - can we remove it?

probes/pkg/ping/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+		start := time.Now()
+		if err := ping(config.pingServerAddress); err != nil {
+			klog.Infof("Got error: %v", err)
```
wojtek-t: warning maybe?

probes/pkg/ping/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+			// TODO(mm4tt): Increment server not available gauge metric.
+			continue
+		}
+		end := time.Now()
+		latency := end.Sub(start)
+		klog.Infof("Request took: %v\n", latency)
```
wojtek-t: V(4) ?

probes/pkg/ping/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+		time.Sleep(config.pingSleepDuration)
+		klog.Infof("ping -> %s...\n", config.pingServerAddress)
+		start := time.Now()
+		if err := ping(config.pingServerAddress); err != nil {
+			klog.Infof("Got error: %v", err)
+			// TODO(mm4tt): Increment server not available gauge metric.
+			continue
+		}
+		end := time.Now()
+		latency := end.Sub(start)
```
wojtek-t: latency = time.Since(startTime)

probes/pkg/ping/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+		pingSleepDuration: *pingSleepDuration,
+	}
+}
+
+// Run runs the ping client probe that periodically pings the ping server and exports latency metric.
+func Run(config *PingClientConfig) {
+	for {
+		time.Sleep(config.pingSleepDuration)
+		klog.Infof("ping -> %s...\n", config.pingServerAddress)
+		start := time.Now()
```
wojtek-t: startTime maybe?

probes/pkg/ping/server/server.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Run runs the ping server.
+func Run(config *PingServerConfig) {
+	klog.Infof("Listening on %s \n", config.pingServerBindAddress)
+	http.HandleFunc("/", pong)
+	klog.Fatal(http.ListenAndServe(config.pingServerBindAddress, nil))
+}
+
+func pong(w http.ResponseWriter, r *http.Request) {
+	klog.Infof("pong -> %s\n", r.RemoteAddr)
```
wojtek-t: V(4) or something like that maybe?

probes/pkg/ping/server/server.go
```diff
@@ -0,0 +1,55 @@
...
+	w.Write([]byte("pong"))
```
wojtek-t: I think we should explicitly set status 200.

probes/Dockerfile
mm4tt: Every probe is different and requires different arguments, that's why I set only the image **entrypoint** (which is exactly the same for every probe) and didn't set the image **cmd** (which will be different).
See https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

Currently, if you execute 
```
docker run gcr.io/k8s-testimages/probes:v0.0.1
```
You'll get
```
F0408 13:52:18.015111       1 main.go:63] --mode not set!
```

Then you can execute the same with `--help` or with `-mode=XXX` to see what probes are supported and what other flags are required.
If I removed the entrypoint then executing the docker run would produce nothing. I'm quite sure it would be confusing for some people (e.g. I'd find it confusing if I didn't know the set-up :))

probes/pkg/ping/client/client.go
mm4tt: Done.

probes/Dockerfile
wojtek-t: I though that if default doesn't work for you (which is the case here, because you need to pass flags), then you need to set the command too - aparently it's enough to pass args.

So I'm fine with this - thanks!

probes/pkg/ping/server/server.go
mm4tt: Done.

probes/cmd/main.go
```diff
@@ -0,0 +1,65 @@
...
+	case "ping-server":
+		pingserver.Run(pingserver.NewDefaultPingServerConfig())
+	default:
+		klog.Fatalf("Unrecognized mode: '%s'", *mode)
+	}
+}
+
+func verifyFlags() {
+	if *metricAddress == "" {
+		klog.Fatal("--metric-bind-address not set!")
```
wojtek-t: Do we need metrics for server? It doesn't sound like "fatal" to me..

probes/pkg/ping/server/server.go
mm4tt: Done.

probes/cmd/main.go
mm4tt: Yes, we use it to determine whether the probe is up and running. See my reply in the other PR, the one which introduces probes framework.

wojtek-t: Hmm - I'm not huge fan if that. Why can't we set liveness and/or readiness probes on those pods (by exposing dedicated /healthz handler) rather than hacking with /metrics?

mm4tt: I didn't know about liveness and readiness probes.
They're nice features, but after reading more about them, I think they solve a bit different problem then I need to solve.

Liveness probe will restart the container if it's not alive, readiness probe will make sure that no traffic goes to the pod via service as long as it's not ready. 
Both of them are something we may want to consider in the future, but for now I need something else, i.e. an easy way of determining if all probes are up, before starting the actual measurement.

I don't see an easy way of achieving that via liveness/readiness probes, other than querying all the probe pods via api and extracting the readiness probe status. 
And even if I do that I still need to wait for the some of the prometheus targets to become ready (e.g. ping-client). 
So, in my opinion, doing everything via prometheus it's just the simplest solution available right now. 

I agree that using prometheus targets to determine whether a pod is ready is a bit hacky, but I think the same applies to using liveness/readiness probes as they were not designed for that purpose. 

LMK WDYT, I'm happy to discuss that in person when I'm in the office tomorrow.

wojtek-t: in Endpoints object:
https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3795
the Addresses field contain only ready pods - so if you add readiness probe, Endpoints object will not contain not-ready pods.
So you may simply track the Endpoints object contents.

WDYT?

mm4tt: As discussed personally, added comments and todos to implement readiness probes and to consider using something more intuitive to check whether probes are ready. 
Thanks!

PR #475, opened 2019-04-02
Framework for running Probes in ClusterLoader2

mm4tt:
Probes definition can be found in https://github.com/kubernetes/perf-tests/pull/474

The measurement is currently disabled, will be enabling it gradually.

### DIFF
--- a/clusterloader2/pkg/measurement/common/probes/manifests/ping-server-service.yaml
+++ b/clusterloader2/pkg/measurement/common/probes/manifests/ping-server-service.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: probes
+  name: ping-server
+  labels:
+    probe: ping-server
+spec:
+
+  ports:
+    - name: metrics
+      port: 8080
+    - name: http
+      port: 8081
+  selector:
+    probe: ping-server
--- a/clusterloader2/pkg/measurement/common/probes/manifests/ping-server-serviceMonitor.yaml
+++ b/clusterloader2/pkg/measurement/common/probes/manifests/ping-server-serviceMonitor.yaml
@@ -1,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  namespace: probes
+  name: ping-server
+spec:
+  endpoints:
+    - interval: 30s
+      port: metrics
+  namespaceSelector:
+    matchNames:
+      - probes
+  selector:
+    matchLabels:
+      probe: ping-server
--- a/clusterloader2/pkg/measurement/common/probes/probes.go
+++ b/clusterloader2/pkg/measurement/common/probes/probes.go
@@ -1,0 +1,281 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package probes
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/prometheus"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	name            = "Probes"
+	probesNamespace = "probes"
+
+	manifestGlob = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/*.yaml"
+
+	checkProbesReadyInterval = 15 * time.Second
+	checkProbesReadyTimeout  = 5 * time.Minute
+
+	currentProbesMetricsVersion = "v1"
+)
+
+var (
+	serviceMonitorGvr = schema.GroupVersionResource{Group: "monitoring.coreos.com", Version: "v1", Resource: "servicemonitors"}
+)
+
+func init() {
+	measurement.Register(name, createProbesMeasurement)
+}
+
+func createProbesMeasurement() measurement.Measurement {
+	return &probesMeasurement{
+		probeNameToPrometheusQueryTmpl: map[string]string{
+			"in_cluster_network_latency": "quantile_over_time(0.99, probes:in_cluster_network_latency:histogram_quantile[%v])",
+		},
+	}
+}
+
+type probesMeasurement struct {
+	// probeNameToPrometheusQueryTmpl defines a config of this measurement. Updating the config in the
+	// createProbesMeasurement method is the only place in go code that needs to be changed while
+	// adding a new probe.
+	// Each query template should accept a single %v placeholder corresponding to the query window
+	// length. See the 'in_cluster_network_latency' as an example.
+	probeNameToPrometheusQueryTmpl map[string]string
+
+	framework        *framework.Framework
+	replicasPerProbe int
+	templateMapping  map[string]interface{}
+	startTime        time.Time
+}
+
+// Execute supports two actions:
+// - start - starts probes and sets up monitoring
+// - gather - Gathers and prints metrics.
+func (p *probesMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	if config.CloudProvider == "kubemark" {
+		klog.Info("Probes cannot work in Kubemark, skipping the measurement execution")
+		return nil, nil
+	}
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	switch action {
+	case "start":
+		return nil, p.start(config)
+	case "gather":
+		summaries, err := p.gather(config.Params)
+		if err != nil && !errors.IsMetricViolationError(err) {
+			return nil, err
+		}
+		return summaries, err
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+// Dispose cleans up after the measurement.
+func (p *probesMeasurement) Dispose() {
+	if p.framework == nil {
+		klog.Info("Probes weren't started, skipping the Dispose() step")
+		return
+	}
+	klog.Info("Stopping probes...")
+	k8sClient := p.framework.GetClientSets().GetClient()
+	if err := client.DeleteNamespace(k8sClient, probesNamespace); err != nil {
+		klog.Errorf("error while deleting %s namespace: %v", probesNamespace, err)
+	}
+	if err := client.WaitForDeleteNamespace(k8sClient, probesNamespace); err != nil {
+		klog.Errorf("error while waiting for %s namespace to be deleted: %v", probesNamespace, err)
+	}
+}
+
+// String returns string representation of this measurement.
+func (p *probesMeasurement) String() string {
+	return name
+}
+
+func (p *probesMeasurement) initialize(config *measurement.MeasurementConfig) error {
+	replicasPerProbe, err := util.GetInt(config.Params, "replicasPerProbe")
+	if err != nil {
+		return err
+	}
+	p.framework = config.ClusterFramework
+	p.replicasPerProbe = replicasPerProbe
+	p.templateMapping = map[string]interface{}{"Replicas": replicasPerProbe}
+	return nil
+}
+
+func (p *probesMeasurement) start(config *measurement.MeasurementConfig) error {
+	klog.Info("Starting probes...")
+	if !p.startTime.IsZero() {
+		return fmt.Errorf("measurement %s cannot be started twice", name)
+	}
+	if err := p.initialize(config); err != nil {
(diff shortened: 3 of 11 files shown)


clusterloader2/pkg/measurement/common/probes/manifests/ping-server-serviceMonitor.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
```
wojtek-t: Why prometheus need to monitor ping-server?

clusterloader2/pkg/measurement/common/probes/manifests/ping-server-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: ping-server
+  labels:
+    probe: ping-server
+spec:
+
+  ports:
+    - name: metrics
```
wojtek-t: Does server expose metrics? I thought it isn't

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+	checkProbesReadyInterval = 30 * time.Second
+	checkProbesReadyTimeout  = 5 * time.Minute
+
+	currentProbesMetricsVersion = "v1"
+)
+
+var probesEnabled bool
+
+func init() {
+	flags.BoolEnvVar(&probesEnabled, "enable-probes", "ENABLE_PROBES", false, "Whether to run the probes in the clusterloader2 tests.")
```
wojtek-t: Why is this needed?
You need to explicitly perform actions on this measurement, so I don't think this additional guard really provides value.

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+	// createProbesMeasurement method is the only place in go code that needs to be changed while
+	// adding a new probe.
+	// Each query template should accept a single %v placeholder corresponding to the query window
+	// length. See the 'in_cluster_network_latency' as an example.
+	probeNameToPrometheusQueryTmpl map[string]string
+
+	framework        *framework.Framework
+	replicasPerProbe int
+	templateMapping  map[string]interface{}
+	nServiceMonitors int
```
wojtek-t: What is nServiceMonitors?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+	startTime        time.Time
+	mutex            sync.Mutex
```
wojtek-t: What is mutex protecting?

i think that in all other measurements there was silent assumption that the Execute method is not thread-safe and we don't do any locking. So if this was the motivation here, I would suggest removing this for now (just for consistency).

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	switch action {
+	case "start":
+		klog.Info("Starting probes...")
+		err := p.start(config)
+		return nil, err
```
wojtek-t: return nil, p.start(config)

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+		klog.Info("Probes are disabled, skipping the measurement execution")
+		return nil, nil
+	}
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	switch action {
+	case "start":
+		klog.Info("Starting probes...")
```
wojtek-t: Can you move this to start() method?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+		err := p.start(config)
+		return nil, err
+	case "gather":
+		klog.Info("Gathering metrics from probes...")
+		summary, err := p.gather(config.Params)
+		if err != nil && !errors.IsMetricViolationError(err) {
+			return nil, err
+		}
+		return []measurement.Summary{summary}, nil
```
wojtek-t: s/nil/err/ ?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	klog.Info("Stopping probes...")
+	k8sClient := p.framework.GetClientSets().GetClient()
+	if err := client.DeleteNamespace(k8sClient, probesNamespace); err != nil {
+		klog.Errorf("error while deleting %s namespace: %v", probesNamespace, err)
+	}
+	if err := client.WaitForDeleteNamespace(k8sClient, probesNamespace); err != nil {
+		klog.Errorf("error while waiting for %s namespace to be deleted: %v", probesNamespace, err)
+	}
+	*p = *createProbesMeasurement().(*probesMeasurement) // Reset measurement state.
```
wojtek-t: I don't think that's needed - we don't need to reinitialize it becasue the same measurement instantiation won't be reused.

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+				err = errors.NewMetricViolationError(probe.name, err.Error())
+				klog.Errorf("%s: %v", p, err)
+			}
+		}
+		probeSummaries = append(probeSummaries, probe)
+	}
+	return &probesSummary{probeSummaries: probeSummaries}, err
+}
+
+func (p *probesMeasurement) createProbesObjects() error {
```
wojtek-t: Seems really similar to this one:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/prometheus.go#L134

Can we refactor and reuse?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,338 @@
...
+				"Perc99": float64(probe.latency.Perc99) / 1000000,
+			},
+			Unit:   "ms",
+			Labels: map[string]string{"Metric": name},
+		})
+	}
+	return util.PrettyPrintJSON(perfData)
+}
+
+// TODO(mm4tt): Remove the method below and start using the one from common util once it's available.
```
wojtek-t: What is this blocked on? Can we do that now?

clusterloader2/testing/density/config.yaml
```diff
@@ -54,6 +54,11 @@ steps:
       action: start
       nodeMode: {{$NODE_MODE}}
       resourceConstraints: {{$DENSITY_RESOURCE_CONSTRAINTS_FILE}}
+  - Identifier: Probes
```
wojtek-t: Can we enable in a separate PR? In case of problems it's smaller thing to revert then :-)

clusterloader2/pkg/measurement/common/probes/manifests/ping-server-serviceMonitor.yaml
wojtek-t: OK - let's follow up on this on the other PR.

clusterloader2/pkg/measurement/common/probes/probes.go
wojtek-t: OK - I got your point. Kubemark is definitely a valid example.

I think it's a different thing that chaos-monkey - it's a measurement that is needed for SLO measuring. Let me think about it a bit more.

clusterloader2/testing/density/config.yaml
wojtek-t: Still - even if unused I would prefer to add it in a separate PR, you still register it, even if it doesn't perform any non-trivial code, it does something.

clusterloader2/pkg/measurement/common/probes/probes.go
mm4tt: Done in https://github.com/kubernetes/perf-tests/pull/482

clusterloader2/testing/density/config.yaml
mm4tt: Done

clusterloader2/pkg/measurement/common/probes/probes.go
wojtek-t: I was thinking a bit about it and my current thinking about it is:

Kubemark is quite specific setup - some things just doesn't make sense. I tend to think that at some point we may be forced to fork our tests for kubemark, but obviously I would like to delay it by as much as possible.

This is the same case for measurements - some just may not make sense in Kubemark. And this is exactly one example.
And in my opinion such things should be handle directly in measurement - if measurement doesn't make sense in Kubemark, it should just print some warning or sth like that and not start. This will solve the problem of kubemark.

Going to the problem of disabling (or gradually enabling) stuff - I don't think it's the right approach. We should try to keep all the tests consistent. There is a question how we should enable stuff, but how I see that is that we may create a temporary fork of config and have an "experimental" job that will be using that config. And once we believe it works, we will be graduating it to the main config (for all setups).

Thoughts?

mm4tt: Sounds good. Added a code that disables probes in kubemark and will enable the probes gradually via templates. 
Thanks!

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,326 @@
...
+		return err
+	}
+	k8sClient := p.framework.GetClientSets().GetClient()
+	if err := client.CreateNamespace(k8sClient, probesNamespace); err != nil {
+		return err
+	}
+	if err := p.createProbesObjects(); err != nil {
+		return err
+	}
+	if err := p.waitTillProbesReady(); err != nil {
```
wojtek-t: waitForProbesReady ?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,326 @@
...
+			if err != nil {
+				return nil, err
+			}
+			latency := time.Duration(float64(sample.Value) * float64(time.Second))
+			probe.latency.SetQuantile(quantile, latency)
+		}
+		if threshold, ok := thresholds[probeName]; ok {
+			if err := probe.latency.VerifyThreshold(threshold); err != nil {
+				err = errors.NewMetricViolationError(probe.name, err.Error())
+				klog.Errorf("%s: %v", p, err)
```
wojtek-t: I think we don't want this error - this will be logged when checking violations above, right?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+type probeSummary struct {
+	name    string
+	latency measurementutil.LatencyMetric
+}
+
+// SummaryName returns name of the summary.
+func (p *probesSummary) SummaryName() string {
+	return name
```
wojtek-t: This isn't very intuitive - I think that different probes are conceptually different things.
As a user, I would more expect to be separate summaries.

Why not simply have just probeSummary and return a separate one for every single probe?

clusterloader2/pkg/measurement/common/probes/probes.go
wojtek-t: What is the status of it?

clusterloader2/pkg/measurement/common/probes/probes.go
mm4tt: Done.

clusterloader2/pkg/measurement/common/probes/probes.go
mm4tt: The motivation here was to group all the probes together and export in a single, unified form, so they can be easily (and generically) imported into PerfDash. My aim was to make adding a new probe as seamless as possible (in this case no changes required in the PerfDash at all). 

But I can change it and export each probe separately if you think that's better. PLMK.

clusterloader2/pkg/measurement/common/probes/probes.go
mm4tt: Good catch, now should be good.

clusterloader2/pkg/measurement/common/probes/probes.go
mm4tt: @krzysied, any updates?

clusterloader2/pkg/measurement/common/probes/probes.go
wojtek-t: I think the goal is great as mentioned, and I think you're already roughly doing what I want:
- probe name is already in probeNameToPrometheusQueryTmpl and is propagate down to probeSummary
- you already have this "probeSummary" struct

I'm just suggesting to:
- remove `probesSummary` and leave only `probeSummary`
- the methods on probesSummary will now become methos on probeSummary
- return []probeSummary as the result (separate summary for every probe)

Does that make sense?

mm4tt: Done.

wojtek-t: Partially - I thought it should be p.Name here.

clusterloader2/pkg/measurement/common/probes/probes.go
krzysied: I haven't touch it yet.
Assuming that for some reason it is blocking I will look into this today.

clusterloader2/pkg/measurement/common/probes/probes.go
mm4tt: Doh! Thanks, fixed.

mm4tt:
/hold

A lot has changed since I've tested it locally last time. Putting a hold on this PR to remember to retest before merging.

clusterloader2/pkg/measurement/common/probes/probes.go
krzysied: https://github.com/kubernetes/perf-tests/pull/502

mm4tt: Thanks, will wait till https://github.com/kubernetes/perf-tests/pull/502 gets merged and remove this code.

mm4tt: Done.

wojtek-t:
/lgtm
/approve

Please cancel hold once you will be ready.

mm4tt:
> New changes are detected. LGTM label has been removed.

I haven't changed anything... :)

mm4tt:
/hold cancel
Run some tests, everything seems to be working fine.

@wojtek-t, I haven't changed anything but Prow Robot removed your lgtm, could you lgtm again? :)

PR #480, opened 2019-04-05
Log prometheus-k8s events on stack set up failures

mm4tt:
Ref. https://github.com/kubernetes/kubernetes/issues/76182

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -17,6 +17,7 @@ limitations under the License.
 package prometheus
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -104,6 +105,10 @@ func (pc *PrometheusController) SetUpPrometheusStack() error {
 		}
 	}
 	if err := pc.waitForPrometheusToBeHealthy(); err != nil {
+		klog.Info("Dumping monitoring/prometheus-k8s events...")
+		list, _ := client.ListEvents(k8sClient, namespace, "prometheus-k8s")
```
wojtek-t: Let's not ignore errors. Here and below.

PR #481, opened 2019-04-05
Paginate listerWatcher's in slo-monitor, upgrade to modern k8s version, use go modules

### LINKED ISSUE #479: slo-monitor: Full lists of events can impact apiserver
Currently the slo-monitor performs a full list of all events when initializing or relisting a watch: https://github.com/kubernetes/perf-tests/blob/386d9f97061e81a33a0e321eff9e07359c876509/slo-monitor/src/monitors/pod_monitor.go#L175-L181

On large scale kubernetes clusters, there can be large volumes of events (100k+) and the list requests can timeout and be retried by the slo-monitor, resulting is significant load to the apiserver and etcd.

This can be fixed by paginating the list requests, but pagination is not available in the version of client-go used by the slo-monitor, so it will need to have it's dependencies updated first.

cc @wojtek-t

jpbetz:
Fixes #479 by paginating the listerWatchers, which prevents the slo-monitor listing arbitrarily large #s of events from etcd. The bulk of the changes here are to get the slo-monitor to depend on a recent version of kubernetes. Rather than attempting to upgrade the Godeps dependencies, I've switched over to go modules.

cc @wojtek-t


jpbetz:
@gmarek, @ihmccreery Do you know who would be an appropriate reviewer/approver for this? Any pointers on how to test?

wojtek-t:
@jpbetz - seems like gofmt is failing.
This is great change - I think the right way to test that is to build an experimental image and run it against some tests (ideally scalability, but I think as a first step any random test suite is fine).

jpbetz:
Thanks @wojtek-t I'll sort out the fmt issues and the get the test image created and start testing that.

jpbetz:
Testing done:

- Sanity check
  - published this change as `gcr.io/jpbetz-public-images/slo-monitor:0.11.1`
  - Spin up a cluster, start the slo-monitor pod
  - Create a pod
  - curl -s localhost:9090/metrics
  - Verify `slomonitor_pod_e2e_startup_latency_seconds_bucket` and `slomonitor_pod_full_startup_latency_seconds_bucket` metrics both record the pod startup latency via `curl -s localhost:<port>/metrics`
  - Verify `docker logs <container>` contains no errors (it just has `watch of *v1.Event ended with: The resourceVersion for the provided watch is too old.` and `Saw negative startup time for <podname>`
- kubetest e2e test suite

wojtek-t:
/lgtm
/approve

Thanks Joe!

Issue #413, opened 2019-02-13
Simplify load test


wojtek-t:
With migration to CL2 being roughly done, I would like to simplify and evolve load test a bit to:
1. create a service in a bundle with RC
2. that creates a problem that we currently have a service only for half of RCs, but fortunately in new enough releases we can solve that by using the mechanism introduced here: https://github.com/kubernetes/kubernetes/pull/71355
3. switch to using Deployment instead of RC

/assign @krzysied

wojtek-t:
@kubernetes/sig-scalability-feature-requests

krzysied:
The problem I see here is that running this simplified load test for older k8s versions will not work. I think we need to split tests into legacy/ head/.
This change will also be required when we switch to new api responsiveness metric.
Wdyt?

- [ ] Creating legacy and head directories with copies of test configs
- [ ] Changing scalability jobs config
- [ ] Removing all test configs

wojtek-t:
I'm fine with legacy/ and head/ for other reasons (e.g. responsiveness metric).

But I don't understand why you need it for the purpose of this issue?

krzysied:
IIUC we want to use kubernetes/kubernetes#71355 which will not be available for older k8s version, right?

wojtek-t:
Ahh - that's what you meant. Yeah, though given that we only run older releases on 100-node tests, even if that would have no effect, it should still be fine...

wojtek-t:
Actually, I think that PR that was submitted wasn't exactly what we want.
We should create head/ and legacy/ subdirectories within individual tests, as needed.

Or rather, I would say we should even go one step further:
- create a legacy/ directory (within tests that we need to modify, so probably both actually)
- migrate old branches to use configs from that directory
- leave everything else as is
Once this is done, we may start modifying the current configs.

wojtek-t:
Point (2) from original description is mostly done in #488

However, as discussed offline with @krzysied while this fixes kube-proxy issues, it still leads to much higher resource usage of DNS which leads to OOMs. So in order to make this change, we would need to adjust DNS resource limits based on size of the cluster somehow.

So the remaining stuff to do is:
- adjust DNS resource limits
- bundle service together with RC
- switch from RC to Deployment

wojtek-t:
With https://github.com/kubernetes/kubernetes/pull/77918 adjusting DNS resource limits will be possible.

@krzysied - once we get out of regressions, it's an important thing to look at.

[...]

PR #488, opened 2019-04-10
Prepare for adding service-proxy-name label for half of the services

wojtek-t:
Ref https://github.com/kubernetes/perf-tests/issues/413

### DIFF
--- a/clusterloader2/pkg/config/template_functions.go
+++ b/clusterloader2/pkg/config/template_functions.go
@@ -46,6 +46,8 @@ func GetFuncs() template.FuncMap {
 		"MinInt":        minInt,
 		"MaxFloat":      maxFloat,
 		"MinFloat":      minFloat,
+		"IsEven":        isEven,
+		"IsOdd":         isOdd,
 		"DefaultParam":  defaultParam,
 	}
 }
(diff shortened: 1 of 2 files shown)


wojtek-t:
/hold

This requires some testing.

clusterloader2/testing/load/service.yaml
```diff
@@ -2,6 +2,10 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{.Name}}
+{{if IsEven 1}}
```
krzysied: `{{if IsEven .Index}}`?

clusterloader2/pkg/config/template_functions.go
```diff
@@ -168,6 +170,14 @@ func minFloat(numbers ...interface{}) float64 {
 	return min
 }
 
+func isEven(number interface{}) bool {
+	return int(toFloat64(number)) % 2 == 0
+}
+
+func isOdd(a interface{}) bool {
```
krzysied: `func isOdd(number interface{}) bool {`

wojtek-t: done

clusterloader2/testing/load/service.yaml
wojtek-t: Long-term yes. But if I would do it now, I would basically remove half of existing services from kube-proxy - I don't want to do this now.

krzysied: Ok.

wojtek-t:
Verified that it's no op now and will work when changed to .Index

/hold cancel

@krzysied - PTAL since I fixed gofmt.

PR #494, opened 2019-04-11
Rebase slo-monitor to distroless

yuwenma:
This is part of the effort described in kep kubernetes/enhancements#900
Compariing to alpine, distroless can mitigate the influence due to CVE issues.

### DIFF
--- a/slo-monitor/Dockerfile
+++ b/slo-monitor/Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.8
+FROM gcr.io/distroless/static:latest
 MAINTAINER Marek Grabowski <gmarek@google.com>
 ADD build/slo-monitor slo-monitor
 


wojtek-t:
@yuwenma - how this was tested (I'm fine with that but I would like to ensure that this won't break things)?

yuwenma:
> @yuwenma - how this was tested (I'm fine with that but I would like to ensure that this won't break things)?

I built the slo-monitor under /build dir, and create a new image via the Dockerfile. The new image couldn't support entrypoint "/bin/sh" which indicates it's no longer using alpine but distroless.

wojtek-t:
> I built the slo-monitor under /build dir, and created a new image from the Dockerfile. The new image no longer supports --entrypoint "/bin/sh" which indicates it's using distroless now.
I don't think the PR needs to be tested in a real cluster. Let me know if there's any further questions.

That's cool, though it's not exactly what i was asking for.
I was more asking, whether we tested if slo-monitor wasn't relying on something from alpine (it shouldn't, but..) so if when run it still work (reports metrics, etc.)

yuwenma:
> > I built the slo-monitor under /build dir, and created a new image from the Dockerfile. The new image no longer supports --entrypoint "/bin/sh" which indicates it's using distroless now.
> > I don't think the PR needs to be tested in a real cluster. Let me know if there's any further questions.
> 
> That's cool, though it's not exactly what i was asking for.
> I was more asking, whether we tested if slo-monitor wasn't relying on something from alpine (it shouldn't, but..) so if when run it still work (reports metrics, etc.)

I built the image (gcr.io/yuwenma-gke-playground/slo-monitor@sha256:bc7cc73bf8cbaed41bfed8ee099919e85e66c0013d5038400211ceeda6cccb0c) and used that image in the slo-monitor-deployment.yaml and slo-monitor-pod.yaml.
And then successfully created the slo-monitor pod    

```bash
slo-monitor                               1/1     Running   0          53s
slo-monitor-5bc54bfcc8-fpkmj              1/1     Running   0          45s
```

PR #499, opened 2019-04-15
ClusterLoader - Adding generic summary

krzysied:
- adding generic summary
- adding extension to summary

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/etcd_metrics.go
+++ b/clusterloader2/pkg/measurement/common/simple/etcd_metrics.go
@@ -56,18 +56,17 @@ type etcdMetricsMeasurement struct {
 // - start - Starts collecting etcd metrics.
 // - gather - Gathers and prints etcd metrics summary.
 func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
-	var summaries []measurement.Summary
 	action, err := util.GetString(config.Params, "action")
 	if err != nil {
-		return summaries, err
+		return nil, err
 	}
 	provider, err := util.GetStringOrDefault(config.Params, "provider", config.ClusterFramework.GetClusterConfig().Provider)
 	if err != nil {
-		return summaries, err
+		return nil, err
 	}
 	host, err := util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().MasterIP)
 	if err != nil {
-		return summaries, err
+		return nil, err
 	}
 
 	switch action {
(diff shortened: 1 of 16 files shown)


clusterloader2/pkg/measurement/common/simple/metrics_for_e2e.go
```diff
@@ -126,9 +131,7 @@ func (*metricsForE2EMeasurement) String() string {
 	return metricsForE2EName
 }
 
-type metricsForE2E metrics.MetricsCollection
-
-func (m *metricsForE2E) filterMetrics() {
+func filterMetrics(m *metrics.MetricsCollection) {
```
wojtek-t: This should be a method on MetricsCollection

clusterloader2/pkg/measurement/common/simple/etcd_metrics.go
```diff
@@ -83,9 +83,13 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig)
 		if err = e.stopAndSummarize(host, provider); err != nil {
 			return summaries, err
 		}
-		summaries := append(summaries, e.metrics)
+		content, err := util.PrettyPrintJSON(e.metrics)
+		if err != nil {
+			return summaries, err
+		}
+		summary := measurement.CreateSummary(etcdMetricsMetricName, "json", content)
+		summaries := append(summaries, summary)
```
wojtek-t: Given you're touching this code anyway, let's maybe change that.
Both here and in pretty much every other measurement, there is exactly 1 place where you return non-empty summaries. We should switch to:
- not define summaries variable at all
- return nil, if we return an error (all places but this)
- here simply to:
return []measurement.Summary{summary}, nil

Same in every other measurement, given you're touching it anyway.

clusterloader2/pkg/measurement/common/simple/scheduling_throughput.go
```diff
@@ -135,23 +135,23 @@ func (s *schedulingThroughputMeasurement) gather() ([]measurement.Summary, error
 	s.stop()
 	klog.Infof("%s: gathering data", s)
 
-	summary := &schedulingThroughput{}
+	st := &schedulingThroughput{}
```
wojtek-t: I actually prefer summary - st isn't very reader-friendly

clusterloader2/pkg/measurement/common/simple/metrics_for_e2e.go
krzysied: The problem is that MetricsCollection is vendor so I can't add method to it.
I can either changed it other repo, update vendor or copy the whole code here, however I'm not sure if it will be worth the effort.

clusterloader2/pkg/measurement/common/simple/etcd_metrics.go
krzysied: Done. (separate commit)

clusterloader2/pkg/measurement/common/simple/scheduling_throughput.go
krzysied: Renamed to "throughputSummary"

PR #501, opened 2019-04-16
Perfdash - e2e metric fix

krzysied:
Correctly merging result of e2e metric with different "contentType" label.


wojtek-t:
/assign @mborsz 

Maciek - can you please take a look into it? I don't know anything about perf-dash anyway :)

PR #502, opened 2019-04-16
ClusterLoader - extracting prometheus query execution to util

krzysied:
Extracting prometheus query execution from api_responsiveness_proemtheus.go to pkg/measurement/util.

### DIFF
--- a/clusterloader2/pkg/measurement/util/prometheus.go
+++ b/clusterloader2/pkg/measurement/util/prometheus.go
@@ -20,11 +20,19 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"strings"
 	"time"
 
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/common/model"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+const (
+	queryTimeout  = 5 * time.Minute
+	queryInterval = 30 * time.Second
 )
 
 // ExtractMetricSamples unpacks metric blob into prometheus model structures.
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/util/prometheus.go
```diff
@@ -74,6 +88,73 @@ type promResponseData struct {
 	v model.Value
 }
 
+// ExecutePrometheusQuery executes given promethous query.
```
mm4tt: s/promethous/prometheus/

clusterloader2/pkg/measurement/util/prometheus.go
```diff
@@ -74,6 +88,73 @@ type promResponseData struct {
...
+// If not config is provided the default one is used.
```
mm4tt: If *no* config

clusterloader2/pkg/measurement/util/prometheus.go
```diff
@@ -74,6 +88,73 @@ type promResponseData struct {
...
+	for _, sample := range samples {
+		if !math.IsNaN(float64(sample.Value)) {
+			resultSamples = append(resultSamples, sample)
+		}
+	}
+	return resultSamples, nil
+}
+
+// PrometheusConfig contains data required to access promethous server.
+type PrometheusConfig struct {
```
mm4tt: TBH, I don't see anything here that clients of this util would like to customize in the near future. IMHO we can hide all these setting inside the implementation for now.

wojtek-t: +1

clusterloader2/pkg/measurement/util/prometheus.go
krzysied: Done.

clusterloader2/pkg/measurement/util/prometheus.go
krzysied: Line removed (there is no config now).

clusterloader2/pkg/measurement/util/prometheus.go
krzysied: Removed.

PR #506, opened 2019-04-18
Allow enabling probes via test config override.

mm4tt:


wojtek-t:
Given what is happening now in tests, I would prefer to wait until we at least know how many regressions there currently are :-)

wojtek-t:
@oxddr - you're working on this now. Can this one now be closed?

oxddr:
Equivalent to this code has been merged in some of my recent PRs. I think we can close it. 

/close

PR #507, opened 2019-04-18
ClusterLoader - Fixing api responsiveness prometheus log

krzysied:
Adding missing argument (measurement name) to log.

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -61,11 +61,14 @@ type apiResponsivenessMeasurementPrometheus struct {
 	apiCalls  map[string]*apiCall
 }
 
+// Execute supports two actions:
+// - start - Stores the measurement starting time.
+// - gather - Gathers and prints current api server latency data based on data collected by prometheus server.
 func (a *apiResponsivenessMeasurementPrometheus) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
 	var summaries []measurement.Summary
 
 	if config.PrometheusFramework == nil {
-		klog.Errorf("%s: prometheus framework is not provided!")
+		klog.Warning("%s: Prometheus is disabled, skipping the measurement!", a)
 		// TODO(#498): for the testing purpose metric is not returning error.
 		return summaries, nil
 	}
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -65,7 +65,7 @@ func (a *apiResponsivenessMeasurementPrometheus) Execute(config *measurement.Mea
 	var summaries []measurement.Summary
 
 	if config.PrometheusFramework == nil {
-		klog.Errorf("%s: prometheus framework is not provided!")
+		klog.Errorf("%s: prometheus framework is not provided!", a)
```
mm4tt: Can we change it to warning, and make the warning message more informative for people not familiar with our test setup, i.e. let's explain what it means/what happens here by saying something like: Prometheus is disabled, skipping the measurement.

oxddr: +1 
In addition a link/information how to fix it would be appreciated.

krzysied: Changed from error to warning. Log message updated.

oxddr: It still doesn't tell or point you to any info on how to enable it. If there is no such documentation at this point, I'd add a todo to update this place once we have it.

krzysied: I just added it to README.
I'm not sure if there should explicit reference in log. The definition of this measurement is that it uses promethues server.

krzysied: Info about how to enable promethues server added to measurement log.

mm4tt: IMHO this is a bit of an overkill to paste the exact flag here. 

We should only explain what happens, i.e. that the measurement will be skipped due to disabled prometheus server and that the test will proceed without it. Users can find more details either in the Measurement README.md or in the Prometheus README.md

wojtek-t: I think I agree with that - we shouldn't provide full help in every single error message :-)

krzysied: I changed the message so it describes only the reason of failure.
I guess we should put more focus on updating the REAME file then...

wojtek-t: I'm in general not a fan of keeping all documentation in readmes - people don't read this.
In general, logs, errors and code should be self-explonatory wherever possible.

PR #509, opened 2019-04-23
Do not use klog.Fatalf as it prints stacktrace. Use os.Exit(1) instead.

oxddr:


clusterloader2/cmd/clusterloader.go
```diff
@@ -260,6 +260,7 @@ func main() {
 		}
 	}
 	if suiteSummary.NumberOfFailedSpecs > 0 {
-		klog.Fatalf("%d tests have failed!", suiteSummary.NumberOfFailedSpecs)
+		klog.Errorf("%d tests have failed!", suiteSummary.NumberOfFailedSpecs)
+		os.Exit(1)
```
wojtek-t: I disagree with this change actually - we're using klog.Fatalf in many other places in different kubernetes repos, and I prefer to be consistent with the rest of the ecosystem.

oxddr: Not using Fatalf here has two benefits 1) doesn't pollute already spammy logs 2) doesn't print stacktrace (stacktrace printer from test runner is very confusing. 
What's the added value of keeping those things consistent? I personally doesn't see consistency as a value per-se, especially in the situation when another approach has some (clear to me at least) advantages.

wojtek-t: > 1) doesn't pollute already spammy logs

I wouldn't say that the log "tests have failed" is a polluting :-) it's one of the most important information that the test produces

> 2) doesn't print stacktrace (a stacktrace printed from a test runner is very confusing).

I actually find stacktraces super useful in maaany cases. Why do you find it confusing?

> What's the added value of keeping those things consistent? 

The added value is that the logs I don't need to think whether in this file we are using this convention or the other - i want to be able to read test logs the same way as component logs and the same way as logs from anything else.
Especially, given it's something it's not printed tens or hundreds times during the test, but once :)

oxddr: > > 1. doesn't pollute already spammy logs
> 
> I wouldn't say that the log "tests have failed" is a polluting :-) it's one of the most important information that the test produces
> 

Information "x tests have failed" is still going to be printed. No changes on that has been made here. 

> > 1. doesn't print stacktrace (a stacktrace printed from a test runner is very confusing).
> 
> I actually find stacktraces super useful in maaany cases. Why do you find it confusing?

Stacktraces are useful if they are coming from actual tests or other places, where you don't expect code to fail. Here we have a rudimentary check, which is expected to fail regularly - after all this file is part of a testing framework. I don't see any value of printing this stacktraces from the framework itself. To me it's confusing, as it suggest, that there is problem with clusterloader itself, not a test it happened to run. 
 
> > What's the added value of keeping those things consistent?
> 
> The added value is that the logs I don't need to think whether in this file we are using this convention or the other - i want to be able to read test logs the same way as component logs and the same way as logs from anything else.
> Especially, given it's something it's not printed tens or hundreds times during the test, but once :)

You still able to do so. The only change here is that unnecessary stacktrace is not printed. Everything else stays.

Issue #503, opened 2019-04-17
Start monitoring master's components (kubelet, kube-scheduler, kube-controller-manager) via Prometheus


mm4tt:
This can be easily achieved with the Prometheus stack with have and can be extremely helpful when debugging kubemark issues.

mm4tt:
We should do the same for non-kubemark clusters as well

mm4tt:
Scraping metrics from kube-scheduler and kube-controller-manager also seems like a good idea.

PR #511, opened 2019-04-23
Unify MasterIp variable used in prometheus templates.

mm4tt:
Currently we have KubemarkMasterIp used in kubemark clusters, but we will need the same variable to monitor master's kubemark in non-kubemark cluster.

Ref. https://github.com/kubernetes/perf-tests/issues/503

PR #514, opened 2019-04-23
Start monitoring master's kubelet (non-kubemark)

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/503

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/default/master-kubelet-endpoints.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/default/master-kubelet-endpoints.yaml
@@ -1,0 +1,14 @@
+# Endpoints object for the kubelet running on master node.
+apiVersion: v1
+kind: Endpoints
+metadata:
+  namespace: kube-system
+  name: master-kubelet
+  labels:
+    k8s-app: master-kubelet
+subsets:
+  - addresses:
+      - ip: {{.MasterIp}}
+    ports:
+      - name: https-metrics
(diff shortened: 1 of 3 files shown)


mm4tt:
/hold
Wait until https://github.com/kubernetes/perf-tests/pull/511 gets merged

clusterloader2/pkg/prometheus/manifests/default/master-kubelet-endpoints.yaml
```diff
@@ -0,0 +1,14 @@
+# Endpoints object for the kubelet running on master node.
+apiVersion: v1
+kind: Endpoints
+metadata:
+  namespace: kube-system
+  name: master-kubelet
+  labels:
+    k8s-app: master-kubelet
```
wojtek-t: What interesting metrics kubelet exposes?

mm4tt: To be honest I'm not sure. I believe that majority of the useful metrics come from cadvisor.
Nevertheless it doesn't cost much to scrape kubelet and it always may turn out to be useful in future.
When it comes to metrics I adhere the principle - better more than less :) 

LMK WDYT

wojtek-t: So the answer is "just in case we will need them" - I'm fine with that, just wanted to understand it.

PR #516, opened 2019-04-23
Store provider instead of isKubemark in PrometheusController

mm4tt:


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -206,12 +211,15 @@ func dumpAdditionalLogsOnPrometheusSetupFailure(k8sClient kubernetes.Interface)
 		klog.Warningf("Error while listing monitoring/prometheus-k8s events: %v", err)
 		return
 	}
-	s, err := json.MarshalIndent(list, "" /*=prefix*/, "  " /*=indent*/)
+	klog.Info(toJson(list))
```
wojtek-t: I think this is not exactly the same - previously non-nil error was printed as warning, with this change, I don't fully know what it will write for function returning two values.
I would personally revert this change - I don't think it buys us much.

mm4tt: Changed to keep the previous behavior.
In case of MarshalIndent errors it will print warning and ignore the error (as it does currently).

I'll need exactly the same code in a code that I'm currently writing, that's why I'm extracting it to a separate method.

wojtek-t: I think it's still not the same, because now on error you will also log the additional empty line.

If you really need that (though I think the overhead of writing it every time is actually negligible, you may want to add a function like "klogJsoinified" or sth like that...

mm4tt: True, removed and left only the isKubemark -> provider change.

PR #518, opened 2019-04-23
Experimental snapshotting of Prometheus's PD (GCP only).

mm4tt:
Automatic snapshot of the prometheus disk after the tests might be extremely helpful in our large scale tests. Especially with the new debug dashboard that @mborsz is working on.
Two main use-cases are:

1. Manual testing, by setting this flag you can start your manual scale tests and analyze the prometheus graphs later, without worrying that it would be wiped by the next run
2. CI testing, we could start storing snapshots for few recent runs of scale tests for comparison / debugging purposes. Last time I checked these snapshots are only a few hundred MBs per run so they basically cost nothing comparing to the gcs storage we use per run. We'd have to come up with some automatic clean-up mechanism eventually, but we could start with manual clean-up every month or so. This scenarion might require more work, but it's something we should consider as a temporary solution given that it requires virtually no work to enable once this PR gets merged.

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prometheus
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"os/exec"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+)
+
+var (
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -0,0 +1,83 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"os/exec"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+)
+
+var (
+	shouldSnapshotPrometheusDisk = pflag.Bool("experimental-gcp-snapshot-prometheus-disk", false, "(experimental, provider=gce|gke only) whether to snapshot Prometheus disk before Prometheus stack is torn down")
```
wojtek-t: This is something that I mentioned to you at some point in the past:
https://github.com/kubernetes/enhancements/issues/177

Can you please maybe ping the author directly over email and ask about the status of it whether it's something we can use here? If that already exists, I would prefer using something existing, especially that it won't introduce restrictions.

mm4tt: The feature is currently in alpha, was supposed to be in alpha in 1.14, but didn't get in.

As discussed personally, leaving it as is for now. It's a hacky, but we don't really have capacity to do anything better now and it should do as a temporary solution.
In future we should investigate other options and rewrite this to something better, e.g. exporting prometheus data to some db or maybe there is already existing tool that we could use that addresses our requirements. 
Opened an issue to track this - https://github.com/kubernetes/perf-tests/issues/523

wojtek-t:
I'm not the most happy person from that solution, but given how important it seems to be for our debugging, let's have this merged.
It's explicitly marked as experimental (both via flag name and via filename), so I can live with it.

@mborsz - FYI

/lgtm
/approve

PR #524, opened 2019-05-06
Increase fluentd-gcp memory constraint.

mm4tt:
This is to deflake the pull-kubernetes-e2e-gce-100-performance, see https://github.com/kubernetes/kubernetes/issues/73884#issuecomment-489542295

### DIFF
--- a/clusterloader2/testing/density/100_nodes/constraints.yaml
+++ b/clusterloader2/testing/density/100_nodes/constraints.yaml
@@ -10,7 +10,7 @@ coredns:
   memoryConstraint: 31457280 #30 * (1024 * 1024)
 fluentd-gcp:
   cpuConstraint: 1
-  memoryConstraint: 241172480 #230 * (1024 * 1024)
+  memoryConstraint: 314572800 #300 * (1024 * 1024)
 heapster:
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/density/100_nodes/constraints.yaml
```diff
@@ -10,7 +10,7 @@ coredns:
   memoryConstraint: 31457280 #30 * (1024 * 1024)
 fluentd-gcp:
   cpuConstraint: 1
-  memoryConstraint: 241172480 #230 * (1024 * 1024)
+  memoryConstraint: 314572800 #300 * (1024 * 1024)
```
wojtek-t: I can live with increasing it, but I would like to understand what caused it.
This seems like a regression.

Did we start logging something more excessively? Did we changed fluentd version?

mm4tt: Agreed, opened an issue to investigate - https://github.com/kubernetes/kubernetes/issues/77492

wojtek-t:
Closing - I don't think it's relevant now.

PR #525, opened 2019-05-06
Switch to apps/v1

liggitt:
Required to get CI green in advance of disabling deprecated apps and extensions APIs

xref https://github.com/kubernetes/kubernetes/issues/43214 https://github.com/kubernetes/kubernetes/pull/70672


liggitt:
fixes the manifests added in https://github.com/kubernetes/perf-tests/pull/419/files#diff-be22106d2c77f9cec5546fb92da5c101 to use apps/v1. prior to that PR, the perf jobs were green against https://github.com/kubernetes/kubernetes/pull/70672, and we need to keep CI green against that PR to ensure we are not adding new dependencies on deprecated APIs

shyamjvs:
Thanks!

Issue #522, opened 2019-04-29
Scrape etcd server metrics via Prometheus


mm4tt:
Currently we scrape only kube-apiserver and some other master components. Scraping etcd server can gives us some additional metrics that may be useful in debugging.

mm4tt:
With https://github.com/kubernetes/test-infra/pull/12774, this is done.

/close

PR #526, opened 2019-05-09
Gather etcd metrics to prometheus

mborsz:
/assign @mm4tt

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-endpoints.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-endpoints.yaml
@@ -18,3 +18,5 @@ subsets:
         port: 10251
       - name: kube-controller-manager
         port: 10252
+      - name: etcd
+        port: 2379
--- a/clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-serviceMonitor.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-serviceMonitor.yaml
@@ -34,3 +34,5 @@ spec:
     port: kube-scheduler
   - interval: 5s
     port: kube-controller-manager
+  - interval: 5s
+    port: etcd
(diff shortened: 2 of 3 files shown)


clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-serviceMonitor.yaml
```diff
@@ -34,3 +34,5 @@ spec:
     port: kube-scheduler
   - interval: 5s
     port: kube-controller-manager
+  - interval: 5s
+    port: etcd
```
mm4tt: NIT: Add newline

clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-endpoints.yaml
```diff
@@ -18,3 +18,5 @@ subsets:
         port: 10251
       - name: kube-controller-manager
         port: 10252
+      - name: etcd
+        port: 2379
```
mm4tt: Unfortunately, it won't work just like that.
It looks like the etcd listens on port 2379 only on localhost - https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/manifests/etcd.manifest#L26
```
--advertise-client-urls http://127.0.0.1:{{ port }}
```

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/522

clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-endpoints.yaml
mborsz: As discussed offline, it works in kubemark as kubemark uses differet etcd config with --listen-client-urls=http://0.0.0.0:2379

mm4tt: After discussing this personally, looks like it will work for kubemark. When setting up etcd for kubemark master we use
```
-listen-client-urls=http://0.0.0.0:2379 
```

PR #533, opened 2019-05-20
Update prometheus operator to v0.30.0

mm4tt:
Ref. https://github.com/kubernetes/kubernetes/pull/70672


mm4tt:
I've run clusterloader2 locally in order to test that the change doesn't break anything.

PR #534, opened 2019-05-21
Migrate prometheus dashboards to separate files

mborsz:
Before this PR, changing dashboards is quite painful: they are a part of bigger files, requires aligning and escaping "{{" and "}}". After the change they are simply json files.


mm4tt:
Nice, in future we should invest in expressing the dashboard in Grafonnet and generate the json files that way. See examples at https://github.com/grafana/grafonnet-lib

/lgtm
/approve

Issue #523, opened 2019-04-29
Prometheus data from tests should be retained


mm4tt:
I've sent https://github.com/kubernetes/perf-tests/pull/518 to add experimental snapshot support (GCP only). It's hacky but should work as a temporary solution for now.

Long term, we should investigate other options and rewrite this to something better, e.g. exporting Prometheus data from tests to some db or maybe there is already existing tool that we could use that addresses our requirements.

mm4tt:
I believe we can close this one. 
While the original idea was to use Cortex, we ended up implementing a "snapshot-base" solution that seems to be working nicely. 
Putting it in other words, as of now the Prometheus test data is retained for all continuous tests. The goal of this issue has been achieved. 

/close

PR #535, opened 2019-05-22
Upgrade prometheus and grafana to the newest versions

mm4tt:

PR #536, opened 2019-05-22
Experimental Prometheus remote write to Cortex.

mm4tt:
Inspired by the Cortex talks on KubeCon, I allowed myself to setup an experimental [Cortex](https://github.com/cortexproject/cortex) cluster.
The cluster is hosted in my dev project, the grafana for it is available at http://35.237.38.211:3000.

The advantages of using Cortex in addition to Prometheus are:
  * It supports multi-tenancy (we can have a separate tenant per test)
  * Horizontally scalable / highly available - it should be easily able to support the load from our tests (if what authors of Cortex are promising is true)
  * Faster / better query enging - we wouldn't run into prometheus "too many samples" limitations anymore
  * Long term storage - we should be able to store metrics from all our place in single place
  * We'll able to have live dashboards for currently running tests
  * And many more :)

I'd like to experiment with it in our scalability tests.
If anything goes wrong it shouldn't break anything, given how the remote write is implemented in prometheus.

Ref. https://github.com/kubernetes/perf-tests/issues/523

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -30,6 +30,7 @@ import (
 
 var (
 	shouldSnapshotPrometheusDisk = pflag.Bool("experimental-gcp-snapshot-prometheus-disk", false, "(experimental, provider=gce|gke only) whether to snapshot Prometheus disk before Prometheus stack is torn down")
+	cortexTenantId               = pflag.String("experimental-cortex-tenant-id", "", "(experimental) if set the remote write to cortex will be enabled with the given tenant id")
 )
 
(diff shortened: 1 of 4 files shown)


mm4tt:
/hold
until https://github.com/kubernetes/perf-tests/pull/535 gets merged

mm4tt:
/assign @wojtek-t 

@wojtek-t, could you take a look? I can give a short demo to the team on Monday.

wojtek-t:
I would like to first understand what exactly this is doing - let's make a demo and we will see how to proceed with this.

wojtek-t:
To follow up - @mm4tt showed us how it works. It looks great but there are couple issues to be solve to productionize it. @mm4tt is going to work in his free cycles on productioninzing it a bit more before we proceed with this PR.

PR #538, opened 2019-05-24
Scrape etcd metrics via prometheus.

mm4tt:
This is possible now the https://github.com/kubernetes/kubernetes/pull/77657 has merged.

Gated behind PROMETHEUS_SCRAPE_ETCD variable in order to enable only in head as otherwise we'd have to cherrypick https://github.com/kubernetes/kubernetes/pull/77657 to older releases.


clusterloader2/pkg/prometheus/manifests/default/master-serviceMonitor.yaml
```diff
@@ -1,19 +1,25 @@
+{{$PROMETHEUS_SCRAPE_ETCD := DefaultParam .PROMETHEUS_SCRAPE_ETCD false}}
```
wojtek-t: Do we really gate it?
I think it's useful enough that I think we should simply switch it on...

mm4tt: Discussed personally, we are gating it to avoid long process of cherrypicking.

wojtek-t: I have those "if"s, but ok - let's make it.

clusterloader2/testing/prometheus/scrape-etcd.yaml
```diff
@@ -0,0 +1 @@
+PROMETHEUS_SCRAPE_ETCD: true
```
wojtek-t: Does this enable it? If so, we shouldn't do that, right?

mm4tt: It doesn't, but we will use it to enable etcd scraping in the test-infra repo by having:
```
      - --test-cmd-args=--testoverrides=./testing/prometheus/scrape-etcd.yaml
```

wojtek-t:
/lgtm
/approve

/hold
Do we want to enable presubmits before we merge that? I think they're close to be ready?
[But feel free to hold cancel if you're confident about this]

mm4tt:
/hold cancel

As discussed, the presubmit has been already enabled, we're good to go :)

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/522

clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-service.yaml
```diff
@@ -14,6 +14,8 @@ spec:
   ports:
     - name: apiserver
       port: 443
+    - name: etcd
+      port: 2379
```
roycaihw: [pull-kubernetes-kubemark-e2e-gce-big](https://testgrid.k8s.io/presubmits-kubernetes-blocking#pull-kubernetes-kubemark-e2e-gce-big) is failing and the starting time seems to align with this PR

this change seems to conflict with the latter duplicate `etcd` port:

from https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-kubemark-e2e-gce-big/1133114154450685954
```
W0527 21:23:36.157] F0527 21:23:36.156399   85477 clusterloader.go:233] Error while setting up prometheus stack: error while applying (/go/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/kubemark/kubemark-master-service.yaml): Service "kubemark-master" is invalid: [spec.ports[5].name: Duplicate value: "etcd", spec.ports[5]: Duplicate value: core.ServicePort{Name:"", Protocol:"TCP", Port:2379, TargetPort:intstr.IntOrString{Type:0, IntVal:0, StrVal:""}, NodePort:0}]
```

Issue #527, opened 2019-05-13
Perf-dash reports the same values for 90/99/100 pctl for small tests


mm4tt:
[Example dashboard](http://perf-dash.k8s.io/#/?jobname=gce-100Nodes&metriccategoryname=E2E&metricname=DensityResources&PodName=fluentd-gcp%2Ffluentd-gcp&Resource=CPU)
![Rx4rfHvtGbo](https://user-images.githubusercontent.com/2604887/57626595-792c4780-7596-11e9-8d9a-464c8b31f4c1.png)

This is due to the way we compute percentiles, which doesn't work well for short tests where we don't have enough metric samples.

The way we compute the percentiles is as follows:

1. We scrape each node every ~1min (e.g. for gce-100 density test, which takes ~10 mins we have about ~6 samples)
2. For each node/container we calculate the 100th/99th/90th percentile (continuing the example, in case of gce-100, where we only have 6 samples, 100th/99th/90th percentile will be the same = max of all samples)
3. Percentiles per node/container are exported to perf-dash
4. Perf Dash aggregates the percentiles using the max function, e.g. it reports the 99th percentile for the whole run as max of all 99th percentiles per node/container. While it may estimate the actual 99th percentile well when there is a lot of samples, it doesn't work at all in small tests. For gce-100 what we see in perf-dash reported as 100th/99th/90th percentile is in fact a max value for all gathered samples.

The problem is that 
1. It's misleading
2. It makes the graphs really noisy due to the fact that max is noisy (that's why we use percentiles)

It makes really hard to debug increases in metrics, especially pinpointing the exact commit where something started.

*Long term*, we should fix how the percentiles are computed, e.g. consider computing real 99th percentile across all samples and not using the max of 99th pctl per node heuristic.  

Short term, we could try adding 50th percentile which will computed correctly as long as we have more than 2 samples. This should be relatively simple and should help us identifying regressions.

mm4tt:
With https://github.com/kubernetes/perf-tests/pull/540 the short-term fix has been implemented. Thanks, @krzysied!

oxddr:
This seems to be fixed, as per @mm4tt comment. 

/close

PR #540, opened 2019-05-27
Perfdash - Label fix

krzysied:
Perfdash plot labels should be set based on last run configuration.


mm4tt:
/lgtm

Thanks!

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/527

PR #541, opened 2019-05-27
failing-test: remove duplicate ports and endpoints

roycaihw:
duplicate `ServicePort` names make `Service` object fail validation during e2e test. The other two APIs (`Endpoints` and `ServiceMonitor`) might tolerate duplication, but de-duped here for clarity.

ref: https://github.com/kubernetes/kubernetes/issues/78401, https://github.com/kubernetes/perf-tests/pull/538#discussion_r287891512

/cc @mm4tt @wojtek-t


mm4tt:
@roycaihw, thanks for fixing!

https://github.com/kubernetes/perf-tests/pull/538 shouldn't have touched the kubemark files, looks like it was result of a bad merge. My bad, sorry for the problems and thanks again!

cwdsuzhou:
@roycaihw thank you for the fix

PR #542, opened 2019-05-28
Make (stripped) prometheus work on gke.

mm4tt:
Please note that this doesn't address the regional clusters. Prometheus will still won't work, or work incorrectly, when there is more than 1 master.
Replicated masters will be addressed in a separate PR.


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -186,9 +193,10 @@ func (pc *PrometheusController) waitForPrometheusToBeHealthy() error {
 
 func (pc *PrometheusController) isPrometheusReady() (bool, error) {
 	// TODO(mm4tt): Re-enable kube-proxy monitoring and expect more targets.
-	// This is a safeguard from a race condition where the prometheus server is started before targets
-	// are registered.
-	expectedTargets := 8 // prometheus, prometheus-operator, grafana, master-kubelet, master-cadvisor, apiserver, controller-manager, scheduler
+	// This is a safeguard from a race condition where the prometheus server is started before
+	// targets are registered. These 4 targets are always expected, in all possible configurations:
+	// prometheus, prometheus-operator, grafana, master-kubelet, apiserver
+	expectedTargets := 4
```
wojtek-t: Seems like should be 5 from the above comment :)

mm4tt: Good catch, forgot to remove master-kublet from the comment.
Done.

PR #544, opened 2019-05-28
Make CL2 accept multiple master ips.

mm4tt:
This change should be no-op, it only changes the CL2 api (flags/envs) to accept multiple values as the `master ip` - both internal and external. 
The api change should be backward compatible, i.e. if flag/env was set to a single master-ip it will continue to work in exactly the same way.

Actual changes supporting multi-master clusters will come in separate PRs.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -57,8 +57,9 @@ func initClusterFlags() {
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
-	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "masterip", "MASTER_IP", "", "Hostname/IP of the masternode")
-	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterInternalIP, "master-internal-ip", "MASTER_INTERNAL_IP", "", "Cluster internal/private IP of the master vm")
+	// TODO(#595): Change the name of the MASTER_IP and MASTER_INTERNAL_IP flags and vars to plural
+	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
+	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterInternalIPs, "master-internal-ip", "MASTER_INTERNAL_IP", nil /*defaultValue*/, "Cluster internal/private IP of the master vm, supports multiple values when separated by commas")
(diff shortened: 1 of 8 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -57,8 +57,8 @@ func initClusterFlags() {
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
-	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIP, "masterip", "MASTER_IP", "", "Hostname/IP of the masternode")
-	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterInternalIP, "master-internal-ip", "MASTER_INTERNAL_IP", "", "Cluster internal/private IP of the master vm")
+	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
```
wojtek-t: I understand that this needs to be done in steps, but we should also change the flag name and env-var name to plural.
At least a TODO for that please.

clusterloader2/pkg/flags/types.go
```diff
@@ -17,9 +17,9 @@ limitations under the License.
 package flags
 
 import (
-	"strconv"
-
 	"github.com/spf13/pflag"
+	"strconv"
```
wojtek-t: Please revert - the convention is that the first group of important are builtin go libraries, exactly how it was before.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -228,9 +228,9 @@ func dumpAdditionalLogsOnPrometheusSetupFailure(k8sClient kubernetes.Interface)
 }
 
 func getMasterIp(clusterConfig config.ClusterConfig) (string, error) {
-	if clusterConfig.MasterInternalIP != "" {
-		klog.Infof("Using internal master ip (%s) to monitor master's components", clusterConfig.MasterInternalIP)
-		return clusterConfig.MasterInternalIP, nil
+	if clusterConfig.GetMasterIp() != "" {
```
wojtek-t: internal

clusterloader2/pkg/flags/flags.go
```diff
@@ -65,6 +66,18 @@ func StringArrayVar(s *[]string, flagName string, defaultValue []string, descrip
 	pflag.StringArrayVar(s, flagName, defaultValue, description)
 }
 
+// StringSliceEnvVar creates a string slice flag with the given parameters.
```
wojtek-t: BTW - this support for string-slice flags, is orthogonal to the overal change, so if you could split that into separate PR, we could merge it now :)

mm4tt: Was done some time ago :)

clusterloader2/pkg/flags/types.go
mm4tt: Outdated

clusterloader2/cmd/clusterloader.go
mm4tt: Added TODO.

mm4tt:
@mborsz, @wojtek-t, PTAL

wojtek-t:
lgtm

PR #547, opened 2019-05-31
Perfdash - Fixing apiserver profile getter

krzysied:
Changing apiserver profile getting from accessing 8080 to using client access.


clusterloader2/pkg/measurement/common/simple/profile.go
```diff
@@ -201,22 +202,32 @@ func (*cpuProfileMeasurement) String() string {
...
 
-	// Get the profile data over SSH.
-	getCommand := fmt.Sprintf("curl -s localhost:%v/debug/pprof/%s", profilePort, p.config.kind)
-	sshResult, err := measurementutil.SSH(getCommand, p.config.host+":22", p.config.provider)
-	if err != nil {
-		if p.config.provider == "gke" {
-			// Only logging error for gke. SSHing to gke master is not supported.
-			klog.Errorf("%s: failed to execute curl command on master through SSH: %v", p.name, err)
-			return nil, nil
+	var profile string
```
wojtek-t: The above won't work - getPortForComponent will return an error for kube-apiserver.
You need to handle kube-apiserver first, and only the getPortForComponent and the rest (btw you also won't have "else" which is always good :)

krzysied: Done.
For some reason I was positive that I moved getPortForComponent to else section...

Issue #546, opened 2019-05-31
Make perfdash configurable


krzysied:
Perfdash should be more configurable, so it can be set up by user to display custom runs data.

- [x] Allowing array flags #548 
- [x] Extracting job config url to flags #548 
- [x] Allowing job configs from files #560 
- [x] Replacing contrib/utils with storage client. #563 
- [x] Making storage access configurable #564 
- [x] Bump up new version and update http://perf-dash.k8s.io #566

krzysied:
Done!

PR #548, opened 2019-05-31
Perfdash - Extracting config paths to flags

krzysied:
- Adding support for multi flags
- Extracting hardcoded config paths to deployment params.

ref #546

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
-	"k8s.io/contrib/test-utils/utils"
 )
 
 // To add new e2e test support, you need to:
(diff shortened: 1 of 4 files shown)


perfdash/config.go
```diff
@@ -239,12 +233,15 @@ type periodic struct {
...
-func getProwConfig() (Jobs, error) {
+func getProwConfig(configPaths []string) (Jobs, error) {
 	fmt.Fprintf(os.Stderr, "Fetching prow config from GitHub...\n")
 	jobs := Jobs{}
-	yamlLinks := []string{
-		"https://raw.githubusercontent.com/kubernetes/test-infra/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml",
-		"https://raw.githubusercontent.com/kubernetes/test-infra/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml",
+	yamlLinks := []string{}
+	for _, configPath := range configPaths {
+		//Currently perfdash supports only urls.
```
wojtek-t: nit: whitespace after "//"

perfdash/config.go
```diff
@@ -239,12 +233,15 @@ type periodic struct {
...
+		if strings.HasPrefix(configPath, "http") {
```
wojtek-t: Maybe we should check extension...

perfdash/config.go
krzysied: Done.

perfdash/config.go
krzysied: Done.

PR #555, opened 2019-06-04
Use strings.Replace instead strings.ReplaceAll to keep compatibility with go1.11.5

### LINKED ISSUE #554: strings.ReplaceAll causes pull-kubernetes-kubemark-e2e-gce-big to fail on 1.13
This was reported on slack: https://kubernetes.slack.com/archives/C09QZTRH7/p1559596507001600

strings.ReplaceAll was [added in go1.12](https://golang.org/doc/go1.12#strings), but I suspect we still build clusterloader2 for stable versions using older version of golang. 

/assign @oxddr
/cc @lbernail

oxddr:
ReplaceAll function has been added in go1.12. We still compile clusterloader2 against go1.11 to run tests against stable Kubernetes versions.

Until we figure out how to compile clusterloader2 with newer version of go and Kubernetes with the older one, we should test against go1.11.

/assign @mborsz 
/fixes #554

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ language: go
 
 matrix:
     include:
-        - go: 1.12
+        - go: 1.11
 
 install:
 # Create and move build under the go path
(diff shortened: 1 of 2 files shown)


mborsz:
I don't have approval here. /assign @mm4tt

PR #557, opened 2019-06-04
Recording rule for Network Programming Latency

oxddr:
This is based on work by @mm4tt in https://github.com/kubernetes/perf-tests/pull/430.


clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubeProxy.yaml
```diff
@@ -0,0 +1,27 @@
...
+  name: kube-proxy
+  namespace: monitoring
+spec:
+  endpoints:
+    # We modify interval depending on the cluster size to avoid collecting too many samples for
+    # large clusters. Also because the tests run longer in bigger clusters we don't need to
+    # collect them as often as in smaller clusters. We use 30s interval for small clusters
+    # (# nodes <= 1000) and 1min interval in big clusters (# nodes > 1000)
+    # TODO(mm4tt): Once we prove the interval works in big clusters, simplify the expression.
+    {{if $PROMETHEUS_SCRAPE_KUBE_PROXY}}
```
mm4tt: I wonder it anything won't fail if it's disabled, e.g. if there isn't a validation that endpoints list shouldn't be empty.
We'll see in the presubmit :)

mm4tt:
You'll also need a kube-proxy service that was temporarily removed in https://github.com/kubernetes/perf-tests/pull/528

https://github.com/kubernetes/perf-tests/blob/d2d330760c0061bd80d3987a03e5c5f34025f727/clusterloader2/pkg/prometheus/manifests/default/kube-proxy-service.yaml

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubeProxy.yaml
oxddr: Good point. Alternatively I think we can put whole object inside this if-statement. 

I'd much rather prefer to enable this is scalability tests without changing code of clusterloader2.

oxddr:
@mm4tt PTAL

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -19,17 +19,23 @@ package prometheus
...
 	"encoding/json"
 	"fmt"
-	"github.com/spf13/pflag"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"os/exec"
 	"time"
 
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/wait"
+
```
mm4tt: NIT: Remove blank line?

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -19,17 +19,23 @@ package prometheus
...
 	"k8s.io/klog"
 )
 
+const (
+	kubeProxyManifest = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/default/experimental/prometheus-serviceMonitorKubeProxy.yaml"
+)
+
 var (
 	shouldSnapshotPrometheusDisk = pflag.Bool("experimental-gcp-snapshot-prometheus-disk", false, "(experimental, provider=gce|gke only) whether to snapshot Prometheus disk before Prometheus stack is torn down")
+	enableKubeProxyScrapping     = pflag.Bool("enable-kube-proxy-scrapping", false, "whether to scrape kube-proxy")
```
mm4tt: Two things:

1. I don't like, having it in the "experimental" file. It was already enabled in 5K and worked, it's most likely gonna work again. Having it in experimental doesn't bring us any gain, and on the other hand we will have to move it away from here once we confirm it works. Let's move it to the prometheus.go file

2. I don't like having it flag gated. In my opinion, whether we monitor kube proxy or not is specific to the test scenario itself, not the the mechanics of clusterloader2. Similar to whether we monitor etcd or whether we run some measurement or not.  I believe it would be better if we introduced a variable, e.g. PROMETHEUS_SCRAPE_KUBEPROXY, that user can set in his test scenario (similar to already existing PROMETHEUS_SCRAPE_ETCD). I'm aware of the problem of failures when applying empty manifests, but I think it could be solved some other way. E.g. by checking if manifests is empty before sending request via k8s-client. It would be a more generic solution, as it would allow us to gate other features in future without introducing any new code.

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -19,17 +19,23 @@ package prometheus
...
 
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/wait"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 )
 
+const (
+	kubeProxyManifest = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/default/experimental/prometheus-serviceMonitorKubeProxy.yaml"
```
mm4tt: To make it safe and make sure it doesn't introduce any change when disabled we should gate applying both serviceMonitor and service. Creating such a big service has a performance implications.

oxddr:
PTAL

mm4tt:
Thanks!

/lgtm
/approve

Issue #498, opened 2019-04-12
Testing api responsiveness prometheus measurement


krzysied:
- Comparing api responsiveness prometheus measurement vs api responsiveness measurement.
  - Api responsiveness prometheus measurement should be enabled in density and load test.
  - Api responsiveness prometheus measurement should not cause test to fail.
- After tests are done api responsiveness prometheus measurement should be changed to test-failing.

oxddr:
It seems that Prometheus-based metrics has been enable for quite some time. What are criteria for switching measurements of API calls latency to Prometheus exclusively?

oxddr:
After offline discussion with @krzysied, it seems that there is discrepancy between old (apiserver-based) and new (Prometheus-based) calculations and the newer one being more strict. It seems however that the Prometheus-based approach is closer to definition of SLO.

wojtek-t:
They are "kind-of" more strict, they are different. But yes - they reflect the SLO and we want to start using them.

oxddr:
The old measurement has been replaced by the one using newer metric. 

/close

PR #567, opened 2019-06-12
Enable reporting SLO violations from Prometheus API responsiveness.

oxddr:
It's enable via flag, so we can slowly ramp it up in tests.

/assign @krzysied


oxddr:
ref: https://github.com/kubernetes/perf-tests/issues/498

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -21,6 +21,7 @@ TODO(krzysied): This measurement should replace api_responsiveness.go.
 package slos
 
 import (
+	"flag"
```
krzysied: nit: "github.com/spf13/pflag" is used by clusterloader. The "flag" will work, but I would suggest using "pflag" for consistency.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -51,6 +53,10 @@ const (
 	latencyWindowSize = 5 * time.Minute
 )
 
+var (
+	enableViolations = pflag.Bool("enable-violations-api-responsiveness-prometheus", false, "Whether to enable violations in the API call responsivness measurment done via Prometheus")
```
mm4tt: s/measurment/measurement/

verify/verify-flags/known-flags.txt
```diff
@@ -24,3 +24,4 @@ ping-server-address
 ping-server-bind-address
 ping-sleep-duration
 master-internal-ip
+enable-violations-api-responsiveness-prometheus
```
mm4tt: NIT, can be done in other PR, could you sort the entries alphabetically in this file?

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
mm4tt: Open question, are we going to have a flag per each prometheus measurement? 

Have you considered moving this to the measurment parameters, e.g. something like
```
  - Identifier: APIResponsivenessPrometheus
    Params:
      action: start
      failTestOnViolation: true|false
```

This could be implemented in a generic way for each prometheus based measurment, wdyt?

oxddr: Good questions. Assuming we want to gate somehow all major changes to clusterloader2, I'd say yes. 

I like your idea or having this in the configuration file. However I think more thought needs to be put into that (e.g. an open question is whether it will support well gating (overrides?)) and I'd rather have it in following PR. 

Let's submit this one and start to enable the violations from this measurement and in the meanwhile let's figure out something more elegant and scalable in terms of num of measurements.

oxddr:
/retest
It builds locally...

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
mm4tt: Removing a flag is a painful, multi-step process.  I'd prefer to avoid it, especially given that I think we're not yet in a place where we can enable these violations, right? 
We'll first need to get a few runs of this reported into perf-dash to see and understand what are the differences. So I guess we should focus on implementing the perf-dash integration first. 

You can also run the tests manually with the violations enabled if you want to check if/which (resource, verb, scope) triples would fail.

oxddr: 1. This is non-production code, so removing flags is not a big deal (two PR at most). 
2. Differences are mostly due to current measurement being less strict (further away from SLO definition). But I agree it make sense to integrate perf-dash first. I assumed this would work out-of-the-box, but it turned out it's not the case. 
3. I'll close this PR and run those tests manually. I'll follow up with another PR that would try to solve this problem in a generic way as you've proposed.

PR #568, opened 2019-06-12
Extract common code for Prometheus-based measurements.

oxddr:
By no means this is the final form of how the library for Prometheus-based measurements should look like. I've simply extracted common code, that I am using to create measurements for networking programming latency. I'd rather have an iterative approach here. 

Also changed small bits here and there.

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -30,7 +30,7 @@ import (
 	"github.com/prometheus/common/model"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
-	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -236,15 +196,12 @@ func getMetricKey(resource, subresource, verb, scope string) string {
...
-	isListCall := (call.Verb == "LIST")
-	isClusterScopedCall := (call.Scope == "cluster")
-	latencyThreshold := apiCallLatencyThreshold
-	if isListCall {
-		latencyThreshold = apiListCallLatencyThreshold
-		if isClusterScopedCall {
-			latencyThreshold = apiClusterScopeListCallThreshold
-		}
+func getSLOThreshold(verb, scope string) time.Duration {
+	if verb != "LIST" {
```
krzysied: Assuming that there is "return" used, "else" is not needed?

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -49,89 +49,49 @@ const (
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
 	latencyWindowSize = 5 * time.Minute
+
+	// number of metrics with highest latency to print
```
krzysied: nit: Comment should start with capital letter and end with ".".

PR #570, opened 2019-06-13
Add support for StringSliceEnvVar flag.

mm4tt:
It will be needed to support multiple masters in cl2.

### DIFF
--- a/clusterloader2/pkg/flags/types.go
+++ b/clusterloader2/pkg/flags/types.go
@@ -18,6 +18,7 @@ package flags
 
 import (
 	"strconv"
+	"strings"
 
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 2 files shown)


mm4tt:
@oxddr, could you please lgtm again? I needed to fix gofmt issue

clusterloader2/pkg/flags/types.go
```diff
@@ -57,6 +58,36 @@ func (s *stringFlagFunc) Type() string {
...
+func (*stringSliceFlagFunc) String() string {
+	return "false"
+}
+
+// Set handles flag value setting.
+func (s *stringSliceFlagFunc) Set(val string) error {
+	if val != "" {
+		*s.valPtr = strings.Split(val, ",")
+	} else {
+		*s.valPtr = nil
```
krzysied: Do we need this `else`? Isn't `nil` a default value?

mm4tt: The flag supports setting both via env var and via flag. So let's say it's MY_STRING_LIST env or --my_string_list flag.

Let's say the MY_STRING_LIST is set to "1,2,3" in some upstream code that we cannot really change, but we want to set it to empty list via flag override. This else clause allows it, I can empty the list by setting --my_string_list=""

oxddr:
and again :D 

/lgtm

clusterloader2/pkg/flags/types.go
krzysied: Ok. :+1:

PR #574, opened 2019-06-19
Create an interface for querying Prometheus from measurements.

oxddr:
This refactoring simplifies unit testing of Prometheus-based measurements.

/assign @mborsz @krzysied 
/kind cleanup

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	"github.com/prometheus/common/model"
-	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -91,7 +89,7 @@ func (a *apiResponsivenessGatherer) Gather(config *measurement.MeasurementConfig
 			if isBad {
 				prefix = "WARNING "
 			}
-			klog.Infof("%s: %vTop latency metric: %+v; threshold: %v", a, prefix, apiCall, sloThreshold)
+			klog.Infof("%s: %vTop latency metric: %+v; threshold: %v", apiResponsivenessMeasurementName, prefix, apiCall, sloThreshold)
```
krzysied: I prefer using `a` instead of `apiResponsivenessMeasurementName` and add `String()` method to `apiResponsivenessGatherer`. This will allow us to have name for a single instance of measurement. Currently instance name is same as measurement name, however I can imagine we want to change it in the future.

oxddr:
ref #575

oxddr:
PTAL

PR #578, opened 2019-06-27
Update clusterloader2/vendor

mborsz:
The goal is to pickup https://github.com/kubernetes/kubernetes/pull/78983

Two commands:
$ govendor fetch k8s.io/kubernetes/pkg/ssh
$ govendor fetch k8s.io/klog@19f20ccff689fcb6e12c00586180dd012481281c # To revert klog update to the old version


wojtek-t:
Do we know that this will still work with go1.11 (which we're using in clusterloader due to older branches?)

mborsz:
I think it does compile with go1.11: after https://github.com/kubernetes/perf-tests/pull/555 travis is using go1.11 and logs say:
```
The command go build (...) k8s.io/perf-tests/clusterloader2/cmd (...) exited with 0
```

PR #579, opened 2019-06-27
ClusterLoader - Adding service_creation_latency measurement

krzysied:
Adding measurement that measures service creation latency.

### DIFF
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -1,0 +1,305 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/perf-tests/clusterloader2/pkg/execservice"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/checker"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/workerqueue"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	serviceCreationLatencyName           = "ServiceCreationLatency"
+	serviceCreationLatencyWorkers        = 10
+	defaultServiceCreationLatencyTimeout = 10 * time.Minute
+	defaultCheckInterval                 = 10 * time.Second
+	pingBackoff                          = 1 * time.Second
+	pingChecks                           = 10
+
+	creatingPhase     = "creating"
+	ipAssigningPhase  = "ipAssigning"
+	reachabilityPhase = "reachability"
+)
+
+func init() {
+	measurement.Register(serviceCreationLatencyName, createServiceCreationLatencyMeasurement)
+}
+
+func createServiceCreationLatencyMeasurement() measurement.Measurement {
+	return &serviceCreationLatencyMeasurement{
+		selector:      measurementutil.NewObjectSelector(),
+		queue:         workerqueue.NewWorkerQueue(serviceCreationLatencyWorkers),
+		creationTimes: measurementutil.NewObjectTransitionTimes(serviceCreationLatencyName),
+		pingCheckers:  checker.NewCheckerMap(),
+	}
+}
+
+type serviceCreationLatencyMeasurement struct {
+	selector      *measurementutil.ObjectSelector
+	waitTimeout   time.Duration
+	stopCh        chan struct{}
+	isRunning     bool
+	queue         workerqueue.Interface
+	client        clientset.Interface
+	creationTimes *measurementutil.ObjectTransitionTimes
+	pingCheckers  checker.CheckerMap
+}
+
+// Execute executes service startup latency measurement actions.
+// Services can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+// "start" action starts observation of the services.
+// "waitForReady" waits until all services are reachable.
+// "gather" returns service created latency summary.
+// This measurement only works for services with LoadBalancer type.
+func (s *serviceCreationLatencyMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	s.client = config.ClusterFramework.GetClientSets().GetClient()
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+
+	switch action {
+	case "start":
+		if s.selector.Parse(config.Params); err != nil {
+			return nil, err
+		}
+		s.waitTimeout, err = util.GetDurationOrDefault(config.Params, "waitTimeout", defaultServiceCreationLatencyTimeout)
+		if err != nil {
+			return nil, err
+		}
+		return nil, s.start()
+	case "waitForReady":
+		return nil, s.waitForReady()
+	case "gather":
+		return s.gather()
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+// Dispose cleans up after the measurement.
+func (s *serviceCreationLatencyMeasurement) Dispose() {
+	if s.isRunning {
+		s.isRunning = false
+		close(s.stopCh)
+	}
+	s.queue.Stop()
+	s.pingCheckers.Dispose()
+}
+
+// String returns a string representation of the metric.
+func (s *serviceCreationLatencyMeasurement) String() string {
+	return serviceCreationLatencyName + ": " + s.selector.String()
+}
+
+func (s *serviceCreationLatencyMeasurement) start() error {
+	if s.isRunning {
+		klog.Infof("%s: wait for controlled pods measurement already running", s)
+		return nil
+	}
+	klog.Infof("%s: starting wait for controlled pods measurement...", s)
+
+	s.isRunning = true
+	s.stopCh = make(chan struct{})
+
+	i := informer.NewInformer(
+		s.client,
+		"services",
+		s.selector,
+		func(oldObj, newObj interface{}) {
+			f := func() {
+				s.handleObject(oldObj, newObj)
+			}
+			s.queue.Add(&f)
+		},
+	)
+	return informer.StartAndSync(i, s.stopCh, informerSyncTimeout)
+}
+
+func (s *serviceCreationLatencyMeasurement) waitForReady() error {
+	return wait.Poll(defaultCheckInterval, s.waitTimeout, func() (bool, error) {
+		reachable := s.creationTimes.Count(reachabilityPhase)
+		ipAssigned := s.creationTimes.Count(ipAssigningPhase)
+		created := s.creationTimes.Count(creatingPhase)
+		klog.Infof("%s: %d created, %d ipAssigned, %d reachable", s, created, ipAssigned, reachable)
+		return created == reachable, nil
+	})
+}
+
+func (s *serviceCreationLatencyMeasurement) gather() ([]measurement.Summary, error) {
+	klog.Infof("%s: gathering service created latency measurement...", s)
+	if !s.isRunning {
+		return nil, fmt.Errorf("metric %s has not been started", s)
+	}
+
+	serviceCreationLatency := s.creationTimes.CalculateTransitionsLatency(map[string]measurementutil.Transition{
+		"create_to_assigned": {
+			From: creatingPhase,
+			To:   ipAssigningPhase,
+		},
+		"assigned_to_available": {
+			From: ipAssigningPhase,
+			To:   reachabilityPhase,
+		},
+		"create_to_available": {
+			From: creatingPhase,
+			To:   reachabilityPhase,
+		},
+	})
+
+	content, err := util.PrettyPrintJSON(measurementutil.LatencyMapToPerfData(serviceCreationLatency))
+	if err != nil {
+		return nil, err
+	}
+	summary := measurement.CreateSummary(serviceCreationLatencyName, "json", content)
+	return []measurement.Summary{summary}, nil
+}
+
+func (s *serviceCreationLatencyMeasurement) handleObject(oldObj, newObj interface{}) {
+	var oldService *corev1.Service
+	var newService *corev1.Service
+	var ok bool
+	oldService, ok = oldObj.(*corev1.Service)
+	if oldObj != nil && !ok {
+		klog.Errorf("%s: uncastable old object: %v", s, oldObj)
+		return
+	}
+	newService, ok = newObj.(*corev1.Service)
+	if newObj != nil && !ok {
+		klog.Errorf("%s: uncastable new object: %v", s, newObj)
+		return
+	}
+	if isEqual := oldService != nil &&
+		newService != nil &&
+		equality.Semantic.DeepEqual(oldService.Spec, newService.Spec) &&
+		equality.Semantic.DeepEqual(oldService.Status, newService.Status); isEqual {
+		return
+	}
+
+	// TODO(#680): Make it thread-safe.
+	if !s.isRunning {
+		return
+	}
+	if newObj == nil {
+		if err := s.deleteObject(oldService); err != nil {
+			klog.Errorf("%s: delete checker error: %v", s, err)
+		}
+		return
+	}
+	if err := s.updateObject(newService); err != nil {
+		klog.Errorf("%s: create checker error: %v", s, err)
+	}
+}
+
+func (s *serviceCreationLatencyMeasurement) deleteObject(svc *corev1.Service) error {
+	key, err := runtimeobjects.CreateMetaNamespaceKey(svc)
+	if err != nil {
+		return fmt.Errorf("meta key created error: %v", err)
+	}
+	s.pingCheckers.DeleteAndStop(key)
+	return nil
+}
+
+func (s *serviceCreationLatencyMeasurement) updateObject(svc *corev1.Service) error {
+	// This measurement only works for services with LoadBalancer type.
+	if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		return nil
+	}
+	key, err := runtimeobjects.CreateMetaNamespaceKey(svc)
+	if err != nil {
+		return fmt.Errorf("meta key created error: %v", err)
+	}
+	if _, exists := s.creationTimes.Get(key, creatingPhase); !exists {
+		s.creationTimes.Set(key, creatingPhase, svc.CreationTimestamp.Time)
+	}
+	if len(svc.Status.LoadBalancer.Ingress) < 1 {
+		return nil
+	}
+	if _, exists := s.creationTimes.Get(key, ipAssigningPhase); exists {
+		return nil
+	}
+	s.creationTimes.Set(key, ipAssigningPhase, time.Now())
+
+	pc := &pingChecker{
+		callerName:    s.String(),
+		svc:           svc,
+		creationTimes: s.creationTimes,
+		stopCh:        make(chan struct{}),
+	}
+	pc.run()
+	s.pingCheckers.Add(key, pc)
+
+	return nil
+}
+
+type pingChecker struct {
+	callerName    string
+	svc           *corev1.Service
+	creationTimes *measurementutil.ObjectTransitionTimes
+	stopCh        chan struct{}
+}
+
+func (p *pingChecker) run() {
+	key, err := runtimeobjects.CreateMetaNamespaceKey(p.svc)
+	if err != nil {
+		klog.Errorf("%s: meta key created error: %v", p.callerName, err)
+		return
+	}
+	success := 0
+	for {
+		select {
+		case <-p.stopCh:
+			return
+		default:
+			if _, exists := p.creationTimes.Get(key, reachabilityPhase); exists {
+				return
+			}
+			// TODO(#679): Current implementation handles only load balancers.
+			// TODO(#685): Make ping checks less communication heavy.
+			_, err := execservice.RunCommand(
+				fmt.Sprintf("curl %s:%d", p.svc.Status.LoadBalancer.Ingress[0].IP, p.svc.Spec.Ports[0].Port))
+			if err != nil {
+				success = 0
+				time.Sleep(pingBackoff)
+				continue
+			}
+			success++
+			if success == pingChecks {
+				p.creationTimes.Set(key, reachabilityPhase, time.Now())
+			}
+		}
+	}
+}
+
+func (p *pingChecker) Stop() {
+	close(p.stopCh)
+}


clusterloader2/pkg/measurement/common/simple/wait_for_services.go
```diff
@@ -0,0 +1,147 @@
...
+	for {
+		select {
+		case <-time.After(w.waitTimeout):
+			return fmt.Errorf("timeout while waiting for %d services to be available in namespace '%v' with labels '%v' and fields '%v' - only %d found running",
+				desiredCount, w.namespace, w.labelSelector, w.fieldSelector, available)
+		case <-time.After(defaultWaitForServicesInterval):
+			services := w.store.List()
+			created = len(services)
+			available = 0
+			// TODO: Measurement waits only for load balancers.
```
prameshj: Can we add another LoadBalancer service specific measurement that will run "resp, err := http.Get("http://<LoadBalancerIP>:servicePort")" and check that err is nil and resp is valid?

wojtek-t: You mean probe that service?

There is a separate work trying to start probes and I would like to have those consistent.

But my main question is: is that probe going to measure:
(a) whether the service is healthy over longer period of time?
or
(b) ping the service as long as it doesn't respond to get the latency of when we get the first successful response

clusterloader2/pkg/measurement/common/simple/wait_for_services.go
```diff
@@ -0,0 +1,147 @@
...
+			// It should be extended to all kinds of services.
+			for i := 0; i < len(services); i++ {
+				if len(services[i].Status.LoadBalancer.Ingress) > 0 {
```
wojtek-t: @prameshj @bowei - one thing that I'm not sure about and would need your clarification is:
- does the fact that LoadBalancerStatus has the Ingress set already means that the LB is programmed?
- or maybe (which is what I thought) it means that we chose the ip address, and we started programming it

krzysied:
I've re-implemented the measurement. Currently it measures time of service creation: creation->IP assigned->available.
@wojtek-t PTAL.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,538 @@
+/*
```
wojtek-t: This confirms my initial claim months ago, that this directory shouldn't have been called "simple/" :D

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,538 @@
...
+	})
+	defer timeoutTimer.Stop()
+	if !cache.WaitForCacheSync(timeoutCh, s.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+
+	return nil
+}
+
+func (s *serviceCreationLatencyMeasurement) setup() error {
```
wojtek-t: Personally, I wouldn't do that part of this measurement.
Similarly as we create prometheus stack, I would natively add to CL an ability to setup our "maintainance" namespace and start this pod there. [If we agree on that, please move it to a dedicate PR.]

@mm4tt - for thoughts

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,538 @@
...
+	return measurementutil.LatencyMetric{Perc50: perc50, Perc90: perc90, Perc99: perc99}
+}
+
+type serviceCreationLatency struct {
+	CreateToAssignedLatency    measurementutil.LatencyMetric `json:"createToAssignedLatency"`
+	AssignedToAvailableLatency measurementutil.LatencyMetric `json:"assignedToAvailableLatency"`
+	CreateToAvailableLatency   measurementutil.LatencyMetric `json:"createToAvailableLatency"`
+}
+
+func podStartupLatencyToPerfData(latency *serviceCreationLatency) *measurementutil.PerfData {
```
wojtek-t: It shouldn't be called podStartup... :)

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,538 @@
...
+	Latency time.Duration
+}
+
+type serviceCreationLatencySlice []serviceCreationLatencyData
+
+func (a serviceCreationLatencySlice) Len() int           { return len(a) }
+func (a serviceCreationLatencySlice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a serviceCreationLatencySlice) Less(i, j int) bool { return a[i].Latency < a[j].Latency }
+
+func extractLatencyMetrics(latencies []serviceCreationLatencyData) measurementutil.LatencyMetric {
```
wojtek-t: The exact same function already exists in pod_startup_slo - it should be extracted somewhere.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,538 @@
...
+func podStartupLatencyToPerfData(latency *serviceCreationLatency) *measurementutil.PerfData {
+	const nsToMsRatio = 1000000.0
+	perfData := &measurementutil.PerfData{Version: "1.0"}
+	perfData.DataItems = append(perfData.DataItems, latency.CreateToAssignedLatency.ToPerfData("create_to_assigned", nsToMsRatio))
+	perfData.DataItems = append(perfData.DataItems, latency.AssignedToAvailableLatency.ToPerfData("assigned_to_available", nsToMsRatio))
+	perfData.DataItems = append(perfData.DataItems, latency.CreateToAvailableLatency.ToPerfData("create_to_available", nsToMsRatio))
+	return perfData
+}
+
+func (s *serviceCreationLatencyMeasurement) newInformer() (cache.SharedInformer, error) {
```
wojtek-t: It's pretty clear already that we may need different of those.
Instead of adding it here, you should create a file (probably something like pkg/framework/informers.go) and add helpers to create informers there (also move the existing one for pods) - majority of code is shared between those and can be extracted.

clusterloader2/pkg/measurement/util/latency_metric.go
```diff
@@ -54,3 +54,18 @@ func (metric *LatencyMetric) VerifyThreshold(threshold *LatencyMetric) error {
 	}
 	return nil
 }
+
+// ToPerfData converts latency metric to PerfData.
+func (metric *LatencyMetric) ToPerfData(name string, ratio float64) DataItem {
```
wojtek-t: This and extractLatencyMetrics to a separate PR please :)

krzysied: Moved to #623.

clusterloader2/pkg/measurement/common/simple/wait_for_services.go
prameshj: Option b). Goal is to measure the latency from issuing the service creation command to having the load balancer respond.

clusterloader2/pkg/measurement/common/simple/wait_for_services.go
prameshj: Since the IP has been populated in status, the loadbalancer has been setup at this point. I have seen there being some delay after this in actually getting a response when pinging the LB ip. Hence the need to measure this latency.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
mm4tt: I second Wojtek's idea, this "maintenance" namespace and pod(s) could be reused for other stuff in future. And this measurement may require the feature to be on, similarly to the ApiCallLatencyPrometheus measurement requiring Prometheus to be enabled.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Fixed.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+
+// String returns a string representation of the metric.
+func (*serviceCreationLatencyMeasurement) String() string {
+	return serviceCreationLatencyName
+}
+
+type serviceCreationTimes struct {
+	creation  time.Time
+	assigned  time.Time
+	available time.Time
```
wojtek-t: reachable?

[In general naming is strange here - you have:
- creation - noun
- assigned/available - not nouns
Should be unified.]

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+	timeoutCh := make(chan struct{})
+	timeoutTimer := time.AfterFunc(informerSyncTimeout, func() {
+		close(timeoutCh)
+	})
+	defer timeoutTimer.Stop()
+	if !cache.WaitForCacheSync(timeoutCh, s.informer.HasSynced) {
+		return fmt.Errorf("timed out waiting for caches to sync")
+	}
+
+	return nil
```
wojtek-t: This whole function looks so boiler-platy - can't we have some "meta-measurement" where we will do this once, instead of separately for every single measurement that needs informer on some resource?

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+func (s *serviceCreationLatencyMeasurement) waitTillReady() error {
+	for {
+		created, assigned, available := 0, 0, 0
+		select {
+		case <-time.After(defaultWaitTimeout):
+			klog.Errorf("%s: waiting for services timeout")
```
wojtek-t: Don't log if you're returning an error.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+	queue            workerqueue.Interface
+	informer         cache.SharedInformer
+	clusterFramework *framework.Framework
+
+	lock          sync.Mutex
+	creationTimes map[string]*serviceCreationTimes
+	pingCheckers  map[string]*pingChecker
+}
+
+// Execute waits until services are ready.
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+		s.fieldSelector, err = util.GetStringOrDefault(config.Params, "fieldSelector", "")
+		if err != nil {
+			return nil, err
+		}
+		s.waitTimeout, err = util.GetDurationOrDefault(config.Params, "waitTimeout", defaultServiceCreationLatencyTimeout)
+		if err != nil {
+			return nil, err
+		}
+		return nil, s.start()
+	case "waitTillReady":
```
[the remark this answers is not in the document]
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+	defer s.lock.Unlock()
+	for _, pc := range s.pingCheckers {
+		pc.stop()
+	}
+	s.pingCheckers = make(map[string]*pingChecker)
+}
+
+// String returns a string representation of the metric.
+func (*serviceCreationLatencyMeasurement) String() string {
+	return serviceCreationLatencyName
```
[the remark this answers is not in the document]
krzysied: Added selector string.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+type serviceCreationTimes struct {
+	creation  time.Time
+	assigned  time.Time
```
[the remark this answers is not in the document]
krzysied: Changed to: created, ipAssigned, reachable

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/simple/service_creation_latency.go
krzysied: I want to have failure logged when timeout happens - it very convenient for debugging.
Returning error is required to change measurement status.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,413 @@
...
+type serviceCreationLatencyMeasurement struct {
+	namespace        string
+	labelSelector    string
+	fieldSelector    string
+	waitTimeout      time.Duration
+	stopCh           chan struct{}
+	isRunning        bool
+	selectorsString  string
+	queue            workerqueue.Interface
+	informer         cache.SharedInformer
```
wojtek-t: You removed informer from other measurements - I don't think this one is different.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,413 @@
...
+	clusterFramework *framework.Framework
```
wojtek-t: I don't think we need clusterFramework - client is all you need.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+type serviceCreationLatencyMeasurement struct {
+	namespace        string
+	labelSelector    string
+	fieldSelector    string
+	waitTimeout      time.Duration
+	stopCh           chan struct{}
+	isRunning        bool
+	selectorsString  string
```
wojtek-t: Do we really need to make it part of measurement?

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,413 @@
...
+	queue            workerqueue.Interface
+	informer         cache.SharedInformer
+	clusterFramework *framework.Framework
+
+	lock          sync.Mutex
+	creationTimes map[string]*serviceCreationTimes
```
wojtek-t: I like it much more than those 5 maps:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go#L66

if you could fix that in a separate PR, it would be great.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: IIUC I can have CreateSelectorsString called in String method. However assuming that the selector will not change after "start" action is executed, it's better to store the value.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,411 @@
...
+		queue:         workerqueue.NewWorkerQueue(serviceCreationLatencyWorkers),
+		creationTimes: make(map[string]*serviceCreationTimes),
+		pingCheckers:  make(map[string]*pingChecker),
+	}
+}
+
+type serviceCreationLatencyMeasurement struct {
+	namespace       string
+	labelSelector   string
+	fieldSelector   string
```
wojtek-t: There are already couple measurements, that are using both namespace, labelSelector and fieldSelector.
Can we opaque those 3 into some type (like objectSelector?) and add a common parsing method for those to avoid that boilerplate everywhere.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,411 @@
...
+// Dispose cleans up after the measurement.
+func (s *serviceCreationLatencyMeasurement) Dispose() {
+	if s.isRunning {
+		s.isRunning = false
+		close(s.stopCh)
+	}
+	s.queue.Stop()
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	for _, pc := range s.pingCheckers {
```
wojtek-t: This is also a common pattern.
I propose that we do the following:
- introduce Checker interface, that will only have a Stop() method
- on top of that we introduce type CheckersMap, that will have methods: NewCheckersMap and StopAllCheckers()

I would like to unify it further later, but what I mentioned above is simple and we can do that very quickly and will already introduce a common pattern for the future.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+		assignedToAvailableLatency = append(assignedToAvailableLatency, serviceCreationLatencyData{
+			Key:     key,
+			Latency: ct.reachable.Sub(ct.ipAssigned),
+		})
+		createToAvailableLatency = append(createToAvailableLatency, serviceCreationLatencyData{
+			Key:     key,
+			Latency: ct.reachable.Sub(ct.created),
+		})
+	}
```
wojtek-t: This whole stuff is boilerplate - I've been already thinking how to make it cleaner and avoid doing the same stuff in every single measurement, and I think that I have some idea now.

So here is the idea:
"creationTimes" (and similar things in other measurements like podStartupTime) should be of the type of:
map[string]time.Time [means phase->timestamp]
[Then this type is common for every single measurement, only names of phases differ]

- in the individual measurement you only define names of phases and how to compute timestamps of them for individual object

- then you have a generic "Compute" function (well, we need better name) as following:
func (*objectTransitionTimes) Compute(map[string]pair<string,string>) map[string]measurementutil.LatencyMetric {}

that basically computes for every given string->pair<phase1,phase2>
- first list of latencies of all objects between phase1->phase2
- sorts them
- prints latencies
- and computes percentiles

Then you basically write this logic once and will be able to reuse it in every measurement producing latency.
[This will also introduce a common pattern for writing such measurements, which I'm really looking for]

wojtek-t: You can first do it for pod_startup_latency to get framework (in a separate PR) and then use it here.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,411 @@
...
+	success := 0
+	for {
+		select {
+		case <-p.stopCh:
+			return
+		default:
+			if !p.creationTimes.reachable.IsZero() {
+				return
+			}
+			if len(p.svc.Status.LoadBalancer.Ingress) < 1 || len(p.svc.Spec.Ports) < 1 {
```
wojtek-t: This isn't needed - you're already checking that in updateObjectLocked.

The contract of run() function should be that it takes only services that are pingable.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,411 @@
...
+			stopCh:        make(chan struct{}),
+		}
+		go s.pingCheckers[key].run()
+	}
+	return nil
+}
+
+type pingChecker struct {
+	caller        string
+	key           string
```
wojtek-t: Why do we need key? We can compute namespace/name when it's really neded (it's exaclty one place for looging purpose).

The smaller objects are, the easier to understand it is.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: https://github.com/kubernetes/perf-tests/pull/654 created.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,366 @@
...
+func (s *serviceCreationLatencyMeasurement) deleteObjectLocked(svc *corev1.Service) error {
+	key, err := runtimeobjects.CreateMetaNamespaceKey(svc)
+	if err != nil {
+		return fmt.Errorf("meta key created error: %v", err)
+	}
+	if pc, exists := s.pingCheckers[key]; exists {
+		pc.stop()
+		delete(s.pingCheckers, key)
+	}
+	s.creationTimes.DeleteKey(key)
```
wojtek-t: I don't understand it - I don't think we should ever remove anything from statistics.

clusterloader2/pkg/measurement/util/phase_latency.go
```diff
@@ -65,6 +65,24 @@ func (o *ObjectTransitionTimes) Get(key, phase string) (time.Time, bool) {
 	return time.Time{}, false
 }
 
+// DeleteKey removes all entries for given key.
+func (o *ObjectTransitionTimes) DeleteKey(key string) {
```
wojtek-t: I don't think this function is needed - I don't see any case where we should remove statistics of something that we're measuring.

clusterloader2/pkg/measurement/util/phase_latency.go
```diff
@@ -65,6 +65,24 @@ func (o *ObjectTransitionTimes) Get(key, phase string) (time.Time, bool) {
...
+	if _, exists := o.times[key]; exists {
+		delete(o.times, key)
+	}
+}
+
+// Count returns number of key having given phase entry.
+func (o *ObjectTransitionTimes) Count(phase string) int {
```
wojtek-t: I was wondering if we can avoid that by making things differently, but didn't come up with anything reasonable enough :(

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,366 @@
...
+
+	content, err := util.PrettyPrintJSON(measurementutil.LatencyMapToPerfData(serviceCreationLatency))
+	if err != nil {
+		return nil, err
+	}
+	summary := measurement.CreateSummary(serviceCreationLatencyName, "json", content)
+	return []measurement.Summary{summary}, err
+}
+
+func (s *serviceCreationLatencyMeasurement) printLatencies(latencies []measurementutil.LatencyData, header string) {
```
wojtek-t: No longer needed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,366 @@
...
+	metrics := measurementutil.NewLatencyMetric(latencies)
+	index := len(latencies) - 100
+	if index < 0 {
+		index = 0
+	}
+	klog.Infof("%s: %d %s: %v", s, len(latencies)-index, header, latencies[index:])
+	klog.Infof("%s: perc50: %v, perc90: %v, perc99: %v", s, metrics.Perc50, metrics.Perc90, metrics.Perc99)
+}
+
+type serviceCreationLatencyData struct {
```
wojtek-t: This one too.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,366 @@
...
+		s.creationTimes.Set(key, creatingPhase, time.Now())
+	}
+	if len(svc.Status.LoadBalancer.Ingress) < 1 {
+		return nil
+	}
+	if _, exists := s.creationTimes.Get(key, ipAssigningPhase); !exists {
+		s.creationTimes.Set(key, ipAssigningPhase, time.Now())
+		klog.Infof("%s: %s ipAssigned", s, key)
+	}
+	if _, exists := s.pingCheckers[key]; !exists {
```
wojtek-t: I think you don't need this if - this should happen exactly when you first observe ingress address.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Logging moved from pingChecker to service update handler.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/util/phase_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Removed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Removed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: In this implementation this line is the definition of "first observed ingress address".

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,345 @@
...
+		creationTimes: measurementutil.NewObjectTransitionTimes(serviceCreationLatencyName),
+		pingCheckers:  make(map[string]*pingChecker),
+	}
+}
+
+type serviceCreationLatencyMeasurement struct {
+	namespace       string
+	labelSelector   string
+	fieldSelector   string
+	waitTimeout     time.Duration
```
wojtek-t: Doesn't seem to be used anywhere.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,345 @@
...
+	if newObj != nil && !ok {
+		klog.Errorf("%s: uncastable new object: %v", s, newObj)
+		return
+	}
+	isEqual := oldService != nil &&
+		newService != nil &&
+		equality.Semantic.DeepEqual(oldService.Spec, newService.Spec) &&
+		equality.Semantic.DeepEqual(oldService.Status, newService.Status)
+	if isEqual {
+		// Skip updates without changes in the spec.
```
wojtek-t: This doesn't match check above - you also check status changes.

Remove this comment - please don't add comments that only describe what the code is doing.

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: Yes - but the line 287:
"if _, exists := s.creationTimes.Get(key, ipAssigningPhase); !exists"

is also definition of the same thing :)

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,345 @@
...
+		s.creationTimes.Set(key, creatingPhase, time.Now())
+	}
+	if len(svc.Status.LoadBalancer.Ingress) < 1 {
+		return nil
+	}
+	if _, exists := s.creationTimes.Get(key, ipAssigningPhase); !exists {
+		s.creationTimes.Set(key, ipAssigningPhase, time.Now())
+		klog.Infof("%s: %s ipAssigned", s, key)
+	}
+	if len(svc.Spec.Ports) < 1 {
```
wojtek-t: This is specified by user - so if this is unset, it won't be set randomly.
Take a look at this validation: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L3778

You should remove this check and instead at the beginning of this function ignore services that doesn't have port set at all (before you even add something to creationTimes).

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Fixed. Now is used in `waitForReady`.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Removed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,331 @@
...
+		pingCheckers:  make(map[string]*pingChecker),
+	}
+}
+
+type serviceCreationLatencyMeasurement struct {
+	selector        *measurementutil.ObjectSelector
+	waitTimeout     time.Duration
+	stopCh          chan struct{}
+	isRunning       bool
+	selectorsString string
```
wojtek-t: Remove selectorString

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: this is still valid comment

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,331 @@
...
+				success = 0
+				time.Sleep(pingBackoff)
+				continue
+			}
+			success++
+			if success == pingChecks {
+				func() {
+					klog.Infof("%v: %s is reachable", p.owner, p.key)
+					p.owner.lock.Lock()
+					defer p.owner.lock.Unlock()
```
wojtek-t: Please make measurementutil.ObjectTransitionTimes - this will simplify couple things in this file.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,331 @@
...
+	return informer.StartAndSync(i, s.stopCh, informerSyncTimeout)
+}
+
+func (s *serviceCreationLatencyMeasurement) waitForReady() error {
+	for {
+		created, ipAssigned, reachable := 0, 0, 0
+		select {
+		case <-time.After(s.waitTimeout):
+			klog.Errorf("%s: waiting for services timeout")
+			return fmt.Errorf("waiting for services timeout")
```
wojtek-t: Don't log error if you return it - if you return an error it's responsibility of a called to handle it (whether they want to log it or not).

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return nil, s.start()
+	case "waitForReady":
+		err = s.waitForReady()
+		if err != nil {
+			klog.Errorf("%s: %v", s, err)
+		}
+		return nil, err
```
wojtek-t: don't log error if you return it - it's responsibility of a caller to handle it.
This whole case should be:
```
return nil, s.waitForReady()
```

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+	)
+	return informer.StartAndSync(i, s.stopCh, informerSyncTimeout)
+}
+
+func (s *serviceCreationLatencyMeasurement) waitForReady() error {
+	for {
+		created, ipAssigned, reachable := 0, 0, 0
+		select {
+		case <-time.After(s.waitTimeout):
+			return fmt.Errorf("waiting for services timeout")
```
wojtek-t: Please add context to that it's for "s" measurement.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+			To:   reachabilityPhase,
+		},
+	})
+
+	content, err := util.PrettyPrintJSON(measurementutil.LatencyMapToPerfData(serviceCreationLatency))
+	if err != nil {
+		return nil, err
+	}
+	summary := measurement.CreateSummary(serviceCreationLatencyName, "json", content)
+	return []measurement.Summary{summary}, err
```
wojtek-t: s/err/nil/ 
[there is no way for it to be non-nil in this codepath]

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+	newService, ok = newObj.(*corev1.Service)
+	if newObj != nil && !ok {
+		klog.Errorf("%s: uncastable new object: %v", s, newObj)
+		return
+	}
+	isEqual := oldService != nil &&
+		newService != nil &&
+		equality.Semantic.DeepEqual(oldService.Spec, newService.Spec) &&
+		equality.Semantic.DeepEqual(oldService.Status, newService.Status)
+	if isEqual {
```
wojtek-t: if isEqual := ...; isEqual {
  return
}

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+		return
+	}
+
+	if !s.isRunning {
```
wojtek-t: This isn't thread-safe - you're accessing this value and can modify it when calling Dispose()
This isn't specific to this measurement - it's a problem for many existing measurements (and it works only because we're cleaning ourselves in test, so nothing is happening when Dispose is called).

Please file an issue for this.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+	key, err := runtimeobjects.CreateMetaNamespaceKey(svc)
+	if err != nil {
+		return fmt.Errorf("meta key created error: %v", err)
+	}
+	s.pingCheckers.DeleteAndStop(key)
+	return nil
+}
+
+func (s *serviceCreationLatencyMeasurement) updateObject(svc *corev1.Service) error {
+	if len(svc.Spec.Ports) < 1 {
```
wojtek-t: Please add a comment why we ignore it.
It's because for now we're only making this measurement to work for services with LoadBalancer type and services of this type has to have port specified.

[This should be a comment both here and at the measurement level]

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+		return nil
+	}
+	key, err := runtimeobjects.CreateMetaNamespaceKey(svc)
+	if err != nil {
+		return fmt.Errorf("meta key created error: %v", err)
+	}
+	if _, exists := s.creationTimes.Get(key, creatingPhase); !exists {
+		s.creationTimes.Set(key, creatingPhase, time.Now())
```
wojtek-t: Why time.Now()?
It should be service.CreationTimestamp.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	if len(svc.Status.LoadBalancer.Ingress) < 1 {
+		return nil
+	}
+	if _, exists := s.creationTimes.Get(key, ipAssigningPhase); exists {
+		return nil
+	}
+	s.creationTimes.Set(key, ipAssigningPhase, time.Now())
+	klog.Infof("%s: %s ipAssigned", s, key)
```
wojtek-t: Do we really want to log it?

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: Still valid.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func (p *pingChecker) run() {
+	success := 0
+	for {
+		select {
+		case <-p.stopCh:
+			return
+		default:
+			if _, exists := p.creationTimes.Get(p.key, reachabilityPhase); exists {
```
wojtek-t: I don't understand how can this happen.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+				return
+			}
+			_, err := execservice.RunCommand(
```
wojtek-t: Please add a comment (at the measurement level) to what I mentioned above (that we only support LoadBalancer services for now.
And here please add a TODO to handle more ingress ips and more ports.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+				fmt.Sprintf("curl %s:%d", p.svc.Status.LoadBalancer.Ingress[0].IP, p.svc.Spec.Ports[0].Port))
+			if err != nil {
+				success = 0
+				time.Sleep(pingBackoff)
+				continue
+			}
+			success++
+			if success == pingChecks {
+				func() {
```
wojtek-t: func not needed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,317 @@
...
+					klog.Infof("%v: %s is reachable", p.callerName, p.key)
```
wojtek-t: I'm not convinced we really would like to log it.

clusterloader2/pkg/measurement/util/phase_latency.go
```diff
@@ -71,6 +71,19 @@ func (o *ObjectTransitionTimes) Get(key, phase string) (time.Time, bool) {
 	return time.Time{}, false
 }
 
+// Count returns number of key having given phase entry.
+func (o *ObjectTransitionTimes) Count(phase string) int {
+	o.lock.Lock()
+	defer o.lock.Unlock()
+	n := 0
```
wojtek-t: s/n/count/

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Shouldn't this be done by the caller?

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: We are doing similar thing when waiting for pods. I found it convenient when debugging, so I want to have similar logic here.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Sanity check.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Removed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Same answer as for logging ipAssigned.

clusterloader2/pkg/measurement/util/phase_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: key removed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: We have similar logic on all other measurements. If I'm changing this, I have to change it in all other places.
I'll keeping it this way (to keep style coherent) and I'll add the issue to fix the problem on cluster loader level.

krzysied: #682

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// Execute executes service startup latency measurement actions.
+// Services can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+// "start" action starts observation of the services.
+// "waitForReady" waits until all services are reachable.
+// "gather" returns service created latency summary.
+// This measurement only works for services with LoadBalancer type.
+// Service of this type has to have port specified.
```
wojtek-t: Please remove - this is part of validation that we do in k8s apiserver, so it's always the case.

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: Given you're inconsistent already anyway, let's change this one now.

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: No - you should return a meaningful error for yourself.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (s *serviceCreationLatencyMeasurement) waitForReady() error {
+	for {
+		created, ipAssigned, reachable := 0, 0, 0
+		select {
+		case <-time.After(s.waitTimeout):
+			return fmt.Errorf("waiting for services timeout")
+		case <-time.After(defaultCheckInterval):
+			func() {
```
wojtek-t: func no longer needed.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+	)
+	return informer.StartAndSync(i, s.stopCh, informerSyncTimeout)
+}
+
+func (s *serviceCreationLatencyMeasurement) waitForReady() error {
+	for {
+		created, ipAssigned, reachable := 0, 0, 0
+		select {
+		case <-time.After(s.waitTimeout):
```
wojtek-t: The semantics of this waitTimeout is strange, ebcause it starts from scratch on every iteration.

Why can't you simply use wait.Poll() function that is so widely used in k8s?

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: And as a result, you no longer need to define those variables above.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return fmt.Errorf("meta key created error: %v", err)
+	}
+	s.pingCheckers.DeleteAndStop(key)
+	return nil
+}
+
+func (s *serviceCreationLatencyMeasurement) updateObject(svc *corev1.Service) error {
+	// This measurement only works for services with LoadBalancer type.
+	// Service of this type has to have port specified.
```
wojtek-t: Remove it - this checked by validation in k8s anyway.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(svc.Spec.Ports) < 1 {
```
wojtek-t: BTW - when I know think about it, we can simply check svc.Spec.Type instead - it will be consistent with what we're saying we support.

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: We're logging every change state for every pod? I highly doubt.

I looked into wait_for_pods, wait_for_controlled_pods and pod_startup_slo and in neither of those you log it.
If you really log it, please point me to it.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,323 @@
...
+	for {
+		select {
+		case <-p.stopCh:
+			return
+		default:
+			if _, exists := p.creationTimes.Get(key, reachabilityPhase); exists {
+				return
+			}
+			// TODO(#679): Current implementation handles only load balancers.
+			_, err := execservice.RunCommand(
```
wojtek-t: How long this command takes? In general I would like to understand what we will report if e.g. service will be programmed and available immediately.

clusterloader2/pkg/measurement/util/phase_latency.go
wojtek-t: Probably for simplicity you can move it to a separate PR :)

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: It's quite fast, although it need to pass 10 times. There will always be slight delay.
However if there is a time change in service becoming available, the measurement will spot it.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: not yet

clusterloader2/pkg/measurement/common/service_creation_latency.go
wojtek-t: 10 times if sth takes 50ms is fine, 10 times if it takes 500ms doesn't sounds fine.

At the very least I would like to see an issue opened with a TODO here that we should validate if calling RunCommand for every single ping isn't too expensive (e.g. maybe after first success, we should run second that will do ping 9 times or sth like that?). But TODO is fine.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Now it is, I didn't push newer version for some reason.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done. https://github.com/kubernetes/perf-tests/issues/685

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,306 @@
...
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/workerqueue"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	serviceCreationLatencyName           = "ServiceCreationLatency"
+	serviceCreationLatencyWorkers        = 10
+	defaultServiceCreationLatencyTimeout = 1 * time.Minute
+	defaultWaitTimeout                   = 10 * time.Minute
```
wojtek-t: This is not used anywhere - please remove.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -0,0 +1,306 @@
...
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/workerqueue"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	serviceCreationLatencyName           = "ServiceCreationLatency"
+	serviceCreationLatencyWorkers        = 10
+	defaultServiceCreationLatencyTimeout = 1 * time.Minute
```
wojtek-t: 1m seems to low to me. 10m?

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/service_creation_latency.go
krzysied: Done.

Issue #575, opened 2019-06-25
[slo] Measure and expose Networki Programming Latency in tests


oxddr:
*SLI = Network Programming Latency*

What needs to be done:
1. Enable kube-proxy scrapping gradually in tests
2. Add a measurement of SLI
3. Integrated perf-dash so it can display SLI

Nice-to-have:
1. Refactor clusterloader2 so adding new Prometheus-based measurements is easy
1. Write down documentation on how to add new measurement end-to-end (implementing measurement, integrating perf-dash). 

/area slo
/area clusterloader
/sig scalability

PR #586, opened 2019-06-28
Fix typo in measurement name.

oxddr:
/assign @krzysied @mm4tt


oxddr:
ref https://github.com/kubernetes/perf-tests/issues/575

wojtek-t:
This PR broke scalability tests. The fix is in https://github.com/kubernetes/perf-tests/pull/587, but we should really understand why clusters2 and clusterload2-kubemark passed.
Opened an issue to figure it out: https://github.com/kubernetes/perf-tests/issues/588

PR #587, opened 2019-06-29
Fix scalability tests

wojtek-t:


wojtek-t:
Self-applying lgtm label to fix scalability tests.

Issue #588, opened 2019-06-29
Figure out why presubmit didn't catch breaking scalability tests in https://github.com/kubernetes/perf-tests/pull/586


wojtek-t:
@kubernetes/sig-scalability-bugs 
@mm4tt @krzysied @mborsz

mm4tt:
My guess is that the "NetworkProgrammingLatency" is a gated feature that has to be explicitly enabled via config override and it's not enabled in presubmits. 
I guess we should enable it in presubmit once we get confidence that it works (which I think we got already).

wojtek-t:
Makes sense - but that implies that we should never enable sth in blocking tests without enabling it in presubmits too.

oxddr:
@wojtek-t yes, we should have enabled the experiment in the presubmit first. I'll create a short doc about how to rollout experiments in clusterloader. Hopefully with clear and explicit  instructions, we can avoid situations like that.

PR #590, opened 2019-07-01
Document process for rolling out clusterloader2 experiments.

oxddr:
Let's have a discussion about how to rollout experiments to clusterloader2. By no means I insist this is the final form of this document (and practices it describes). 

fixes https://github.com/kubernetes/perf-tests/issues/588

/assign @mm4tt @wojtek-t 
/cc @mborsz @krzysied

### DIFF
--- a/clusterloader2/docs/experiments.md
+++ b/clusterloader2/docs/experiments.md
@@ -1,0 +1,100 @@
+# Clusterloader2 - experiment rollout
+
+In this doc any change to the behavior of clusterloader2 that
+
+    - enables new measurement
+    - changes semantic of an existing measurement
+    - changes how clusterloader2 setups cluster and run tests
+
+is referred as an "experiment".
+
+## Motivation
+
+clusterloader2 is a tool used by all scalability, performance tests. Tests
+compile clusterloader2 at [HEAD], thus introducing breaking changes to
+clusterloader2 will stop scalability tests from passing at all. They have a
+large blast radius: every PR to k/k needs to pass. Also while for smaller and
+faster tests, breakages aren't that costly (unless they happen on the weekend,
+see https://github.com/kubernetes/perf-tests/pull/586), they are expensive for
+large, rarely run tests (e.g. [ci-kubernetes-e2e-gce-scale-performance]).
+
+For this reason, all new changes/features added to clusterloader2 shall be gated
+and rolled out gradually. This way, we can minimize blast radius of breaking
+changes. We can even stop some kinds of them to happen at all, as they should be
+caught at presubmit time and not allowed to merge at all.
+
+### General principles
+
+1. Allow at least 24h between changes to test configs to ensure the experiment
+   is stable. We don't want to block PR to k/k because of a flaky experiment in
+   clusterloader2.
+
+1. Check with sig-scalability whether there is a regression in the test you want
+   to enable experiment for. We don't want new features in clusterloader2 to
+   interfere with regression' debugging.
+
+1. For tests not listed below it's fine to enable experiments at your
+   convenience.
+
+### Step-by-step process
+
+_Each step should be a separate PR_
+
+1. Add a "knob" to turn on the future experiment. Usually this means adding a
+   new environmental variable (PR to [test-infra]) or a new override file (PR to
+   [perf-tests]). At this point the "knob" is not used by any code path.
+
+   Since the knob is not used anywhere it's no-op and should be safe.
+
+1. Enable experiment in [perf-tests presubmits],
+   [ci-kubernetes-e2e-gci-gce-scalability] and [ci-kubernetes-kubemark-100-gce]
+
+   Again since the knob is not used anywhere it's no-op and should be safe.
+
+   Perf-test presubmit runs two jobs: [pull-perf-tests-clusterloader2] and
+   [pull-perf-tests-clusterloader2-kubemark]. Primary role of those presubmits
+   is to catch bugs in code from perf-tests, so we enable the experiment for
+   both of those jobs first. Enabling experiment in
+   [ci-kubernetes-e2e-gci-gce-scalability] should give you enough data points to
+   determine whether experiments work, once we add new code path or config. It
+   also runs frequent enough, so in case of problems you can revert quickly.
+   Before enabling the experiment in [ci-kubernetes-e2e-gci-gce-scalability]
+   make sure there is no ongoing regression affecting this test. If we are at
+   the code freeze of thaw, you should wait with updating
+   [ci-kubernetes-e2e-gci-gce-scalability] until the freeze is suspended.
+
+1. Add a new code path or config that uses "knob" added in the first PR.
+
+   We've already enable it in the first step, so the PR will be only merged if
+   the new code path or configuration passes [perf-tests presubmits].
+
+1. Enable experiment in [pull-kubernetes-e2e-gce-100-performance] and
+   [pull-kubernetes-kubemark-e2e-gce-big]
+
+   Changing presubmit definitions in test-infra has an ability to break k/k
+   presubmits. PRs to test-infra don't trigger presubmit in k/k. Once you enable
+   the experiment in the presubmit you need to watch next 3 runs after your PR
+   is merged to detect breakages.
+
+1. Enable experiment in [ci-kubernetes-e2e-gce-scale-performance] and rest of Kubemark
+   tests ([ci-kubernetes-kubemark-500-gce], [ci-kubernetes-kubemark-gce-scale],
+   and [ci-kubernetes-kubemark-high-density-100-gce])
+
+   If feasible, please test experiment locally (it is outside of jobs run on
+   Prow) first, as those tests run once a day and expensive. Double-check with
+   sig-scalability that there is no ongoing regression in big clusters.
+
+[head]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L121
+[ci-kubernetes-e2e-gce-scale-performance]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L44
+[ci-kubernetes-e2e-gci-gce-scalability]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L98
+[ci-kubernetes-kubemark-100-gce]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L258
+[ci-kubernetes-kubemark-500-gce]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L307
+[ci-kubernetes-kubemark-gce-scale]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L355
+[ci-kubernetes-kubemark-high-density-100-gce]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L406
+[perf-tests presubmits]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L267
+[pull-kubernetes-e2e-gce-100-performance]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L3
+[pull-kubernetes-kubemark-e2e-gce-big]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L149
+[pull-perf-tests-clusterloader2-kubemark]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L317
+[pull-perf-tests-clusterloader2]: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L268
+[test-infra]: https://github.com/kubernetes/test-infra
+[perf-test]: https://github.com/kubernetes/perf-tests


clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,44 @@
...
+
+For this reason, all new changes features added to clusterloader2, shall be gated and rolled out gradually. This way we can minimize blast radius of breaking changes. We can even remove some kinds of them to happen at all, as they should be caught at presubmit time and not allowed to merge at all.
+
+## Experiment rollout
+
+### General principles
+
+1. If for some reason it doesn't make sense to enable an experiment in particular test job indicate the reason clearly in the PR (e.g. measuring kube-proxy metrics in kubemark can be skipped as kube-proxy is not running there).
```
mm4tt: NIT: It's not true that kube-proxy doesn't run there, a hollow kube-proxy runs there

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. Check with sig-scalability whether there is a regression in the test you want to enable experiment for. We don't want new features in clusterloader2 to interfere with regression' debugging.
+
+1. For tests not listed below it's fine to enable experiments
+
+### Step-by-step process
+
+_Each step should be a separate PR_
+
+1. Enable experiment in [perf-tests presubmits](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L267) and [ci-kubernetes-e2e-gci-gce-scalability](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L98)
```
mm4tt: So overall the steps are

1. perf-tests presubmits and gce-100 CI
2. k/k presubmits
3. kubemark CI
4. gce-scale CI

Questions:
1. Can we split the first step and have a gce-100 CI as a separate step?
2. What about GKE tests using CL2?

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. Enable experiment in [pull-kubernetes-e2e-gce-100-performance](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L3) and [pull-kubernetes-kubemark-e2e-gce-big](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L149)
+
+   Block changes to k/k if they don't work with the experiment enabled.
```
mm4tt: Could you rewrite this to make it clear what is the intention? 

I think there should be some additional procedure here, this step is the most prone to blocking k/k changes, due to the fact that you modify the presubmit via PR in test-infra, which doesn't trigger the k/k presubmit to run. It's very easy to  break the k/k presubmit if you don't keep an eye on k/k presubmits directly after your test-infra PR has merged

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,44 @@
+# clusterloader2 - experiment rollout
+
+In this doc any change to the behavior of clusterloader2 is referred as an "experiment".
+
+## Motivation
+
+clusterloader2 is a tool used a by all scalability (performance) tests. Tests compile clusterloader2 at [HEAD](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L121). Introducing breaking changes to clusterloader2 will stop tests from passing at all. While for smaller and faster tests, this is not a huge costs (unless this happens on the weekend, see https://github.com/kubernetes/perf-tests/pull/586), this can be costly in rarely run, expensive tests (e.g. [ci-kubernetes-e2e-gce-scale-performance](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-release-blocking-jobs.yaml#L44)).
```
mm4tt: s/used a by/used by/

mm4tt: It can be also costly if we break k/k presubmit, as it blocks all k/k PRs from being merged.

clusterloader2/docs/experiments.md
oxddr: TIL! 
However I still expect it to not do real work, so it doesn't make sense to scrape it in this case. I'll rephrase this sentences.

clusterloader2/docs/experiments.md
oxddr: It's mentioned in point 2 under "general principles", that we don't want to break k/k presubmits. I'll be more explicit in this section about it too.

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,44 @@
...
+
+For this reason, all new changes features added to clusterloader2, shall be gated and rolled out gradually. This way we can minimize blast radius of breaking changes. We can even remove some kinds of them to happen at all, as they should be caught at presubmit time and not allowed to merge at all.
```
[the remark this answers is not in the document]
wojtek-t: I actually started wondering if "gradual rollout" is really what we would like to have.

I think the principle should be:
- we have some "staging" job that we update first
- once this work, we believe the change is safe, and that means we can enable it anywhere else.

clusterloader2/docs/experiments.md
oxddr: ad. 1) Yes, we can. However enabling an experiment only for perf-test presubmit doesn't add much value IMO. It's only executed when PR for this repo is opened, which is not-frequent. gce-100 CI should give us enough data points and on the other hands, if things go south we can revert an experiment quickly. 

ad. 2) GKE tests are not release-blocking and used by PRs. IMO you can enable experiment at any point you like, assuming you will take care if you break them. Do we want more formalized approach?

clusterloader2/docs/experiments.md
wojtek-t: nit: I would prefer to split it into lines, instead of having such long lines where it's even hard to comment on specific things (I would have more spefic comments)

oxddr: Done, PTAL.

clusterloader2/docs/experiments.md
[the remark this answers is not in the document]
oxddr: What's your specific argument against gradual rollout/why you think an approach suggested by you is better?

clusterloader2/docs/experiments.md
oxddr: Added more details. WDYT?

clusterloader2/docs/experiments.md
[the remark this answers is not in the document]
wojtek-t: It's much more work (it's easier to forget about one of 5 steps than one of 2 steps).

clusterloader2/docs/experiments.md
wojtek-t: > ad. 1) Yes, we can. However enabling an experiment only for perf-test presubmit doesn't add much value IMO.

I disagree - it is critical. It would actually prevent the issue that happened last week.
The flow in my opinion should be:
- change test-infra config for perf-tests presubmit
- introduce the env-var in perf-tests - that would automatically test whether the config is correct
- enable it everywhere else

Regarding gradual rollout - I'm not sure how much it buys us. We generally should be easily able to assess the performance impact of the chances in clusterloader from the code. So as long as we enable first the experiments in "non-blocking" mode, I don't think it actually provides much value.

clusterloader2/docs/experiments.md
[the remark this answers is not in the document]
oxddr: re: more work: Interestingly, we rollout some (most?) changes to clusterloader2 gradually already (e.g. support for Promethues or new SLO measurements). I agree it is more work that rolling changes out in a single PR. I'd argue whether it's really that much more work. Enabling experiments would be a single line PRs.

re forgetting: The main reason I wanted to document the process it that we minimize chances that one forgets about certain steps. We could either link to it or check it, when not sure what step should go next. That's the reason it's so prescriptive.

clusterloader2/docs/experiments.md
oxddr: Wojtek, I think the import part of that sentence is the word **only**. By no means I wanted to say, that updating perf-tests presubmit jobs is not important. It is! But on the other hand it gives us little signal, as those jobs are not run continuously. 

As mentioned in another comment - we already rollout changes gradually. This doc simply tries to codify this practice in an opinionated way. One reason we don't want to enable experiment everywhere at once is regressions. It happen much more often that we have problems with gce5000, while gce100 is doing just fine. To not interfere with debugging, we wait with enabling in gce5000, while the new code is doing just fine in gce100. While waiting for resolving the regression, we shouldn't allow new code to k/k that would break with the experiment. I'd rather have this reflected in the process, than making exceptions to it few times a quarter. 

Also, I disagree, that's easy to access the performance impact. It seems that our tests are prone to even few more API calls (like the current situation with extra lists from kubelet).

wojtek-t: >  It is! But on the other hand it gives us little signal, as those jobs are not run continuously.

My main motivation now is to avoid breaking the test because of bad push. And that solves this problem.

> It happen much more often that we have problems with gce5000, while gce100 is doing just fine. To not interfere with debugging, we wait with enabling in gce5000, while the new code is doing just fine in gce100.

If it's doing just fine in 100-node, what's the point with waiting for it?
I actually claim that we will know how it behaves on 100-node cluster just from presubmit.

> Also, I disagree, that's easy to access the performance impact. It seems that our tests are prone to even few more API calls (like the current situation with extra lists from kubelet).

Listing all objects of a given type (which was the case from kubelet) is a know thing that will cause issues - it is not surprising at all.
I agree that there may be some subtle things, but I actually claim that we won't catch them by gradual rollout too (they won't be visible in 100-node clusters).

oxddr: 1. A good addition to this discussion is the issue from this week, when innocent change by Maciek broke all our tests and kubernetes presubmit. Gradual rollout in this case would reduce the blast radius only to our test only. Personally I think it's worth add a little work on us to prevent breaking k/k presubmit. That's afterall one of my main motivation for this doc - to avoid breakages of k/k presubmit because of bugs on our side. 

2. re: waiting: we sometimes want to wait, because there is an ongoing regression, not because there is something wrong with experiment. This has happened few time last quarter, that we wait with experiments, until regressions are resolved. 

3. I guess we both know that listing object can be source of issues. But it's no obvious for everyone. But anyway, this was just an example. I can imagine other, more subtle problems, which may not be easy to spot.

wojtek-t: > A good addition to this discussion is the issue from this week, when innocent change by Maciek broke all our tests and kubernetes presubmit. Gradual rollout in this case would reduce the blast radius only to our test only. Personally I think it's worth add a little work on us to prevent breaking k/k presubmit. That's afterall one of my main motivation for this doc - to avoid breakages of k/k presubmit because of bugs on our side.

That my motivation too (and actually that's the only thing that I really would like to achieve).
And to do that you need to:
1. first submit all the code to actually use it (including test-infra)
2. then send out PR ti perf-tests to start using  that env-var/config-file or sth to perf-tests (test-definition) so that the presubmit on perf-tests repo will run against full change
3. once that passes, (assuming our presubmit is identical that the one in k/k repo, which hopefuly is), we know we won't break that and 100-node tests (because presubmit in perf-tests repo passed), and we can enable it everywhere else

My main point here is: it's point 1 and 2 being done in the correct order that will prevent those issues - not the gradual rollout itself.

oxddr: I've changed the content according to our offline discussion. 

I've also remove step with enabling change in Kubemark.

oxddr:
PTAL - I've changed this doc as it was discussed offline with Wojtek.

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
+# Clusterloader2 - experiment rollout
+
+In this doc any change to the behavior of clusterloader2 is referred as an
+"experiment".
```
wojtek-t: Maybe it's actually nit-picking, but I'm wondering what "change to behavior" really means.
It's very vague and I would like to avoid a pathological case when we will be gating every single change to CL2.

I think we should be more explicit and talk about:
- new measurements?
- semantic changes to existing measurements?
- something else?

@mm4tt for thoughts

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+
+clusterloader2 is a tool used by all scalability, performance tests. Tests
+compile clusterloader2 at [HEAD], thus introducing breaking changes to
+clusterloader2 will stop scalability tests from passing at all. They large blast
```
wojtek-t: s/They/they have/ ?

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
...
+large, rarely run tests (e.g. [ci-kubernetes-e2e-gce-scale-performance]).
+
+For this reason, all new changes/features added to clusterloader2 shall be gated
+and rolled out gradually. This way, we can minimize blast radius of breaking
+changes. We can even stop some kinds of them to happen at all, as they should be
+caught at presubmit time and not allowed to merge at all.
+
+## Experiment rollout
+
+### Experiment design
```
wojtek-t: Remove this?

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
...
+   new environmental variable (PR to [test-infra]) or a new override file (PR to
+   [perf-tests]). At this point the "knob" is not used by any code path.
+
+   Since the knob is not used anywhere it's no-op and should be safe.
+
+
+1. Enable experiment in [perf-tests presubmits] and
+   [ci-kubernetes-e2e-gci-gce-scalability]
+
+   Again cince the knob is not used anywhere it's no-op and should be safe.
```
wojtek-t: since

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
...
+
+   Perf-test presubmit runs two jobs: [pull-perf-tests-clusterloader2] and
+   [pull-perf-tests-clusterloader2-kubemark]. Primary role of those presubmits
+   is to catch bugs in code from perf-tests, so we enable the experiment for
+   both of those jobs first. Enabling experiment in
+   [ci-kubernetes-e2e-gci-gce-scalability] should give you enough data points to
+   determine whether experiments work, once we add new code path or config. It
+   also run frequent enough, so in case of problems you can revert quickly.
```
wojtek-t: s/run/runs/

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. Enable experiment in [pull-kubernetes-e2e-gce-100-performance] and
+   [pull-kubernetes-kubemark-e2e-gce-big]
+
+   Changing presubmit definitions in test-infra has an ability to break k/k
+   presubmits. PRs to test-infra don't trigger presubmit in k/k. Once you enable
+   the experiment in the presubmit you need to watch next 3 runs after your PR
+   is merged to detect breakages.
+
+1. Enable experiment in [ci-kubernetes-e2e-gce-scale-performance]
```
wojtek-t: And all other kubemarks?

oxddr: No strong opinion on this. @mm4tt?

clusterloader2/docs/experiments.md
oxddr: I left it deliberately vague, so if you wonder whether your change is actually changing behavior or not, you should default to enable it gradually. However I am happy to be more precise here. 

I agree with your list. I'd add to the list major changes to clusterloader that alters how test are setup and run (such us enabling Prometheus or enabling node exporter).

mm4tt: +1 to being more precise and including things that we're 100% sure that should be gated. I believe for the rest, we (and others) should just use common sense.

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,97 @@
...
+1. Enable experiment in [perf-tests presubmits] and
+   [ci-kubernetes-e2e-gci-gce-scalability]
+
+   Again cince the knob is not used anywhere it's no-op and should be safe.
+
+   Perf-test presubmit runs two jobs: [pull-perf-tests-clusterloader2] and
+   [pull-perf-tests-clusterloader2-kubemark]. Primary role of those presubmits
+   is to catch bugs in code from perf-tests, so we enable the experiment for
+   both of those jobs first. Enabling experiment in
+   [ci-kubernetes-e2e-gci-gce-scalability] should give you enough data points to
```
mm4tt: It's a release blocking job, so we should avoid making any changes there during code-freeze/thaw

clusterloader2/docs/experiments.md
wojtek-t: Given that there is a very limited set of people who can approve changes to it, I would start with what there is (I'm fine what you suggested in addition to my examples) and we can extend this list when we better-defined things have a potential to break stuff.

clusterloader2/docs/experiments.md
mm4tt: Not sure if I follow, I don't see any step modifying kubemark tests (except presubmits) and I think it should be listed here (especially if we recommend changing kubemark presubmit)

clusterloader2/docs/experiments.md
oxddr: Added info, that one should wait until the freeze is suspended.

clusterloader2/docs/experiments.md
oxddr: I've re-added Kubemark test as part of the last step.

oxddr:
PTAL

clusterloader2/docs/experiments.md
wojtek-t: I'm sorry - by "what there is", I though about my list with your addition, not with there is in the doc now :)

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,101 @@
...
+
+For this reason, all new changes/features added to clusterloader2 shall be gated
+and rolled out gradually. This way, we can minimize blast radius of breaking
+changes. We can even stop some kinds of them to happen at all, as they should be
+caught at presubmit time and not allowed to merge at all.
+
+### General principles
+
+1. Change to clusterloader2 behavior is understood as:
+
```
wojtek-t: nit: remove this empty line?

clusterloader2/docs/experiments.md
```diff
@@ -0,0 +1,101 @@
...
+   - enabling new measurement
+   - changing semantic of an existing measurement
+   - changing how clusterloader2 setups cluster and run tests
```
wojtek-t: Ohh - I see it's here - I would suggest moving it to the beginning and say that this is what we later call experiment.

PR #592, opened 2019-07-01
Show Network programming latency in perf-dash

oxddr:
Rename parsing function more clearly (it's actually not specific to responsiveness data).

fixes: https://github.com/kubernetes/perf-tests/issues/575

Tested locally: b9a2U3ZYNEd

/assign @mm4tt

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -69,18 +69,18 @@ var (
 				{
 					Name:             "density",
 					OutputFilePrefix: "PodStartupLatency",
-					Parser:           parseResponsivenessData,
+					Parser:           parsePerfData,
 				},
 				{
 					Name:             "density",
 					OutputFilePrefix: "PodStartupLatency_PodStartupLatency",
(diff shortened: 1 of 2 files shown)


mm4tt:
Nice!

Would be nice to have all %-iles in the perf-dash, but I think we can tackle it when working on the generic Prometheus measurement.
What I had in mind would be to ultimately have something like:

```
Identifier: NetworkProgrammingLatency
    Method: PrometheusMetric
    Params:
      query: quantile_over_time(%v, kubeproxy:kubeproxy_network_programming_duration:histogram_quantile{quantile='%v'}[%v])
      percentiles: [50, 90, 99]
      thresholds: # thresholds is a map, it can be empty = no assertions or have different values for different percentiles. 
        90: 12s
        99: 16s
```

Which would automatically export all %-iles to perf-dash. It could be reused for other SLIs that we're going to implement in cl2.

In the meantime approving this one. 

/lgtm
/approve

PR #596, opened 2019-07-03
Add node-exporter to our tests

mborsz:
Add support for node-exporter in our tests.

It will run only on registered nodes with kubernetes.io/node-exporter=true label (none right now, it will be possible to add this when https://github.com/kubernetes/kubernetes/pull/79703 is merged.

* It does work for both kubemark and regular tests *
/assign @mm4tt

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -19,17 +19,21 @@ package prometheus
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"time"
 
+	"golang.org/x/sync/errgroup"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/util/system"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
 
 const (
@@ -41,6 +45,8 @@ const (
 	checkPrometheusReadyInterval = 30 * time.Second
 	checkPrometheusReadyTimeout  = 15 * time.Minute
 	numK8sClients                = 1
+	nodeExporterPod              = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/exporters/node-exporter.yaml"
+	nodeExporterTemplateName     = "PROMETHEUS_SCRAPE_NODE_EXPORTER"
 )
 
 // PrometheusController is a util for managing (setting up / tearing down) the prometheus stack in
@@ -93,6 +99,11 @@ func (pc *PrometheusController) SetUpPrometheusStack() error {
 	}
 	if err := pc.applyManifests(coreManifests); err != nil {
 		return err
+	}
+	if scrapeNodeExporter, ok := pc.templateMapping[nodeExporterTemplateName].(bool); ok && scrapeNodeExporter {
+		if err := pc.runNodeExporter(); err != nil {
+			return err
+		}
 	}
 	if pc.isKubemark() {
 		if err := pc.exposeKubemarkApiServerMetrics(); err != nil {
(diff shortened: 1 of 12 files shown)


wojtek-t:
/hold

Can we wait with merging that before it will be possible? And rerun tests when all depending PRs will be merged?

mborsz:
I modified the code, it should work for both kubemark and regular tests as long as masters are registered, have kubernetes.io/node-exporter=true label and testoverride=scrape-node-exporter is passed to the clusterloader.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -183,6 +192,38 @@ func (pc *PrometheusController) exposeKubemarkApiServerMetrics() error {
...
+	if err != nil {
+		return err
+	}
+
+	numMasters := 0
+	for _, node := range nodes {
+		if system.IsMasterNode(node.Name) {
+			numMasters++
+			if label := node.Labels[nodeExporterNodeLabel]; label != "true" {
+				return fmt.Errorf("Master %q doesn't have required %q label which is required by node-expoter", node.Name, nodeExporterNodeLabel)
```
mm4tt: NIT: error strings in go shouldn't start with capital letter
NIT: s/node-expoter/node-exporter/

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -183,6 +192,38 @@ func (pc *PrometheusController) exposeKubemarkApiServerMetrics() error {
...
+			}
+		}
+	}
+
+	if numMasters == 0 {
+		return fmt.Errorf("node-exporter requires master to be regirstered nodes")
```
mm4tt: masters to be registered

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -183,6 +192,38 @@ func (pc *PrometheusController) exposeKubemarkApiServerMetrics() error {
...
+	}
+
+	return kubemarkFramework.ApplyTemplatedManifests(
+		kubemarkNodeExporterDS, pc.templateMapping, client.Retry(apierrs.IsNotFound))
```
mm4tt: I don't think we need to retry IsNotFound here, it's needed for Prometheus CRDs to deflake race conditions, but here we shouldn't expect 404 when creating DaemonSets.

clusterloader2/pkg/prometheus/manifests/default/node-exporter-serviceMonitor.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: node-exporter
+  name: node-exporter
+  namespace: monitoring
+spec:
+  endpoints:
+  - interval: 15s
```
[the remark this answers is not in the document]
mborsz: Done

mborsz:
Good point. I added gates for non-kubemark as well and fixed all comments. PTAL

mm4tt:
Please fix the tests and ping me when it's done :)

mborsz:
I had to change the approach a little bit: Instead of using daemonset (which doesn't work in kubemark as masters have NetworkUnavaiable condition) I ssh and use static pods instead. PTAL

clusterloader2/pkg/util/ssh.go
```diff
@@ -0,0 +1,40 @@
...
+	"io"
+	"os/exec"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+// SSH executes command on a given node with stdin provided.
+// If stdin is nil, the process reads from null device.
+func SSH(command string, node *v1.Node, stdin io.Reader) error {
```
mm4tt: Similar logic/function is used in nodes.go (node-killer), could we change it to use this util?

Can be done as a part of separate PR

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -183,6 +194,44 @@ func (pc *PrometheusController) exposeKubemarkApiServerMetrics() error {
...
+		node := node
+		if system.IsMasterNode(node.Name) {
+			numMasters++
+			g.Go(func() error {
+				f, err := os.Open(os.ExpandEnv(nodeExporterPod))
+				if err != nil {
+					return fmt.Errorf("Unable to open manifest file: %v", err)
+				}
+				defer f.Close()
+				return util.SSH("sudo tee /etc/kubernetes/manifests/node-exporter.yaml > /dev/null", &node, f)
```
wojtek-t: I'm not a huge fan of this approach - can we at least add a TODO to figure out if we can do that in a better way in the future?

mborsz: Added TODO.

mborsz:
PTAL

Issue #600, opened 2019-07-03
Scalability and Performance Test Cases for Ephemeral Volumes


mucahitkurt:
Related enhancement document; https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190204-scale-testing.md

Ephemeral Volume Types; EmptyDir, ConfigMap, Secret, DownwardAPI

/sig storage
/assign @mucahitkurt

PR #601, opened 2019-07-03
Scalability and Performance Test Cases for Ephemeral Volumes

mucahitkurt:
Related issue: #600 

Different override files are used to test different ephemeral volume types.
For test cases and parameters you can check README file.

/sig storage
/assign @msau42


msau42:
/ok-to-test
cc @hantaowang

msau42:
cc @wojtek-t

wojtek-t:
@mucahitkurt - thanks a lot for this PR
Sorry, I'm a bit overloaded - I will take a look into it early next week. Sorry for delay.

/assign

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+{{$VOLUMES_PER_POD := 4}}
+{{$VOLUME_TYPE := DefaultParam .VOLUME_TYPE "EmptyDir"}}
+{{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+
+{{$APP_NAME := "pod-load"}}
+{{$GROUP := "pod-with-ephemeral-volume-startup-latency"}}
+
+#Variables
+# TODO think about this!
```
mm4tt: Ownerless/issue-less todo, please fix

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
```diff
@@ -0,0 +1,97 @@
...
+{{$podStartupTimeout := $VOLUMES_PER_POD}}
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$podsPerNamespace := MultiplyInt $NODES_PER_NAMESPACE $PODS_PER_NODE}}
+{{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
+{{$volumesPerNamespace := MultiplyInt $podsPerNamespace $VOLUMES_PER_POD}}
+
+name: pod-with-ephemeral-volume-startup-latency
+automanagedNamespaces: {{$namespaces}}
+tuningSets:
+# TODO different options?
```
mm4tt: Same here

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
```diff
@@ -0,0 +1,97 @@
...
+        VolumesPerPod: {{$VOLUMES_PER_POD}}
+        VolumeType: {{$VOLUME_TYPE}}
+        AppName: {{$APP_NAME}}
+- measurements:
+  - Identifier: WaitForRunningPodsWithStorage
+    Method: WaitForRunningPods
+    Params:
+      desiredPodCount: {{$totalPods}}
+      labelSelector: group = {{$GROUP}}
+      # TODO how can i determine this?
```
mm4tt: Same here

Good approach is to start with something big enough, run the test a few times and see what is the distribution of podWithStorage startup-latency and then adjust the timeout making it big enough that it should never time out if everything goes well but small enough that it would catch issues in a reasonable time.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
```diff
@@ -0,0 +1,97 @@
+#Constants
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 1}}
```
mm4tt: QQ, should the test scale with the size of the cluster and what was the cluster size you had in your mind when designing this test?

We're running scalability tests on clusters up to 5K nodes, if this test was run on such cluster as is, it would create 5K namespaces with 1 pod in each namespace? Is it want you wanted to achieve?

mucahitkurt: Actually there will be three different cluster usage scenarios, 2 of them will use 1 node and one of them will be cluster wide.

There will be a [test rollout phase](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190204-scale-testing.md#test-rollout) before these tests are run regularly, to determine the thresholds, but I don't how that rollout process will be executed, before PR merge or after?, maybe @msau42 knows. I thought as these parameters can be changed manually during rollout phase, so I didn't create different override files for 1 node case and cluster wide case. Because the same thing is also valid for parameters like how many volumes will be created per node, or how many volumes will be created per node. Of course I can create different override files with the [max limits](https://kubernetes.io/docs/setup/best-practices/cluster-large/) for each volume type.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
mucahitkurt: It's not clear in comments but I added this and other TODOs because I need feedbacks about these parameters and their values.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
mucahitkurt: I think this one is also determined with the [test rollout process](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190204-scale-testing.md#test-rollout).

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
wojtek-t: There will be a test rollout phase before these tests are run regularly, to determine the thresholds, but I don't know how that rollout process will be executed, before PR merge or after?

We should mark this test as more experimental (maybe under experimental/ directory) and we can do that after the PR is merged. But I'm fine either way too.

@mm4tt - the first two cases, are more about performance than scalability, and will be run against very small (1-node?) cluster. That is probably important part of the answer to the question.

@mucahitkurt - I think that we should focus on the first 2 case first (as the usecase there is completely different). And leave the third case (where the tests may potentially be visibly different) as a second thing. WDYT?

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
wojtek-t: See my comment above.
I actually believe that we need two separate tests:
1. first focus only on the first two cases from the KEP (which by design are running on a very small cluster)
2. separately address the third case, which I think has to be written a bit differently

Assuming we're focuing on (1) here, I would vote for simplifying this case and not think about namespaces, etc. at all here. I would be much easier to understand then,

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
wojtek-t: uniformqps make sense to me

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
```diff
@@ -0,0 +1,97 @@
...
+- name: UniformQPS
+  qpsLoad:
+    qps: {{$LOAD_TEST_THROUGHPUT}}
+steps:
+- measurements:
+  - Identifier: APIResponsiveness
+    Method: APIResponsiveness
+    Params:
+      action: reset
+{{ if $PROVISION_VOLUME }}
```
wojtek-t: Why do you want to gate provisioning volumes?

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
```diff
@@ -0,0 +1,97 @@
...
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: {{$volumesPerNamespace}}
+    tuningSet: UniformQPS
+    objectBundle:
+    - basename: vol-{{$APP_NAME}}
+      objectTemplatePath: {{$VOLUME_TYPE}}.yaml
+{{ end }}
+# Start measurements
```
wojtek-t: I vote for moving this up and doing measurement operations together.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/Secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{.Name}}
+type: Opaque
+data:
+  username: dXNlcm5hbWVfCg==
+  password: cGFzc3dvcmRfCg==
```
wojtek-t: nit: can you add end-of-lines in all files?

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/pod.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: {{.Name}}
+  labels:
+    app: {{.Name}}
+    group: {{.Group}}
+spec:
+  containers:
+  - name: {{.Name}}
+    image: k8s.gcr.io/pause:3.1
+    imagePullPolicy: IfNotPresent
+    ports:
```
wojtek-t: nit: do we need ports?

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/pod.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - containerPort: 80
+      name: {{.Name}}
+    volumeMounts:
+    {{ range $volumeIndex, $vol := Seq .VolumesPerPod }}
+    - name: vol-{{$volumeIndex}}
+      mountPath: /usr/share/{{$volumeIndex}}
+    {{ end }}
+  volumes:
+  {{ range $volumeIndex, $vol := Seq .VolumesPerPod }}
```
wojtek-t: I'm wondering if this really makes it easier to understand.

I agree that it would result in some duplication of code (or I should say config), but wouldn't having:
pod_with_secrets, pod_with_configmaps, pod_with_emptydirs as separate yamls, be simpler?

Longer-term, I filed: https://github.com/kubernetes/perf-tests/issues/619, but I don't know how easy that is to do.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
mucahitkurt: Covering only first 2 cases makes sense to me, I can create a different PR for the 3rd case after this PR is merged. I'll put these test cases under an `experimental` directory and `1_node` directory.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
mucahitkurt: I'll delete the parameters that I need for the (2) tests

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
mucahitkurt: I added the gate because for some volumes, like configmap and secret, I should pre-provision the volumes to use it inside the pod, but for some volumes, like downwardapi and emptyDir, I don't need volume pre-provisioning, volume settings are directly written inside pod template.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/pod.yaml
mucahitkurt: no, I'll delete

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
mucahitkurt: I'm not sure about this, because KEP says that [`assuming that the volumes have already been provisioned`](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190204-scale-testing.md#pod-startup).

If I move this up, then measurement is started before volume provisioning.

msau42: I think we want to separate out provisioning time measurements from pod startup measurements because provisioning is generally a one-time bootstrapping operation (for an application).

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/pod.yaml
mucahitkurt: I'll do it, also will move the volume specific yaml files under the specific volume directory

mucahitkurt:
@wojtek-t thanks a lot for your comments! I tried to reply & fix them, PTAL.

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
wojtek-t: > If I move this up, then measurement is started before volume provisioning.

yes - but the measurement only measures pod startup time. the fact that you create some other objects doesn't affect that at all.

We want to separate out provisioning time - but that means creating those volumes before we create a pod - which is happening in this test no matter when you start the measurement.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
```diff
@@ -0,0 +1,95 @@
...
+      objectTemplatePath: {{$VOLUME_TEMPLATE_PATH}}
+{{ end }}
+# Start measurements
+- measurements:
+  - Identifier: PodWithMultiVolumeStartupLatency
+    Method: PodStartupLatency
+    Params:
+      action: start
+      labelSelector: group = {{$GROUP}}
+      threshold: {{$podStartupTimeout}}s
```
wojtek-t: Medium-term (or maybe even not that medium), we should merge the logic of detecting whether the pod is stateful or stateless into the measurement and use thresholds automatically.

@mm4tt @oxddr - do you know if we have an issue opened for that?

clusterloader2/testing/storage/pod-startup/ephemeral-volumes/config.yaml
msau42: When does pod startup measurement begin?  For ephemeral volumes, kubelet cannot mount API volumes until the API object is created, and containers cannot start until all volumes are mounted. For PVCs, scheduler cannot schedule the pod until the pvs are provisioned and bound to the PVC.

wojtek-t: The "measurement start" action in the config is basically "start watching for pod creations and for those measure pod startup time".
Measuring pod startup time for individual pod starts when that particular pod is created, which in case of this config are created after all volumes already exist.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
```diff
@@ -0,0 +1,95 @@
...
+{{ if $PROVISION_VOLUME }}
+# Provision volumes
+- phases:
+  - namespaceRange:
```
wojtek-t: @krzysied - I was thinking about it in the past, but I don't know if we ever did that (and if I ever shared with you that I was thinking about it).
In majority of cases, we're setting "namespaceRange" to basically all automanaged-namespaces.
What I would like to do is to remove all of those and say:
"if you don't specify namespaceRange, we default to all automanaged namespaces"

Is that done? If not, would you be able to add that (should be quite simple to do).

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
```diff
@@ -0,0 +1,95 @@
...
+        AppName: {{$APP_NAME}}
+- measurements:
+  - Identifier: WaitForRunningPodsWithStorage
+    Method: WaitForRunningPods
+    Params:
+      desiredPodCount: {{$TOTAL_PODS}}
+      labelSelector: group = {{$GROUP}}
+      # TODO decide this after test roll-out phase
+      timeout: 15m
+- measurements:
```
wojtek-t: Similarly as for starting measurement, can you also move it to the bottom?

[As I mentioned somewhere else, medium term, we will simply bake all the necessary thresholds into the measurement itself and we will bundle all SLO-related ones together to limit boilerplate. It will just be simpler to do then :)]

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/configmap/override.yaml
```diff
@@ -0,0 +1,3 @@
+PROVISION_VOLUME: true
+POD_TEMPLATE_PATH: "configmap/pod_with_ConfigMaps.yaml"
+VOLUME_TEMPLATE_PATH: "configmap/ConfigMap.yaml"
```
wojtek-t: Do you somewhere use the assumption that the file names are capitalized ConfigMap (I don't see that, but maybe I'm missing something).
If not, it would be more natural to have:
configmap/config_map.yaml
configmap/pod_with_config_map.yaml

[Same for secrets and others]

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
krzysied: https://github.com/kubernetes/perf-tests/blob/459486e6ac3a7e11411e230ff90041f8f66e9104/clusterloader2/pkg/test/simple_test_executor.go#L312-L315
If namespaceRange is unset, the cluster level object will be created. IIRC currently we are not using this feature. The question is if we want to create non-namespace objects in the future...

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
mucahitkurt: ok., I think deleting pods before the gather action of PodStartupLatency measurement has no effect?

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/configmap/override.yaml
mucahitkurt: There was but no longer, I'll rename it, but volume names are not used as separate words generally, usages are like `ConfigMap`, `EmptyDir` etc. I'll rename them as like `configmap/configmap.yaml`, `configmap/pod_with_configmap.yaml` etc.

wojtek-t: Sure that sounds fine, thanks!

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
wojtek-t: PodStartupLatency measurement can only change if when pods are being created - so yes, that will have no effect.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
wojtek-t: Ahh ok - we definitely would like to be able to create non-namespaced objects (e.g. CRD is non-namespaced).

We should probably come up with some "allNamespaces" thing to reduce amount of boilerplate...

mucahitkurt:
Hi @wojtek-t, @msau42 ,

Is there anything that I should do with this PR, including test roll out phase, I don't know how that will be executed and will be anything that I can contribute?

I'll open another PR for the ephemeral volume test cases that will be run on cluster wide. Thanks!

msau42:
/lgtm
Thanks for taking this on!  @wojtek-t once this merges, how do you want to proceed on establishing a baseline?

wojtek-t:
@msau42 

> Thanks for taking this on! @wojtek-t once this merges, how do you want to proceed on establishing a baseline?

I think we should create a test-suite similar to this one:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L2
[It should probably live in the same place, we may want to add it under experimental tab for sig-scalability.]
After couple days, we should go over tests and see how the results look like (may be useful to expose them in perf-dash, @krzysied may help with that).

Can someone some sig-storage pick it?

/approve
/hold

Holding to get agreement on next steps - feel free to cancel if you ~agree with the above.

mucahitkurt:
> @msau42
> 
> > Thanks for taking this on! @wojtek-t once this merges, how do you want to proceed on establishing a baseline?
> 
> I think we should create a test-suite similar to this one:
> https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L2
> [It should probably live in the same place, we may want to add it under experimental tab for sig-scalability.]
> After couple days, we should go over tests and see how the results look like (may be useful to expose them in perf-dash, @krzysied may help with that).
> 
> Can someone some sig-storage pick it?
> 
> /approve
> /hold
> 
> Holding to get agreement on next steps - feel free to cancel if you ~agree with the above.

@wojtek-t , @msau42 

I have some questions about periodic job configuration, we have 2 test cases, one of them is max volumes for a pod case, other one is max pod-volume(1-1) for a node case. Currently I have one config file, and it seems for first test case, and volumes for pod value is 4(I added randomly).

First question is, should I create two different override files for the 2 different test cases? And periodic job definitions will have two different override files, one for test case specific(max volume for a pod or max volume for a node), other one for volume type?

Second question is, what should my initial values be for the max volumes for a pod case(current value is 4), and max pod-volume(1-1) for a node case? Max limits is 100 pods per node, can I use this for both scenarios?

msau42:
This plan sounds good to me. For max pod-volume, 100 pods with 1 volume each sounds good to me. For max volumes in 1 pod, 100 volumes also sounds fine.  Although, for the 2nd case, how is it measured? Does the test only measure 1 pod? Or does it measure multiple pods serially?

wojtek-t:
> First question is, should I create two different override files for the 2 different test cases? And periodic job definitions will have two different override files, one for test case specific(max volume for a pod or max volume for a node), other one for volume type? So there are 4 ephemeral volume types, 2 test cases, and there will be 8 periodic job definitions?

I think that sounds most reasonable for now.
[Medium term, (assuming we already don't - @krzysied - what is the current state?) we should allow running the same test with different overrides as part of the same test job.]

> This plan sounds good to me. For max pod-volume, 100 pods with 1 volume each sounds good to me. For max volumes in 1 pod, 100 volumes also sounds fine. Although, for the 2nd case, how is it measured? Does the test only measure 1 pod? Or does it measure multiple pods serially?

+1
I think that for now we only create 1 pod and measure startup for it. This has a drawback that it's not statistically valuable. We should try changing to "create+delete" N times and compute percentiles from that. But I'm fine with moving those changes to subsequent PR.

krzysied:
> > First question is, should I create two different override files for the 2 different test cases? And periodic job definitions will have two different override files, one for test case specific(max volume for a pod or max volume for a node), other one for volume type? So there are 4 ephemeral volume types, 2 test cases, and there will be 8 periodic job definitions?

>I think that sounds most reasonable for now.
[Medium term, (assuming we already don't - @krzysied - what is the current state?) we should allow running the same test with different overrides as part of the same test job.]

Yes, currently it requires 8 periodic job definitions. Overrides are passed to the clusterloader as flags so it applies to the job scope, not the test scope.

mucahitkurt:
@wojtek-t @msau42 I added a new override file for the max volumes per node case, and also updated the default volume number to 100.

I'll open a PR to test-infra for the 8 periodic job definitions.

wojtek-t:
@mucahitkurt - next time please squash the commits before merging
[other than that lgtm]

mucahitkurt:
> @mucahitkurt - next time please squash the commits before merging
> [other than that lgtm]

Ok., sorry about that, I forgot.

PR #603, opened 2019-07-04
Update perfdash to 2.2

oxddr:
Also switch perfdash to use debian images (not Google ones).

ref https://github.com/kubernetes/perf-tests/issues/575

/assign @krzysied


krzysied:
/approve
/lgtm

Thanks!

Issue #597, opened 2019-07-03
Measure and expose In-cluster Network Latency SLI


oxddr:
This SLI is defined here: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/network_latency.md

/area slo
/assign @oxddr

PR #609, opened 2019-07-04
Display in-cluster network latency in perf-dash.

oxddr:
ref https://github.com/kubernetes/perf-tests/issues/579

Also rename "KubeProxy" section to "Network".

/hold 
wait until metrics is enable in gce100 and we have some results to display and test this upon

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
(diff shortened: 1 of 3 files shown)


oxddr:
M2p33vFRZXw

/hold cancel
/assign @krzysied

oxddr:

krzysied:
You can bump up perfdash version in the same PR, assuming that you want to push this change to http://perf-dash.k8s.io.

PR #613, opened 2019-07-05
Add probe periodically performing DNS lookups

oxddr:
Also move existing ping prober to a separate directory. 

/assign @wojtek-t @mm4tt


mm4tt:
Please fix Travis CI build

mm4tt:
Why cannot you just add new mode to the exist probe binary? By doing that you'd avoid over 50% of the boilerplate in this PR

oxddr:
Personally I prefer binaries with a single responsibility, mostly because they are easier to understand and reason about. I can discuss this aspect in more details. 

Which pieces of boilerplate are you worried most? Go? Makefiles? Dockefiles? It can be reduced, by re-writing files in more reusable way instead of copying them.

mm4tt:
> Personally I prefer binaries with a single responsibility, mostly because they are easier to understand and reason about. I can discuss this aspect in more details.

I agree with single-responsibility rule, but having probes in the current form, where main is:
```
switch *mode {
	case "ping-client":
		pingclient.Run(pingclient.NewDefaultPingClientConfig())
	case "ping-server":
		pingserver.Run(pingserver.NewDefaultPingServerConfig())
	case "dns":
		dns.Run(pingserver.NewDefaultDNSConfig())
```

and each probe is isolated in its own directory IMHO fulfills a single-responsibility rule. 

There is boilerplate that will be the same for each probe + custom code, that's why we have this framework, what is the benefit in writing everything from scratch for each probe?


> Which pieces of boilerplate are you worried most? Go? Makefiles? Dockefiles? It can be reduced, by re-writing files in more reusable way instead of copying them.

All of them + some assumptions we make about the probes in the clusterloader2 framework, e.g. that every probe exposes metrics. 
I understand that it can be reduced, but IMHO why solve new problems if we can avoid them in a first place?

Summarizing, I believe that a single binary with mode switch is a win-win situation, as we still have single responsibility without any boilerplate overhead.
On the other hand the binary per probe will give us single responsibility, but there will be a lot of costs associated, as a lot of boilerplate (think ~5 new probes this year) + new problem of how to manage this boilerplate + no common denominator of every probe, would be harder to make assumptions in clusterloader / easier to break something.

+wojtekt@ to be a judge here and make a verdict :)

oxddr:
I disagree there are lot of costs associated with having single binary per-prober. 

1. There is very little boilerplate (I'd hardly say this is boilerplate, but's another topic): code needed for a new probe, that's not prober-specific logic is main.go (5 loc) and Dockerfile (17 loc, not shareable by design AFAIK), maybe a line of two in Makefile, assuming will make it more generic. I am not sure what you understand by "managing boilerplate"
2. Adding new probers is to me actually an argument to keep things separate, as 5 new probers increase the affect of shared fate and will make shared main even more complicated. 
3. Probers are simple programs, I'm not sure how putting them into a single binary adds a common denominator (why do we need it in the first place?)
4. clusterloader2 doesn't really make any assumption about probers. It run some binaries and in the end expects some metrics in Prometheus (well, that maybe the only assumption it makes). 

Those probes, that we run are extremely simple programs. I am not sure, what's the added value by sticking them together and thus making them more complicated. Especially if we plan to add that many new probes this year.

Benefits of having separations:
1. Probes could be actually simple, single-line programs. I have a small refactoring in-flight to show this. 
2. No shared fate. 

Neutral:
1. Amount of config needed in clusterloader is the same no matter whether we put probers in the same binary of not.

mm4tt:
> I disagree there are lot of costs associated with having single binary per-prober.
> 
> 1. There is very little boilerplate (I'd hardly say this is boilerplate, but's another topic): code needed for a new probe, that's not prober-specific logic is main.go (5 loc) and Dockerfile (17 loc, not shareable by design AFAIK), maybe a line of two in Makefile, assuming will make it more generic. I am not sure what you understand by "managing boilerplate"

Probe code (probe.go): 73 lines
Changes (excluding refactoring you did) that would be needed for each new probe: Dockerfile (17) + Makefile (17) + main.go (33) + go.mod (8) + go.sum (autogenerated, 53)  = 128 lines 
For me that's a lot of boilerplate, especially if you multiply it by a number of probes we'll end up with hundreds of lines of boilerplate code.
I understand that we could potentially reduce that by some templating and stuff (that's what I meant by "managing boilerplate") but what's the point if we can avoid the boilerplate?

> 2. Adding new probers is to me actually an argument to keep things separate, as 5 new probers increase the affect of shared fate and will make shared main even more complicated.

Sorry, but what is complicated in adding a 2 line  "case" in a switch statement?

Other argument here is that imagine some other time want to write their probe, it would be much easier to explain that they should just write one file with their logic and add a case in the main file. Explaining that they need to create their own: Dockerfile, Makefile, main.go, mod.go would be much complicated especially if they don't have experience in go modules or makefiles or docker.

> 3. Probers are simple programs, I'm not sure how putting them into a single binary adds a common denominator (why do we need it in the first place?)

We'll want to implement liveness and readiness probes for them. We already expose metrics in every probe even if it doesn't expose anything useful (e.g. ping-server) as this is an assumption we use in clusterloader2 to check whether probes are successfully up and the test can begin.

> 4. clusterloader2 doesn't really make any assumption about probers. It run some binaries and in the end expects some metrics in Prometheus (well, that maybe the only assumption it makes).

It does, see the point above. 

> Those probes, that we run are extremely simple programs. I am not sure, what's the added value by sticking them together and thus making them more complicated. Especially if we plan to add that many new probes this year.

We may also want to re-use the code in prod and have a mode in which we're run few probes as a part of single binary to minimize number of pods.
Also in the current state, there is probes framework handling boilerplate and each probe is a very short separate file, I think this is a good isolation. 

> 
> Benefits of having separations:
> 
> 1. Probes could be actually simple, single-line programs. I have a small refactoring in-flight to show this.
> 2. No shared fate.
> 
> Neutral:
> 
> 1. Amount of config needed in clusterloader is the same no matter whether we put probers in the same binary of not.

Overall, I see that we have a different vision here, I think exchanging comments it's not the most efficient way in such case. Let' set up a meeting tomorrow to discuss and find a solution.

oxddr:
Discussed offline. In the future, we would like those probes to run in production clusters, so we want to minimize resource consumption - hence keeping them in a single binary and serving more type from it.

PR #616, opened 2019-07-08
ClusterLoader2: switch RCs to Deployments

mm4tt:
Experimental change to switch deprecated ReplicationControllers to Deployments. 
We'll see if the today's CI/CD scale test passes with this change. If no we'll revert, if yes we'll cleanup and continue using Deployments.

### DIFF
--- a/clusterloader2/testing/density/deployment.yaml
+++ b/clusterloader2/testing/density/deployment.yaml
@@ -1,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  labels:
+    group: {{.Group}}
+spec:
+  replicas: {{.Replicas}}
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        name: {{.Name}}
+        group: {{.Group}}
+    spec:
+      containers:
+      - image: k8s.gcr.io/pause:3.1
+        imagePullPolicy: IfNotPresent
+        name: {{.Name}}
+        ports:
+        resources:
+          requests:
+            cpu: {{.CpuRequest}}
+            memory: {{.MemoryRequest}}
+      # Add not-ready/unreachable tolerations for 15 minutes so that node
+      # failure doesn't trigger pod deletion.
+      tolerations:
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 900
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 900
--- a/clusterloader2/testing/load/deployment.yaml
+++ b/clusterloader2/testing/load/deployment.yaml
@@ -1,0 +1,38 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  labels:
+    group: load
+    svc: {{.SvcName}}-{{.Index}}
+spec:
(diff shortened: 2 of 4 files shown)


clusterloader2/testing/load/deployment.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
```
wojtek-t: Shouldn't this be:
apps/v1
?

clusterloader2/testing/density/deployment.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: v1
```
wojtek-t: Shouldn't this be:
apps/v1
?

mm4tt:
> I would also vote for removing rc.yaml files.

As I said in the description, there would be a cleanup PR tomorrow if the tests pass. There are a few other things we should change, e.g. removing rc from object names, etc.

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/618

krzysied:
WaitForControlledPodsRunning measurements should also be update:
```
      apiVersion: apps/v1
      kind: Deployment
```

mm4tt:
> WaitForControlledPodsRunning measurements should also be update:
> 
> ```
>       apiVersion: apps/v1
>       kind: Deployment
> ```

Thanks, done!

oxddr:
This change seems to affect all of our tests at once, I'd much rather have this rolled out gradually.

mm4tt:
> This change seems to affect all of our tests at once, I'd much rather have this rolled out gradually.

We've discussed it during team meeting today, it's an experiment that we'd rollback tomorrow in case of any issues. 
Monday it's an ideal day to do it, as there will be almost no differences comparing to the yesterday's run.

oxddr:
This change also affects k/k presubmit, not only scalability tests. I'd much rather have it tested on solely our tests first, before it could affect unrelated PRs.

mm4tt:
> This change also affects k/k presubmit, not only scalability tests. I'd much rather have it tested on solely our tests first, before it could affect unrelated PRs.

k/perf-tests presubmit runs the same presubmit as would be run for k/k master. I think it should catch any issue there, if not we'll have a AI to address that.

oxddr:
This PR has broken some k/k presubmits (e.g. [1](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-100-performance/1148480520233947136), [2](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-100-performance/1148480520233947136), [3](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-100-performance/1148473352638173184)) despite passing presubmits here. We really should rollout such changes gradually, without breaking unrelated PRs.

mm4tt:
> This PR has broken some k/k presubmits (e.g. [1](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-100-performance/1148480520233947136), [2](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-100-performance/1148480520233947136), [3](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/directory/pull-kubernetes-e2e-gce-100-performance/1148473352638173184)) despite passing presubmits here. We really should rollout such changes gradually, without breaking unrelated PRs.

Agreed, this was a specific case, we needed to run on monday minimize impact of other changes, but in general we should have a clear rollout policy and stick to it.

PR #622, opened 2019-07-09
Measure pod startup time in load test

wojtek-t:

PR #624, opened 2019-07-11
ClusterLaoder - Adding group label to load test pods

krzysied:
Adding group label to pods, so they can be observed by pod startup latency measurement.


krzysied:
/assign @wojtek-t 
ref #622

krzysied:
/hold cancel
I realized that I updated rc instead of deployment...

PR #629, opened 2019-07-11
ClusterLoader - Informers

krzysied:
Extracting informer and dynamic informer.

### DIFF
--- a/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
@@ -27,13 +27,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/workerqueue"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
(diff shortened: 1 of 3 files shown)


krzysied:
/assign @wojtek-t 
This PR will be rebased after #628 is merged.

clusterloader2/pkg/measurement/util/workerqueue/workerqueue.go
```diff
@@ -0,0 +1,68 @@
+/*
```
wojtek-t: This has merged already, can you please rebase?

krzysied: Done.

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
```diff
@@ -205,11 +165,15 @@ func (w *waitForControlledPodsRunningMeasurement) start() error {
...
-		klog.Infof("%v: create dynamic informer failed.", w)
-		return err
-	}
+	w.informer = informer.NewDynamicInformer(
+		w.clusterFramework.GetDynamicClients().GetClient(),
+		w.gvr,
+		w.namespace,
+		w.fieldSelector,
+		w.labelSelector,
+		w.queue,
```
wojtek-t: Why do we need to pass w.queue? I would assume that it's enough to pass handleObject function (and that already has access to everything it needs).

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -274,14 +276,14 @@ func (p *podStartupLatencyMeasurement) gatherScheduleTimes(c clientset.Interface
 	return nil
 }
 
-func (p *podStartupLatencyMeasurement) checkPod(obj interface{}) {
+func (p *podStartupLatencyMeasurement) checkPod(_, obj interface{}) {
```
wojtek-t: Does this handle nil correctly?

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
wojtek-t: To go back to that - I think that w.queue shouldn't be part of this interface.
If you need to add to the queue as part of handler, handler function should be doing that.
But I can imagine certain cases where we don't want to do that and we shouldn't force users to it.

clusterloader2/pkg/measurement/util/informer/informer.go
```diff
@@ -0,0 +1,99 @@
...
+	queue workerqueue.Interface,
+	handleObj func(interface{}, interface{}),
+) cache.SharedInformer {
+	optionsModifier := func(options *metav1.ListOptions) {
+		options.FieldSelector = fieldSelector
+		options.LabelSelector = labelSelector
+	}
+	listerWatcher := cache.NewFilteredListWatchFromClient(c.CoreV1().RESTClient(), kind, namespace, optionsModifier)
+	informer := cache.NewSharedInformer(listerWatcher, nil, 0)
+	appendHandler(informer, queue, handleObj)
```
wojtek-t: Why not call it addEventHandler (at least it's consistent with the name what it is doing).

clusterloader2/pkg/measurement/common/simple/wait_for_controlled_pods.go
krzysied: Queue removed.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
krzysied: There is a obj cast to pod interface. IIRC casting nil always returns (nil, false).
Anyway, I've just added explicit nil check in method passed to informer.

clusterloader2/pkg/measurement/util/informer/informer.go
krzysied: Renamed.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -129,20 +135,22 @@ func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
...
+		p.fieldSelector,
+		p.labelSelector,
+		func(_, newObj interface{}) {
+			if newObj == nil {
+				return
+			}
+			f := func() {
+				p.checkPod(newObj)
+			}
+			p.queue.Add(&f)
```
wojtek-t: Why do we need queue here?
We didn't have it and I would prefer not to change the logic in refactoring PRs.

krzysied: checkPod method is grabbing lock. I though that maybe when there is a burst of event it can cause watch to lag. Adding a queue would solve the problem - it behaves as a buffer.

wojtek-t: That's fine - but please avoid changing logic in larger refactoring PR. If you want to do that, please move that change to a separate PR.

krzysied: Ok. I removed the queue.

PR #631, opened 2019-07-12
ClusterLoader - Adding exec service

krzysied:


clusterloader2/pkg/proxypod/manifest/proxy_pod.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Pod
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+spec:
+  containers:
+  - name: ubuntu
+    image: ubuntu:latest
+    command: [ "/bin/bash", "-c", "--" ]
+    args: [ "apt-get update; apt-get install --assume-yes curl; while true; do sleep 30; done;" ]
```
wojtek-t: Can't we really take something existing?

clusterloader2/pkg/proxypod/manifest/proxy_pod.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+spec:
+  containers:
+  - name: ubuntu
+    image: ubuntu:latest
```
wojtek-t: I don't think we need ubuntu.
Also, latest is always a bad idea.

clusterloader2/cmd/clusterloader.go
```diff
@@ -80,6 +81,7 @@ func validateClusterFlags() *errors.ErrorList {
 func initFlags() {
 	flags.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
 	flags.BoolEnvVar(&clusterLoaderConfig.EnablePrometheusServer, "enable-prometheus-server", "ENABLE_PROMETHEUS_SERVER", false, "Whether to set-up the prometheus server in the cluster.")
+	flags.BoolEnvVar(&clusterLoaderConfig.EnableProxyPod, "enable-proxy-pod", "ENABLE_PROXY_POD", false, "Whether to set-up the proxy pod in the cluster.")
```
wojtek-t: I would personally say that we should call it "maintainance", and instead of having a single pod (which is always a bad idea), create a deployment with a service underneath.

@mm4tt - wdyt about "maintainance" name?

clusterloader2/pkg/proxypod/proxy_pod.go
```diff
@@ -0,0 +1,93 @@
...
+		proxyPodPath,
+		mapping,
+		client.Retry(apierrs.IsNotFound)); err != nil {
+		return fmt.Errorf("pod %s creation error: %v", ProxyPodName, err)
+	}
+	if err := wait.Poll(
+		proxyPodCheckInterval,
+		proxyPodCheckTimeout,
+		func() (bool, error) {
+			obj, err := f.GetObject(schema.GroupVersionKind{Version: "v1", Kind: "pod"}, ProxyPodNamespace, ProxyPodName)
```
wojtek-t: This is a bit strange - why we don't use typed client?

clusterloader2/pkg/proxypod/proxy_pod.go
```diff
@@ -0,0 +1,93 @@
...
+	return nil
+}
+
+// TearDownProxyPod deletes proxy pod.
+func TearDownProxyPod(f *framework.Framework) error {
+	klog.Info("Tearing down proxy pod")
+	if err := f.DeleteObject(schema.GroupVersionKind{Version: "v1", Kind: "pod"}, ProxyPodNamespace, ProxyPodName); err != nil {
+		return fmt.Errorf("deleting %s pod error: %v", ProxyPodName, err)
+	}
+	if err := client.DeleteNamespace(f.GetClientSets().GetClient(), ProxyPodNamespace); err != nil {
```
wojtek-t: DeleteNamespace should remove everything in it.
So unless we won't create "externally load-balanced" service, you don't need to remove pod.

clusterloader2/pkg/proxypod/manifest/proxy_pod.yaml
krzysied: I changed image to "agnhost:2.2" which is used to execute curl in kubernetes e2e tests.

clusterloader2/pkg/proxypod/proxy_pod.go
krzysied: Fixed.

clusterloader2/pkg/proxypod/proxy_pod.go
krzysied: Removed.

clusterloader2/cmd/clusterloader.go
mm4tt: Hmmm, what about "support" or "test-util" ?

krzysied: I changed pod to deployment. It's named "exec-pod" in "cluster-loader" namespace.

clusterloader2/cmd/clusterloader.go
```diff
@@ -80,6 +81,7 @@ func validateClusterFlags() *errors.ErrorList {
 func initFlags() {
 	flags.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
 	flags.BoolEnvVar(&clusterLoaderConfig.EnablePrometheusServer, "enable-prometheus-server", "ENABLE_PROMETHEUS_SERVER", false, "Whether to set-up the prometheus server in the cluster.")
+	flags.BoolEnvVar(&clusterLoaderConfig.EnableExecPod, "enable-exec-pod", "ENABLE_EXEC_POD", false, "Whether to set-up the exec pod in the cluster.")
```
wojtek-t: It's no longer a pod - it's a deployment.
But I'm not convinced it should even be called "enable-exec-deployment".

Maybe even more like "enable-exec-service"? [I mean more high-level service meaning, than specific Kubernetes service meaning.]

@mm4tt - thought?

clusterloader2/pkg/execpod/exec_pod.go
```diff
@@ -0,0 +1,104 @@
...
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+)
+
+const (
+	execPodDeploymentNamespace = "cluster-loader"
+	execPodDeploymentName      = "exec-pod"
+	execPodReplicas            = 3
+	execPodSelector            = "pod = exec"
```
wojtek-t: pod is a bad label.
type? feature?

clusterloader2/pkg/execpod/exec_pod.go
```diff
@@ -0,0 +1,104 @@
...
+	execPodPath                = "pkg/execpod/manifest/exec_pod.yaml"
+
+	execPodCheckInterval = 10 * time.Second
+	execPodCheckTimeout  = 2 * time.Minute
+)
+
+// SetUpExecPod creates exec pod.
+func SetUpExecPod(f *framework.Framework) error {
```
wojtek-t: You're not setting up a pod - you're setting something higher-level.

I'm leaning towards saying SetUpExecService, but this collides with k8s Service, which is not my goal.
But definitely not a Pod.

clusterloader2/pkg/execpod/exec_pod.go
```diff
@@ -0,0 +1,104 @@
...
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+)
+
+const (
+	execPodDeploymentNamespace = "cluster-loader"
```
wojtek-t: execDeploymentNamespace?

clusterloader2/pkg/execpod/exec_pod.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := client.CreateNamespace(f.GetClientSets().GetClient(), execPodDeploymentNamespace); err != nil {
+		return fmt.Errorf("namespace %s creation error: %v", execPodDeploymentNamespace, err)
+	}
+	if err := f.ApplyTemplatedManifests(
+		execPodPath,
+		mapping,
+		client.Retry(apierrs.IsNotFound)); err != nil {
+		return fmt.Errorf("pod %s creation error: %v", execPodDeploymentName, err)
+	}
+	ps, err := measurementutil.NewPodStore(f.GetClientSets().GetClient(), execPodDeploymentNamespace, execPodSelector, "")
```
wojtek-t: Can't we reuse some functionality of WaitForPods measumrent? Maybe we can extract this code somehow?

clusterloader2/pkg/execpod/exec_pod.go
krzysied: IMHO the feature we are setting is a pod that can be used for `kubectl exec`. The deployment (and many replicas of the pod) are only to ensure pod availability.

wojtek-t: I disagree - the feature that we enable is to allow "running some custom command from within a cluster".

@mm4tt

clusterloader2/pkg/execpod/exec_pod.go
krzysied: Done.

clusterloader2/pkg/execpod/exec_pod.go
krzysied: Changed to "feature".

clusterloader2/pkg/execpod/exec_pod.go
krzysied: IIUC it means that `execpod` instead of providing pod, it should rather execute given command?

wojtek-t: Kind-of.
The helper function that you expose should be "RunCommand" (or sth like that) instead of "GetExecPod".

clusterloader2/cmd/clusterloader.go
krzysied: Changed to "enable-exec-service".

clusterloader2/pkg/execpod/exec_pod.go
krzysied: Done.

clusterloader2/pkg/execpod/exec_pod.go
krzysied: Done. #638

clusterloader2/cmd/clusterloader.go
```diff
@@ -80,6 +81,7 @@ func validateClusterFlags() *errors.ErrorList {
 func initFlags() {
 	flags.StringVar(&clusterLoaderConfig.ReportDir, "report-dir", "", "Path to the directory where the reports should be saved. Default is empty, which cause reports being written to standard output.")
 	flags.BoolEnvVar(&clusterLoaderConfig.EnablePrometheusServer, "enable-prometheus-server", "ENABLE_PROMETHEUS_SERVER", false, "Whether to set-up the prometheus server in the cluster.")
+	flags.BoolEnvVar(&clusterLoaderConfig.EnableExecService, "enable-exec-service", "ENABLE_EXEC_SERVICE", false, "Whether to enable exec service for the measurements.")
```
mm4tt: "for the measurements" - this is an implementation detail that this service is going to be used from measurements, but doesn't explain the purpose of the exec service. Could you rewrite? Something like, ... exec service that allows executing arbitrary commands from a pod running in the cluster, e.g. calling curl on an IP address to check connectivity.

clusterloader2/pkg/execservice/exec_service.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+)
+
+const (
+	execDeploymentNamespace = "cluster-loader"
+	execDeploymentName      = "exec-pod"
+	execDeploymentPath      = "pkg/execservice/manifest/exec_deployment.yaml"
+	execPodReplicas         = 3
```
serathius, line 40: How did we select 3 replicas?

PR #636, opened 2019-07-15
Clean up and finalize switching RCs to Deployments in performance tests

mm4tt:
For the justification behind the new threshold and in general going forward with Deployments instead of RCs see https://github.com/kubernetes/kubernetes/issues/79898#issuecomment-511393121


clusterloader2/testing/density/config.yaml
```diff
@@ -136,15 +136,15 @@ steps:
     replicasPerNamespace: {{$latencyReplicas}}
     tuningSet: Uniform5qps
     objectBundle:
-    - basename: latency-pod-rc
+    - basename: latency-pod-deployment
```
wojtek-t: latency-deployment?

clusterloader2/testing/density/config.yaml
```diff
@@ -157,10 +157,10 @@ steps:
     replicasPerNamespace: 0
     tuningSet: Uniform5qps
     objectBundle:
-    - basename: latency-pod-rc
+    - basename: latency-pod-deployment
```
wojtek-t: latency-deployment?

mm4tt: Done.

clusterloader2/testing/density/config.yaml
mm4tt: Done.

wojtek-t:
/lgtm
/approve

/hold
@mm4tt - let's merge it tomorrow morning, so that we don't break anything unintentially.

PR #639, opened 2019-07-16
Old style exceptions -> new style for Python 3

cclauss:
Old style exceptions are syntax errors in Python 3 but new style exceptions work as expected in both Python 2 and Python 3.

Also fix __reduce()__ and __printf()__ ;-)

### DIFF
--- a/dns/py/data.py
+++ b/dns/py/data.py
@@ -100,6 +100,7 @@ class Parser(object):
 
   def _compute_derived(self):
     # Note: not very efficient, but functional
+    from functools import reduce
     histogram = reduce(
         list.__add__,
         [[rtt]*count for rtt, count in self.histogram],
--- a/dns/py/runner.py
+++ b/dns/py/runner.py
@@ -271,7 +271,7 @@ class Runner(object):
         results['data']['max_kubedns_cpu'] = res_usage.get()
         results['data']['max_kubedns_memory'] = res_usage.get()
         results['data']['histogram'] = parser.histogram
-      except Exception, exc:
+      except Exception as exc:
         _log.error('Error parsing results: %s', exc)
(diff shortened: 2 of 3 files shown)


clusterloader/vendor/github.com/ugorji/go/codec/test.py
```diff
@@ -98,15 +99,15 @@ def myStopRpcServer():
 def doRpcClientToPythonSvc(port):
     address = msgpackrpc.Address('localhost', port)
     client = msgpackrpc.Client(address, unpack_encoding='utf-8')
-    print client.call("Echo123", "A1", "B2", "C3")
-    print client.call("EchoStruct", {"A" :"Aa", "B":"Bb", "C":"Cc"})
+    print(client.call("Echo123", "A1", "B2", "C3"))
```
wojtek-t: You can't change whatever is in vendor/ directories. Please remove changes from this and the next file.

cclauss: I will change but...

This is the problem with vendoring in...  When you do that, you _own_ the code... Syntax Errors and all.  If you do not want to own the code, then do not vendor it in but pip install it instead.

cclauss: Also note that these changes are already landed in upstream.

wojtek-t: Also note that this whole directory was removed completely in the meantime :)

cclauss: Good news.  Thanks.

wojtek-t:
/ok-to-test

@cclauss - please also squash commits

PR #643, opened 2019-07-16
Increase fluentd memoryConstraint to 260 MB

mm4tt:
Fixing https://github.com/kubernetes/kubernetes/issues/77492#issuecomment-511785738

This should deflake the test, the increase was most likely caused by fluentd version upgrade.
The 260MB seems as a reasonable limit currently:
wbKugbOALxy

### DIFF
--- a/clusterloader2/testing/density/100_nodes/constraints.yaml
+++ b/clusterloader2/testing/density/100_nodes/constraints.yaml
@@ -10,7 +10,7 @@ coredns:
   memoryConstraint: 31457280 #30 * (1024 * 1024)
 fluentd-gcp:
   cpuConstraint: 1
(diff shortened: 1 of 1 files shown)


wojtek-t:
/hold

Where is this coming from? We should know why we're bumping it.

mm4tt:
@wojtek-t, see https://github.com/kubernetes/kubernetes/issues/77492#issuecomment-511785738

wojtek-t:
ok - I thought it's something new - this should probably say "Fix that issue" in the first description of this PR
[feel free to cancel once fixed]

mm4tt:
Thanks, updated the PR description.

/hold cancel

mm4tt:
/assign @wojtek-t 
As you've already taken a look :)

PR #645, opened 2019-07-16
ClusterLoader - Moving common/simple/ content to common/

### LINKED ISSUE #642: Move all contents of clusterloader2/pkg/measurement/common/simple/ down to common/
/assign @krzysied

krzysied:
/fixes #642

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -37,9 +37,9 @@ import (
 	"k8s.io/perf-tests/clusterloader2/pkg/test"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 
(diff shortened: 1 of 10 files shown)


mm4tt:
/assing @oxddr 

Janek said he had some pending PRs in this directory, so moving it may currently create more harm than gain :)

wojtek-t:
> Janek said he had some pending PRs in this directory, so moving it may currently create more harm than gain :)

I'm fine with waiting couple days, but there will always be open PRs against it.
And note that if you move them via git mv - git rebase just picks it, so it shouldn't add much more pain.

oxddr:
Thanks for the cleanup 

However let me cleanup backlog of PR touching this directory and then merge this PR, so 
/hold

PR #647, opened 2019-07-16
ClusterLoader - Extracting informer start and sync logic

krzysied:


clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -179,18 +179,7 @@ func (w *waitForControlledPodsRunningMeasurement) start() error {
...
-	timeoutTimer := time.AfterFunc(informerSyncTimeout, func() {
-		close(timeoutCh)
-	})
-	defer timeoutTimer.Stop()
-	if !cache.WaitForCacheSync(timeoutCh, w.informer.HasSynced) {
-		return fmt.Errorf("timed out waiting for caches to sync")
-	}
-
-	return nil
+	return informer.StartAndSync(w.informer, w.stopCh, informerSyncTimeout)
```
wojtek-t: We need to be consistent - either you use "w.informer" in all measurements to detect if it's running (then change the measurement above) or you introduce a dedicated isRunning (and then this one has to be changed).

[And potentially other in both options.]

krzysied: I unified it. Now it uses `isRunning` in both metrics.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -131,31 +129,23 @@ func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
...
-	}
-	return nil
+	return informer.StartAndSync(i, p.stopCh, informerSyncTimeout)
 }
 
 func (p *podStartupLatencyMeasurement) stop() {
 	if p.isRunning {
 		p.isRunning = false
-		close(p.stopCh)
+		if p.stopCh != nil {
```
wojtek-t: p.isRunning is set together with p.stopCh.

The measurement is not thread-safe anyway, so I don't think this if is needed for anything.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -134,7 +132,9 @@ func (w *waitForControlledPodsRunningMeasurement) Execute(config *measurement.Me
 
 // Dispose cleans up after the measurement.
 func (w *waitForControlledPodsRunningMeasurement) Dispose() {
-	close(w.stopCh)
+	if w.stopCh != nil {
```
wojtek-t: if w.isRunning

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -150,23 +150,21 @@ func (*waitForControlledPodsRunningMeasurement) String() string {
 }
 
 func (w *waitForControlledPodsRunningMeasurement) start() error {
+	if w.isRunning {
+		klog.Infof("%v: wait for controlled pods measurement already running", w)
+		return nil
+	}
+	klog.Infof("%v: starting wait for controlled pods measurement...", w)
+	w.isRunning = true
```
wojtek-t: initialize w.isRunning together with w.stopCh

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
krzysied: Done.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go

PR #655, opened 2019-07-18
Add flag to override the prometheus snapshot name

mm4tt:
This will be useful when running multiple similar tests locally to better organize the snaphots.
E.g. when running golang1.13 test we've already tried over 5 custom k8s versions, it'd be easier if I can give a snapshot a meaningful name (e.g. golang1.12.6-basline) rather than using the auto-generated pd name (e.g. golang-kubemark-dynami-pvc-1a775eaa-22bb-405a-81a5-eb2dd545168c)


mm4tt:
> /lgtm
> But it still requires entry in known-flags...

Doh, thanks, fixed

mm4tt:
@krzysied, PTAL

PR #656, opened 2019-07-18
add function to wait for bound PVCs

hantaowang:
Adds a function `WaitForBoundPVCs` that works very similarly to `WaitForRunningPods` and waits until the desired number of PVCs becomes "Bound". 

The code is very similar to the code for `WaitForRunningPods` - I played with experimenting on a way to generalize the similar parts into a generic `WaitForResource` but I am not sure how easy or feasible it will be to actually complete it. In the mean time, this does it with just some repeated code and logic.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_pvcs.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_pvcs.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"time"
+
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	defaultWaitForPVCsTimeout       = 60 * time.Second
+	defaultWaitForPVCsInterval      = 5 * time.Second
+	waitForBoundPVCsMeasurementName = "WaitForBoundPVCs"
+)
+
+func init() {
+	if err := measurement.Register(waitForBoundPVCsMeasurementName, createWaitForBoundPVCsMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", waitForBoundPVCsMeasurementName, err)
+	}
+}
+
+func createWaitForBoundPVCsMeasurement() measurement.Measurement {
+	return &waitForBoundPVCsMeasurement{}
+}
+
+type waitForBoundPVCsMeasurement struct{}
+
+// Execute waits until desired number of PVCs are bound or until timeout happens.
+// PVCs can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+func (w *waitForBoundPVCsMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	desiredPVCCount, err := util.GetInt(config.Params, "desiredPVCCount")
+	if err != nil {
+		return nil, err
+	}
+	selector := measurementutil.NewObjectSelector()
+	if err := selector.Parse(config.Params); err != nil {
+		return nil, err
+	}
+	timeout, err := util.GetDurationOrDefault(config.Params, "timeout", defaultWaitForPVCsTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	stopCh := make(chan struct{})
+	time.AfterFunc(timeout, func() {
+		close(stopCh)
+	})
+	options := &measurementutil.WaitForPVCOptions{
+		Selector:            selector,
+		DesiredPVCCount:     desiredPVCCount,
+		EnableLogging:       true,
+		CallerName:          w.String(),
+		WaitForPVCsInterval: defaultWaitForPVCsInterval,
+	}
+	return nil, measurementutil.WaitForPVCs(config.ClusterFramework.GetClientSets().GetClient(), stopCh, options)
+}
+
+// Dispose cleans up after the measurement.
+func (*waitForBoundPVCsMeasurement) Dispose() {}
+
+// String returns a string representation of the measurement.
+func (*waitForBoundPVCsMeasurement) String() string {
+	return waitForBoundPVCsMeasurementName
+}
--- a/clusterloader2/pkg/measurement/util/pvcs.go
+++ b/clusterloader2/pkg/measurement/util/pvcs.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// PVCStartupStatus represents phase of a pvc group.
+type PVCStartupStatus struct {
+	Pending  int
+	Bound    int
+	Lost     int
+	Expected int
+	Created  int
+}
+
+// String returns string representation for PVCsStartupStatus.
+func (s *PVCStartupStatus) String() string {
+	return fmt.Sprintf("PVCs: %d out of %d created, %d bound, %d pending, %d lost", s.Created, s.Expected, s.Bound, s.Pending, s.Lost)
+}
+
+// ComputePVCStartupStatus computes PVCsStartupStatus for a group of PVCs.
+func ComputePVCStartupStatus(pvcs []*corev1.PersistentVolumeClaim, expected int) PVCStartupStatus {
+	startupStatus := PVCStartupStatus{
+		Expected: expected,
+	}
+	for _, p := range pvcs {
+		startupStatus.Created++
+		if p.Status.Phase == corev1.ClaimPending {
+			startupStatus.Pending++
+		} else if p.Status.Phase == corev1.ClaimBound {
+			startupStatus.Bound++
+		} else if p.Status.Phase == corev1.ClaimLost {
+			startupStatus.Lost++
+		}
+	}
+	return startupStatus
+}
+
+type pvcInfo struct {
+	oldPhase string
+	phase    string
+}
+
+// PVCDiff represets diff between old and new group of pvcs.
+type PVCDiff map[string]*pvcInfo
+
+// Print formats and prints the give PVCDiff.
+func (p PVCDiff) String(ignorePhases sets.String) string {
+	ret := ""
+	for name, info := range p {
+		if ignorePhases.Has(info.phase) {
+			continue
+		}
+		if info.phase == nonExist {
+			ret += fmt.Sprintf("PVC %v was deleted, had phase %v\n", name, info.oldPhase)
+			continue
+		}
+		msg := fmt.Sprintf("PVC %v ", name)
+		if info.oldPhase != info.phase {
+			if info.oldPhase == nonExist {
+				msg += fmt.Sprintf("in phase %v ", info.phase)
+			} else {
+				msg += fmt.Sprintf("went from phase: %v -> %v ", info.oldPhase, info.phase)
+			}
+			ret += msg + "\n"
+		}
+	}
+	return ret
+}
+
+// DeletedPVCs returns a slice of PVCs that were present at the beginning
+// and then disappeared.
+func (p PVCDiff) DeletedPVCs() []string {
+	var deletedPVCs []string
+	for pvcName, pvcInfo := range p {
+		if pvcInfo.phase == nonExist {
+			deletedPVCs = append(deletedPVCs, pvcName)
+		}
+	}
+	return deletedPVCs
+}
+
+// AddedPVCs returns a slice of PVCs that were added.
+func (p PVCDiff) AddedPVCs() []string {
+	var addedPVCs []string
+	for pvcName, pvcInfo := range p {
+		if pvcInfo.oldPhase == nonExist {
+			addedPVCs = append(addedPVCs, pvcName)
+		}
+	}
+	return addedPVCs
+}
+
+// DiffPVCs computes a PVCDiff given 2 lists of PVCs.
+func DiffPVCs(oldPVCs []*corev1.PersistentVolumeClaim, curPVCs []*corev1.PersistentVolumeClaim) PVCDiff {
+	pvcInfoMap := PVCDiff{}
+
+	// New PVCs will show up in the curPVCs list but not in oldPVCs. They have oldhostname/phase == nonexist.
+	for _, pvc := range curPVCs {
+		pvcInfoMap[pvc.Name] = &pvcInfo{phase: string(pvc.Status.Phase), oldPhase: nonExist}
+	}
+
+	// Deleted PVCs will show up in the oldPVCs list but not in curPVCs. They have a hostname/phase == nonexist.
+	for _, pvc := range oldPVCs {
+		if info, ok := pvcInfoMap[pvc.Name]; ok {
+			info.oldPhase = string(pvc.Status.Phase)
+		} else {
+			pvcInfoMap[pvc.Name] = &pvcInfo{phase: nonExist, oldPhase: string(pvc.Status.Phase)}
+		}
+	}
+	return pvcInfoMap
+}
--- a/clusterloader2/pkg/measurement/util/wait_for_pvcs.go
+++ b/clusterloader2/pkg/measurement/util/wait_for_pvcs.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+// WaitForPVCOptions is an options used by WaitForPVCs methods.
+type WaitForPVCOptions struct {
+	Selector            *ObjectSelector
+	DesiredPVCCount     int
+	EnableLogging       bool
+	CallerName          string
+	WaitForPVCsInterval time.Duration
+}
+
+// WaitForPVCs waits till disire nuber of PVCs is running.
+// PVCs are be specified by namespace, field and/or label selectors.
+// If stopCh is closed before all PVCs are running, the error will be returned.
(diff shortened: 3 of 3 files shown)


hantaowang:
/assign @wojtek-t @davidz627
cc @jingxu97

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
```diff
@@ -0,0 +1,96 @@
...
+	EnableLogging       bool
+	CallerName          string
+	WaitForPVCsInterval time.Duration
+}
+
+// WaitForPVCs waits till disire nuber of PVCs is running.
+// PVCs are be specified by namespace, field and/or label selectors.
+// If stopCh is closed before all PVCs are running, the error will be returned.
+func WaitForPVCs(clientSet clientset.Interface, stopCh <-chan struct{}, options *WaitForPVCOptions) error {
+	// TODO(#269): Change to shared PVCStore.
```
wojtek-t: It's fine to remove this todo.

clusterloader2/pkg/measurement/util/pvc_store.go
```diff
@@ -0,0 +1,89 @@
...
+
+// PVCStore is a convenient wrapper around cache.Store that returns list of v1.PVC instead of interface{}.
+type PVCStore struct {
+	cache.Store
+	stopCh    chan struct{}
+	Reflector *cache.Reflector
+}
+
+// NewPVCStore creates PVCStore based on given namespace, label selector and field selector.
+func NewPVCStore(c clientset.Interface, namespace string, labelSelector string, fieldSelector string) (*PVCStore, error) {
```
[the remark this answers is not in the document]
wojtek-t: In the meantime, please use ObjectSelector introduced in https://github.com/kubernetes/perf-tests/pull/660/files (hopefully it will merge before you will read this comment).

clusterloader2/pkg/measurement/util/pvcs.go
```diff
@@ -0,0 +1,129 @@
...
+		} else if p.Status.Phase == corev1.ClaimBound {
+			startupStatus.Bound++
+		} else if p.Status.Phase == corev1.ClaimLost {
+			startupStatus.Lost++
+		}
+	}
+	return startupStatus
+}
+
+type pvcInfo struct {
```
wojtek-t: I'm wondering if we need that.

The reason why we added that for pods is that pod may be running and turn into not-running after that.

@msau42 - Is it the same case for PVCs?
I mean, is it possible that claim went to Bound state and then it will go to Lost state?

If not, I would say we should remove it, as this whole logic for diffs, isn't really needed.
[For the context: we did this for pods to avoid silently skipping the situations that a lot of pods are silently dying, but because new ones are created by RS, we are eventually converging to the desired state.]

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
```diff
@@ -0,0 +1,94 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	defaultWaitForPVCsTimeout       = 60 * time.Second
```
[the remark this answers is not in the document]
hantaowang: My testing scripts set a timeout rather than use the default. I think 60 sec is a good upper bound for non stress situations.

clusterloader2/pkg/measurement/util/pvc_store.go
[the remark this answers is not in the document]
hantaowang: OK, that can be a different PR. For this one i can incorporate the object selector.

clusterloader2/pkg/measurement/util/pvcs.go
hantaowang: It is possible if the underlying PV is lost. See: https://github.com/kubernetes/kubernetes/issues/60298#issuecomment-367997711

msau42: Lost is a possible state for a PVC to be in, but I wouldn't expect that to happen in a normal test run.  Unless there is some bug (and in that case it would be good to catch too).

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
hantaowang: ack

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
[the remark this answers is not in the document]
wojtek-t: Well - the purpose of scale tests is actually stress, so I would rather be quite conservative in setting this timeout.

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := measurement.Register(waitForBoundPVCsMeasurementName, createWaitForBoundPVCsMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", waitForBoundPVCsMeasurementName, err)
+	}
+}
+
+func createWaitForBoundPVCsMeasurement() measurement.Measurement {
+	return &waitForBoundPVCsMeasurement{}
+}
+
+type waitForBoundPVCsMeasurement struct{}
```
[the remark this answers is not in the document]
wojtek-t: @krzysied - we should do that for all existing measurements.

clusterloader2/pkg/measurement/util/pvc_store.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// NewPVCStore creates PVCStore based on given namespace, label selector and field selector.
+func NewPVCStore(c clientset.Interface, selector *ObjectSelector) (*PVCStore, error) {
+	lw := &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			options.LabelSelector = selector.LabelSelector
+			options.FieldSelector = selector.FieldSelector
+			obj, err := c.CoreV1().PersistentVolumeClaims(selector.Namespace).List(options)
+			return runtime.Object(obj), err
```
wojtek-t: Why is this cast needed?

clusterloader2/pkg/measurement/util/pvcs.go
wojtek-t: Got it - thanks helps. Thanks!

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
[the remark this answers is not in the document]
krzysied: All measurements register create function: func() -> Measurement. IMHO this should trigger type check during compilation.

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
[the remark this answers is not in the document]
hantaowang: That hard part about this is the wait time for PVCs depend heavily on the # of PVCs, since the backend can only provision disks so fast. I dont think its possible to set a good default value that wont be too low or too high. In my tests, what I do is set this to be `MAX(10 sec, NUM_PVCS * PER_PVC_SLO)`

clusterloader2/pkg/measurement/util/pvc_store.go
hantaowang: I don't think  it is. Happy to change it in this PR or open a new one since the PodStore is done this way too.

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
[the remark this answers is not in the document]
hantaowang: The create measurement func should do the type check because it returns that waitForBoundPVCsMeasurement struct as the desired Measurement interface.

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Execute waits until desired number of PVCs are bound or until timeout happens.
+// PVCs can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all-namespace scope is assumed.
+func (w *waitForBoundPVCsMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	desiredPVCCount, err := util.GetInt(config.Params, "desiredPVCCount")
```
davidz627: is this the desired amt of PVC's created or bound? I feel like it might be bound in which case your should specify that

clusterloader2/pkg/measurement/util/pvc_store.go
```diff
@@ -0,0 +1,89 @@
...
+	reflector := cache.NewNamedReflector(name, lw, &v1.PersistentVolumeClaim{}, store, 0)
+	go reflector.Run(stopCh)
+	if err := wait.PollImmediate(50*time.Millisecond, 2*time.Minute, func() (bool, error) {
+		if len(reflector.LastSyncResourceVersion()) != 0 {
+			return true, nil
+		}
+		return false, nil
+	}); err != nil {
+		close(stopCh)
+		return nil, err
```
davidz627: add context to `err` I don't think `PollImmediate` will return an error with good enough context

clusterloader2/pkg/measurement/util/pvc_store.go
```diff
@@ -0,0 +1,89 @@
...
+			return c.CoreV1().PersistentVolumeClaims(selector.Namespace).Watch(options)
+		},
+	}
+	store := cache.NewStore(cache.MetaNamespaceKeyFunc)
+	stopCh := make(chan struct{})
+	name := fmt.Sprintf("PVCStore: %s", selector.String())
+	reflector := cache.NewNamedReflector(name, lw, &v1.PersistentVolumeClaim{}, store, 0)
+	go reflector.Run(stopCh)
+	if err := wait.PollImmediate(50*time.Millisecond, 2*time.Minute, func() (bool, error) {
+		if len(reflector.LastSyncResourceVersion()) != 0 {
```
davidz627: is this the best way to find out whether reflector is synced. no `IsSynced` type func?

clusterloader2/pkg/measurement/util/pvc_store.go
```diff
@@ -0,0 +1,89 @@
...
+		close(stopCh)
+		return nil, err
+	}
+	return &PVCStore{Store: store, stopCh: stopCh, Reflector: reflector}, nil
+}
+
+// List returns to list of PVCs (that satisfy conditions provided to NewPVCStore).
+func (s *PVCStore) List() []*v1.PersistentVolumeClaim {
+	objects := s.Store.List()
+	pvcs := make([]*v1.PersistentVolumeClaim, 0)
```
davidz627: why are we specifically making a list of `0` size? why not len(objects)

clusterloader2/pkg/measurement/util/pvcs.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// ComputePVCStartupStatus computes PVCsStartupStatus for a group of PVCs.
+func ComputePVCStartupStatus(pvcs []*corev1.PersistentVolumeClaim, expected int) PVCStartupStatus {
+	startupStatus := PVCStartupStatus{
+		Expected: expected,
+	}
+	for _, p := range pvcs {
+		startupStatus.Created++
+		if p.Status.Phase == corev1.ClaimPending {
```
davidz627: nit: switch statement?

clusterloader2/pkg/measurement/util/pvcs.go
```diff
@@ -0,0 +1,131 @@
...
+		} else if p.Status.Phase == corev1.ClaimLost {
+			startupStatus.Lost++
+		}
+	}
+	return startupStatus
+}
+
+type pvcInfo struct {
+	oldPhase string
+	phase    string
```
davidz627: create a type for these "phases" and define const's to represent the different phases instead of doing a bunch of string things

davidz627: or use the phase type from k8s

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
```diff
@@ -0,0 +1,92 @@
...
+// If stopCh is closed before all PVCs are running, the error will be returned.
+func WaitForPVCs(clientSet clientset.Interface, stopCh <-chan struct{}, options *WaitForPVCOptions) error {
+	ps, err := NewPVCStore(clientSet, options.Selector)
+	if err != nil {
+		return fmt.Errorf("PVC store creation error: %v", err)
+	}
+	defer ps.Stop()
+
+	var pvcsStatus PVCStartupStatus
+	scaling := uninitialized
```
davidz627: what is `uninitialized`? Why is this not a variable declaration i.e. `var scaling *Type`

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
```diff
@@ -0,0 +1,92 @@
...
+				options.DesiredPVCCount, options.Selector.Namespace, options.Selector.LabelSelector, options.Selector.FieldSelector, pvcsStatus.Bound)
+		case <-time.After(options.WaitForPVCsInterval):
+			pvcs := ps.List()
+			pvcsStatus = ComputePVCStartupStatus(pvcs, options.DesiredPVCCount)
+			if scaling != uninitialized {
+				diff := DiffPVCs(oldPVCs, pvcs)
+				deletedPVCs := diff.DeletedPVCs()
+				if scaling != down && len(deletedPVCs) > 0 {
+					klog.Errorf("%s: %s: %d PVCs disappeared: %v", options.CallerName, options.Selector.String(), len(deletedPVCs), strings.Join(deletedPVCs, ", "))
+					klog.Infof("%s: %v", options.CallerName, diff.String(sets.NewString()))
```
davidz627: why error and log immediately after this. This info looks the same as the one in the other if statement. logic could probably be cleaned up a bit here

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
```diff
@@ -0,0 +1,92 @@
...
+				}
+				addedPVCs := diff.AddedPVCs()
+				if scaling != up && len(addedPVCs) > 0 {
+					klog.Errorf("%s: %s: %d PVCs appeared: %v", options.CallerName, options.Selector.String(), len(deletedPVCs), strings.Join(deletedPVCs, ", "))
+					klog.Infof("%s: %v", options.CallerName, diff.String(sets.NewString()))
+				}
+			} else {
+				switch {
```
davidz627: why don't we set this much earlier before the for loop instead of using uninitialized and having an else statement do initialization once in a branch

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
```diff
@@ -0,0 +1,92 @@
...
+				case len(pvcs) == options.DesiredPVCCount:
+					scaling = none
+				case len(pvcs) < options.DesiredPVCCount:
+					scaling = up
+				case len(pvcs) > options.DesiredPVCCount:
+					scaling = down
+				}
+			}
+			if options.EnableLogging {
```
davidz627: what about all the other times we log things, why is only this one gated by `EnableLogging`? There should probaly be a wrapper function for this.

clusterloader2/pkg/measurement/util/pvc_store.go
wojtek-t: It's returning *v1.PodList:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/kubernetes/typed/core/v1/pod.go#L46

I'm think that it implements runtime.Object interface (otherwise a bunch of other things won't work).

wojtek-t: And we use it in a bunch of places without casting, e.g.:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/secret/secret_manager.go#L140

so it has to work.

clusterloader2/pkg/measurement/util/pvc_store.go
wojtek-t: Yes - it's the only way of doing that :)

clusterloader2/pkg/measurement/util/pvc_store.go
wojtek-t: should be 0, len(objects)

[to allocate all needed capacity, but not allocate objects themself]

clusterloader2/pkg/measurement/util/pvcs.go
wojtek-t: was copied from pods.go - may probably be cleaned up, but that's not directly related to this PR

clusterloader2/pkg/measurement/util/pvcs.go
wojtek-t: As above - this was copied from other already existing files.
While I agree we can improve stuff, I wouldn't block on that (in fact I actually prefer consistency).

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
wojtek-t: The same as above (and I won't comment on other).
This is not the problem with this PR, but the fact that it was already done like this before. I think consistency is more important that cleaning this up in just one place.

@krzysied - btw, those are valid comments that we may consider applying

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
```diff
@@ -0,0 +1,83 @@
+/*
```
wojtek-t: High-level comment:
The review latency of a code is exponential to its size. So more you split it into smaller chunks, the faster you will have it done.

clusterloader2/pkg/measurement/util/pvc_store.go
hantaowang: ack, just changing for pvc_store then

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
hantaowang: I thought it would be intuitive because the name of the measurement is "WaitForBoundPVCs"

clusterloader2/pkg/measurement/util/pvc_store.go
hantaowang: ack

clusterloader2/pkg/measurement/util/pvc_store.go
hantaowang: @davidz627, can you see if this is what you mean?

clusterloader2/pkg/measurement/util/pvcs.go
hantaowang: ok, this can be a separate PR then

clusterloader2/pkg/measurement/util/pvcs.go
hantaowang: ok, this can be a separate PR then

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
hantaowang: ok, this can be a separate PR then

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
davidz627: understand the thought; however, it wasn't immediately clear to me so maybe not that intuitive :)

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
hantaowang: How does this look? The downsides are that 1) the stop chan does not work until after the first list() and 2) the list and ComputePVCStartupStatus lines are repeated.

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
hantaowang: The other are error logs, and this one particular log info. Not only that, it is an info that is once per iteration of the for loop so it can become quite annoying if youre waiting for a long time so it makes sense that this one should have an option to be disabled.

clusterloader2/pkg/measurement/util/pvc_store.go
```diff
@@ -0,0 +1,88 @@
+/*
```
wojtek-t: Did that for you in https://github.com/kubernetes/perf-tests/pull/678 already :)

clusterloader2/pkg/measurement/common/wait_for_pvcs.go
wojtek-t: hmm - it was intuitive for me based on the name. But if you have better suggestions for the name, I'm open.

wojtek-t:
Requires rebasing now. Also please squash the commits.
[I think it's in good enough shape to be approved then - I will just take a final pass once this is done.]

clusterloader2/pkg/measurement/util/wait_for_pvcs.go
```diff
@@ -0,0 +1,92 @@
...
+				klog.Infof("%s: %s: %s", options.CallerName, options.Selector.String(), pvcsStatus.String())
+			}
+			// We wait until there is a desired number of PVCs bound and all other PVCs are pending.
+			if len(pvcs) == (pvcsStatus.Bound+pvcsStatus.Pending) && pvcsStatus.Bound == options.DesiredPVCCount {
+				return nil
+			}
+			oldPVCs = pvcs
+			pvcs = ps.List()
```
wojtek-t: I don't like this, because you are processing pvs that were there WaitForPVCsInterval ago (which sounds wrong).
But I don't want to block this PR for a long - I will send out a PR to fix it and unify with wait_for_pods a bit.

PR #657, opened 2019-07-19
add persistent volume scale tests

hantaowang:
Experimental storage tests to test startup latency of booting pods with PVs. I tried staying as close to @masu's KEP as possible.

This PR depends on functionality from #656. I have it open so we can parallelize the review for both. 

**1_node_scale_by_pod**
This test runs on 1 node, 100 pods per node, and 1 PV per pod. This stresses # of PVs per node and # of pods per node.

**1_node_scale_by_pvcs_per_pod**
This test runs on 1 node, 1 pod per node, and 100 PVs per pod. This stresses # of PVs per node and # of PVs per pod.

**cluster_load_scale_by_nodes**
This test runs on X nodes (X from 1 to 5000+). It has a fixed 10 pods per node and 1 PV per pod. This test simulates a realistic large cluster with a normal amount of PVs per node. 

**TODO (In future PR)**:
- Implement the ability to boot the GCE PD CSI Driver to use as the storage backend. 
- Write test to boot a large cluster with a large amount of PVs, and the boot a couple more pods w/ PVs in this saturated cluster.

/assign @wojtek-t @davidz627
cc @msau42 @jingxu97 @verult


hantaowang:
Which tuning set do you recommend? And for the parallelism limited ts, what parallel limit should I set? I wonder if setting it equal to the # of namespaces is reasonable.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
```diff
@@ -0,0 +1,142 @@
...
+# Constants
+  {{$NODE_MODE := DefaultParam .NODE_MODE "allnodes"}}
+  {{$DEFAULT_NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+
+# Variables
+  {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 50}}
+  {{$PVC_PER_POD := DefaultParam .PVC_PER_POD 2}}
+  {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+  {{$TEST_NAME := DefaultParam .TEST_NAME "generic-test"}}
+  {{$PVC_SIZE := DefaultParam .PVC_SIZE "8Gi"}}
```
davidz627: Why did we pick `8Gi`

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
```diff
@@ -0,0 +1,142 @@
...
+  {{$PER_PVC_BOUND_SLO := DefaultParam .PVC_BOUND_SLO 2}}
+  {{$POD_STARTUP_SLO := DefaultParam .POD_STARTUP_SLO 100}}
```
davidz627: why default to `100` then set to `300` in every test?

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
```diff
@@ -0,0 +1,142 @@
...
+  {{$NODE_MODE := DefaultParam .NODE_MODE "allnodes"}}
+  {{$DEFAULT_NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+
+# Variables
+  {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 50}}
+  {{$PVC_PER_POD := DefaultParam .PVC_PER_POD 2}}
+  {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+  {{$TEST_NAME := DefaultParam .TEST_NAME "generic-test"}}
+  {{$PVC_SIZE := DefaultParam .PVC_SIZE "8Gi"}}
+  {{$PER_PVC_BOUND_SLO := DefaultParam .PVC_BOUND_SLO 2}}
```
davidz627: is this in seconds? 2 seconds binding seems low. Not sure if this includes provisioning time or not though

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
```diff
@@ -0,0 +1,142 @@
...
+          max: {{$namespaces}}
+        replicasPerNamespace: {{$podsPerNamespace}}
+        tuningSet: UniformQPS
+        objectBundle:
+          - basename: pod
+            objectTemplatePath: pod.yaml
+            templateFillMap:
+              PVCsPerPod: {{$PVC_PER_POD}}
+  - measurements:
+      - Identifier: WaitForRunningPVCOwningPods
```
davidz627: naming nit: `s/PVCOwningPods/PodsWithPVC`?

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
```diff
@@ -0,0 +1,142 @@
...
+      - Identifier: APIResponsivenessPrometheus
+        Method: APIResponsivenessPrometheus
+        Params:
+          action: start
+      - Identifier: TestMetrics
+        Method: TestMetrics
+        Params:
+          action: start
+          nodeMode: {{$NODE_MODE}}
+  # Create PVCs
```
wojtek-t: I'm wondering if we really need this file.
This test is extremely similar to: https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/cluster-wide/config.yaml

The only difference is that instead of creating ephemeral storage we create PVCs and wait for them.
But it should be possible to unify them (if you wait for PVC in ephemeral-storage scenario, it will be no-op anyway).

So instead of having two separate files to maintain, I would much rather prefer unifying those two together - I don't think the differences justify having a separate one.

@mucahitkurt - as the one who added the test I mentioned above.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: Its a nice factor of 2 :)

In honestly there is not a reason behind it other than it seems like a reasonable choice for a single mid sized PV to be used as the default.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: I changed it, but this number is not set yet until I figure out a reasonable SLO for each test. I override it to 300 because that seemed reasonable enough as an upperbound for testing purposes.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: Provisioning is done in parallel to a degree. 2 sec is low if you are no provisioning that many, which is why I use a PVCBoundTime of min 60 sec.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: ack

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: I would be happy to merge the two configs (in a separate PR). There definitely are similarities although I think the PV config will be a little bit more config due to the WaitForBoundPVCs and such.

wojtek-t: I think the WaitForBoundPVCs is the only difference. And just adding it won't break ephemeral volumes, because simple there won't be anything to wait for there.
So unless I'm not missing anything, I'm against having two separate configs and merging them later - we should just do the right thing from the beginning.

hantaowang: I would much prefer the base config be the one I wrote. There is a difference in that my config has an option to scale automatically based on # of nodes and overrides is used for different test variations (defined by the KEP), whereas @mucahitkurt’s work operates similar to configs for load and density where overrides is used to define the size of the cluster.

Its ok imo to not merge it yet, but it would take some time to work together and refactor how the configs work. Idealy, it should be the same base woth maybe a flag for pv vs ephemeral.

mucahitkurt: > @mucahitkurt’s work operates similar to configs for load and density where overrides is used to define the size of the cluster.

Sorry If I got it wrong, but actually I don't use overrides for the cluster size, e.g. 1_node and cluster_wide case, I use different configs under `1_node` and `cluster-wide`, I started with one config but got feedback on the related PR to separate the configs for those cases for the sake of simplicity over duplication. But of course we can think that cluster-wide config already covers the 1 node case, as you already did.

mucahitkurt: I'm not sure if it will cause to some limitations, but may be it would be better if we can get rid of root `ephemeral-volumes` directory and use the PVC as a different volume type

wojtek-t: > I'm not sure if it will cause to some limitations, but may be it would be better if we can get rid of root ephemeral-volumes directory and use the PVC as a different volume type

Yes - we probably should.

> I would much prefer the base config be the one I wrote. There is a difference in that my config has an option to scale automatically based on # of nodes and overrides is used for different test variations (defined by the KEP), whereas @mucahitkurt’s work operates similar to configs for load and density where overrides is used to define the size of the cluster.

I'm not trying to say you can't modify @mucahitkurt config - we definitely need to. And I'm completely fine with modify something there.
I'm also fine with replacing the one added by @mucahitkurt with this one (though I need to review it more carefully at that point). I just don't see any reason to have both at this point.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
jingxu97: sorry still confuse about the meaning of this parameter and the set of "2"?

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+# Variables
+  {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 50}}
+  {{$PVC_PER_POD := DefaultParam .PVC_PER_POD 2}}
+  {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+  {{$TEST_NAME := DefaultParam .TEST_NAME "generic-test"}}
+  {{$PVC_SIZE := DefaultParam .PVC_SIZE "8Gi"}}
+  {{$PER_PVC_BOUND_SLO := DefaultParam .PVC_BOUND_SLO 2}}
+  {{$POD_STARTUP_SLO := DefaultParam .POD_STARTUP_SLO 300}}
+  {{$OBJ_CREATION_QPS := DefaultParam .OBJ_CREATION_QPS 10}}
```
jingxu97: what is for?

hantaowang: OBJ_CREATION_QPS is a configurable variable that describes how fast to create API objects at. The default value of 10 QPS means to create it 10 queries per second. It is used to configure the UniformQPS tuning set (a thing that is responsible for creating k8s objects). 

@msau42 @wojtek-t any thoughts on what the most appropriate tuning set to test with is? I am actually not sure if sequence of QPS means in a single namespace or across all namespaces.

hantaowang:
Once #656 is merged, please look at #688 so I can rebase this PR to use the same config as the ephemeral tests (Although I expect there will be 1 more PR after #688 to change that config into something that is more of a middle group between the current one and mine).

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
wojtek-t: With https://github.com/kubernetes/perf-tests/pull/688 merged, I think this file is no longer needed - just that one requires changes. Can you please update this PR?

hantaowang: ack

hantaowang:
rebased and ready for review 
/hold cancel

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: It just means that because the backend can only provision a certain number of PVs at a time and thus is to a degree serial, I set an SLO for the per PV provision time at 2 seconds. So 100 PVs has a total SLO of 2 * 100 = 200 seconds.

msau42: I think 2 seconds is too fast, especially while we're in an experimental phase and are testing our limits. Let's do something like 30 seconds?

We're also limited by qps of k8s api server and gce api

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
 
-{{$POD_TEMPLATE_PATH := DefaultParam .POD_TEMPLATE_PATH "../volume-types/emptydir/pod_with_emptydir.yaml"}}
-{{$LOAD_TEST_THROUGHPUT := DefaultParam .LOAD_TEST_THROUGHPUT 10}}
+# Framework Variables
+  {{$OBJ_CREATION_QPS := DefaultParam .OBJ_CREATION_QPS 10}}
+  {{$PARALLELISM_LIMITED := DefaultParam .PARALLELISM_LIMITED 1}}
+  {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+  {{$TEST_NAME := DefaultParam .TEST_NAME "generic-test"}}
+  {{$GROUP := "pod-with-volumes-startup-latency"}}
+  {{$TUNING_SET := DefaultParam .TUNING_SET "UniformQPS"}}
```
wojtek-t: Let's remove this - I don't think we want to allow such significant configuration

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
 
-name: pod-with-ephemeral-volume-startup-latency
+
+name: pv-storage-{{$TEST_NAME}}-{{$VOLUME_TYPE}}
 automanagedNamespaces: {{$namespaces}}
 tuningSets:
-- name: UniformQPS
-  qpsLoad:
-    qps: {{$LOAD_TEST_THROUGHPUT}}
+  - name: Sequence
```
wojtek-t: This isn't used anywhere - let's remove it.
Allowing to many knobs to configure is never good - the more opinionated we are the better.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
-{{$VOLUMES_PER_POD := DefaultParam .VOLUMES_PER_POD 1}}
-{{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
-{{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+# Test Variales
+  {{$PODS_PER_NODE := .PODS_PER_NODE}}
+  {{$POD_TEMPLATE_PATH := .POD_TEMPLATE_PATH }}
+  {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
+  {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
+  {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+  {{$VOL_SIZE := DefaultParam .VOL_SIZE "8Gi"}}
```
wojtek-t: Do we really want to expose an option for it? I vote for being opinionated and setting it just to 8G.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
+  {{$WAIT_FOR_PVC := DefaultParam .WAIT_FOR_PVC false}}
```
wojtek-t: We don't need it - if there won't be PVC, the measurement will be no-op.
The less configuration options we have the easier to use it is.

Let's just run this measurement in every test - sometime it will just be no-op.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
+  {{$POD_STARTUP_TIMEOUT := DefaultParam .POD_STARTUP_TIMEOUT "15m"}}
 
-{{$APP_NAME := "pod-load"}}
-{{$GROUP := "pod-with-ephemeral-volume-startup-latency"}}
+# SLO Variables
+  {{$POD_STARTUP_SLO := DefaultParam .POD_STARTUP_SLO 300}}
+  {{$PER_PVC_BOUND_SLO := DefaultParam .PVC_BOUND_SLO 2}}
```
wojtek-t: We don't have any SLO for pvc binding - let's not call it that way.
PVC_BINDING_TIME maybe?

The other comment - do we really need it? Can't we hardcode it?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
+  {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 
-{{$POD_TEMPLATE_PATH := DefaultParam .POD_TEMPLATE_PATH "../volume-types/emptydir/pod_with_emptydir.yaml"}}
-{{$LOAD_TEST_THROUGHPUT := DefaultParam .LOAD_TEST_THROUGHPUT 10}}
+# Framework Variables
+  {{$OBJ_CREATION_QPS := DefaultParam .OBJ_CREATION_QPS 10}}
+  {{$PARALLELISM_LIMITED := DefaultParam .PARALLELISM_LIMITED 1}}
+  {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+  {{$TEST_NAME := DefaultParam .TEST_NAME "generic-test"}}
+  {{$GROUP := "pod-with-volumes-startup-latency"}}
```
wojtek-t: Let's not introduce unnecessary things.
Please use some shorter name group = volume seems good enough, and hardcode it in necessary places.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
+{{ end }}
 steps:
 # Start measurements
 - measurements:
   - Identifier: APIResponsiveness
     Method: APIResponsiveness
     Params:
       action: reset
-  - Identifier: PodWithMultiVolumeStartupLatency
+  - Identifier: APIResponsivenessPrometheus
```
wojtek-t: If you introduce this, you also need to gather metrics at the end (otherwise doesn't make sense).

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pod_with_pvc.yaml
```diff
@@ -0,0 +1,27 @@
+{{$index := .Index}}
+{{$appName := .AppName}}
+{{$volumesPerPod := .VolumesPerPod}}
+apiVersion: v1
+kind: Pod
```
mucahitkurt: I think the same concern about [using raw volumes instead of a controller](https://github.com/kubernetes/perf-tests/issues/666) is valid for this one.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
jingxu97: so it limits how many objects in total to create in one second, which includes PVC/PV, Pods etc?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,96 +1,151 @@
...
+  {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE | MaxInt 1}}
+  {{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
+  {{$podsPerNamespace := MultiplyInt $NODES_PER_NAMESPACE $PODS_PER_NODE}}
+  {{$volumesPerNamespace := MultiplyInt $podsPerNamespace $VOLUMES_PER_POD}}
+  {{$totalVols := MultiplyInt $volumesPerNamespace $namespaces}}
+  {{$PVCBoundTime := MultiplyInt $totalVols $PER_PVC_BOUND_SLO | MaxInt 60}}
 
-name: pod-with-ephemeral-volume-startup-latency
+
+name: pv-storage-{{$TEST_NAME}}-{{$VOLUME_TYPE}}
```
mucahitkurt: since `pv` doesn't cover `ephemeral`, it might be better just starting with `storage-` or `volume-`

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: Most removed, I left test-name but I can remove that too.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: Ok, I defaulted it to sequence with PL=1.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: Spoke with team, and we want to keep this know because different storage options have different min sizes (and will fail if you try to provision less). Also, different sizes have different provision times.

Zonal PV: min 5G
Regional PV: min 200 G
Filestore: min 1T

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: It will still try to wait for totalVol PVCs right, which will time out and fail the test. How can I configure this to be a no-op?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: Depending on the PV storage option, hard coding will break the tests. Filestore takes forever to provision.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: ack

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: ack

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pod_with_pvc.yaml
hantaowang: can you wait on that PR until this one is merged?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: ack, set to storage-

hantaowang:
will squash later

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: OK - I initially thought that the measurement simply looks into what PVCs exist and waits for all of them to be bounded. It's not exactly what we do, so in the current implementation I think we need to "if" that.
I would like to clean this up, but not necessary block this PR on it.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,100 +1,143 @@
...
+  {{$PODS_PER_NODE := .PODS_PER_NODE}}
+  {{$POD_TEMPLATE_PATH := .POD_TEMPLATE_PATH }}
+  {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
+  {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
+  {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+  {{$VOL_SIZE := DefaultParam .VOL_SIZE "8Gi"}}
+  {{$WAIT_FOR_PVC := DefaultParam .WAIT_FOR_PVC false}}
+  {{$POD_STARTUP_TIMEOUT := DefaultParam .POD_STARTUP_TIMEOUT "15m"}}
+  {{$POD_STARTUP_SLO := DefaultParam .POD_STARTUP_SLO 300}}
+  {{$PVC_BINDING_TIME := DefaultParam .PVC_BOUND_SLO 2}}
```
wojtek-t: Do we really need to expose this?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,100 +1,143 @@
...
+  {{$TEST_NAME := DefaultParam .TEST_NAME "generic-test"}}
+  {{$PODS_PER_NODE := .PODS_PER_NODE}}
+  {{$POD_TEMPLATE_PATH := .POD_TEMPLATE_PATH }}
+  {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
+  {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
+  {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+  {{$VOL_SIZE := DefaultParam .VOL_SIZE "8Gi"}}
+  {{$WAIT_FOR_PVC := DefaultParam .WAIT_FOR_PVC false}}
+  {{$POD_STARTUP_TIMEOUT := DefaultParam .POD_STARTUP_TIMEOUT "15m"}}
+  {{$POD_STARTUP_SLO := DefaultParam .POD_STARTUP_SLO 300}}
```
wojtek-t: Please add a TODO to remove it one we figure out the right value.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -1,100 +1,143 @@
...
     objectBundle:
-    - basename: {{$APP_NAME}}
+    - basename: pod-{{$TEST_NAME}}
       objectTemplatePath: {{$POD_TEMPLATE_PATH}}
       templateFillMap:
-        Group: {{$GROUP}}
+        Group: volume-test
         VolumesPerPod: {{$VOLUMES_PER_POD}}
-        AppName: {{$APP_NAME}}
+        AppName: Sequence
```
wojtek-t: Why is this AppName used for?

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pod_with_pvc.yaml
wojtek-t: I agree - let's have this one merged and then migrate to deployments all in one shot.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
```diff
@@ -0,0 +1,15 @@
...
+    app: {{.Name}}
+    group: {{.Group}}
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{.VolSize}}
+  persistentVolumeReclaimPolicy: Delete
+  volumeBindingMode: Immediate
```
wojtek-t: nit: add end-of line

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: Can't we just use Name instead of it?

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
msau42: Reclaim policy and binding mode are fields on the storageclass, not PVC.

Until we have a way to measure and subtract out provisioning time, a prerequisite of running this test using default storageclass is that the default storageclass needs to have binding mode = immediate.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
msau42: This discussion also brought up an interesting point.  We are not measuring provisioning time because we said that is bootstrapping behavior that only happens the first time.  

We also format the filesystem the first time we mount the volume on any node and that is something this test will currently measure during pod startup.  Is it something we want to include? If not, then we would need to bring up pods twice and only measure the second time.

clusterloader2/testing/experimental/storage/pod-startup/persistent-volumes/config.yaml
hantaowang: yes thats right

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: ack, this shouldnt even be here

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: I can remove this know for now, but we may need it later if we do other types of PD.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
hantaowang: I removed those two lines and added an assumption to the top.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: ack

hantaowang:
@wojtek-t I've squashed the commits because I think this is in a good place to merge now. Please give your final comments or approval.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: Great point - I opened https://github.com/kubernetes/perf-tests/issues/706 to not block this PR on it.

PR #658, opened 2019-07-19
Gather etcd profiles

wojtek-t:
/hold


mm4tt:
For posterity, from discussion with @wojtek-t on chat, gathering etcd profiles requires running etcd with --debug flag, which currently is not set in cluster-up scripts (it may have some performance implications)
Presubmit fo this PR should verify that this PR is no-op and nothing breaks when --debug is not set in etcd.

/lgtm

wojtek-t:
The profiles are there

/hold cancel

PR #660, opened 2019-07-19
ClusterLoader - Extracting object selector logic

krzysied:


wojtek-t:
/lgtm

@krzysied - we should propagate it further (e.g. change the NewInformer to use it, change podStore to use it). I'm fine with a separate PR to unblock you though.
/hold
Please cancel hold if you acknowledge the above.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -108,7 +99,7 @@ func (s *schedulingThroughputMeasurement) start(clientSet clientset.Interface, n
 
 	go func() {
 		defer ps.Stop()
-		selectorsString := measurementutil.CreateSelectorsString(namespace, labelSelector, fieldSelector)
+		selectorsString := selector.String()
```
oxddr: nit: Why is this variable needed? Can't you just pass selector to the klog directly? I don't see any other usages.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -121,16 +111,16 @@ func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
 		klog.Infof("%s: pod startup latancy measurement already running", p)
 		return nil
 	}
-	p.selectorsString = measurementutil.CreateSelectorsString(p.namespace, p.labelSelector, p.fieldSelector)
+	p.selectorsString = p.selector.String()
```
oxddr: Is it necessary to keep the string? Cannot we just compute it every time it's needed?

krzysied:
@wojtek-t I've added selector to podStore, Informer, ...

clusterloader2/pkg/measurement/common/scheduling_throughput.go
krzysied: I don't want to calculate every time. It will not change, so I just store it.

oxddr: That's three strings to concatenate. I don't think that would improve performance.

wojtek-t: It could visibly improve if we would be calling that often. But in this case we are not, so I agree with @oxddr that this isn't worth it.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
krzysied: I'm not a fan of computing every time the same value - I prefer to store it.

oxddr: Please see the other comment - the same principle applies here. It's not used often enough to have any performance gain.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
krzysied: selectorString removed.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
krzysied: selectorString removed.

oxddr:
/lgtm

Thanks!

PR #661, opened 2019-07-19
fix X worst latency log

hantaowang:
The log prints in the wrong order. This simple Pr fixes that.


wojtek-t:
Thanks!

/ok-to-test
/lgtm
/approve

PR #662, opened 2019-07-21
Cluster Wide Scalability and Performance Test Cases for Ephemeral Volumes

mucahitkurt:
Test case to help determine a limit for max number of volumes on a cluster.

Related issue: #600
Related PR: https://github.com/kubernetes/perf-tests/pull/601

I need support for the final values of the parameters, `$NODES_PER_NAMESPACE`, `$PODS_PER_NODE`, tuning set parameter.

/sig storage
/assign @msau42 
/assign @wojtek-t


clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/volume-types/configmap/override.yaml
```diff
@@ -0,0 +1,3 @@
+PROVISION_VOLUME: true
```
wojtek-t: Moving this files will clearly break your tests.
I think it's what we would like to do, but it would be good to open a PR in test-infra that would fix that so that we could merge them ~at the same time, to mitigate the amount of time when it will be broken.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/cluster-wide/config.yaml
```diff
@@ -0,0 +1,101 @@
...
+# Create pods
+- phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: {{$podsPerNamespace}}
+    tuningSet: UniformQPS
+    objectBundle:
+    - basename: {{$APP_NAME}}
+      objectTemplatePath: {{$POD_TEMPLATE_PATH}}
```
wojtek-t: I'm fine with it as a starting phase, but we should move them to use deployments (using individual pods is more like anti-pattern, and startup-time is measured from pod creation anyway, so it shouldn't cause any additional problems).

mucahitkurt: When I use a controller like deployments I can't control the volume names of the pod that controller creates. 

I tried it with statefulset,  for replication number 2, statefulset creates pod likes app-0-pod-0, app-0-pod-1, if I want to use volumes with these pod I should create volumes likes pod-0-vol-0, pod-0-vol-1, pod-1-vol-0, pod-1-vol-1. But I can't create volume names like 2 indexed values for pod-[podIndex]-vol-[volumeIndex], because clusterloader2 framework doesn't allow this.

I know that it's an anti-pattern working directly with the pods without a controller, but I couldn't solve the above problem and used directly pod.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/volume-types/configmap/override.yaml
mucahitkurt: https://github.com/kubernetes/test-infra/pull/13544

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/cluster-wide/config.yaml
wojtek-t: I opened https://github.com/kubernetes/perf-tests/issues/666 for it - let's continue the discussion there. I think I still don't understand where the problem is, but I don't want to block this PR on it.

PR #664, opened 2019-07-22
ClusterLoader - pod startup event polling

### LINKED ISSUE #663: ClusterLoader - 80% of logs are useless
Currently 80% of the logs gather by cluster loader looks like this:
```
I0719 22:58:11.023759  199702 phase_latency.go:76] PodStartupLatency: failed to find schedule time for test-ahw1lj-13/big-deployment-2-788bf9864f-jktzx
I0719 22:58:11.023786  199702 phase_latency.go:76] PodStartupLatency: failed to find schedule time for test-ahw1lj-37/small-deployment-199-66b4cc59b7-kqz5b
I0719 22:58:11.023795  199702 phase_latency.go:76] PodStartupLatency: failed to find schedule time for test-ahw1lj-41/medium-deployment-3-5fb6668747-gkbrn
```

krzysied:
Listing scheduler events every 30 minutes.

/fixes #663

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -37,7 +37,7 @@ const (
 	defaultPodStartupLatencyThreshold = 5 * time.Second
 	podStartupLatencyMeasurementName  = "PodStartupLatency"
 	informerSyncTimeout               = time.Minute
-	successfulStartupRatioThreshold   = 0.99
+	eventPollingInterval              = 30 * time.Minute
 
 	createPhase   = "create"
 	schedulePhase = "schedule"
@@ -113,6 +113,7 @@ func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
 	klog.Infof("%s: starting pod startup latency measurement...", p)
 	p.isRunning = true
 	p.stopCh = make(chan struct{})
+	go p.eventPolling(c)
 	i := informer.NewInformer(
 		c,
 		"pods",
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -113,6 +114,11 @@ func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
 	klog.Infof("%s: starting pod startup latency measurement...", p)
 	p.isRunning = true
 	p.stopCh = make(chan struct{})
+	wait.Until(func() {
+		if err := p.gatherScheduleTimes(c); err != nil {
```
wojtek-t: What's the point of listing events immediately on startup - the list is expensive and will not given any results anyway.

krzysied: Fixed.

krzysied:
> Before we do this change, I would like to see hw much it is impacting performance (given that this data is only used for debugging, not really computing any SLO, etc.)

I guess I have to run 5k test with this change.. We shouldn't see any impact in 100 nodes test. IIRC each 100 nodes test takes less then 30 minutes to be run.

wojtek-t:
> I guess I have to run 5k test with this change.. We shouldn't see any impact in 100 nodes test. IIRC each 100 nodes test takes less then 30 minutes to be run.

Correct. Also with prometheus metrics and probably with 2x (or even more) speedup.
Kubemark is fine though.

krzysied:
I found ```GET /api/v1/events: (2.201830901s) 200 [eventer/v0.0.0 (linux/amd64) kubernetes/$Format 35.227.27.137:51564]```. Eventer is making those calls every 10 minutes.
Cluster loader is listing only subset of events, so it should have big impact. Plus we can change clusterloader to list every 1h instead of 30min, to reduce number of calls.

Btw, why eventer is listing events instead of watching?

wojtek-t:
> Eventer is making those calls every 10 minutes. Cluster loader is listing only subset of events, so it should have big impact.

I disagree with the impact argument - for the purpose of listing, we still need to list *everything* from etcd, and filter all of those (we have non-trivial filter function, which means it's more work). So it only impacts amount of data to serialize and send. So I wouldn't say it's actually the same order of magnitude work to do.

> Plus we can change clusterloader to list every 1h instead of 30min, to reduce number of calls.

1h may mean we will still loose some stuff

> Btw, why eventer is listing events instead of watching?

IIRC it is watching. I guess it's falling out of the history window (due to to small resource limits or sth?)

Issue #667, opened 2019-07-23
Summaries are not generated for new storage tests


wojtek-t:
In some of the newly introduced storage tests, pod_startup_latency summaries aren't generated.

E.g. in this run:
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability-max-configmap-vol-per-pod/1153307037698887684/
clearly it was measured, it was even logged as violated, but no .json file exists

What is interesting, in other tests sharing the same test-config (just different overrides), it works fine, e.g.:
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability-max-emptydir-vol-per-pod/1153307037698887682/

wojtek-t:
@mucahitkurt - this is the problem you had

wojtek-t:
hmm - I think I see what is happening - let me send a PR.

PR #668, opened 2019-07-23
Fix error handling in pod startup time metric

wojtek-t:
This basically means that our tests could never failing in pod-startup-latency SLO. !!!

Fix: https://github.com/kubernetes/perf-tests/issues/667

@kubernetes/sig-scalability-bugs 
@mm4tt @mborsz @oxddr @krzysied


oxddr:
/lgtm 

I am not sure whether this would help in this scenario, but I think we should enable [errcheck](https://github.com/kisielk/errcheck) in our presubmit. This tools warns you when unchecked error is returned.

PR #676, opened 2019-07-23
Clean up steps in CL2 test configs.

mm4tt:
The change adds name for each steps describing what it does. I believe this will make the test configs easier to understand to people not familiar with clusterloader2.
In addition we'll be logging when step starts and ends, so it's easier to identify particular steps in the log.

### DIFF
--- a/clusterloader2/pkg/test/simple_test_executor.go
+++ b/clusterloader2/pkg/test/simple_test_executor.go
@@ -96,6 +96,9 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
 
 // ExecuteStep executes single test step based on provided step configuration.
 func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) *errors.ErrorList {
+	if step.Name != "" {
+		klog.Infof("Step %q started", step.Name)
+	}
 	var wg wait.Group
 	errList := errors.NewErrorList()
 	if len(step.Measurements) > 0 {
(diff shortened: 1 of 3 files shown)


mm4tt:
It's not directly related, but I sent this PR when doing https://github.com/kubernetes/perf-tests/issues/413.
This PR is no simplifying per se, but is kind of a clean up.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -96,6 +96,9 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
 
 // ExecuteStep executes single test step based on provided step configuration.
 func (ste *simpleTestExecutor) ExecuteStep(ctx Context, step *api.Step) *errors.ErrorList {
+	if step.Name != "" {
+		klog.Infof("Step \"%s\" started", step.Name)
```
oxddr: nit: please use %q instead - it adds quotes around string for you.

oxddr:
Nice one! LGTM except a small nit.

mm4tt:
> Nice one! LGTM except a small nit.

Done.

clusterloader2/testing/density/config.yaml
```diff
@@ -167,19 +174,20 @@ steps:
...
       objectTemplatePath: deployment.yaml
-- measurements:
+
+- name: Waiting for latency pods to be deleted
+  measurements:
   - Identifier: WaitForRunningLatencyDeployments
     Method: WaitForControlledPodsRunning
     Params:
       action: gather
-- measurements:
```
wojtek-t: This one is tricky - it works, but technically it's more intuitive to first wait for deletion of pods and then report latencies, rather than doing that in-parallel.

mm4tt: That's a really good point. This should be a separate step, thanks for catching!
Done.

Issue #641, opened 2019-07-16
Allow for running the same ClusterLoader2 test with different overrides as part of the same job


wojtek-t:
More context in: https://github.com/kubernetes/perf-tests/pull/601#issuecomment-511776754

wojtek-t:
This has already been fixed.

PR #677, opened 2019-07-23
ClusterLoader - Test suite design

krzysied:
ref https://github.com/kubernetes/perf-tests/issues/641

### DIFF
--- a/clusterloader2/docs/test_suite_design.md
+++ b/clusterloader2/docs/test_suite_design.md
@@ -1,0 +1,41 @@
+# Cluster loader - TestSuite
+
+## Background
+
+Since 2019 cluster loader become main tool for kubernetes e2e performance tests.
+It's very convenient for running single test case, but currently it's impossible
+to run the same test with two different sets of parameters in a single job run.
+This forces developers to create separate job configs for almost each test.
+
+## TestSuite
+
+
+```
+type TestSuite []TestScenario
+
+type TestScenario struct {
+	// Identifier is a unique test scenario name across test suite.
+	Identifier string
+	// ConfigPath defines path to the config file that satisfy already
+	// existing cluster loader API.
+	ConfigPath string
+	// OverridePaths defines what override files should be applied
+	// to the config specified by the ConfigPath.
+	OverridePaths []string
+}
+```
+
+Test scenario level overrides benefits:
+- Ability to use the same template with multiple overrides in a single job.
+- Less error prone - variables with same names across different configs
+could have been accidentally overwritten before.
+
+To make using same test configs in one test suite more user-friendly,
+the Identifier field was introduced. This identifier will be used to match
+run results to the corresponding test.
+
+## Other
+
+Path to the test suite yaml file will be passed to cluster loader as a flag.
+After TestSuite is supported by cluster loader, passing single test config paths
+and overrides using flags will be disabled.


clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+TestSuite is a test bundle.
+Overrides is a set of override file paths. Overrides provided on the TestSuite level will be applied
+to all of the test, unless they are overidden on the test level.
+
+```
+type TestSuite struct {
+	Overrides []string
+	Tests     []Test
+}
+
```
wojtek-t: nit: remove this empty line

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+Since 08/2018 cluster loader become main tool for kubernetes e2e performance tests.
+It's is a very convenient tool for running single test case, but not for running
+test bundles. Currently it is impossible to run the same test
+with two different parameters sets in a single job run.
+This forces developers to create separate job configs for
+almost each tests.
+
+## TestSuite
+
+TestSuite is a test bundle.
```
wojtek-t: TestSuite is a set of tests running in a single test job.

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+Overrides is a set of override file paths. Overrides provided on the TestSuite level will be applied
```
wojtek-t: If someone doesn't know about CL, they won't know what are "override file paths".

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+to all of the test, unless they are overidden on the test level.
```
wojtek-t: How they can be overridden? Is it decided individually for every single env-var?

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+
+## Test
+
+Test defines a single test scenario. Config defines path to the config file
+that satisfy already existing cluster loader API. Overrides defines what
+override files should be applied to the config file.
+To enable using same test config in one test suite,
+the Identifier field was introduced. This indetifier will be used to match
```
wojtek-t: So is it only used for reporting or for something else?

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+type Test struct {
+	Overrides  []string
+	Config     string
+	Identifier string
+}
+```
+
+## Other
+
+User should either use a single test + global overrides or a test suite. Using those two options
```
wojtek-t: So technically, those will be two different flags?

Why do we need that (for more than short-period of time to do migration)? Why can't we just support TestSuite?

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
+# Cluster loader - TestSuite
+
+## Background
+
+Since 08/2018 cluster loader become main tool for kubernetes e2e performance tests.
+It's is a very convenient tool for running single test case, but not for running
```
mm4tt: NIT: "It's is" - please fix

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+test bundles. Currently it is impossible to run the same test
+with two different parameters sets in a single job run.
+This forces developers to create separate job configs for
+almost each tests.
```
mm4tt: NIT: s/each tests/each test/

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
+# Cluster loader - TestSuite
+
+## Background
+
+Since 08/2018 cluster loader become main tool for kubernetes e2e performance tests.
+It's is a very convenient tool for running single test case, but not for running
+test bundles. Currently it is impossible to run the same test
+with two different parameters sets in a single job run.
```
mm4tt: NIT: s/parameters sets/parameter sets/

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+
+## TestSuite
+
+TestSuite is a test bundle.
+Overrides is a set of override file paths. Overrides provided on the TestSuite level will be applied
+to all of the test, unless they are overidden on the test level.
+
+```
+type TestSuite struct {
+	Overrides []string
```
mm4tt: So this is basically duplicating the --testoverrides flag functionality, isn't it?
Do we need it?

clusterloader2/docs/test_suite_design.md
mm4tt: +1

And in such case, wouldn't we want to have Config to be of Config type instead of string, so we won't need at least 2 files per suite (one with suite def, the other with config)?

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+Test defines a single test scenario. Config defines path to the config file
+that satisfy already existing cluster loader API. Overrides defines what
+override files should be applied to the config file.
+To enable using same test config in one test suite,
+the Identifier field was introduced. This indetifier will be used to match
+run results to the corresponding test.
+
+```
+type Test struct {
+	Overrides  []string
```
mm4tt: Would be good to add a word or two about override priorities. 
If we have two level of overrides it's easy to imagine a situation where I override something at the suite level and test level at the same time. E.g. I set LOAD_TEST_THROUGHPUT to 20 at suite level and to 50 at some test level. I assume in such case the test level override will take precedence over suite level, right?

clusterloader2/docs/test_suite_design.md
mm4tt: Actually giving it a second thought, looks like the only thing that is missing is Config level overrides. Have you considered just adding an Override map/list to the Config struct? Wouldn't it solve the problem in a simpler way?

wojtek-t: > Actually giving it a second thought, looks like the only thing that is missing is Config level overrides. Have you considered just adding an Override map/list to the Config struct? Wouldn't it solve the problem in a simpler way?

i don't think so, because the Config file is what we would like to share between tests. And if you did that, you would have to duplicate the config file.

clusterloader2/docs/test_suite_design.md
wojtek-t: After thinking a bit more about it, I vote for removing it.
The situation that you will need a common override for all tests in the job is fairly rare (well, from the top of my head I can't really think of any).

And having:

type TestSuite []Test

is conceptually much simpler

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,46 @@
...
+
+Test defines a single test scenario. Config defines path to the config file
+that satisfy already existing cluster loader API. Overrides defines what
+override files should be applied to the config file.
+To enable using same test config in one test suite,
+the Identifier field was introduced. This indetifier will be used to match
+run results to the corresponding test.
+
+```
+type Test struct {
```
wojtek-t: Maybe this should be called TestScenario instead?

clusterloader2/docs/test_suite_design.md
wojtek-t: I added a comment that I vote for removing the "top-level" overrides.

clusterloader2/docs/test_suite_design.md
mm4tt: What I had in mind would to be change the semantics of testconfig flag to be richer and basically take over the Test/TestSuite responsibility.

Instead of having
```
--testconfig=testing/load/density.yaml
--testconfig=testing/load/config.yaml
--testoverrides=./testing/load/kubemark/throughput_override.yaml
```

We'd have 
```
--test=name=load,config=testing/load/config.yaml,overrides=testing/load/kubemark/throughput_override.yaml,testing/load/myoverride.yaml
--test=name=density,config=testing/density/config.yaml,overrides=...
```

And we wouldn't need any additional structs and abstractions for TestSuite and Test.
WDYT?

wojtek-t: As discussed offline - I think this is conceptually exactly what would be a result of applying all of my comments. And I don't have strong opinion if we want to have additional yaml per test-job or much longer flags.

However, after thinking about it a bit more, I think my personal preference would be to:
- have those "job definitions" sit in test-infra repo (in our directory) and actually be yamls as they are more readable than those your funny strings :)
- in test job reference those yamls

clusterloader2/docs/test_suite_design.md
krzysied: IMHO user will only see it as something for matching results. We can still implement it without identifier, however user will have to match results with tests based on timestamp. This doesn't look user-friendly.

clusterloader2/docs/test_suite_design.md
krzysied: > Why do we need that (for more than short-period of time to do migration)? Why can't we just support TestSuite?

The main reason for keeping old falgs was the migration. That's why I wanted to keep both sets of flags. Removing old flags should be an independent step IMHO.

> have those "job definitions" sit in test-infra repo (in our directory) and actually be yamls 

+1 for yamls

clusterloader2/docs/test_suite_design.md
krzysied: I've removed top level overrides.

clusterloader2/docs/test_suite_design.md
krzysied: Changed to ```type TestSuite []TestScenario```

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
+# Cluster loader - TestSuite
+
+## Background
+
+Since 08/2018 cluster loader become main tool for kubernetes e2e performance tests.
```
wojtek-t: We actually migrated to it in 2019 - so until then it wasn't :)

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
...
+TestSuite is a set of test scenarios running in a single test job.
+
+```
+type TestSuite []TestScenario
+```
+
+## Test
+
+TestScenario is a representation of single test scenario.
+Config defines path to the config file
```
wojtek-t: Is there any reason why those lines are so short? :)

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
...
+Since 08/2018 cluster loader become main tool for kubernetes e2e performance tests.
+It is a very convenient tool for running single test case, but not for running
+test bundles. Currently it is impossible to run the same test
+with two different parameter sets in a single job run.
+This forces developers to create separate job configs for
+almost each test.
+
+## TestSuite
+
+TestSuite is a set of test scenarios running in a single test job.
```
wojtek-t: I don't think this comment brings any value :)

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
...
+
+```
+type TestSuite []TestScenario
+```
+
+## Test
+
+TestScenario is a representation of single test scenario.
```
wojtek-t: This sentence also doesn't bring any value :)

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
...
+Config defines path to the config file
+that satisfy already existing cluster loader API.
```
wojtek-t: This should probably be a comment for the field in the type.

clusterloader2/docs/test_suite_design.md
wojtek-t: > The main reason for keeping old falgs was the migration. That's why I wanted to keep both sets of flags. Removing old flags should be an independent step IMHO.

Disagree. For migration we probably need it, but intermittent state shouldn't be defined in the design.

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
...
+Overrides defines what override files should used.
+Each pair <variable, value> from the override file will by applied to
+the config template.
+To enable using same test configs in one test suite,
+the Identifier field was introduced. This identifier will be used to match
+run results to the corresponding test.
+
+```
+type TestScenario struct {
+	Overrides  []string
```
wojtek-t: OverrideFilePaths ?

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,43 @@
...
+	Config     string
```
wojtek-t: ConfigFilePath?
ConfigPath?

clusterloader2/docs/test_suite_design.md
krzysied: Trying to have no more than 80 characters in each line.

clusterloader2/docs/test_suite_design.md
krzysied: `set` -> `list`. List of scenarios suggests that we will be running tests in the same order as provided.

clusterloader2/docs/test_suite_design.md
krzysied: changed to ConfigPath.

clusterloader2/docs/test_suite_design.md
krzysied: Changed to OverridePaths

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Background
+
+Since 2019 cluster loader become main tool for kubernetes e2e performance tests.
+It is a very convenient tool for running single test case, but not for running
+test bundles. Currently it is impossible to run the same test
+with two different parameter sets in a single job run.
+This forces developers to create separate job configs for almost each test.
+
+## TestSuite
```
oxddr: Any chance to have an example of how tests will look like before and after the change. I am having troubles to understand the benefit of this change.

krzysied: This will allow us to have same test template multiple times in a run. For example:
```
- ConfigPath: storage_test.yaml
  OverridePaths:
  - use_storage_type1.yaml
- ConfigPath: storage_test.yaml
  OverridePaths:
  - use_storage_type2.yaml
- ConfigPath: storage_test.yaml
  OverridePaths:
  - use_storage_type3.yaml
```
Previously we had to create separate ci job for each test. IMHO the gains are:
- It's more or less that same thing we define we current cluster loader flags, however we will able to assign overrides to specific test.
- We can bundle test together, so we don't have to create and tear down cluster for each test case. For short tests this will be a huge gain.

clusterloader2/docs/test_suite_design.md
krzysied: I changed structure of the document a little bit. I moved parts of the description to the corresponding fields.

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,50 @@
...
+	ConfigPath string
+	// OverridePaths defines what override files should be applied
+        // to the config specified by the ConfigPath.
+	OverridePaths []string
+}
+```
+
+Assigning overrides on test scenario level has two benefits:
+- Allows to use same template more than once.
+- Allows users to be more flexible with variable naming.
```
wojtek-t: I don't understand this one.

wojtek-t: Ohh - it's explained. So should be sth like:
```
Per-scenario overrides benefits:
- ability to use the same template with multiple overrides in a single job
- less error prone - variables with same names across different configs could have been accidentally overwritten before
```

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,50 @@
...
+Since 2019 cluster loader become main tool for kubernetes e2e performance tests.
+It is a very convenient tool for running single test case, but not for running
+test bundles. Currently it is impossible to run the same test
+with two different parameter sets in a single job run.
+This forces developers to create separate job configs for almost each test.
+
+## TestSuite
+
+TestSuite is a list of test scenarios running in a single test job.
+Tests will be executed by cluster loader in the order as provided.
```
wojtek-t: Do we really need this section? This explanation doesn't bring any value.
I would change to sth like

# Test suite definition
We define test suite as following:

```
type TestSuite []TestScenario

type TestScenario struct {
  ...
```

Those two lines doesn't provide any value. In general - the shorter the doc is the better if it contains all the information needed.

clusterloader2/docs/test_suite_design.md
krzysied: Done.

clusterloader2/docs/test_suite_design.md
krzysied: Done.

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,41 @@
+# Cluster loader - TestSuite
+
+## Background
+
+Since 2019 cluster loader become main tool for kubernetes e2e performance tests.
+It is a very convenient tool for running single test case, but not for running
+test bundles. Currently it is impossible to run the same test
```
wojtek-t: nit:

It's very convenient for running single test case, but currently it's impossible to run the same test with two different sets of parameters in a single job run.

krzysied: Done.

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,40 @@
...
+
+
+```
+type TestSuite []TestScenario
+
+type TestScenario struct {
+	// Identifier is a unique test scenario name across test suite.
+	Identifier string
+	// ConfigPath defines path to the config file that satisfy already
+        // existing cluster loader API.
```
mm4tt: NIT: Fix indent

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,40 @@
...
+	ConfigPath string
+	// OverridePaths defines what override files should be applied
+        // to the config specified by the ConfigPath.
```
mm4tt: Same here

clusterloader2/docs/test_suite_design.md
```diff
@@ -0,0 +1,40 @@
...
+could have been accidentally overwritten before.
+
+To make using same test configs in one test suite more user-friendly,
+the Identifier field was introduced. This identifier will be used to match
+run results to the corresponding test.
+
+## Other
+
+After TestSuite is support by cluster loader, passing config paths and overrides
+using flags will be disabled.
```
mm4tt: NIT: After TestSuite is **supported**

Also, what does it mean in terms of passing testsuite to the clusterloader?  Where the discussion landed? Will there be a new flag for testsuite?
Could you clarify it here so it's clear what we should expect?

wojtek-t: My personal preference would be that:
- we define yamls in test-infra (because they basically describe job config)
- we pass path to yaml to job config

krzysied: > we pass path to yaml to job config

+1

> we define yamls in test-infra (because they basically describe job config)

My assumption is that if test config changes, sometimes you want to update test suite in the same PR. IMHO configs should be in perf-tests.

wojtek-t: As discussed offline - this should live in test-infra, because that doesn't define test - it's only defining job config.

krzysied:
I think I've resolved all comments. @wojtek-t, is there anything to be done before lgtm?

clusterloader2/docs/test_suite_design.md
wojtek-t: Please mention that the supported way will be passing (via flag) a path to config file with TestSuite.

krzysied: Done.

PR #683, opened 2019-07-24
Simplified load test - deployments and services bundled together

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/413

This change is no-op for CI/CD tests.
The simplified load test is is added in a new file `experimental_config.yaml` that will eventually replace the old `config.yaml` once the new config is tested properly.

I run the tests on 100 node cluster and the results look promising:

zb7NsBk7ns5 (1)

### DIFF
--- a/clusterloader2/testing/load/experimental_config.yaml
+++ b/clusterloader2/testing/load/experimental_config.yaml
@@ -1,0 +1,283 @@
+# TODO(https://github.com/kubernetes/perf-tests/issues/413): Rename to config.yaml and delete the old one once tested.
+
+# ASSUMPTIONS:
+# - Underlying cluster should have 100+ nodes.
+# - Number of nodes should be divisible by NODES_PER_NAMESPACE (default 100).
+# - The number of created SVCs is half the number of created Deployments.
+# - Only half of Deployments will be assigned 1-1 to existing SVCs.
+
+#Constants
+{{$NODE_MODE := DefaultParam .NODE_MODE "allnodes"}}
+{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+{{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
+{{$LOAD_TEST_THROUGHPUT := DefaultParam .LOAD_TEST_THROUGHPUT 10}}
+{{$BIG_GROUP_SIZE := 250}}
+{{$MEDIUM_GROUP_SIZE := 30}}
+{{$SMALL_GROUP_SIZE := 5}}
+{{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+{{$PROMETHEUS_SCRAPE_KUBE_PROXY := DefaultParam .PROMETHEUS_SCRAPE_KUBE_PROXY false}}
+{{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/load/config2.yaml
```diff
@@ -0,0 +1,276 @@
+# TODO(https://github.com/kubernetes/perf-tests/issues/413): Rename to config.yaml and delete the old one once tested.
```
oxddr: nit: can we have this file renamed more descriptively? It's hard to tell difference between config and config2. What about config_with_svc_bundle or sth similar?

wojtek-t: experimental_config maybe?

[I'm not sure people should understand it up until now]

clusterloader2/testing/load/config2.yaml
```diff
@@ -0,0 +1,276 @@
...
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      apiVersion: apps/v1
+      kind: Deployment
+      labelSelector: group = load
+      operationTimeout: 15m
+
+- name: Creating Deployments and Services
```
wojtek-t: "Creating bundles" maybe?

Soon we will add something else and will be renaming again.

clusterloader2/testing/load/config2.yaml
```diff
@@ -0,0 +1,276 @@
...
+    replicasPerNamespace: {{$bigDeploymentsPerNamespace}}
+    tuningSet: RandomizedSaturationTimeLimited
+    objectBundle:
+    - basename: big-deployment
+      objectTemplatePath: deployment.yaml
+      templateFillMap:
+        ReplicasMin: {{$BIG_GROUP_SIZE}}
+        ReplicasMax: {{$BIG_GROUP_SIZE}}
+        SvcName: big-service
+    - basename: big-service
```
wojtek-t: Service should be created first (within bundle we are (hopefully) creating objects in serial).

clusterloader2/testing/load/config2.yaml
mm4tt: Changed to "Creating objects" and same for deleting

clusterloader2/testing/load/config2.yaml
mm4tt: Yes, we do it in serial. Done.

clusterloader2/testing/load/config2.yaml
mm4tt: Changed to experimental_config.

clusterloader2/testing/load/experimental_config.yaml
```diff
@@ -0,0 +1,276 @@
...
+  phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: RandomizedSaturationTimeLimited
+    objectBundle:
+    - basename: big-deployment
+      objectTemplatePath: deployment.yaml
+    - basename: big-service
```
wojtek-t: What I started wondering here is that previously, we were sending all endpoints updates of "currently-being-deleted" deployment and only then removing the service and endpoints object.

With this change, for deletion, we will delete service immediately after deleting deployment (and deleting deployment doesn't mean the pods are also deleted - the are only oprhaned and being deleted by GC). So we most probably first delete the service and only then start deleting pods which would trigger endpoint updates.

I'm not sure if it's necessary bad thing to do, but it;s visibly different, so at the very lease it deserves an explicit comment that we're aware of it and this is on purpose.

mm4tt: Done, PTAL

PR #686, opened 2019-07-24
Storage perf tests decrease volumes and pod number

mucahitkurt:
We want to run our experimental tests with the new volumes and pod numbers;
- 10 volumes per pod, for the max volumes per pod test case,
- 25 pods per node, for the max volumes per node test case

and compare with the current test cases' pod_startup time results.

Current pod_startup time for p90 changes between 70 seconds-90 seconds for max volumes per node case.

Current pod_startup time for p90 changes between 22 seconds-24 seconds for configMap and secret, 2.5-3.75 seconds for emptyDir and Downward API for max volumes per pod case.

/sig storage
cc @wojtek-t 
cc @msau42


msau42:
/ok-to-test

IIUC:
* 1 volume per pod, 100 pods takes 70-90 seconds avg to startup?
* 1 pod with 100 volumes takes 20-30 seconds to startup?

I'm very surprised that the 2nd case is faster than the multiple pods case.

msau42:
I just looked at the results at http://perf-dash.k8s.io/#/?jobname=storage-max-emptydir-vol-per-pod&metriccategoryname=E2E&metricname=PodStartup&Metric=pod_startup

So those numbers above are for configmap and secret volume types.

Downward API and emptydir are closer to:
* 70-80 seconds for 100 pods with 1 volume each
* 2-3 seconds for 1 pod with 100 volumes

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_node/override.yaml
```diff
@@ -1,2 +1,2 @@
-TOTAL_PODS: 100
+TOTAL_PODS: 25
```
msau42: should we have different configs for different number of volumes/pods, so we can compare side by side?

mucahitkurt:
> I just looked at the results at http://perf-dash.k8s.io/#/?jobname=storage-max-emptydir-vol-per-pod&metriccategoryname=E2E&metricname=PodStartup&Metric=pod_startup
> 
> So those numbers above are for configmap and secret volume types.
> 
> Downward API and emptydir are closer to:
> 
> * 70-80 seconds for 100 pods with 1 volume each
> * 2-3 seconds for 1 pod with 100 volumes

Sorry I read it wrong for the downward API and emptyDir, they are under 5 seconds for the 100 volumes per pod case, and tests are green, I edited the description.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_node/override.yaml
mucahitkurt: yes, it would be better, it needs additional override files and job definitions. I can do it if it won't cause any side effect for the test infra, cc @wojtek-t

wojtek-t: total pods =100 actually I think should stay because it's roughly how many pods per node we support.
This test case is fairly useful in my opinion.

If we want to test 25 too, this should be a separate config.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_pod/override.yaml
```diff
@@ -1,2 +1,2 @@
 TOTAL_PODS: 1
-VOLUMES_PER_POD: 100
+VOLUMES_PER_POD: 10
```
wojtek-t: I don't have strong opinion about this one. I would like to avoid having donzens of jobs and growing this matrix of scenarios, but I don't know if we should have 10 or 100 here.

wojtek-t: @msau42 - WDYT?

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_node/override.yaml
msau42: Agree, I mainly want to test lower numbers for debugging purposes and find the threshhold where pod startup latency starts to fall off the cliff.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_pod/override.yaml
msau42: Realistically I don't think a single pod will have 100 volumes.  5-10 seems more realistic, but the initial purpose of this experiment is to find our current scaling thresholds.  

I think emptydir and downward API volumes performance is acceptable.  Unsure about secrets and configmaps.  Again I want to find the threshold where performance starts to fall off.

msau42:
@mucahitkurt this initial data is very useful! As we start to change/add new thresholds, can we keep a spreadsheet that summarizes the results?

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_pod/override.yaml
wojtek-t: > I think emptydir and downward API volumes performance is acceptable. Unsure about secrets and configmaps. Again I want to find the threshold where performance starts to fall off.

I think at least part of the problem is in QPS limits - we need to GET each secret/configmap from apiserer before we even start the pod. So with 5qps limit, 100 volumes just means 20s at start.

I think that what we should seriously consider (at least as an experiment) is to bump it to something large (say 100+) and see what will happen. I would even say that it's probably first thing to do.

wojtek-t:
> @mucahitkurt this initial data is very useful! As we start to change/add new thresholds, can we keep a spreadsheet that summarizes the results?

+100
Not sure what the best format is, it can even be some .md file somewhere (maybe even in this repo) - I guess I can live with anything.
But before we make this change, we definitely need to have this data somewhere.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/max_volumes_per_pod/override.yaml
mucahitkurt: Is it meaningful to add the wait time for configmap and secret volumes be available to pod startup time? Should we add some barrier to wait they are available before starting to pod creation?

mucahitkurt:
> > @mucahitkurt this initial data is very useful! As we start to change/add new thresholds, can we keep a spreadsheet that summarizes the results?
> 
> +100
> Not sure what the best format is, it can even be some .md file somewhere (maybe even in this repo) - I guess I can live with anything.
> But before we make this change, we definitely need to have this data somewhere.

I'm looking into how I can do this

mucahitkurt:
@wojtek-t , @msau42 

For test results data, I wrote [a script like this](https://gist.github.com/mucahitkurt/7a40c5b38440473cced0ceb4a00090f0) and fetched the test results for ephemeral volumes podstartup latencies. I created a json file for the test results and added it to the PR, I thought that we can use it easily for further analysis, it's size is 1MB, I'm not sure whether this's a problem or not.

I created a different override file for the `PODS_PER_NODE: 25` case.

I'm not sure about the final decision about `VOLUMES_PER_POD: 10` case, should we add a different override file, or changing existing one, or should we test 100+ volumes case first, as @wojtek-t mentioned?

clusterloader2/testing/experimental/storage/pod-startup/max_volumes_per_node/25_pods_per_node/override.yaml
```diff
@@ -0,0 +1,3 @@
+PODS_PER_NODE: 25
```
mucahitkurt: I keep the default as 100, since this one is experimental, didn't create a different directory for the 100 pods case.

clusterloader2/testing/experimental/storage/pod-startup/max_volumes_per_node/25_pods_per_node/override.yaml
```diff
@@ -0,0 +1,3 @@
...
+VOLUMES_PER_POD: 1
+NODES_PER_NAMESPACE: 1
```
wojtek-t: I couldn't comment on the file below - so let me comment here.
After looking into that, I don't think we should commit this file into this repo.
In fact, I just realized that it only contains data that we already have and can actually be viewed in perf-dash.k8s.io
[well, something is broken as of now it doesn't show results since uyesterday morning, but I'm pretty sure it's minor issue]

So I think we should just get rid of that json file.

mucahitkurt: So I think I got it wrong your and @msau42 comment about keeping initial data.

I understand it as keeping test results, yes this data can be seen from dashboard but I thought it only shows some windowed data, and at the future we cant see initial results.

So could you give me more detail about what data we want to keep, is it like a test scenario - test result summary pairs?

wojtek-t: > I understand it as keeping test results, yes this data can be seen from dashboard but I thought it only shows some windowed data, and at the future we cant see initial results.

yes - it only shows some window (1month? or sth like that).

> So could you give me more detail about what data we want to keep, is it like a test scenario - test result summary pairs?

My personal feeling is that all we need is:
scenario -> [pod startup time percentiles] (just 3 numbers)
mapping.

But let's wait for @msau42 to get her opinion too.

msau42: Yes I agree I just meant keep a summary of our experiments while we're adjusting all the parameters.  i'm interested in:
* Scenario: max-per-pod, max-per-node
* Important parameters: # pods, # volumes per pod
* Pod startup time percentiles
* Link to one of the test logs? How long do we keep these test run artifacts?

wojtek-t: Yes - by scenario I meant both what it does and it's params.

Re link to logs - I think we have TTL of 2months.

mucahitkurt: ok, I'll prepare that kind of doc, I think @wojtek-t qps change is another parameter in here.

mucahitkurt: I replaced json file with a yaml file for test summary information

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
```diff
@@ -0,0 +1,119 @@
+pod-with-volumes-startup-latency-scenarios:
+  scenario: max-volumes-per-pod
+    number-of-volumes-per-pod: 100
+    number-of-pods-per-node: 1
+    number-of-node: 1
+    volume-types:
+      volume-type: configmap
+        before-qps-limit-bump: true
```
wojtek-t: instead of "before-qps-limit-bump" I would add specific value for qps-limit (default is 5, I bumped to 100)

The reason is that we may want to experiment with other values too.

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
```diff
@@ -0,0 +1,119 @@
...
+        pod-startup:
+          perc50: 22240.349315
+          perc90: 22240.349315
+          perc99: 22240.349315
+          test-artifacts: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability-max-configmap-vol-per-pod/1156085407868260353/artifacts/
```
wojtek-t: This one is tricky - because this dir (depending on the time) contains results from before-qps-bump, after it (and once new changes appear additional configurations).

So technically it would be more correct to move it to the level of "volume-type".
[So what i have on my mind is the structure:
volume-types:
```
- volume type: configmap
  test-artifacts: ...
  results:
    param1: value1 (e.g. qps)
    param2: value2
    pod-startup:
```

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
mucahitkurt: 100 or 10?

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
mucahitkurt: Do you think test artifacts link points to all test run artifacts of the volume type? If so, actually I take a specific test run for the percentiles, because it seems that most of them have almost similar results, and the artifact link points to that build, e.g. a specific test run.

So if it's not correct taking one test run instead of an average of many, can it be like below?
```
- volume type: configmap
  results:
    test-artifacts: ...
      param1: value1 (e.g. qps)
      param2: value2
      pod-startup:
```

mucahitkurt: I think I got it wrong, I'll take the volume entity as root and add different test scenario results under it, like below;
```
pod-with-volumes-startup-latency-scenarios:
  volume-type: configmap
    test-artifacts: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability-max-configmap-vol-per-pod/
    results:
      scenario-1:
        number-of-volumes-per-pod: 100
        number-of-pods-per-node: 1
        number-of-node: 1
        qps-limit: 5
        pod-startup:
          perc50: 22240.349315
          perc90: 22240.349315
          perc99: 22240.349315
      scenario-2:
        number-of-volumes-per-pod: 100
        number-of-pods-per-node: 1
        number-of-node: 1
        qps-limit: 10
        pod-startup:
          perc50: 2205.065107
          perc90: 2205.065107
          perc99: 2205.065107
```

wojtek-t: Yes - that's exactly what I had on my mind - sorry for not being clear

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
```diff
@@ -0,0 +1,170 @@
...
+          qps-limit: 5
+          pod-startup:
+            perc50: 22240.349315
+            perc90: 22240.349315
+            perc99: 22240.349315
+        scenario-2:
+          number-of-volumes-per-pod: 100
+          number-of-pods-per-node: 1
+          number-of-node: 1
+          qps-limit: 10
```
wojtek-t: 100

Let me clarify, because in fact I was changing two different things separately:
- first I bumped kube-api-qps limit in kubelet from 5 to 100: https://github.com/kubernetes/test-infra/pull/13679
- then I reduced speed of test from creating pods from ~100/s to 10/s: https://github.com/kubernetes/perf-tests/pull/720

These two things are mixed together in this file.

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
```diff
@@ -0,0 +1,170 @@
+pod-with-volumes-startup-latency-scenarios:
+  volume-type: configmap
```
wojtek-t: nit: is this a valid yaml?

I think that for slices/arrays, you have to have "-" in yaml. I mean:
- volume-type

[not that this is particularly important]

mucahitkurt: yes that's not valid, I try to create a valid one, thanks for heads up!

clusterloader2/testing/experimental/storage/pod-startup/test-result/volume_podstartup_time_results.yaml
```diff
@@ -0,0 +1,159 @@
...
+      number-of-node: 1
+      kube-api-qps-limit: 5
+      pod-startup:
+        perc50: 74109.157201
+        perc90: 89109.291651
+        perc99: 93112.022213
+    - number-of-volumes-per-pod: 1
+      number-of-pods-per-node: 100
+      number-of-node: 1
+      kube-api-qps-limit: 100
```
wojtek-t: We also switched from generating pods ~100pods/s to generating pods 10pods/s.

I think this should also be catched in this file.

mucahitkurt: agree, I added that info as `qps-pod-throughput`

clusterloader2/testing/experimental/storage/pod-startup/max_volumes_per_node/25_pods_per_node/override.yaml
wojtek-t: please also add TEST_NAME (as it is here):
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/experimental/storage/pod-startup/max_volumes_per_node/override.yaml

wojtek-t:
one last comment - lgtm once applied

thanks!

clusterloader2/testing/experimental/storage/pod-startup/max_volumes_per_node/25_pods_per_node/override.yaml
mucahitkurt: Added

mucahitkurt:
@wojtek-t Should I wait for test-suite implementation to add this test case to periodic jobs? Otherwise I should define additional 4 jobs(I think PVC side is not activated yet).

wojtek-t:
> @wojtek-t Should I wait for test-suite implementation to add this test case to periodic jobs? Otherwise I should define additional 4 jobs(I think PVC side is not activated yet).

Let's wait - I think we're pretty close.

/lgtm
/approve

Issue #666, opened 2019-07-23
Don't use raw pods in storage experimental tests


wojtek-t:
Forked from https://github.com/kubernetes/perf-tests/pull/662

@wojtek-t : I'm fine with it as a starting phase, but we should move them to use deployments (using individual pods is more like anti-pattern, and startup-time is measured from pod creation anyway, so it shouldn't cause any additional problems).

@mucahitkurt : When I use a controller like deployments I can't control the volume names of the pod that controller creates.

I tried it with statefulset, for replication number 2, statefulset creates pod likes app-0-pod-0, app-0-pod-1, if I want to use volumes with these pod I should create volumes likes pod-0-vol-0, pod-0-vol-1, pod-1-vol-0, pod-1-vol-1. But I can't create volume names like 2 indexed values for pod-[podIndex]-vol-[volumeIndex], because clusterloader2 framework doesn't allow this.

I know that it's an anti-pattern working directly with the pods without a controller, but I couldn't solve the above problem and used directly pod.

wojtek-t:
> When I use a controller like deployments I can't control the volume names of the pod that controller creates

@mucahitkurt - I think I don't fully understand it.
What I was suggesting is that instead of creating N pods, create N deployments, with 1 pod each.
Given that deployment contains PodSpec which is exactly the same as PodSpec in Pod, you should basically be able to do exactly the same in both cases.

What am I missing?

mucahitkurt:
@wojtek-t when I use N deployments instead of N pods there should't be problem, I thought/tried replication properties of the controller to create N pods with one deployment. I thought creating N deployments is not meaningful

I'll try it and will change existing test cases.

wojtek-t:
@mucahitkurt - sorry, I guess I wasn't clear. I had N deployments on my mind since the early beginning - otherwise we don't control speed of creation, etc.

PR #687, opened 2019-07-24
Use deployment instead of raw pod for storage perf tests

mucahitkurt:
Fix https://github.com/kubernetes/perf-tests/issues/666

Raw pod object templates are changed with deployment templates and the same pods definitions are used inside deployments. Deployment-Pod relation is 1-1.

Also test config files are changed according to these new deployment templates.

/sig storage
/assign @wojtek-t 
cc @msau42


msau42:
@wojtek-t are scale tests run on regional clusters?  I'm thinking of when we want to reuse these for PVC tests too and each PVC is restricted to a single zone.  To support multiple PVCs in a Pod case, we would need to either:

* Use delayed volume binding to ensure all pvcs within a pod are provisioned in the same zone as the pod.  This will add provisioning into pod startup latency though.
* Use StatefulSet, which hashes all PVCs within a StatefulSet to provision from the same zone.

hantaowang:
How does this work for config and secret where the name of the config/secret is different for every pod?

mucahitkurt:
> How does this work for config and secret where the name of the config/secret is different for every pod?

@hantaowang I use N deployments for N pods, so the relation is 1-1, and volume matching shouldn't be different for the raw pod case, I create N volumes and match these volumes the N pods inside the N deployments.

clusterloader2/testing/experimental/storage/pod-startup/ephemeral-volumes/1_node/config.yaml
```diff
@@ -3,7 +3,7 @@
 
 #Constants
 {{$TOTAL_PODS := DefaultParam .TOTAL_PODS 1}}
-{{$POD_TEMPLATE_PATH := DefaultParam .POD_TEMPLATE_PATH "../volume-types/emptydir/pod_with_emptydir.yaml"}}
+{{$DEPLOYMENT_TEMPLATE_PATH := DefaultParam .DEPLOYMENT_TEMPLATE_PATH "../volume-types/emptydir/deployment_with_emptydir.yaml"}}
```
wojtek-t: I'm sorry - I was blocking @hantaowang PRs for quite some time and your stuff, so this time I will do it the other way (unfortunately you're both touching the same stuff).
https://github.com/kubernetes/perf-tests/pull/688 is super useful, so I would like that one to merge first and then I will ask you to rebase this one on top.

hantaowang:
Doesn't this then also stress the deployment controller? We may inadvertently become bottle necked by the deployment and replicaset controller despite the fact that so many deployments are rare. Deployments are definitely a dimension to stress on, but imo for storage that should be eliminated as a variable.

wojtek-t:
> Doesn't this then also stress the deployment controller? We may inadvertently become bottle necked by the deployment and replicaset controller despite the fact that so many deployments are rare. Deployments are definitely a dimension to stress on, but imo for storage that should be eliminated as a variable.

Disagree - this number of deployments actually isn't very high (we're testing higher numbers). And we should be using what users are using not something artificial.

hantaowang:
Is it that realistic to use 1 deployment per pod?

I'm thinking more for when the tests become larger scale (say 5000 nodes at 10  pods per node). I don't see anyone using 50000 deployments to manage this as that would definitely overload the controllers.

wojtek-t:
> Is it that realistic to use 1 deployment per pod?

I've seen a bunch of large user clusters with a huge number of those - yes it is realistic.

> I'm thinking more for when the tests become larger scale (say 5000 nodes at 10 pods per node). I don't see anyone using 50000 deployments to manage this as that would definitely overload the controllers.

I wouldn't be that convinced that this will for sure overload controllers.
We will see when we get to this scale - I think we will hit many other limitations before that.

mucahitkurt:
waiting for the [PVC tests PR's](https://github.com/kubernetes/perf-tests/pull/657) merge to continue

wojtek-t:
That PR has just merged - can you please rebase?

mucahitkurt:
@wojtek-t, @hantaowang pod usages are converted to deployment usage

clusterloader2/testing/experimental/storage/pod-startup/volume-types/configmap/deployment_with_configmap.yaml
```diff
@@ -0,0 +1,48 @@
+{{$index := .Index}}
+{{$volumesPerPod := .VolumesPerPod}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  labels:
+    app: {{.Name}}
```
wojtek-t: What do we need "app" label for?

clusterloader2/testing/experimental/storage/pod-startup/volume-types/configmap/deployment_with_configmap.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - name: vol-{{$volumeIndex}}
+          mountPath: /usr/share/{{$volumeIndex}}
+        {{ end }}
+      volumes:
+      {{ range $volumeIndex, $vol := Seq .VolumesPerPod }}
+      - name: vol-{{$volumeIndex}}
+        configMap:
+          name: vol-{{AddInt $volumeIndex (MultiplyInt $index $volumesPerPod)}}
+      {{ end }}
+      dnsPolicy: Default
```
wojtek-t: We don't need to set it - it will be defaulted anyway.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/configmap/deployment_with_configmap.yaml
```diff
@@ -0,0 +1,48 @@
...
+      terminationGracePeriodSeconds: 1
```
wojtek-t: Please remove - I don't think we need it for anything.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/configmap/deployment_with_configmap.yaml
```diff
@@ -0,0 +1,48 @@
...
+      # failure doesn't trigger pod deletion.
+      tolerations:
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 900
+      - key: "node.kubernetes.io/unreachable"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 900
```
wojtek-t: All the comments in this file apply to all other files too.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/configmap/deployment_with_configmap.yaml
mucahitkurt: I think we don't need it

wojtek-t:
/lgtm
/approve

thanks!

PR #688, opened 2019-07-25
Reorganize storage tests

hantaowang:
This PR reorganizes the storage tests so there is no distinction between ephemeral and persistent storage types. It also combines the test config for 1 node and multi node tests, which was largely repeated.

@wojtek-t @davidz627 @jingxu97 @mucahitkurt


hantaowang:
/hold
because untested. I will test tomorrow.

hantaowang:
/hold cancel 
This has been tested.

clusterloader2/testing/experimental/storage/pod-startup/tests/max_volumes_per_node/override.yaml
```diff
@@ -0,0 +1,3 @@
+PODS_PER_NODE: 100
```
wojtek-t: Please remove the "tests/" directory (this whole directory clusterloader2/testing is only about tests).

clusterloader2/testing/experimental/storage/pod-startup/max_volumes_per_node/override.yaml

is enough

clusterloader2/testing/experimental/storage/pod-startup/tests/max_volumes_per_node/override.yaml
```diff
@@ -0,0 +1,3 @@
...
+VOLUMES_PER_POD: 1
+NODES_PER_NAMESPACE: 1
```
wojtek-t: nit: please add new lines at the end of files

wojtek-t: Same in other files you added.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -8,7 +7,7 @@
 {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 
-{{$POD_TEMPLATE_PATH := DefaultParam .POD_TEMPLATE_PATH "../emptydir/pod_with_emptydir.yaml"}}
+{{$POD_TEMPLATE_PATH := DefaultParam .POD_TEMPLATE_PATH "../volume-types/emptydir/pod_with_emptydir.yaml"}}
```
wojtek-t: Obviously this PR is going to break all the existing tests that are running.
Given they are experimental I'm fine with breaking them for a moment, but you should prepare a PR in test-infra that will fix the paths to configs and overrides so that we can merge that immediately after this PR will be merged.

clusterloader2/testing/experimental/storage/pod-startup/tests/max_volumes_per_node/override.yaml
hantaowang: done, i thought it would be confusing if the tests are at the same level as the volume-types

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: With https://github.com/kubernetes/test-infra/pull/13601 opened, I'm fine with approving this one.

PR #691, opened 2019-07-25
Gather etcd mutex profile and refactor profile.go

mm4tt:
Refactoring in profile.go allows adding new profile types much easier, it also removes duplicated and unnecessary code.

### DIFF
--- a/clusterloader2/pkg/measurement/common/profile.go
+++ b/clusterloader2/pkg/measurement/common/profile.go
@@ -18,7 +18,6 @@ package common
 
 import (
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mm4tt:
Ref. https://github.com/kubernetes/kubernetes/issues/80209

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -75,13 +76,20 @@ type profileMeasurement struct {
...
+	return func() measurement.Measurement {
+		return &profileMeasurement{
+			name:   name,
+			config: &profileConfig{kind: kind},
+		}
+	}
+}
+
+func (p *profileMeasurement) start(config *measurement.MeasurementConfig) error {
+	err := p.populateProfileConfig(config)
```
wojtek-t: if err := ...; err != nil {
  return err
}

wojtek-t:
Also travis is failing - golint.

clusterloader2/pkg/measurement/common/profile.go
mm4tt: Done.

Issue #694, opened 2019-07-25
ResourceUsage measurement is essentially broken for short tests


mm4tt:
See https://github.com/kubernetes/kubernetes/issues/80212#issuecomment-515010726
Essentially two same runs in terms of scheduler CPU usage, one reported cpu usage of 0.5 other of 0.1.

Everything comes from the fact that we scrape the resource usage only a few times during short tests. In such case both 99th and 90th are the max observed usage, which doesn't make sense if cpu characteristic is similar to the scheduler in density test. Scheduler CPU has one initial spike (caused by saturation phase) and then a flat area (latency phase). Depending on if we hit that spike or not we'll get completely different results.

mm4tt:
This is not done.

/reopen

PR #697, opened 2019-07-26
Gather resource usage on master every 10s

mm4tt:
This is to address https://github.com/kubernetes/perf-tests/issues/694.

It's enough to scrape more often only on master as when it comes to nodes'components we scrape them much more often on average, e.g. for 100 node cluster we have 100 times more data for nodes components as each node is scrape separately every 60s.

In addition it should be safe to hit cadvisor every 10s on master due to two reasons:

1. We do it already via prometheus every 5s (and it doesn't seem to be impacting master's performance at all)
2. From the tests I run looks like cadvisor updates every ~10-16s and request are fast (~100ms), so it implements caching and requests are cheap and in addition there is no gain in scraping it often than 10s


mm4tt:
/hold 
It's not a good idea to submit this on Friday, let's wait till Monday

wojtek-t:
I'm fine with that, but let's wait with merging until Monday.

/lgtm

mm4tt:
/hold cancel
@oxddr, FYI

PR #702, opened 2019-07-29
Clean up load test experiments

mm4tt:
1. Move and rename experimental_load to bundled_services_and_deployments and add a comment in it that it's currently frozen

2. Create extended_config.yaml which is an exact copy of load/config at this point. This will make reviewing future changes easier.


clusterloader2/testing/load/experimental/extended_config.yaml
```diff
@@ -0,0 +1,313 @@
+# ASSUMPTIONS:
```
wojtek-t: What is the difference between this and the other one that you moved?

mm4tt: This is the exact copy of the current load/config.yaml. Because we cannot have services and deployments bundled together I'm starting with a fresh load/config.yaml. I'll be adding new objects here, starting with secrets and configmaps

PR #703, opened 2019-07-29
Create configmaps and secrets in extended load test

mm4tt:
https://github.com/kubernetes/perf-tests/issues/704

### DIFF
--- a/clusterloader2/pkg/config/template_functions.go
+++ b/clusterloader2/pkg/config/template_functions.go
@@ -52,6 +52,7 @@ func GetFuncs() template.FuncMap {
 		"MinInt":        minInt,
 		"MaxFloat":      maxFloat,
 		"MinFloat":      minFloat,
+		"Mod":           mod,
 		"IsEven":        isEven,
 		"IsOdd":         isOdd,
 		"DefaultParam":  defaultParam,
@@ -181,6 +182,10 @@ func minFloat(numbers ...interface{}) float64 {
 		min = math.Min(min, toFloat64(number))
 	}
 	return min
+}
+
+func mod(a interface{}, b interface{}) int {
+	return int(toFloat64(a)) % int(toFloat64(b))
 }
 
 func isEven(number interface{}) bool {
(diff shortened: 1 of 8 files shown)


clusterloader2/testing/load/deployment.yaml
```diff
@@ -24,6 +27,15 @@ spec:
           requests:
             cpu: 10m
             memory: "10M"
+        volumeMounts:
+          {{if and $EnableConfigMaps (IsModIn .Index 20 8 19) }} # .Index % 20 in {8,19} - 10% deployments will have ConfigMap
```
wojtek-t: Given that we're randomly distributing instances (indexes) over time, I think "IsDivisible" would actually be enough and easier to follow.

mm4tt: My intention was to have 10% of pods with secrets, 10% of pods with configmaps and to  exercise all combinations, i.e. to have:
1. a pod with just secret
2. a pod with just configmap
3. a pod with both

While I agree that it may be hard to parse at the first glance, it allows to easily express arbitrary configurations. 

Let's discuss this tomorrow in person.

wojtek-t: OK - I see your point.
Then my question is why do we need "InSet"? Why "IsModEqual" not enough?

clusterloader2/pkg/config/template_functions.go
```diff
@@ -187,6 +188,15 @@ func isEven(number interface{}) bool {
 	return int(toFloat64(number))%2 == 0
 }
 
+func isModIn(a interface{}, mod interface{}, acceptedValues ...interface{}) bool {
```
wojtek-t: Since you introduce it, can you also remove IsEven and IsOdd ?

clusterloader2/testing/load/deployment.yaml
mm4tt: Realized that built-in eq supports multiple values. In that cases implementing just Mod is enough, which I did.

clusterloader2/pkg/config/template_functions.go
mm4tt: Will do that in a follow up PR, I have some other things to change as well.

mm4tt:
@wojtek-t, PTAL

PR #705, opened 2019-07-29
add functionality to deploy cloud-sa secret and driver

hantaowang:
@davidz627 @wojtek-t @jingxu97 @verult 

This PR adds the ability to test PVs provisioned and managed by the CSI GCP PD Driver. Included is a change to cl2's run-e2e.sh that deploys a secret (if it exists) and a yaml that deploys the driver and all its associates rbacs and SAs (one of which uses this secret).

I will rebase this after https://github.com/kubernetes/perf-tests/pull/657 is merged so that the ability to deploy the driver yaml (if necessary) is build into the storage test config. It will be turned off by default and an additional override file in volume-types/persistentvolume will turn it on.

### DIFF
--- a/clusterloader2/run-e2e.sh
+++ b/clusterloader2/run-e2e.sh
@@ -22,5 +22,16 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
 export KUBEMARK_ROOT_KUBECONFIG="${KUBEMARK_ROOT_KUBECONFIG:-${HOME}/.kube/config}"
 
+# Deploy the GCP PD CSI Driver if required
+if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
+   if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
+      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
+      exit 1
+   fi
+   kubectl create secret generic cloud-sa --from-file="${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}"
+   kubectl apply -f ${CLUSTERLOADER_ROOT}/drivers/gcp-csi-driver-stable.yaml
+   kubectl wait pods -l app=gcp-compute-persistent-disk-csi-driver --for condition=Ready --timeout=300s
+fi
+
 cd ${CLUSTERLOADER_ROOT}/ && go build -o clusterloader './cmd/'
 ./clusterloader --alsologtostderr "$@"
--- a/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
@@ -13,6 +13,7 @@
   {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
   {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
   {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+  {{$USE_CSI := DefaultParam .USE_CSI false}}
   {{$VOL_SIZE := DefaultParam .VOL_SIZE "8Gi"}}
   {{$WAIT_FOR_PVC := DefaultParam .WAIT_FOR_PVC false}}
   {{$POD_THROUGHPUT := DefaultParam .POD_THROUGHPUT 10}}
@@ -70,6 +71,7 @@ steps:
       templateFillMap:
         Group: volume-test
         VolSize: {{$VOL_SIZE}}
+        UseCSI: {{$USE_CSI}}
 {{ end }}
 {{ if $WAIT_FOR_PVC }}
 - measurements:
--- a/clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
@@ -8,6 +8,9 @@ metadata:
 spec:
   accessModes:
     - ReadWriteOnce
+  {{ if .UseCSI }}
+  storageClassName: csi-gce-pd
+  {{ end }}
   resources:
     requests:
       storage: {{.VolSize}}
--- a/clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/usecsi.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/usecsi.yaml
@@ -1,0 +1,1 @@
+USE_CSI: true
--- a/clusterloader2/drivers/gcp-csi-driver-stable.yaml
+++ b/clusterloader2/drivers/gcp-csi-driver-stable.yaml
@@ -1,0 +1,354 @@
+# This config generated from the GCP PD CSI Driver
+# https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver
+# with the command: kustomize build deploy/kubernetes/overlays/stable
+# and an addition storage class from examples/kubernetes/zonal-sc-example.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: csi-controller-sa
+  namespace: default
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: csi-node-sa
+  namespace: default
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: driver-registrar-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: external-attacher-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumes
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - storage.k8s.io
+    resources:
+      - csinodes
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - storage.k8s.io
+    resources:
+      - volumeattachments
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: external-provisioner-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumes
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumeclaims
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+  - apiGroups:
+      - storage.k8s.io
+    resources:
+      - storageclasses
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - list
+      - watch
+      - create
+      - update
+      - patch
+  - apiGroups:
+      - storage.k8s.io
+    resources:
+      - csinodes
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: csi-controller-attacher-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: external-attacher-role
+subjects:
+  - kind: ServiceAccount
+    name: csi-controller-sa
+    namespace: default
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: csi-controller-provisioner-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: external-provisioner-role
+subjects:
+  - kind: ServiceAccount
+    name: csi-controller-sa
+    namespace: default
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: driver-registrar-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: driver-registrar-role
+subjects:
+  - kind: ServiceAccount
+    name: csi-node-sa
+    namespace: default
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  labels:
+    app: gcp-compute-persistent-disk-csi-driver
+  name: csi-gce-pd-controller
+  namespace: default
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: gcp-compute-persistent-disk-csi-driver
+  serviceName: csi-gce-pd
+  template:
+    metadata:
+      labels:
+        app: gcp-compute-persistent-disk-csi-driver
+    spec:
+      containers:
+        - args:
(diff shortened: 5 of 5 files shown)


hantaowang:
/hold 
for rebase

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,10 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
 export KUBEMARK_ROOT_KUBECONFIG="${KUBEMARK_ROOT_KUBECONFIG:-${HOME}/.kube/config}"
 
+# Deploy a secret that contains the e2e Google credentials
+if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
```
wojtek-t: 1. Do we have access to this env-var here? [I hope we do, just asking if you verified it.]

2. Please add :- to avoid using unset var (both here and below)

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -56,6 +57,23 @@ steps:
...
+        max: 1
+      replicasPerNamespace: 1
+      tuningSet: Sequence
+      objectBundle:
+          objectTemplatePath: {{.CSI_DRIVER_PATH}}
+- measurements:
+    - Identifier: WaitForRunningDriverPods
+      Method: WaitForRunningPods
+      Params:
+        desiredPodCount: {{SumInt .Nodes 1}}
```
wojtek-t: Why N+1 ?

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/csi-override.yaml
```diff
@@ -0,0 +1,2 @@
+DEPLOY_CSI_DRIVER: true
+CSI_DRIVER_PATH: "volume-types/persistentvolume/gcp-csi-driver-stable.yaml"
```
wojtek-t: nit: please add end-line here

hantaowang:
Ive rebased it, but as it stands this PR 99% will not work. A couple to consider:
1) what will happen if i hard code default into the driver yaml and the CL2 tries to deploy to a test namespace? The driver and its secrets should be in namespace default.
2) if an error is raised or the test namespace overrides the hard coded one, then should i add functionality to manually specify a namespace in phase? This breaks the abstraction of phase a little because phasr allows multiple possible namespaces, hence a range.
3) if not, then how can i deploy the driver? Possibly via a script to be run pre test, but then this script will be an exact copy of the deploy script in the pd driver repo.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/gcp-csi-driver-stable.yaml
```diff
@@ -0,0 +1,343 @@
+apiVersion: v1
```
wojtek-t: @msau42 @davidz627 - can you please review this file - I don't know enough about what exactly is needed and how to configure it to make it work.

clusterloader2/run-e2e.sh
hantaowang: The env var is set by prow afaik and is inherited for all prow jobs.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: One driver per node plus a cluster wide driver.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/gcp-csi-driver-stable.yaml
hantaowang: Please see my latest comment. Tldr: it wont but raises an interesting problem.

wojtek-t:
> Ive rebased it, but as it stands this PR 99% will not work. A couple to consider:
1) what will happen if i hard code default into the driver yaml and the CL2 tries to deploy to a test namespace? The driver and its secrets should be in namespace default.
2) if an error is raised or the test namespace overrides the hard coded one, then should i add functionality to manually specify a namespace in phase? This breaks the abstraction of phase a little because phasr allows multiple possible namespaces, hence a range.
3) if not, then how can i deploy the driver? Possibly via a script to be run pre test, but then this script will be an exact copy of the deploy script in the pd driver repo.

I missed this initially - those are good question.
I don't have very strong opinion here - I believe we should allow users to overwrite namespaces if they want (so not specifying namespace range would mean that we don't try to use automanaged ones).
OTOH, I would also be fine with running something as part run-e2e.sh before we trigger CL, but then we need some way to wait for waiting for them, so we may need to reimplement some stuff too...

@mm4tt - for thoughts too

hantaowang:
> OTOH, I would also be fine with running something as part run-e2e.sh before we trigger CL, but then we need some way to wait for waiting for them, so we may need to reimplement some stuff too...

I should point out in our tests we do this by just waiting 60 seconds after deploying the driver. But i dont know if that will work for larger scale clusters.

msau42:
Keep in mind that eventually the driver will be bundled in gke. Which can be optin during gcloud container create

msau42:
Regarding waiting for the driver to be up, we can watch csinode and wait for the driver to be registered

clusterloader2/run-e2e.sh
davidz627: log/echo a warning if this couldn't be found. Looks like if the file doesn't exist we just give up

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
davidz627: worth commenting in the code here

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/gcp-csi-driver-stable.yaml
davidz627: this looks like a copy paste of the `stable` driver :) Please comment at top of file a xref to the original files.

This is the correct set of yamls to deploy a driver into a cluster (along with the secret defined in that other file)

hantaowang:
> Regarding waiting for the driver to be up, we can watch csinode and wait for the driver to be registered

If we just deploy with an addition script in run-e2e.sh, we can actually just use `kubectl wait` I believe. (https://www.jeffgeerling.com/blog/2018/updating-kubernetes-deployment-and-waiting-it-roll-out-shell-script) This also makes the config a lot simplier because CSI and non CSI configs will now look the same where as previously I would need an `{{ if $CSI }}` condition.

wojtek-t:
@hantaowang - did you verify that this actually works? I think that @mm4tt yesterday mentioned that having multiple objects in one file will not work with CL...

> If we just deploy with an addition script in run-e2e.sh, we can actually just use kubectl wait I believe. (https://www.jeffgeerling.com/blog/2018/updating-kubernetes-deployment-and-waiting-it-roll-out-shell-script) This also makes the config a lot simplier because CSI and non CSI configs will now look the same where as previously I would need an {{ if $CSI }} condition.

I'm fine with that too. Technically, installing driver is more cluster-setup than test itself, so I would be fine with running something in run-e2e (if we can wait for it)

hantaowang:
> @hantaowang - did you verify that this actually works? I think that @mm4tt yesterday mentioned that having multiple objects in one file will not work with CL...

No I believe with how object bundles work right now, it  won't work. This was just to mock what deploying with CL2 would look like. The default namespace issue also presents a problem, so thats another thing that won't work.

I'll try the run-e2e.sh method and make sure that works, and then ill push here and remove the hold.

hantaowang:
I tested this method and it does work. The deployment is built into run-e2e.sh. There is no tear down, but that should be fine - we just need to not run csi and non csi tests on the same cluster.

To use it, add `--deploy-csi-driver` as the first flag of run-e2e.sh. Then override the `USE_CSI` config variable.

/assign @wojtek-t 
/cc @davidz627 @jingxu97
cc @msau42 @verult
/hold cancel

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,19 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
 export KUBEMARK_ROOT_KUBECONFIG="${KUBEMARK_ROOT_KUBECONFIG:-${HOME}/.kube/config}"
 
+# Deploy the GCP PD CSI Driver is required
+if [ $1 = "--deploy-csi-driver" ]; then
```
davidz627: I don't know how back params work but I feel like theres a better way to do this, doesn't this just check the first arg for an exact string match? What if I use the "single-dash" syntax instead.

Worth investigating other methods, I've seen other things before that seem to work.

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,19 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
...
+   echo "${E2E_GOOGLE_APPLICATION_CREDENTIALS}"
```
davidz627: do we need to print the contents?

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,19 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
...
+   if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
+      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
+      exit 1
+   else
+      kubectl create secret generic cloud-sa --from-file="${E2E_GOOGLE_APPLICATION_CREDENTIALS}"
+   fi
+   kubectl apply -f ${CLUSTERLOADER_ROOT}/gcp-csi-driver-stable.yaml
+   kubectl wait pods -l app=gcp-compute-persistent-disk-csi-driver --for condition=Ready --timeout=300s
```
davidz627: cool I didn't know this existed :)

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -13,6 +13,7 @@
   {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
   {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
   {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
+  {{$USE_CSI := DefaultParam .USE_CSI false}}
```
davidz627: should this param be more specific like `DEPLOY_PD_CSI_DRIVER` or something?

clusterloader2/run-e2e.sh
hantaowang: oh sorry, that was for testing

clusterloader2/run-e2e.sh
hantaowang: I not sure what other methods can work. i didnt want to check every argument although that is a possibility. Its a bit harder since that arg cant be passed to CL2, so it needs to be spliced out. Open to suggestions.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
hantaowang: I think in the future, we may want to support more drivers to test, so its vague here in the config on purpose.

clusterloader2/gcp-csi-driver-stable.yaml
```diff
@@ -0,0 +1,354 @@
+# This config generated from the GCP PD CSI Driver
```
wojtek-t: I'm not a fan of putting this file in the clusterloader2/ main dir.
Can you move it to "clusterloader2/drivers" or "cl2/csi/" or sth like that.

[e.g. cl2/testing/prometheus contains manifests

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,18 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
 export KUBEMARK_ROOT_KUBECONFIG="${KUBEMARK_ROOT_KUBECONFIG:-${HOME}/.kube/config}"
 
+# Deploy the GCP PD CSI Driver is required
+if [ $1 = "--deploy-csi-driver" ]; then
```
wojtek-t: I'm not a fan of this approach - other than supporting different flag, I prefer to be much more explicit.
Can we instead simply introduce a dedicated env var and use it.

e.g. "DEPLOY_GCI_DRIVER" (and just check if it is true).

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,18 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
...
+   if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
```
wojtek-t: :-

otherwise you're risking that this will fail with unbounded variable

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,18 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
...
+      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
+      exit 1
+   else
+      kubectl create secret generic cloud-sa --from-file="${E2E_GOOGLE_APPLICATION_CREDENTIALS}"
```
wojtek-t: same here

clusterloader2/run-e2e.sh
wojtek-t: me neither :)

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,18 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
...
+   fi
+   kubectl apply -f ${CLUSTERLOADER_ROOT}/gcp-csi-driver-stable.yaml
+   kubectl wait pods -l app=gcp-compute-persistent-disk-csi-driver --for condition=Ready --timeout=300s
+   shift
```
wojtek-t: what does it do?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: USE_CSI sounds fine - (in test we're not deploying anything, just potentially using the functionality)

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
 spec:
   accessModes:
     - ReadWriteOnce
+  {{ if .UseCSI }}
+  storageClassName: csi-gce-pd
+  {{ end }}
```
wojtek-t: Just out of curiosity - what will it do with UseCSI is false?

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/usecsi.yaml
```diff
@@ -0,0 +1 @@
+USE_CSI: true
```
wojtek-t: nit: add end of line

clusterloader2/run-e2e.sh
hantaowang: is it ok to configure the test with env vars? I defaulted to flags because everything else was configured by flags but this seems like a more flexible solution.

clusterloader2/run-e2e.sh
hantaowang: shift shifts the arguments by 1, essentially popping off the first arg. I will remove this for using env vars.

clusterloader2/run-e2e.sh
hantaowang: ack

clusterloader2/run-e2e.sh
hantaowang: ack

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
hantaowang: that storage class line is just not added, so the PVC will be identical to the current non csi one.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/usecsi.yaml
hantaowang: ack

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,17 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
 export KUBEMARK_ROOT_KUBECONFIG="${KUBEMARK_ROOT_KUBECONFIG:-${HOME}/.kube/config}"
 
+# Deploy the GCP PD CSI Driver is required
```
wojtek-t: s/is/if/

clusterloader2/run-e2e.sh
```diff
@@ -22,5 +22,17 @@ CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE}")
...
+if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
+   if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
+      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
+      exit 1
+   else
```
wojtek-t: remove else - there is exit 1 above anyway

clusterloader2/run-e2e.sh
hantaowang: ack

clusterloader2/run-e2e.sh
hantaowang: ack

PR #707, opened 2019-07-30
Fix travis build issues

mm4tt:
Vendor dir for probed was generated via `go mod vendor`
Without it the build wasn't hermetic when using go1.11, as our module has requirement of go1.12


.travis.yml
```diff
@@ -28,5 +28,5 @@ script:
     - ./verify/verify-flags-underscore.py
     - ./verify/verify-dashboard-format.sh
     - go get -d ./...
-    - travis_wait go build $(go list ./... | grep -v "/vendor/")
+    - travis_wait go build $(go list ./... | grep -v "/vendor/" | grep -v "clusterloader2/")
```
wojtek-t: Technically, we don't build unit tests. But maybe that's fine, wdyt?

PR #709, opened 2019-07-30
Run DNS probes in tests

oxddr:

This PR also adds parsed overrides as an argument to Measurements. This is needed, so the probes can be started from inside measurement and still benefit from knobs passed via testoverrides to clusterloader2. I think the approach is not ideal, but I am struggling to come up with something better. I am open to other ideas. 

/assign @wojtek-t @mm4tt

### DIFF
--- a/clusterloader2/pkg/measurement/common/probes/manifests/dnsLookup/dns-prober-deployment.yaml
+++ b/clusterloader2/pkg/measurement/common/probes/manifests/dnsLookup/dns-prober-deployment.yaml
@@ -1,0 +1,34 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: probes
+  name: dns
+  labels:
+    probe: dns
+spec:
+  selector:
+    matchLabels:
+      probe: dns
+  replicas: {{.Replicas}}
+  template:
+    metadata:
+      labels:
+        probe: dns
+    spec:
+      containers:
+        - name: dns
+          image: gcr.io/k8s-testimages/probes:v0.0.3
+          args:
+            - --metric-bind-address=0.0.0.0:8080
+            - --mode=dns
+            # Instead of creating dedicated "null-service" use one that's already exists
+            # TODO(oxddr): according to @wojtek-t there are differences between fully and not fully qualified domain names
+            # Investigate it and potentially measure latency for both
+            - --dns-probe-url=ping-server.probes
+          resources:
+            limits:
+              cpu: 100m
+              memory: 100Mi
+          ports:
+            - containerPort: 8080
+              name: metrics
--- a/clusterloader2/pkg/measurement/common/probes/probes.go
+++ b/clusterloader2/pkg/measurement/common/probes/probes.go
@@ -18,10 +18,9 @@ package probes
 
 import (
 	"fmt"
+	"path"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
@@ -36,18 +35,27 @@ import (
 const (
 	probesNamespace = "probes"
 
+	manifestsPathPrefix = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/"
+
 	checkProbesReadyInterval = 15 * time.Second
 	checkProbesReadyTimeout  = 5 * time.Minute
 )
 
 var (
-	serviceMonitorGvr = schema.GroupVersionResource{Group: "monitoring.coreos.com", Version: "v1", Resource: "servicemonitors"}
-
 	networkLatencyConfig = proberConfig{
-		Name:          "InClusterNetworkLatency",
-		MetricVersion: "v1",
-		Query:         "quantile_over_time(0.99, probes:in_cluster_network_latency:histogram_quantile[%v])",
-		Manifests:     "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/*.yaml",
+		Name:             "InClusterNetworkLatency",
+		MetricVersion:    "v1",
+		Query:            "quantile_over_time(0.99, probes:in_cluster_network_latency:histogram_quantile[%v])",
+		Manifests:        "*.yaml",
+		ProbeLabelValues: []string{"ping-client", "ping-server"},
+	}
+
+	dnsLookupConfig = proberConfig{
+		Name:             "DnsLookupLatency",
+		MetricVersion:    "v1",
+		Query:            "quantile_over_time(0.99, probes:dns_lookup_latency:histogram_quantile[%v])",
+		Manifests:        "dnsLookup/*yaml",
+		ProbeLabelValues: []string{"dns"},
 	}
 )
 
@@ -56,13 +64,18 @@ func init() {
 	if err := measurement.Register(networkLatencyConfig.Name, create); err != nil {
 		klog.Errorf("cannot register %s: %v", networkLatencyConfig.Name, err)
 	}
+	create = func() measurement.Measurement { return createProber(dnsLookupConfig) }
+	if err := measurement.Register(dnsLookupConfig.Name, create); err != nil {
+		klog.Errorf("cannot register %s: %v", dnsLookupConfig.Name, err)
+	}
 }
 
 type proberConfig struct {
-	Name          string
-	MetricVersion string
-	Query         string
-	Manifests     string
+	Name             string
+	MetricVersion    string
+	Query            string
+	Manifests        string
+	ProbeLabelValues []string
 }
 
 func createProber(config proberConfig) measurement.Measurement {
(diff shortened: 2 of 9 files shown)


clusterloader2/pkg/measurement/common/probes/manifests/dns-prober-deployment.yaml
```diff
@@ -0,0 +1,36 @@
+{{$ENABLE_DNS_PROBES := DefaultParam .ENABLE_DNS_PROBES false}}
```
wojtek-t: I'm wondering if it wouldn't be better to move it to test configs (similarly as this is done here:
https://github.com/kubernetes/perf-tests/pull/719/files)

But maybe it's not that big difference, because we will remove it in couple weeks anyway, right?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -124,14 +128,15 @@ func (p *probesMeasurement) String() string {
 	return probesMeasurementName
 }
 
-func (p *probesMeasurement) initialize(config *measurement.MeasurementConfig) error {
-	replicasPerProbe, err := util.GetInt(config.Params, "replicasPerProbe")
+func (p *probesMeasurement) initialize(c *measurement.MeasurementConfig) error {
```
wojtek-t: I actually prefer config than c - it's more meaningful.

clusterloader2/pkg/measurement/interface.go
```diff
@@ -37,6 +37,8 @@ type MeasurementConfig struct {
 	// Identifier identifies this instance of measurement.
 	Identifier    string
 	CloudProvider string
+	// Overrides passed to the clusterloader2 via --testoverrides flag
+	Overrides map[string]interface{}
```
wojtek-t: I'm against this - especially that with the new TestSuite design we won't have global testoverrides.

Why do you need this?

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: Sure, that's leftover from a previous iteration.

clusterloader2/pkg/measurement/interface.go
oxddr: I'm not a fan of this either. All I'd like to do is to pass overrides to the measurement, so it can have conditions based on them. If you have another idea how to do it, especially with the new design, I'd be happy to change it.

clusterloader2/pkg/measurement/common/probes/manifests/dns-prober-deployment.yaml
oxddr: Not easily, unfortunately. probes.go is implemented as monolith and doesn't have any knobs to turn on or off specific queries/probes. I've pointed that earlier, when I wanted to refactror it, so we'd have a single measurement per-probe but you were against it at that time.

oxddr: Saying so, I'd rather have a measurement per-prober. This way we can easily enable and disable existing probes. Most of the existing code could be reused, in a way that adding new measuremetn would be a few line of code, similarly to how [Prometheus-based measurements were implemented](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/prometheus_measurement.go)

clusterloader2/pkg/measurement/interface.go
wojtek-t: Hmm - actually I realized that maybe the only bad thing here is the comment :)
Because measurement is create per-test anyway - so we will just be passing different overrides in different tests.

I'm still not a a huge fan of it, but maybe actually it's not that bad.

So I guess I can live with it if you remove "via --testoverrides" flag", because this will soon disappear :)

oxddr:
/hold 

Need to debug tests failures.

oxddr:
The failed test actually showed an issue with probes.go. While it's possible to conditionally enable manifests, it's not possible (in the current state) to stop probes.go from executing certain queries - it runs them all. I've created a PR, which refactors probes.go so there is a single instance per-query. More details in https://github.com/kubernetes/perf-tests/pull/739.

clusterloader2/pkg/measurement/common/probes/manifests/dnsLookup/dns-prober-deployment.yaml
```diff
@@ -0,0 +1,32 @@
...
+        probe: dns
+    spec:
+      containers:
+        - name: dns
+          image: gcr.io/k8s-testimages/probes:v0.0.3
+          args:
+            - --metric-bind-address=0.0.0.0:8080
+            - --mode=dns
+            # Instead of creating dedicated "null-service" use one that's already exists
+            - --dns-probe-url=ping-server.probes
```
wojtek-t: I'm wondering whether we don't need fully qualified name..

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -56,13 +67,18 @@ func init() {
...
+		klog.Errorf("cannot register %s: %v", dnsLookupConfig.Name, err)
+	}
 }
 
 type proberConfig struct {
 	Name          string
 	MetricVersion string
 	Query         string
 	Manifests     string
+	Pods          []string
```
wojtek-t: This name is very misleading - you basically pass the value of labels of this pod, right?

clusterloader2/pkg/measurement/common/probes/manifests/dnsLookup/dns-prober-deployment.yaml
oxddr: According to https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/ this should be fine.

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: Changed to ProbeNameLabels.

clusterloader2/pkg/measurement/common/probes/manifests/dnsLookup/dns-prober-deployment.yaml
wojtek-t: I agree it works - but I remember in the past there were significant performance differences when using fully qualified vs not fully qualified names.
Maybe add a TODO to evalue the impact of having that non-fully-qualified.

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -56,13 +64,18 @@ func init() {
...
 type proberConfig struct {
-	Name          string
-	MetricVersion string
-	Query         string
-	Manifests     string
+	Name            string
+	MetricVersion   string
+	Query           string
+	Manifests       string
+	ProbeNameLabels []string
```
wojtek-t: ProbeLabelValues ?

[in the end it's value of the label named "probe"]

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -202,28 +215,28 @@ func (p *probesMeasurement) gather(params map[string]interface{}) (measurement.S
...
-			err = nil // Retryable error, don't return it.
-		}
-		return false, err
-	}
 	// TODO(mm4tt): Using prometheus targets to check whether probes are up is a bit hacky.
 	//              Consider rewriting this to something more intuitive.
-	expectedTargets := p.replicasPerProbe * len(serviceMonitors.Items)
+	selector := func(t prometheus.Target) bool {
+		for _, pod := range p.config.ProbeNameLabels {
+			if t.Labels["job"] == pod && t.Labels["namespace"] == probesNamespace {
```
wojtek-t: This won't work - the label you're setting is "probe", right?

[We should ensure that it's the same for different probes]

oxddr: It does, I've checked it specifically. Labels returned here by Prometheus are not the same as ones defined on the actual pod that Prometheus is watching.

clusterloader2/pkg/measurement/common/probes/manifests/dnsLookup/dns-prober-deployment.yaml
oxddr: A good idea may be to actually check both of them.

wojtek-t: Agree - can you please add a TODO?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -202,28 +215,28 @@ func (p *probesMeasurement) gather(params map[string]interface{}) (measurement.S
...
-		if client.IsRetryableAPIError(err) || client.IsRetryableNetError(err) {
-			err = nil // Retryable error, don't return it.
-		}
-		return false, err
-	}
 	// TODO(mm4tt): Using prometheus targets to check whether probes are up is a bit hacky.
 	//              Consider rewriting this to something more intuitive.
-	expectedTargets := p.replicasPerProbe * len(serviceMonitors.Items)
+	selector := func(t prometheus.Target) bool {
+		for _, pod := range p.config.ProbeLabelValues {
```
wojtek-t: s/pod/value/

clusterloader2/pkg/measurement/common/probes/probes.go
wojtek-t: Ahh interesting - thanks for checking.
Can you please add a comment that prometheus does some translation here, so that someone else won't be asking the same question in the future?

Issue #710, opened 2019-07-30
Most of experimental storage tests are broken


wojtek-t:
They are failing with errors like this
```
[measurement call WaitForRunningPods - WaitForRunningPodsWithVolumes error: timeout while waiting for 100 pods to be running in namespace '' with labels 'group = volume-test' and fields '' - only 0 found running]
```

@hantaowang @mucahitkurt - can you please take a look?

@msau42 @davidz627 - FYI

wojtek-t:
The error above is coming from this:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability-max-configmap-vol-per-node/1156177767411224577

But majority of these are failing with similar error:
https://testgrid.k8s.io/sig-scalability-experiments#Summary

mucahitkurt:
I'm looking into this issue

PR #713, opened 2019-07-30
Storage tests: Remove app name from pod and volume object template

mucahitkurt:
remove app name parameter from pod and volume object template

Fix: https://github.com/kubernetes/perf-tests/issues/710

/sig storage

/assign @wojtek-t 
cc @hantaowang @msau42

### DIFF
--- a/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
@@ -65,7 +65,7 @@ steps:
     replicasPerNamespace: {{$volumesPerNamespace}}
     tuningSet: Sequence
     objectBundle:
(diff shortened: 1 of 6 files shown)


hantaowang:
What is the app name actually used for? I removed the ability to accept appname from the pod specs.

hantaowang:
Oh i see- its still used to populate the volume name. I reckon we can probably remove the test name form the volume and pod base name itself. There is not gonna be another vol-* or pod-* in the test namespace.

mucahitkurt:
@hantaowang ok., I'll update the pr to not use testname.

wojtek-t:
@mucahitkurt - please ping me when this will be ready

mucahitkurt:
> @mucahitkurt - please ping me when this will be ready

@wojtek-t it's ready

PR #714, opened 2019-07-30
WIP: Extended load test - stateful sets and PVs

mm4tt:


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -79,6 +79,18 @@ func ListRuntimeObjectsForKind(c clientset.Interface, kind, namespace, labelSele
 			}
 			return nil
 		}
+	case "StatefulSet":
```
wojtek-t: Can you move this part to a separate PR - we want this merged no matter what will happen with this PR.

Issue #704, opened 2019-07-29
 Scalability load test extended to exercise Deployments, DaemonSets, StatefulSets, Jobs, PersistentVolumes, Secrets, ConfigMaps, NetworkPolicies


mm4tt:

mm4tt:
I've run a 5K node test yesterday, using the extended load scenario with enabled: Secrets, ConfigMaps, StatefulSets and PVs. 
The test passed, but the new prometheus-based api-call latency measurement failed for a few tuples:
```
W0731 02:58:04.461] I0731 02:58:04.459614   10763 api_responsiveness_prometheus.go:90] APIResponsiveness: WARNING Top latency metric: {Resource:persistentvolumes Subresource: Verb:POST Scope:cluster Latency:perc50: 266.666666ms, perc90: 1.375s, perc99: 2.443099999s Count:553}; threshold: 1s
W0731 02:58:04.461] I0731 02:58:04.459626   10763 api_responsiveness_prometheus.go:90] APIResponsiveness: WARNING Top latency metric: {Resource:leases Subresource: Verb:GET Scope:namespace Latency:perc50: 32.018464ms, perc90: 143.243359ms, perc99: 1.663881526s Count:17274382}; threshold: 1s
W0731 02:58:04.461] I0731 02:58:04.459643   10763 api_responsiveness_prometheus.go:90] APIResponsiveness: WARNING Top latency metric: {Resource:services Subresource: Verb:DELETE Scope:namespace Latency:perc50: 1.215999999s, perc90: 1.472s, perc99: 1.4972s Count:8251}; threshold: 1s
W0731 02:58:04.461] I0731 02:58:04.459655   10763 api_responsiveness_prometheus.go:90] APIResponsiveness: WARNING Top latency metric: {Resource:namespaces Subresource: Verb:GET Scope:cluster Latency:perc50: 26.875234ms, perc90: 48.375422ms, perc99: 1.14s Count:5228}; threshold: 1s
W0731 02:58:04.462] I0731 02:58:04.459668   10763 api_responsiveness_prometheus.go:90] APIResponsiveness: WARNING Top latency metric: {Resource:configmaps Subresource: Verb:POST Scope:namespace Latency:perc50: 28.571428ms, perc90: 88.1ms, perc99: 1.02472s Count:18627}; threshold: 1s
```
I checked Prometheus graphs for that run, and it looks like the Prometheus based api-call was broken by single spikes (happening around log rotate) in all cases:
[...]

PR #716, opened 2019-07-31
Extended load - StatefulSets and PVs.

mm4tt:

As explained in one TODO, PVs are created by StatefulSets but are not deleted until namespaces are deleted (so after test, after all measurements are stopped). This is something we should change.


clusterloader2/testing/load/experimental/statefulset.yaml
```diff
@@ -0,0 +1,55 @@
...
+  {{if $EnablePVs}}
+  # TODO(#704): We should have better control over deleting PVs.
+  # While this StatefulSet's feature is convenient for creating PVs, there is no easy way to delete
+  # them later, as deleting StatefulSet doesn't delete PVs. They are going to be eventually deleted
+  # once namespace is deleted, but ideally we should have a better (more controlled) way of doing
+  # that, so we can delete them during test, e.g.  #             after StatefulSets are deleted, and
+  # measure api call latency for that. There are two options here:
+  # 1. Create a new measurement that will delete PVs created by StatefulSets
+  # 2. Create PVs manually, attach them here (but not via volumeClaimTemplates) and delete them
+  #    manually later
```
wojtek-t: I actually think that there is a third option which is faster and simpler to do (and I would vote for proceeding this path at least for now).
PVCs connected to StatefulSet have naming schema: "<pvc-name-from-template>-statefulset-name>-<index>".

So we can actually create a template file for PVC and we will simply add a step to delete PVCs that would look like:
```
    namespaceRange:
      min: 1
      max: {{$namespaces}}
    replicasPerNamespace: 0
    tuningSet: RandomizedSaturationTimeLimited
    objectBundle:
      - basename: pvname-ssname
        objectTemplatePath: pvc.yaml
```

If at some point we decide for more pvc-per-pod or more statefulset-per-namespace, we can opaque it in `range` (ugly but works - I don't we need it immediately).


The only remaining problem we have is that we enumrate from 1 to N, and StatefulSet from 0 to N-1.
But I think we can change that universally for CL, to actually use 0..N-1 scheme (I think that's reasonable on itself).

clusterloader2/testing/load/experimental/extended_config.yaml
```diff
@@ -126,8 +134,18 @@ steps:
...
       labelSelector: group = load
       operationTimeout: 15m
+  {{if $ENABLE_STATEFULSETS}}
+  - Identifier: WaitForRunningStatefulSets
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      apiVersion: apps/v1
+      kind: StatefulSet
+      labelSelector: group = statefulset
```
wojtek-t: This one is pretty much only selecting `controlling objects` (so statefulsets).

So for consistency with deployments (and probably other in the near future), I vote for making that:
```
group = load
```

This would mean that they will also be measured as part of pod-startup-time and all stuff that was assuming that all pods actually have group = load label.

WDYT?

clusterloader2/testing/load/experimental/statefulset.yaml
mm4tt: Done, but from what I managed to get from code it looks like we're enumerating from 0 in CL2 already. Please correct me if I'm wrong

wojtek-t: I think it used to be 1..N, but apparently it's not:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/test/simple_test_executor.go#L199

clusterloader2/testing/load/config.yaml
```diff
@@ -8,23 +8,36 @@
 {{$NODE_MODE := DefaultParam .NODE_MODE "allnodes"}}
 {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
-{{$LOAD_TEST_THROUGHPUT := DefaultParam .LOAD_TEST_THROUGHPUT 10}}
+{{$LOAD_TEST_THROUGHPUT := DefaultParam .LOAD_TEST_THROUGHPUT 100}}
```
wojtek-t: Revert

clusterloader2/testing/load/config.yaml
```diff
@@ -8,23 +8,36 @@
...
 {{$BIG_GROUP_SIZE := 250}}
 {{$MEDIUM_GROUP_SIZE := 30}}
 {{$SMALL_GROUP_SIZE := 5}}
+{{$STATEFUL_SET_SIZE := 10}}
```
wojtek-t: Apparently one of my comments disappeared in the meantime.
What I was suggesting that with the ability for parallel-pod-management, we may want to create 1 of each: small, medium and large StatefulSet per namespace instead of a corresponding deployment.

This would also simplify the logic below (i.e. you would leave everything as was before and do:
```
if ENABLE_STATEFUL_SETS {
if small-deployments > 0 {
  small-deployments--
  small-statefulsets++
}
if medium-deployments > 0 {
 ...
...
```

clusterloader2/testing/load/config.yaml
```diff
@@ -8,23 +8,36 @@
...
+{{$bigDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 4 $BIG_GROUP_SIZE)}}
 # mediumDeployments - 1/4 of namespace pods should be in medium Deployments.
-{{$mediumDeploymentsPerNamespace := DivideInt (MultiplyInt $NODES_PER_NAMESPACE $PODS_PER_NODE) (MultiplyInt 4 $MEDIUM_GROUP_SIZE)}}
+{{$mediumDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 4 $MEDIUM_GROUP_SIZE)}}
 # smallDeployments - 1/2 of namespace pods should be in small Deployments.
-{{$smallDeploymentsPerNamespace := DivideInt (MultiplyInt $NODES_PER_NAMESPACE $PODS_PER_NODE) (MultiplyInt 2 $SMALL_GROUP_SIZE)}}
+# Number of small deployments is reduced to make space for stateful sets and stick to the $PODS_PER_NODE density.
+{{$smallDeploymentsPerNamespace := DivideInt (SubtractInt (DivideInt $podsPerNamespace 2) $statefulSetPodsPerNamespace) $SMALL_GROUP_SIZE}}
 
+# TODO: Remove
```
wojtek-t: remove

clusterloader2/testing/load/config.yaml
wojtek-t: I'm also fine with (as a first step) getting rid of stateful-sets-per-namespace and explicitly assuming 1 of each.

clusterloader2/testing/load/experimental/statefulset.yaml
mm4tt: As discussed f2f, unfortunately this wouldn't work, as CL2 is not able to delete objects that it didn't create. Long story short it keeps it own state of all objects that were created/modified. When we try to delete the PVCs, CL2 doesn't know about them and assumes that none of them exist and doesn't do anything when deleting. 

Leaving the TODO as it was, will tackle it in a separate PR.

clusterloader2/testing/load/statefulset.yaml
```diff
@@ -0,0 +1,56 @@
...
+        volumeMounts:
+          - name: pv
+            mountPath: /var/pv
+        {{end}}
+        resources:
+          requests:
+            cpu: 10m
+            memory: "10M"
+  {{if $EnablePVs}}
+  # TODO(#704): We should have better control over deleting PVs.
```
wojtek-t: As I mentioned offline - let's maybe first do only StatefulSets and we will extend with PVs in a separate step.

mm4tt:
Closing this one, will add StatefulSets separately in https://github.com/kubernetes/perf-tests/pull/776 and then tackle PVs in a separate PR.

PR #717, opened 2019-07-31
Perfdash - Storage test config fix

krzysied:
Adding config for new test names and new measurement identifiers.


wojtek-t:
/lgtm

Thanks!

PR #719, opened 2019-07-31
Extended Load: NetworkPolicies

mm4tt:
This is most likely no-op until we turn on some Network Policy Provider that will start enfocring these network policies.

It should be pretty straightforward to turn on Calico both in GKE and in GCE.
This should be done separately to isolate any potential performance impact of tuning it just on.


clusterloader2/testing/load/experimental/overrides/enable_networkpolicies.yaml
```diff
@@ -0,0 +1 @@
+ENABLE_NETWORKPOLICIES: true
```
oxddr: You should add this knob in a separate PR, then enable it in perf-tests presubmit and the follow with the rest of this PR. This way, the presubmit will test the codepath/scenarios that's being added as part of this PR preventing breakages.

This has been documented in [here](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/experiments.md)

mm4tt: This is an experimental suite, it's not part of any CI/CD tests. Therefore, I don't think that the rollout precedure applies here :)

oxddr: Make sense.

mm4tt:
Closing this one. @jkaniuk took it over.

PR #721, opened 2019-07-31
ClusterLoader - Adding TestSuite api

krzysied:
ref #641

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -20,6 +20,20 @@ import (
 	"time"
 )
 
+// TestSuite defines list of test scenarios to be run.
+type TestSuite []TestScenario
+
+// TestScenario defines customized test to be run.
+type TestScenario struct {
+	// Identifier is a unique test scenario name across test suite.
+	Identifier string `json: identifier`
+	// ConfigPath defines path to the file containing a single Config definition.
+	ConfigPath string `json: configPath`
+	// OverridePaths defines what override files should be applied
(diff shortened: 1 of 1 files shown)


clusterloader2/api/types.go
```diff
@@ -20,6 +20,21 @@ import (
...
 
+// TestSuite defines list of test scenarios to be run.
+type TestSuite []TestScenario
+
+// TestScenario defines customized test to be run.
+type TestScenario struct {
+	// Identifier is a unique test scenario name across test suite.
+	Identifier string `json: identifier`
+	// ConfigPath defines path to the config file that satisfy already
+	// existing cluster loader API.
```
wojtek-t: What is "existing cluster loader API"?

It should mention something like that this file contains a single Config.

krzysied: Done.

clusterloader2/api/types.go
```diff
@@ -20,6 +20,20 @@ import (
...
 )
 
+// TestSuite defines list of test scenarios to be run.
+type TestSuite []TestScenario
+
+// TestScenario defines customized test to be run.
+type TestScenario struct {
+	// Identifier is a unique test scenario name across test suite.
+	Identifier string `json: identifier`
+	// ConfigPath defines path to the single test config file.
```
wojtek-t: .. to the file containing a single Config definition.

maybe? or sth like that?

krzysied: Done.

PR #723, opened 2019-08-01
add function to wait for available PVs

hantaowang:
Per #722, this adds a function that waits until all PVs are deleted. 

@msau42 @davidz627 @jingxu97 @verult

/assign @wojtek-t

Issue #722, opened 2019-08-01
PV Experimental storage tests fail to delete all resources


hantaowang:
The resource diff tests is failing for all storage tests using PVs. This is likely because while PVCs are deleted, there is nothing to make sure that all PVs and thus the underlying volume is deleted from GCE. We can use WaitForBoundPVCs to wait until PVCs are unbound, but that doesnt guarantee that the PV controller will delete the volumes before the test finishes.

What are your thoughts on how we can approach this? I think there must be new functionality added, but this is going to increase the complexity of both the storage test config and the CL2 framework.

@wojtek-t @davidz627 @jingxu97 @verult @msau42

msau42:
While we figure this out, can we quickly disable the tests?

What we do in our regular e2es is wait until all PVs are gone

hantaowang:
> While we figure this out, can we quickly disable the tests?

Of course: https://github.com/kubernetes/test-infra/pull/13721

> What we do in our regular e2es is wait until all PVs are gone

We can do that too. Its probably the simplest option that just amounts to adding another function WaitForProvisionedPV to CL2. It seems weird because its oddly similar to WaitForBoundPVC although it is different.

wojtek-t:
> We can do that too. Its probably the simplest option that just amounts to adding another function WaitForProvisionedPV to CL2. It seems weird because its oddly similar to WaitForBoundPVC although it is different.

sounds fine to me (yes, it will be similar, but as I mentioned on some other PR, we probably will do some refactoring at some point to share more code)

PR #724, opened 2019-08-01
wait for PV to be deleted in storage tests

hantaowang:
Per #722, this turns on waiting for PVs to be deleted for storage tests. I tested this locally with #723 merged.

Because PVs are created on demand by PVCs, I have not found a way to add a label to the created PVs. Thus this config currently waits for all global PVs to be deleted.

@msau42 @davidz627 @jingxu97 @verult

/assign @wojtek-t

### DIFF
--- a/clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/override.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/override.yaml
@@ -1,4 +1,4 @@
 PROVISION_VOLUME: true
-WAIT_FOR_PVC: true
+WAIT_FOR_PVS: true
 DEPLOYMENT_TEMPLATE_PATH: "volume-types/persistentvolume/deployment_with_pvc.yaml"
 VOLUME_TEMPLATE_PATH: "volume-types/persistentvolume/pvc.yaml"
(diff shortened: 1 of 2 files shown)


wojtek-t:
/ok-to-test

This is blocked on PR adding that measurement.

davidz627:
/lgtm
what are "measurement" are we blocked on?

hantaowang:
> what are "measurement" are we blocked on?

It was #723, which is now merged.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/override.yaml
```diff
@@ -1,4 +1,4 @@
 PROVISION_VOLUME: true
-WAIT_FOR_PVC: true
+WAIT_FOR_PVS: true
```
jingxu97: just to want to make sure we don't need to wait for PVC after we change to wait for PV?

hantaowang: You still do the wait. I just named the name of this variable to be more generic. You can see its now used in two places: waiting for PVC to be bound and waiting for PV to be deleted. But because PVC being bound is essentially waiting for PV to be bound, it is in a sense still waiting for PV to do something, so I feel this name is better.

Issue #615, opened 2019-07-08
Measure and expose Network Programming Latency SLI


oxddr:
For not the measurement is non-violating. As part of this issue we should figure out an initial value, at which we should start to fail tests. This should be an input to finalize SLO definition for Network Programming Latency. 

/assign @oxddr 
/area slo

PR #736, opened 2019-08-02
Disable "NetworkProgrammingLatency" measurement in kubemark

oxddr:
/assign @wojtek-t

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/network_programming.go
+++ b/clusterloader2/pkg/measurement/common/slos/network_programming.go
@@ -47,6 +47,10 @@ func init() {
 
 type netProgGatherer struct{}
 
+func (n *netProgGatherer) IsEnabled(config *measurement.MeasurementConfig) bool {
+	return config.CloudProvider != "kubemark"
+}
+
 func (n *netProgGatherer) Gather(executor QueryExecutor, startTime time.Time) (measurement.Summary, error) {
 	latency, err := n.query(executor, startTime)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/slos/prometheus_measurement.go
```diff
@@ -60,6 +61,10 @@ func (m *prometheusMeasurement) Execute(config *measurement.MeasurementConfig) (
 		return nil, nil
 	}
 
+	if !m.gatherer.IsEnabled(config) {
+		klog.Warningf("%s is disabled, skipping the measuerment", m)
```
wojtek-t: should be some return here :)

wojtek-t: Also, for consistency maybe have the same log as here:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/probes/probes.go#L85

oxddr: Added. 

Left the message as it was - this is general piece of code, not tied a specific measurement. I'd rather not add Kubemark specific message in here.

oxddr:
PTAL

clusterloader2/pkg/measurement/common/slos/network_programming.go
```diff
@@ -47,6 +47,14 @@ func init() {
 
 type netProgGatherer struct{}
 
+func (n *netProgGatherer) IsEnabled(config *measurement.MeasurementConfig) bool {
+	if config.CloudProvider == "kubemark" {
+		klog.Warningf("%s is disabled in Kubemark", n)
```
wojtek-t: I would remove this warning - logging twice that something is disabled doen't make sense to me.

clusterloader2/pkg/measurement/common/slos/network_programming.go
```diff
@@ -47,6 +47,14 @@ func init() {
...
+		return false
+	}
+	return true
```
wojtek-t: in fact, this should just be

return config.CloudProvider != "kubemark"

oxddr:
ref https://github.com/kubernetes/perf-tests/issues/615

PR #737, opened 2019-08-02
ClusterLoader - Adding test scenario to config

krzysied:
ref #641

### DIFF
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -16,13 +16,16 @@ limitations under the License.
 
 package config
 
-// ClusterLoaderConfig represents all flags used by CLusterLoader
+import (
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+// ClusterLoaderConfig represents all single test run parameters used by CLusterLoader.
 type ClusterLoaderConfig struct {
 	ClusterConfig     ClusterConfig
 	ReportDir         string
 	EnableExecService bool
-	TestConfigPath    string
-	TestOverridesPath []string
+	TestScenario      api.TestScenario
 	PrometheusConfig  PrometheusConfig
 }
 
--- a/clusterloader2/pkg/test/simple_test_executor.go
+++ b/clusterloader2/pkg/test/simple_test_executor.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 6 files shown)


krzysied:
/hold
I want to check if summary names remain unchanged.

clusterloader2/pkg/config/cluster.go
```diff
@@ -16,15 +16,18 @@ limitations under the License.
...
-	EnableExecService        bool          `json: enableExecService`
-	TearDownPrometheusServer bool          `json: tearDownPrometheusServer`
-	TestConfigPath           string        `json: testConfigPath`
-	TestOverridesPath        []string      `json: testOverrides`
+	ClusterConfig            ClusterConfig    `json: clusterConfig`
+	ReportDir                string           `json: reportDir`
+	EnablePrometheusServer   bool             `json: enablePrometheusServer`
+	EnableExecService        bool             `json: enableExecService`
+	TearDownPrometheusServer bool             `json: tearDownPrometheusServer`
+	TestScenario             api.TestScenario `json: testScenario`
```
wojtek-t: Can we remove TestOverridesPath now?

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -86,8 +86,12 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
 		if ctx.GetClusterLoaderConfig().ReportDir == "" {
 			klog.Infof("%v: %v", summary.SummaryName(), summary.SummaryContent())
 		} else {
+			id := ""
+			if ctx.GetClusterLoaderConfig().TestScenario.Identifier != "" {
+				id = "_" + ctx.GetClusterLoaderConfig().TestScenario.Identifier
```
wojtek-t: Why the id starts with "_" ?

clusterloader2/pkg/config/cluster.go
krzysied: Yes, it's being removed (or rather moved to test scenario structure) by this PR.

clusterloader2/pkg/test/simple_test_executor.go
krzysied: I'm using it later in file name. I want to add additional separator "_" if id is not empty.
When identifier provided: SummaryName_configName_testId_timstamp.ext.
Without id: SummaryName_configName_timstamp.ext

wojtek-t: well, id is not the most intuitive name then :)

infix? distinctor? (neither of those is good, but less confusing)

krzysied: Changed to "testDistinctor".

clusterloader2/pkg/config/cluster.go
wojtek-t: Wait - TestScenario is not a flag to ClusterLoader (it's the path to config file with test scenario)

krzysied: "testSuiteConfigPath" - flag - Path to the test suite config file.
"TestScenario" - structure - defines test config, overrides and identifier.

clusterloader2/pkg/config/cluster.go
```diff
@@ -16,14 +16,19 @@ limitations under the License.
...
 type ClusterLoaderConfig struct {
-	ClusterConfig     ClusterConfig
-	ReportDir         string
-	EnableExecService bool
-	TestConfigPath    string
-	TestOverridesPath []string
-	PrometheusConfig  PrometheusConfig
+	ClusterConfig            ClusterConfig
+	ReportDir                string
+	EnablePrometheusServer   bool
```
wojtek-t: seems unused

clusterloader2/pkg/config/cluster.go
```diff
@@ -16,14 +16,19 @@ limitations under the License.
...
+	EnableExecService        bool
+	TearDownPrometheusServer bool
```
wojtek-t: this one too

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -86,8 +86,19 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
...
 			klog.Infof("%v: %v", summary.SummaryName(), summary.SummaryContent())
 		} else {
+			testDinstinctor := ""
+			if ctx.GetClusterLoaderConfig().TestScenario.Identifier != "" {
+				testDinstinctor = "_" + ctx.GetClusterLoaderConfig().TestScenario.Identifier
+			}
 			// TODO(krzysied): Remember to keep original filename style for backward compatibility.
-			filePath := path.Join(ctx.GetClusterLoaderConfig().ReportDir, summary.SummaryName()+"_"+conf.Name+"_"+summary.SummaryTime().Format(time.RFC3339)+"."+summary.SummaryExt())
+			filePath := path.Join(ctx.GetClusterLoaderConfig().ReportDir, summary.SummaryName()+
+				"_"+
```
wojtek-t:
```
fileName := strings.Join([]string{summary.SummaryName(), conf.Name+testDistinctor, summary.SummaryTime.Format(...), "_")
filePath := path.Join(reportDir, strings.Join([]string{fileName, extension}, ".")
```

will be a bit more readable what is happening here

krzysied: Done.

PR #739, opened 2019-08-05
Refactor probes.go so that there is a single prober enabled per instance

oxddr:
This allows to enable probes using overrides, so they can be rolled out gradually or rolled back if needed. The previous version was optimized for minimizing code changes needed to enable a new prober. However in light of https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/experiments.md this make gradual rollout impossible without changes to probes.go.

### DIFF
--- a/clusterloader2/pkg/measurement/common/probes/probes.go
+++ b/clusterloader2/pkg/measurement/common/probes/probes.go
@@ -34,42 +34,45 @@ import (
 )
 
 const (
-	probesMeasurementName = "Probes"
-	probesNamespace       = "probes"
-
-	manifestGlob = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/*.yaml"
+	probesNamespace = "probes"
 
 	checkProbesReadyInterval = 15 * time.Second
 	checkProbesReadyTimeout  = 5 * time.Minute
-
-	currentProbesMetricsVersion = "v1"
 )
 
 var (
 	serviceMonitorGvr = schema.GroupVersionResource{Group: "monitoring.coreos.com", Version: "v1", Resource: "servicemonitors"}
+
+	networkLatencyConfig = proberConfig{
+		Name:          "InClusterNetworkLatency",
+		MetricVersion: "v1",
+		Query:         "quantile_over_time(0.99, probes:in_cluster_network_latency:histogram_quantile[%v])",
+		Manifests:     "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/*.yaml",
+	}
 )
 
 func init() {
-	if err := measurement.Register(probesMeasurementName, createProbesMeasurement); err != nil {
-		klog.Errorf("cannot register %s: %v", probesMeasurementName, err)
+	create := func() measurement.Measurement { return createProber(networkLatencyConfig) }
+	if err := measurement.Register(networkLatencyConfig.Name, create); err != nil {
+		klog.Errorf("cannot register %s: %v", networkLatencyConfig.Name, err)
 	}
 }
 
-func createProbesMeasurement() measurement.Measurement {
+type proberConfig struct {
+	Name          string
+	MetricVersion string
+	Query         string
+	Manifests     string
+}
+
+func createProber(config proberConfig) measurement.Measurement {
 	return &probesMeasurement{
-		probeNameToPrometheusQueryTmpl: map[string]string{
-			"in_cluster_network_latency": "quantile_over_time(0.99, probes:in_cluster_network_latency:histogram_quantile[%v])",
-		},
+		config: config,
 	}
 }
 
 type probesMeasurement struct {
-	// probeNameToPrometheusQueryTmpl defines a config of this measurement. Updating the config in the
-	// createProbesMeasurement method is the only place in go code that needs to be changed while
-	// adding a new probe.
-	// Each query template should accept a single %v placeholder corresponding to the query window
-	// length. See the 'in_cluster_network_latency' as an example.
-	probeNameToPrometheusQueryTmpl map[string]string
+	config proberConfig
 
 	framework        *framework.Framework
 	replicasPerProbe int
@@ -93,11 +96,11 @@ func (p *probesMeasurement) Execute(config *measurement.MeasurementConfig) ([]me
 	case "start":
 		return nil, p.start(config)
 	case "gather":
-		summaries, err := p.gather(config.Params)
+		summary, err := p.gather(config.Params)
 		if err != nil && !errors.IsMetricViolationError(err) {
 			return nil, err
 		}
-		return summaries, err
+		return []measurement.Summary{summary}, err
 	default:
 		return nil, fmt.Errorf("unknown action %v", action)
 	}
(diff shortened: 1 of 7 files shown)


clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -157,54 +160,50 @@ func (p *probesMeasurement) start(config *measurement.MeasurementConfig) error {
...
 }
 
-func (p *probesMeasurement) gather(params map[string]interface{}) ([]measurement.Summary, error) {
+func (p *probesMeasurement) gather(params map[string]interface{}) (measurement.Summary, error) {
 	klog.Info("Gathering metrics from probes...")
 	if p.startTime.IsZero() {
 		return nil, fmt.Errorf("measurement %s has not been started", p)
 	}
-	thresholds, err := parseThresholds(params)
+	threshold, err := util.GetDurationOrDefault(params, "threshold", 0)
```
wojtek-t: Unless I'm missing something, this is not going to work with multiple probes that would like to have different thresholds (which is definitely a valid usecase - different probes will have different thresholds).

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -157,54 +160,50 @@ func (p *probesMeasurement) start(config *measurement.MeasurementConfig) error {
...
+	latency, err := measurementutil.NewLatencyMetricPrometheus(samples)
+	if err != nil {
+		return nil, err
+	}
+
+	var violation error
+	prefix, suffix := "", ""
+	if threshold > 0 {
+		suffix = fmt.Sprintf(", expected perc99 <= %v", threshold)
+		if latency.Perc99 > threshold {
```
wojtek-t: I actually liked the previous VerifyThreshold - why can't we create and use it here too?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -228,46 +227,23 @@ func (p *probesMeasurement) checkProbesReady() (bool, error) {
...
-	for name, thresholdVal := range t.(map[string]interface{}) {
-		threshold, err := time.ParseDuration(thresholdVal.(string))
-		if err != nil {
-			return nil, err
-		}
-		thresholds[name] = makeLatencyThreshold(threshold)
+func (p *probesMeasurement) createSummary(latency measurementutil.LatencyMetric) (measurement.Summary, error) {
+	content, err := util.PrettyPrintJSON(&measurementutil.PerfData{
+		Version: currentProbesMetricsVersion,
+		// azdas
```
wojtek-t: ??

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -228,46 +227,23 @@ func (p *probesMeasurement) checkProbesReady() (bool, error) {
...
+		DataItems: []measurementutil.DataItem{latency.ToPerfData("Probes")},
```
wojtek-t: I guess this needs to be specific probes name now?

oxddr: Unfortunately to keep existing summaries compatible with this change we need to use hardcoded "Probes" here.

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: Unfinished comment. Will fix.

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: I didn't like it - it was implicit and unnecessarily complicated for what can be a simple check of two floats.

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: This PR introduces an assumption that there is a single instance of this measurement created for each probe. Based on that we can have a single threshold parameter.

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: Added comment in the code.

clusterloader2/pkg/measurement/common/probes/probes.go
wojtek-t: Hmm - I actually think that we may take advantage of it, and do comparisons on smaller percentiles too.

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -228,46 +229,23 @@ func (p *probesMeasurement) checkProbesReady() (bool, error) {
...
-	for name, thresholdVal := range t.(map[string]interface{}) {
-		threshold, err := time.ParseDuration(thresholdVal.(string))
-		if err != nil {
-			return nil, err
-		}
-		thresholds[name] = makeLatencyThreshold(threshold)
+func (p *probesMeasurement) createSummary(latency measurementutil.LatencyMetric) (measurement.Summary, error) {
+	content, err := util.PrettyPrintJSON(&measurementutil.PerfData{
+		Version: p.config.MetricVersion,
+		// NOTE(oxddr): for backward compatibility we need to pass "Probes" to be used a metric label.
```
wojtek-t: Backward compatibility of data we store in gcs?
Is it so important?

I actually think that it's fine to break it (given we're not relying on it anywhere) to make it readable for other probes added in the future.

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -228,46 +229,23 @@ func (p *probesMeasurement) checkProbesReady() (bool, error) {
...
+		DataItems: []measurementutil.DataItem{latency.ToPerfData("Probes")},
+	})
+	if err != nil {
+		return nil, err
 	}
-	return thresholds, nil
+	return measurement.CreateSummary(p.config.MetricName, "json", content), nil
```
wojtek-t: Why does it have to be MetricName? It would be much more intuitive to make it prober.Name

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: Yes, more precisely with perf-dash configuration. It just makes my life easier, since I'm going to look at this data. But maybe it's not worth to sacrifice long-term benefit for a short-term comfort.

clusterloader2/pkg/measurement/common/probes/probes.go
oxddr: This is the same issue as above. Going to drop that.

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -157,54 +160,50 @@ func (p *probesMeasurement) start(config *measurement.MeasurementConfig) error {
...
+	latency, err := measurementutil.NewLatencyMetricPrometheus(samples)
+	if err != nil {
+		return nil, err
+	}
+
+	var violation error
+	prefix, suffix := "", ""
+	if threshold > 0 {
+		suffix = fmt.Sprintf(", expected perc99 <= %v", threshold)
+		err := latency.VerifyThreshold(threshold)
```
wojtek-t: if err:= ... ; er != nil {

PR #740, opened 2019-08-05
Added a simple scale test for L4ILB.

prameshj:
The test creates 3 ILBs - 1 with 300 backends, 1 with 150 and the third one with 10 backends.
The time taken for an ip address to be assigned to the ILB and the time taken for the IP to serve traffic is also measured.

### DIFF
--- a/clusterloader2/testing/l4ilb/config.yaml
+++ b/clusterloader2/testing/l4ilb/config.yaml
@@ -1,0 +1,156 @@
+#Constants
+{{$LARGE_BACKEND_SIZE := 300}}
+{{$MEDIUM_BACKEND_SIZE := 150}}
+{{$SMALL_BACKEND_SIZE := 10}}
+{{$LARGE_BACKEND_LABEL := "ilb-large"}}
+{{$MEDIUM_BACKEND_LABEL := "ilb-medium"}}
+{{$SMALL_BACKEND_LABEL := "ilb-small"}}
+# adding a fixed value for first version of the test, rate of pod creation not a concern yet.
+{{$ilbQPS := 500}}
+{{$namespaces := 1}}
+
+#Test
+name: ilbload
+automanagedNamespaces: {{$namespaces}}
+tuningSets:
+- name: ILBConstantQPS
+  qpsLoad:
+    averageQps: {{$ilbQPS}}
+steps:
+# Mesure each of the ILB services separately, this will provide insight on how long programming
+# ILB takes as a function of number of backends.
+- measurements:
+  - Identifier: ServiceCreationLatencyLarge
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      labelSelector: size = {{$LARGE_BACKEND_LABEL}}
+  - Identifier: ServiceCreationLatencyMedium
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      labelSelector: size = {{$MEDIUM_BACKEND_LABEL}}
+  - Identifier: ServiceCreationLatencySmall
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      labelSelector: size = {{$SMALL_BACKEND_LABEL}}
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      apiVersion: apps/v1
+      kind: Deployment
+      labelSelector: group = ilb-load
+      operationTimeout: 15m
+# Create ILBs
+- name: Creating ILBs
+  phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 1
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: large-backends-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: large-backends-dep
+        ILBSizeLabel: {{$LARGE_BACKEND_LABEL}}
+    - basename: large-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        NumReplicas: {{$LARGE_BACKEND_SIZE}}
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 1
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: medium-backends-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: medium-backends-dep
+        ILBSizeLabel: {{$MEDIUM_BACKEND_LABEL}}
+    - basename: medium-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        NumReplicas: {{$MEDIUM_BACKEND_SIZE}}
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 1
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: small-backends-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: small-backends-dep
+        ILBSizeLabel: {{$SMALL_BACKEND_LABEL}}
+    - basename: small-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        NumReplicas: {{$SMALL_BACKEND_SIZE}}
+- measurements:
+  - Identifier: ServiceCreationLatencyLarge
+    Method: ServiceCreationLatency
+    Params:
+      action: waitForReady
+  - Identifier: ServiceCreationLatencyMedium
+    Method: ServiceCreationLatency
+    Params:
+      action: waitForReady
+  - Identifier: ServiceCreationLatencySmall
+    Method: ServiceCreationLatency
+    Params:
+      action: waitForReady
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
+# Delete ILBs
+- name: Deleting ILBs
+  phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: large-backends-service
+      objectTemplatePath: service.yaml
+    - basename: large-backends-dep
+      objectTemplatePath: dep.yaml
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: medium-backends-service
+      objectTemplatePath: service.yaml
+    - basename: medium-backends-dep
+      objectTemplatePath: dep.yaml
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: small-backends-service
+      objectTemplatePath: service.yaml
+    - basename: small-backends-dep
+      objectTemplatePath: dep.yaml
+- measurements:
+  - Identifier: ServiceCreationLatencyLarge
+    Method: ServiceCreationLatency
+    Params:
+      action: gather
+  - Identifier: ServiceCreationLatencyMedium
+    Method: ServiceCreationLatency
+    Params:
+      action: gather
+  - Identifier: ServiceCreationLatencySmall
+    Method: ServiceCreationLatency
+    Params:
+      action: gather
--- a/clusterloader2/testing/l4ilb/dep.yaml
+++ b/clusterloader2/testing/l4ilb/dep.yaml
@@ -1,0 +1,22 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  labels:
+    group: ilb-load
+spec:
+  replicas: {{.NumReplicas}}
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        group: ilb-load
+        name: {{.Name}}
+    spec:
+      containers:
+      - name: {{.Name}}
+        image: nginx
+        ports:
+        - containerPort: 8080
--- a/clusterloader2/testing/l4ilb/service.yaml
+++ b/clusterloader2/testing/l4ilb/service.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.Name}}
+  labels:
+    size: {{.ILBSizeLabel}}
+  annotations:
+    cloud.google.com/load-balancer-type: "Internal"
+spec:
+  type: LoadBalancer
+  selector:
+    name: {{.DeploymentBaseName}}-{{.Index}}
+  ports:
+  - port: 8080
+    targetPort: 80


clusterloader2/testing/l4ilb/dep.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+spec:
+  replicas: {{RandIntRange .ReplicasMin .ReplicasMax}}
```
krzysied: Do we want to keep RandIntRange (random number from range [ReplicasMin; ReplizacMax])? If not, we can use a single variable.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+    - basename: large-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        ReplicasMin: {{$LARGE_BACKEND_SIZE}}
+        ReplicasMax: {{$LARGE_BACKEND_SIZE}}
+    - basename: large-backends-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: large-backends-dep
+- phases:
```
krzysied: Do we need a separate step here or can we merge it with the one above?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 1
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: large-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        ReplicasMin: {{$LARGE_BACKEND_SIZE}}
```
krzysied: We are passing always two identical values. Maybe we can pass one value instead?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+    - basename: small-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        ReplicasMin: {{$SMALL_BACKEND_SIZE}}
+        ReplicasMax: {{$SMALL_BACKEND_SIZE}}
+    - basename: small-backends-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: small-backends-dep
+- name: Creating ILBs
```
krzysied: nit: There were some changes in cluster loader regarding `name` field. Currently we can use:
```
- name: Creating ILBs
  phases:
  - ...
  - ...
```
It will log start and end of the step using provided name.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+  qpsLoad:
+    averageQps: {{$ilbQPS}}
+steps:
+# Create ILBs
+- measurements:
+  - Identifier: ServiceCreationLatency
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+  - Identifier: WaitForRunningDeployments
```
krzysied: The `WaitForControlledPodsRunning` is started but never used again. Do we need this?
If so, here is an example of gather method usage: https://github.com/kubernetes/perf-tests/blob/61f39edf3d70731cbf9225dc165ae103ec13e567/clusterloader2/testing/density/config.yaml#L160-L163

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 1
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: medium-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        ReplicasMin: {{$MEDIUM_BACKEND_SIZE}}
+        ReplicasMax: {{$MEDIUM_BACKEND_SIZE}}
```
wojtek-t: I'm wondering what is the point of testing large/medium/small deployments for ILB.
The ILB is build on top of nodes, and based on size of deployments we're only programming iptables, right?

I'm wondering why having just large deployment wouldn't be enough?
It would visibly simplify the test.

prameshj: This is currently the case, we add all nodes to the cluster and program iptables to send the packet to the right deployment pods. However, with ILB subsetting feature, we will create a different backend for each ILB proportional to the size of the deployment. Then this test will be useful to ensure that we pick only nodes where we have a matching endpoint. 

not sure if that can be added as a predefined step:
we would need to list all the matching pods - the nodes they are on and then compare that list with the backend of a given ILB.

wojtek-t: Re sizes: fair enough

Re predefined step - it's is super provider specific to list backends of ILB. We can probably add such measurement, if we make it very specific that it's for gcp only.
Though I wouldn't block this PR on it.

clusterloader2/testing/l4ilb/config.yaml
prameshj: I wanted to keep each deployment + service creation as separate steps so that the latencies are measured individually. don't we need that ?

clusterloader2/testing/l4ilb/config.yaml
prameshj: yes, i agree with you on both points.

clusterloader2/testing/l4ilb/config.yaml
krzysied: - Currently large deployment is on step while medium and small are in the other.
- What do you mean by the latencies being measured individually? Do you want to create deployment+service, wait until it is available and then proceed with the next one? Or do you want to create all deployments+services at once and log latency for each of them separately?

clusterloader2/testing/l4ilb/config.yaml
prameshj: Done

clusterloader2/testing/l4ilb/config.yaml
prameshj: the latter.. create all deployments+services in parallel and log latency for each separately. I put all deployment creation in the same step as phases, looks like the measurement is happening correctly.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      apiVersion: apps/v1
+      kind: Deployment
+      labelSelector: name = ilbload
```
krzysied: Pods from depl.yaml don't have a matching label.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+      operationTimeout: 15m
+- name: Creating ILBs
+- phases:
```
krzysied: nit: `- phases:` -> `  phases`.
Name and phase can be in one step.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+    Method: ServiceCreationLatency
+    Params:
+      action: waitForReady
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
+# Delete ILBs
+- name: Deleting ILBs
+- phases:
```
krzysied: nit: remove  `-`.

clusterloader2/testing/l4ilb/config.yaml
krzysied: The ServiceCreationLatency measurement observes services individually. However the summary it's creating is an aggregation of all observed service latencies. Maybe we want to have separate measurement for each type (size) of deployment?

clusterloader2/testing/l4ilb/config.yaml
prameshj: I do have this in dep.yaml:
  selector:
    matchLabels:
      name: {{.Name}}

and test name is ilbload. Can't the selector be used as "name = ilbload"?

clusterloader2/testing/l4ilb/config.yaml
prameshj: I do see logs like this:
ServiceCreationLatency: 3 worst create_to_available latencies: [{test-fcrs0g-1/small-backends-service-0 1m6.427795413s} {test-fcrs0g-1/large-backends-service-0 1m46.162875508s} {test-fcrs0g-1/medium-backends-service-0 2m27.217369777s}]

assumed this means each service creation is measured individually. Is this not the  case?

clusterloader2/testing/l4ilb/config.yaml
krzysied: In the object yaml, {{.Name}} translates to basename-index, which is an object name - it's the cluster loader convention). It will be different for every object instance.
Moreover, labelSelector should match label from deployment.

clusterloader2/testing/l4ilb/config.yaml
krzysied: They are measured separately.
However, if you look at ServiceCreationLatency_ilblod_<timestamp>.json in directory specified by the 'report-dir' flag, there will be only an aggregation.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,159 @@
...
+    - basename: small-backends-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: small-backends-dep
+        ILBSizeLabel: {{$SMALL_BACKEND_LABEL}}
+- measurements:
+  - Identifier: ServiceCreationLatencyLarge
+    Method: ServiceCreationLatency
+    Params:
+      labelSelector: size = ilblarge
```
krzysied: You can remove labelSelector when action in {"waitForReady", "gather"}.
LabelSelector is only parsed when action: start is called.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
...
+  qpsLoad:
+    averageQps: {{$ilbQPS}}
+steps:
+# Create ILBs
+- measurements:
+  - Identifier: ServiceCreationLatencyLarge
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      labelSelector: size = ilblarge
```
krzysied: nit: `labelSelector: size = {{$LARGE_BACKEND_LABEL}}`
Same for medium and small

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
+#Constants
+{{$LARGE_BACKEND_SIZE := 5}}
```
wojtek-t: saying that 5 is large is a bit of exaggeration :)

Don't we want to test larger services?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
...
+{{$MEDIUM_BACKEND_SIZE := 3}}
+{{$SMALL_BACKEND_SIZE := 1}}
+{{$LARGE_BACKEND_LABEL := "ilblarge"}}
```
wojtek-t: nit:

"ilb-large" "ilb-medium", "ilb-small" would be more readable

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
...
+{{$MEDIUM_BACKEND_LABEL := "ilbmedium"}}
+{{$SMALL_BACKEND_LABEL := "ilbsmall"}}
+# adding a fixed value for first version of the test, rate of pod creation not a concern yet.
+{{$ilbQPS := 500}}
+{{$namespaces := 1}}
+#Test
```
wojtek-t: nit: can you please add an empty line to make the separation clearer for readers?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
...
+    Params:
+      action: start
+      labelSelector: size = {{$SMALL_BACKEND_LABEL}}
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      apiVersion: apps/v1
+      kind: Deployment
+      labelSelector: group = ilbload
```
wojtek-t: nit: "ilb-load" for better readability?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
...
+    averageQps: {{$ilbQPS}}
+steps:
+# Create ILBs
+- measurements:
+  - Identifier: ServiceCreationLatencyLarge
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      labelSelector: size = {{$LARGE_BACKEND_LABEL}}
+  - Identifier: ServiceCreationLatencyMedium
```
wojtek-t: Can you add a short comment why we want to measure those separately, and not together?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,153 @@
...
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 1
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: large-backends-dep
+      objectTemplatePath: dep.yaml
+      templateFillMap:
+        NumReplicas: {{$LARGE_BACKEND_SIZE}}
+    - basename: large-backends-service
```
wojtek-t: We should be creating service before creating deployment [best practices]

wojtek-t: Same below.

clusterloader2/testing/l4ilb/config.yaml
prameshj: Good catch. these were 300, 150 and 10 respectively. I was testing locally with these smaller numbers to make sure the test runs. I committed them by mistake. Will fix.

clusterloader2/testing/l4ilb/config.yaml
prameshj: Curious why this is.. by creating service first, potentially you can have the service VIP with no backends available yet, right?

clusterloader2/testing/l4ilb/config.yaml
prameshj: done

clusterloader2/testing/l4ilb/config.yaml
prameshj: done

clusterloader2/testing/l4ilb/config.yaml
prameshj: done

clusterloader2/testing/l4ilb/config.yaml
prameshj: done

clusterloader2/testing/l4ilb/config.yaml
prameshj: done

clusterloader2/testing/l4ilb/config.yaml
prameshj: done

clusterloader2/testing/l4ilb/config.yaml
wojtek-t: https://kubernetes.io/docs/concepts/configuration/overview/#services is the link

clusterloader2/testing/l4ilb/dep.yaml
```diff
@@ -0,0 +1,22 @@
...
+    group: ilb-load
+spec:
+  replicas: {{.NumReplicas}}
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        group: ilbload
```
wojtek-t: ilb-load

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -0,0 +1,156 @@
...
+# Delete ILBs
+- name: Deleting ILBs
+  phases:
+  - namespaceRange:
+      min: 1
+      max: {{$namespaces}}
+    replicasPerNamespace: 0
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: large-backends-service
```
wojtek-t: nit: Removal should be done in the opposite order - first deployment, second service.

But I'm fine with fixing that in a follow-up PR.

PR #741, opened 2019-08-06
ClusterLoader - Adding prometheus flags

krzysied:
- Moving overrides parameters to prometheus flags.

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/util/system"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/flags"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -46,9 +47,33 @@ const (
 	checkPrometheusReadyTimeout  = 15 * time.Minute
 	numK8sClients                = 1
 	nodeExporterPod              = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/exporters/node-exporter.yaml"
-	nodeExporterTemplateName     = "PROMETHEUS_SCRAPE_NODE_EXPORTER"
 )
 
+var (
```
wojtek-t: Given we're cleaning this up, I'm not a fan of global variables.

Instead, can we have a type with options and make those part of some config (maybe even cluster config) and initialize them with flags more explicitly.
This is one example of doing that:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/options/etcd.go#L112

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -46,9 +47,33 @@ const (
...
+	ScrapeEtcd bool
+	// ScrapeNodeExporter is a flag that enables scraping node exporter.
+	ScrapeNodeExporter bool
+)
+
+func InitFlags() {
+	// TODO: Remove deprecated flags when all ci-jobs are using new flags.
+	flags.BoolEnvVar(&DeprecatedEnableServer, "enable-prometheus-server", "ENABLE_PROMETHEUS_SERVER", false, "(Deprecated) Whether to set-up the prometheus server in the cluster.")
+	flags.BoolEnvVar(&DeprecatedTearDownServer, "tear-down-prometheus-server", "TEAR_DOWN_PROMETHEUS_SERVER", true, "(Deprecated) Whether to tear-down the prometheus server after tests (if set-up).")
+	flags.BoolEnvVar(&EnableServer, "prometheus-enable-server", "PROMETHEUS_ENABLE_SERVER", false, "Whether to set-up the prometheus server in the cluster.")
```
wojtek-t: I really don't like it - this is a very common convension to name vars, flags, etc. with "enable-XXX" or "disable-XXX".

Having "prometheus-enables-server" is a bad choice - what we have currently is much better. Please revert this change.

krzysied: Done.

clusterloader2/pkg/prometheus/prometheus.go
krzysied: Done.

PR #742, opened 2019-08-06
Measure in-cluster network latency with more granularity.

oxddr:
Also:
 * count errors
 * fix golint issues around naming.

/assign @wojtek-t


probes/pkg/ping/client/client.go
```diff
@@ -21,50 +21,91 @@ import (
...
 
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/perf-tests/probes/pkg/common"
+
 	"k8s.io/klog"
-	"k8s.io/perf-tests/probes/pkg/metrics"
 )
 
 var (
+	// inClusterNetworkLatency implements the In-Cluster Network Programming SLI, see
```
wojtek-t: can we move it to metrics.go in this directory?

oxddr: Sure, but why?

wojtek-t: We're generally trying to define metrics in metrics.go files in all kubernetes - it's just easier to find metrics then.

oxddr: sgtm

Is is documented/defined/written somewhere?

wojtek-t: not that I'm aware of

probes/pkg/dns/metrics.go
```diff
@@ -14,24 +14,36 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package metrics
+package dns
```
wojtek-t: Hmm - github suggests you move pkg/metrics/metrics to this one. Is that true? Or is it a copy?

wojtek-t: OK - never mind, I see there are two files now.

PR #744, opened 2019-08-07
Refactoring of Granafa dashboard setup.

oxddr:
This PR is a preparation for adding more dashboards using grafanalib.

/assign @mborsz @wojtek-t

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.json
+++ b/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.json
@@ -5183,7 +5183,7 @@
           "tooltip": {
             "msResolution": true,
             "shared": true,
-            "sort": 0,
+            "sort": 2,
             "value_type": "cumulative"
           },
           "transparent": false,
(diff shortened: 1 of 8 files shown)


clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.json
```diff
@@ -5111,7 +5111,7 @@
           "tooltip": {
             "msResolution": true,
             "shared": true,
-            "sort": 0,
+            "sort": 2,
```
mborsz: IIUC you are just moving some classes around. How is it possible that the generated dashboard changes?

oxddr: I suspect this might be caused by using newer version of grafanalib (earlier on we didn't specify a version which may led to non-reproducible builds). Other than that I don't have a good idea why this has happened.

mborsz:
LGTM. Could you update README.md as well?

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.json
oxddr: Ok, I've found the culprit - I've changed the Graph class used in the Python file fromg.Graph to one with defaults provided.

PR #745, opened 2019-08-07
Show DNS lookup latency in PerfDash.

oxddr:
Also reintroduce Network latency as the metric format has changed.

/hold

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 2.10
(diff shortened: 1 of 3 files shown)


perfdash/Makefile
```diff
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = 2.9
+TAG = 2.10
```
krzysied: Should be a conflict here - 2.10 already taken. Please change to 2.11.

krzysied: Also, please update the deployment file.

PR #748, opened 2019-08-07
Grafana: create SLO dashboard with experimental panels

oxddr:
Experimental panels shows SLO but using different window size (30s and 1m).

/hold


oxddr:
ref https://github.com/kubernetes/perf-tests/issues/498

PR #753, opened 2019-08-12
Create a separate OWNERS file for Grafana dashboards.

oxddr:
/cc @mborsz 
/assign @wojtek-t


clusterloader2/pkg/prometheus/manifests/dashboards/OWNERS
```diff
@@ -0,0 +1,8 @@
+approvers:
+- oxddr
+- mborsz
+- wojtek-t
```
wojtek-t: Please remove myself (I'm on top level anyway) - both here and below.

PR #757, opened 2019-08-21
Update slo-monitor to 0.12 version

wojtek-t:


wojtek-t:
/hold

Holding until the image is promoted to gcr.k8s.io (currently it's only in staging-k8s.gcr.io).

PR #758, opened 2019-08-25
Delete deployment before service in ilbtest

prameshj:
Followup from:
https://github.com/kubernetes/perf-tests/pull/740#discussion_r316988878


krzysied:
I'm not sure if we need this PR https://github.com/kubernetes/perf-tests/blob/a2cebecd55202f2cdd178ff1aa708b621b4df894/clusterloader2/pkg/test/simple_test_executor.go#L182-L187
Assuming that objects are removed in reversed order, deployment are removed before services, so there is no need to change anything, right?

PR #762, opened 2019-08-27
fix: pass provider to errorf in nodekiller

oxddr:
/assign @jkaniuk


clusterloader2/pkg/chaos/nodes.go
```diff
@@ -25,7 +25,7 @@ import (
 	"k8s.io/perf-tests/clusterloader2/api"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 
-	"k8s.io/api/core/v1"
```
krzysied: Why do we change `v1` to `v1`?

oxddr: It was changed automatically by goimports. I am not going to fight robots ;)

PR #763, opened 2019-08-27
 Use a threshold of 1000ms in in-cluster network latency SLI.

oxddr:
Test will fail if the latency exceeds the threshold.


oxddr:
Interestingly, I looked at gce5000 (expecting worse result there) and they seem to be way below this threshold 
yGcbMVCdXAf. 

I'll check what's going on in gce100.

Issue #755, opened 2019-08-20
Ping-client pods restarts simultaneously during 5k scalability test


oxddr:
In 5k scalability tests, during ending phase of density test, most or all ping-client pods are restarting at the very specific moment (18:38-18:39), about half an hour before probe namespace removal (19:09)
Memory leak?

Checked in 5x tests, happening every time:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1156973347335049216
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1159147710117318657
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1159509970870341633
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1162771541545455621
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1163133914735185925

During load test, pods are restarting in 24min interval

/area slo
/assign @oxddr 

Reported by @jkaniuk

oxddr:
My wild guess there is some kind of memory leak in probes.

One action item would be to measure or display more metrics from probes in our Grafana dashboards, so debugging would be easier.

oxddr:
I've run test locally with kubelet scrapping enabled and it seems, that probes' memory usage grows and we have a memory leak in probes. 
![g1gHKQaiFkG](https://user-images.githubusercontent.com/30298753/64125154-62119f80-cda9-11e9-8ebb-3f6dd7391073.png)
I'm going to enable pprof in probes to allow memory profiling.

oxddr:
Based on my local runs, this seems to be fixed: ![sUZWqpTLsnp](https://user-images.githubusercontent.com/30298753/64181726-5389bd80-ce67-11e9-8a5d-77e41569ce40.png)

PR #768, opened 2019-09-02
Fix memory leaks in probes.

oxddr:
Close response body to fix memory leak in ping-client.

Also:
 * bump probes to v0.0.4
 * add pprof handlers to probes binary for ease of debugging

ref https://github.com/kubernetes/perf-tests/issues/755
/assign @mm4tt

### DIFF
--- a/probes/Makefile
+++ b/probes/Makefile
@@ -1,6 +1,6 @@
 PROJECT = k8s-testimages
 IMG = gcr.io/$(PROJECT)/probes
-TAG = v0.0.3
+TAG = v0.0.4
 
 all: push
 
--- a/probes/cmd/main.go
+++ b/probes/cmd/main.go
@@ -20,6 +20,8 @@ import (
 	"flag"
 	"net/http"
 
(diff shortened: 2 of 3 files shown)


probes/pkg/ping/server/server.go
```diff
@@ -50,7 +50,9 @@ func Run(config *PingServerConfig) {
 }
 
 func pong(w http.ResponseWriter, r *http.Request) {
-	klog.V(4).Infof("pong -> %s\n", r.RemoteAddr)
+	// dereference r so it can be closed by http framework
+	addr := r.RemoteAddr
```
wojtek-t: I don't understand why this is needed.

mm4tt: +1, I don't get it either

oxddr: Removed. I am pretty sure it helped in tests yesterday, but now I don't see it.

mm4tt:
/lgtm
/approve

/hold
Please update PR description

PR #771, opened 2019-09-03
Cache Prometheus persistent disk name before running tests

jkaniuk:
/cc @krzysied
/cc @mm4tt

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -28,31 +28,45 @@ import (
 	"k8s.io/klog"
 )
 
+type prometheusDiskMetadata struct {
+	name string
+	zone string
+}
+
 var (
 	shouldSnapshotPrometheusDisk = pflag.Bool("experimental-gcp-snapshot-prometheus-disk", false, "(experimental, provider=gce|gke only) whether to snapshot Prometheus disk before Prometheus stack is torn down")
 	prometheusDiskSnapshotName   = pflag.String("experimental-prometheus-disk-snapshot-name", "", "Name of the prometheus disk snapshot that will be created if snapshots are enabled. If not set, the prometheus disk name will be used.")
 )
 
-func (pc *PrometheusController) snapshotPrometheusDiskIfEnabled() error {
+func (pc *PrometheusController) isEnabled() (bool, error) {
 	if !*shouldSnapshotPrometheusDisk {
-		return nil
+		return false, nil
 	}
 	if pc.provider != "gce" && pc.provider != "gke" && pc.provider != "kubemark" {
-		return fmt.Errorf(
+		return false, fmt.Errorf(
 			"snapshotting Prometheus' disk only available for GCP providers (gce, gke, kubemark), provider is: %s", pc.provider)
+	}
+	return true, nil
+}
+
+func (pc *PrometheusController) cachePrometheusDiskMetadataIfEnabled() error {
+	if enabled, err := pc.isEnabled(); !enabled {
+		return err
 	}
 	return wait.Poll(
 		10*time.Second,
 		2*time.Minute,
-		pc.trySnapshotPrometheusDisk)
+		pc.tryRetrievePrometheusDiskMetadata)
 }
 
-func (pc *PrometheusController) trySnapshotPrometheusDisk() (bool, error) {
-	klog.Info("Trying to snapshot Prometheus' persistent disk...")
+func (pc *PrometheusController) tryRetrievePrometheusDiskMetadata() (bool, error) {
+	klog.Info("Retrieving Prometheus' persistent disk metadata...")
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -75,6 +80,34 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk() (bool, error) {
...
+	}
+	if pc.provider != "gce" && pc.provider != "gke" && pc.provider != "kubemark" {
+		return fmt.Errorf(
+			"snapshotting Prometheus' disk only available for GCP providers (gce, gke, kubemark), provider is: %s", pc.provider)
+	}
+	// Update cache of Prometheus disk metadata
+	err := wait.Poll(
+		10*time.Second,
+		2*time.Minute,
+		pc.tryRetrievePrometheusDiskMetadata)
```
mm4tt: See the comment below first.

Alternatively (to not have to wait.Polls) you can call tryRetrievePrometheusDiskMetadata from trySnaphostPrometheusDisk and only retry the latter.

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -75,6 +80,34 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk() (bool, error) {
...
+	if pc.diskMetadata.name == "" || pc.diskMetadata.zone == "" {
+		klog.Warningf("Missing zone or PD name, aborting snapshot")
+		klog.Infof("PD name=%s, zone=%s", pc.diskMetadata.name, pc.diskMetadata.zone)
+		return err
+	}
+	return pc.trySnapshotPrometheusDisk(pc.diskMetadata.name, pc.diskMetadata.zone)
```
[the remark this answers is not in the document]
jkaniuk: It was not retried before, but I agree - that will be safer. Fixed.

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -75,6 +80,34 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk() (bool, error) {
...
+	pc.diskMetadata.name = pdName
+	pc.diskMetadata.zone = zone
+	return true, nil
+}
+
+func (pc *PrometheusController) snapshotPrometheusDiskIfEnabled() error {
+	if !*shouldSnapshotPrometheusDisk {
+		return nil
+	}
+	if pc.provider != "gce" && pc.provider != "gke" && pc.provider != "kubemark" {
```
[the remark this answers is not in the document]
jkaniuk: done

clusterloader2/pkg/prometheus/experimental.go
jkaniuk: done with separate Poll()

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -75,21 +89,54 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk() (bool, error) {
...
+	pc.diskMetadata.zone = zone
+	return true, nil
+}
+
+func (pc *PrometheusController) snapshotPrometheusDiskIfEnabled() error {
+	if enabled, err := pc.isEnabled(); !enabled {
+		return err
+	}
+	// Update cache of Prometheus disk metadata
+	err := wait.Poll(
```
mm4tt: NIT: You could just call cachePrometheusDiskMetadataIfEnabled here

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -75,21 +89,54 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk() (bool, error) {
...
 		} else {
-			klog.Warningf("Incorrect disk name %v: %v. Using default name: %v", *prometheusDiskSnapshotName, err, pdName)
+			klog.Warningf("Incorrect disk name %v: %v. Using default name: %v", *prometheusDiskSnapshotName, err, snapshotName)
 		}
 	}
-	klog.Infof("Snapshotting PD '%s' into snapshot '%s' in zone '%s'", pdName, snapshotName, zone)
+	// Snapshot Prometheus disk
+	return wait.Poll(
+		10*time.Second,
+		1*time.Minute,
```
mm4tt: The gcloud compute disks snapshot can easily take O(minutes) for big disks (e.g. prometheus disks in 5K node clusters), so the timeout should be bigger than 1 min.

Thanks for improving this code!

jkaniuk: ok, changed to 10min

clusterloader2/pkg/prometheus/experimental.go
jkaniuk: I prefer to leave it as it is. It seems very similar, but error handling is different.

PR #782, opened 2019-09-09
Make GetReplicasFromRuntimeObject work for DaemonSets

mm4tt:


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -288,7 +288,7 @@ func GetReplicasFromRuntimeObject(obj runtime.Object) (int32, error) {
 		}
 		return 0, nil
 	case *appsv1.DaemonSet:
-		return 0, nil
+		return int32(numNodes), nil
```
wojtek-t: I don't think that's a good way of doing this. In DaemonSet you can set node-selector and then less pods can be created.

So instead of that, I think we should pass a ClientSet to this method and here list nodes with a given label-selector and return this number. I know this has a performance implication, but that isn't something that will be called all the time, and it will be more correct.

mm4tt: Done, PTAL

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -299,20 +300,44 @@ func GetReplicasFromRuntimeObject(obj runtime.Object) (int32, error) {
...
+	var numSchedulableNodes int32
+	for _, node := range list {
+		if util.IsNodeSchedulableAndUntainted(&node) {
+			numSchedulableNodes++
+		}
+	}
+	return numSchedulableNodes, nil
+}
+
+// Note: This function assumes each controller has field Spec.Replicas, except Job.
```
wojtek-t: and deaemon set?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -299,20 +300,44 @@ func GetReplicasFromRuntimeObject(obj runtime.Object) (int32, error) {
 	}
 }
 
-// Note: This function assumes each controller has field Spec.Replicas, except Daemonset and Job.
-func getReplicasFromUnstrutured(obj *unstructured.Unstructured) (int32, error) {
+// GetNumSchedulableNodesMatchingSelector returns the number of schedulable nodes matching the provided selector.
+func GetNumSchedulableNodesMatchingSelector(c clientset.Interface, nodeSelector map[string]string) (int32, error) {
```
wojtek-t: please add a TODO to support affinity when it will be needed

PR #795, opened 2019-09-12
Add option to use simple latency query for API latency.

oxddr:
Simple query is intended to be used in short tests, where there is not enough windows to compute two-level SLO. 

Also fix naming as advised by golint (Api -> API)

ref https://github.com/kubernetes/perf-tests/issues/498

/assign @wojtek-t @mm4tt

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -39,11 +39,19 @@ import (
 const (
 	apiResponsivenessPrometheusMeasurementName = "APIResponsivenessPrometheus"
 
+	// TODO(krzysied): figure out why we're getting non-capitalized proxy and fix this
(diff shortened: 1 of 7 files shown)


mm4tt:
/lgtm
/approve

/hold
In addition to adding new functionality it also refactors the code of the old one, so to be on the safe side, let's submit this Monday morning.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -39,11 +39,19 @@ import (
 const (
 	apiResponsivenessPrometheusMeasurementName = "APIResponsivenessPrometheus"
 
+	// TODO(krzysied): figure out why we're getting non-capitalized proxy and fix this
```
wojtek-t: I'm almost sure it was because of https://github.com/kubernetes/kubernetes/issues/81659

PR #802, opened 2019-09-17
Extend load test to cover PersistentVolumes

mm4tt:
The only tricky part here is deleting PVs that are created via StatefulSets. Theses PVs are not automatically deleted when StatefulSets are deleted. 
Because of that I extended the ClusterLoader2 Phase api to allow deleting objects that weren't created directly via CL2. 
The way it works is that once we detect a new object, and if certain option is set, we issue a List request to find number of replicas for that object.

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -18,6 +18,8 @@ package api
 
 import (
 	"time"
+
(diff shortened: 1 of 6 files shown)


mm4tt:
PVs have been disable on kubemark clusters in https://github.com/kubernetes/test-infra/pull/14380
/retest

PR #806, opened 2019-09-20
Restore PROMETHEUS_SCRAPE_KUBE_PROXY guarding

jkaniuk:
Needed back as on GKE kube-proxy listens only on localhost

/cc wojtek-t
/cc mm4tt
/cc oxddr


clusterloader2/testing/load/experimental/overrides/disable_kubeproxy.yaml
```diff
@@ -0,0 +1 @@
+PROMETHEUS_SCRAPE_KUBE_PROXY: false
```
mm4tt: NIT: Rename to include prometheus or scrape, as we're not disabling kube-proxy but scraping kube-proxy

mm4tt: Also move it outside experimental directory, I think there is a dir with prometheus overrides

clusterloader2/testing/load/config.yaml
```diff
@@ -93,10 +94,12 @@ steps:
     Params:
       action: start
       replicasPerProbe: {{AddInt 2 (DivideInt .Nodes 100)}}
+  {{if $PROMETHEUS_SCRAPE_KUBE_PROXY}}
```
wojtek-t: Given that this measurement has no way to work if scraping kube-proxy is disabled, we should simply encode that in the measurement itself (it should be checking if we're scriaping kube-proxy and self-disable if we're not).

mm4tt: I agree, created https://github.com/kubernetes/perf-tests/issues/807 for that.

jkaniuk: Let's do it in another commit.
(or find some solution to get that metric)

clusterloader2/testing/load/experimental/overrides/disable_kubeproxy.yaml
jkaniuk: Done

PR #813, opened 2019-09-25
ClusterLoader - Using unschedulable spec

krzysied:
Filtering deamonset schedulable nodes only by node unschedulable spec.
ref https://github.com/kubernetes/kubernetes/issues/82875


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -316,7 +315,7 @@ func GetNumSchedulableNodesMatchingSelector(c clientset.Interface, nodeSelector
 	}
 	var numSchedulableNodes int32
 	for _, node := range list {
-		if util.IsNodeSchedulableAndUntainted(&node) {
+		if !node.Spec.Unschedulable {
```
wojtek-t: Please make this a private function so that we won't start using it from other places.

krzysied: Done.

PR #818, opened 2019-09-30
Add storage experimental test suite

pierewoj:
This adds a test suite for running storage tests. In the next commit (to
test-infra), a prow job that runs this suite will be defined in
sig-scalability-experimental-periodic-jobs.

The goal of running a suite instead of duplicated jobs is to minimize
config duplication.

There is some redundancy in the suite definition. To be investigated if
it can be reduced using go templates. If it is feasible, it will be done
as a follow-up change once the suite job is working.

Tesing:
* ran clusterloader with
testsuite=testing/experimental/storage/pod-startup/suite.yaml and
verified that tests passed

### DIFF
--- a/clusterloader2/testing/experimental/storage/pod-startup/empty_test_name/override.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/empty_test_name/override.yaml
@@ -1,0 +1,7 @@
+# TODO(pierewoj): Remove this file once all storage tests are migrated to a single suite.
+# This is used for the storage tests migration from separate jobs to a single suite.
+# Old storage tests won't have this override applied, so metric file names will
+# remain the same.
+# New tests will use this override, so the metric file name will be only parametrized
+# by the suite test identifier.
+TEST_NAME: ''
--- a/clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
@@ -1,0 +1,55 @@
+- identifier: emptydir-vol-per-pod
+  configPath: testing/experimental/storage/pod-startup/config.yaml
+  overridePaths:
+    - testing/experimental/storage/pod-startup/volume-types/emptydir/override.yaml
+    - testing/experimental/storage/pod-startup/max_volumes_per_pod/override.yaml
+    - testing/experimental/storage/pod-startup/empty_test_name/override.yaml
+
+- identifier: configmap-vol-per-pod
+  configPath: testing/experimental/storage/pod-startup/config.yaml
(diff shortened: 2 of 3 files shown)


clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
```diff
@@ -0,0 +1,47 @@
+- identifier: emptydir-vol-per-pod
+  configPath: "testing/experimental/storage/pod-startup/config.yaml"
+  overridePaths:
+    - "./testing/experimental/storage/pod-startup/volume-types/emptydir/override.yaml"
```
mborsz: Does it need to be in quotation marks? Here and in all the file?

clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - "./testing/experimental/storage/pod-startup/max_volumes_per_pod/override.yaml"
+
+- identifier: configmap-vol-per-pod
+  configPath: "testing/experimental/storage/pod-startup/config.yaml"
+  overridePaths:
+    - "./testing/experimental/storage/pod-startup/volume-types/configmap/override.yaml"
```
mborsz: Nit: I don't think you need "./" prefix.

clusterloader2/testing/experimental/storage/pod-startup/empty_test_name/override.yaml
```diff
@@ -0,0 +1,6 @@
+# This is used for the storage tests migration from separate jobs to a single suite.
+# Old storage tests won't have this override applied, so metric file names will
+# remain the same.
+# New tests will use this override, so the metric file name will be only parametrized
+# by the suite test identifier.
+TEST_NAME: ''
```
mborsz: Please add TODO(pierewoj): Remove this file once all storage tests are migrated to a single suite.

clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
pierewoj: Fixed

clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
pierewoj: Fixed

clusterloader2/testing/experimental/storage/pod-startup/empty_test_name/override.yaml
pierewoj: Fixed

PR #819, opened 2019-09-30
Skip fetching metrics by perfdash for misconfigured jobs

pierewoj:
Commit https://github.com/kubernetes/test-infra/commit/50615e resulted
in breaking perfdash. This was due to lack of support for
"storage-suite" job type in perfdash.

To prevent such changes from breaking perfdash again, jobs with bad config will
be ignored.

Testing:
* ran perfdash locally with bad config for storage jobs, verified that
other job data is fetched correctly

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -389,48 +389,70 @@ func getProwConfig(configPaths []string) (Jobs, error) {
 			return nil, fmt.Errorf("error unmarshaling prow config from %s: %v", configPath, err)
 		}
 		for _, periodic := range conf.Periodics {
-			var thisPeriodicConfig Tests
-			for _, tag := range periodic.Tags {
-				if strings.HasPrefix(tag, "perfDashPrefix:") {
-					split := strings.SplitN(tag, ":", 2)
-					thisPeriodicConfig.Prefix = strings.TrimSpace(split[1])
-					continue
-				}
-				if strings.HasPrefix(tag, "perfDashJobType:") {
-					split := strings.SplitN(tag, ":", 2)
-					jobType := strings.TrimSpace(split[1])
-					var exists bool
-					if thisPeriodicConfig.Descriptions, exists = jobTypeToDescriptions[jobType]; !exists {
-						fmt.Fprintf(os.Stderr, "warning: unknown job type - %s\n", jobType)
-					}
-					continue
-				}
-				if strings.HasPrefix(tag, "perfDashBuildsCount:") {
-					split := strings.SplitN(tag, ":", 2)
-					i, err := strconv.Atoi(strings.TrimSpace(split[1]))
-					if err != nil {
-						fmt.Fprintf(os.Stderr, "warning: unparsable builds count - %v\n", split[1])
-					}
-					if i < 1 {
-						fmt.Fprintf(os.Stderr, "warning: non-positive builds count - %v\n", i)
-						continue
-					}
-					thisPeriodicConfig.BuildsCount = i
-					continue
-				}
-				if strings.HasPrefix(tag, "perfDash") {
-					fmt.Fprintf(os.Stderr, "warning: unknown perfdash tag name: %q\n", tag)
-				}
-			}
-			if thisPeriodicConfig.Prefix == "" && thisPeriodicConfig.Descriptions == nil {
+			config, err := parsePeriodicConfig(periodic)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "warning: failed to parse config of %q due to: %v\n",
+					periodic.Name, err)
 				continue
 			}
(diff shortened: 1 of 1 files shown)


perfdash/config.go
```diff
@@ -426,7 +426,8 @@ func getProwConfig(configPaths []string) (Jobs, error) {
 				continue
 			}
 			if thisPeriodicConfig.Prefix == "" || thisPeriodicConfig.Descriptions == nil {
-				return nil, fmt.Errorf("invalid perfdash config of periodic %q: none or both of prefix and job type must be specified", periodic.Name)
+				fmt.Fprintf(os.Stderr, "warning: skipping periodic %q due to invalid config\n", periodic.Name)
```
mborsz: I think we need to be more verbose here. "invalid config" can mean a lot of different things.

Please add info that "none or both of prefix and job type..." from the previous error message.

perfdash/config.go
```diff
@@ -389,48 +389,60 @@ func getProwConfig(configPaths []string) (Jobs, error) {
...
+			if config.Prefix == "" && config.Descriptions == nil {
+				// this is expected case for jobs which are not expected to be visible in perfdash
 				continue
 			}
-			if thisPeriodicConfig.Prefix == "" || thisPeriodicConfig.Descriptions == nil {
-				return nil, fmt.Errorf("invalid perfdash config of periodic %q: none or both of prefix and job type must be specified", periodic.Name)
+			if config.Prefix == "" || config.Descriptions == nil {
+				fmt.Fprintf(os.Stderr, "warning: invalid perfdash config of periodic %q: " +
+					"none or both of prefix and job type must be specified\n", periodic.Name)
+				continue
```
mborsz: I think it is better to include this in parsePeriodicConfig.

perfdash/config.go
```diff
@@ -389,48 +389,60 @@ func getProwConfig(configPaths []string) (Jobs, error) {
...
-					fmt.Fprintf(os.Stderr, "warning: unknown perfdash tag name: %q\n", tag)
-				}
+			 config, err := parsePeriodicConfig(periodic)
+			 if err != nil {
+				fmt.Fprintf(os.Stderr, "warning: failed to parse config of %q due to: %v\n",
+					periodic.Name, err)
+				continue
 			}
-			if thisPeriodicConfig.Prefix == "" && thisPeriodicConfig.Descriptions == nil {
+			if config.Prefix == "" && config.Descriptions == nil {
```
mborsz: While you are modifying this, can we move this to parsePeriodicConfig as well?

I was thinking about something like:
config, hasConfig, err := parsePeridicConfig(periodic)
if err != nil {
  // log error
}
if hasConfig {
  jobs[periodic.Name] = config
}

if hasConfig is true if config.Prefix != "" or config.Descriptions != "".

Other way to do that is to say that parsePeriodicConfig can return nil instead of hasConfig.

WDYT?

pierewoj: discussed offline, extracted validation logic to yet another method

perfdash/config.go
```diff
@@ -389,48 +389,70 @@ func getProwConfig(configPaths []string) (Jobs, error) {
...
+		if strings.HasPrefix(tag, "perfDash") {
+			return Tests{}, fmt.Errorf("unknown perfdash tag name: %q", tag)
+		}
+	}
+	return thisPeriodicConfig, nil
+}
+
+func validatePeriodicConfig(config Tests) (shouldUse bool, err error) {
+	if config.Prefix == "" && config.Descriptions == nil {
+		// this is expected case for jobs which are not expected to be visible in perfdash
```
krzysied: nit: Comments should start with capital letter and  end with period.

perfdash/config.go
```diff
@@ -389,48 +389,70 @@ func getProwConfig(configPaths []string) (Jobs, error) {
...
+		return false, nil
+	}
+	if config.Prefix == "" || config.Descriptions == nil {
+		return false, fmt.Errorf("nonee or both of prefix and job type must be specified")
```
krzysied: s/nonee/none/g

perfdash/config.go
```diff
@@ -389,48 +389,70 @@ func getProwConfig(configPaths []string) (Jobs, error) {
...
+			continue
+		}
+		if strings.HasPrefix(tag, "perfDash") {
+			return Tests{}, fmt.Errorf("unknown perfdash tag name: %q", tag)
+		}
+	}
+	return thisPeriodicConfig, nil
+}
+
+func validatePeriodicConfig(config Tests) (shouldUse bool, err error) {
```
krzysied: nit: I'm not sure if named return arguments are popular in golang.
Btw, maybe this function should have different name, so  named return arguments are not needed.

pierewoj:
> Some FYI comments to already merged PR.

Thanks @krzysied, will address those in a follow-up PR

PR #820, opened 2019-10-01
Add support for displaying test suite results in perfdash

pierewoj:
This commit enables perfdash to view metrics recorded by clusterloader
test suite runs.

Before this change, it was not possible to do so, because test suite run
puts measurement results in separate files, for example:
-  APIResponsiveness_storage_configmap-vol-per-node_2019-10-01T06:08:05Z.json
-  APIResponsiveness_storage_downwardapi-vol-per-node_2019-10-01T06:10:00Z.json
In such case, perfdash read metrics data from the first file that
matched the prefix APIResponsiveness_storage (prefix is specified in
config.go file), meaning that other results were ignored.

To solve this issue, support for handling multiple files that match
given prefix is added. For each of the matched files, perfdash extracts
the "suiteId" from the file name and appends it to the result category.
This means that results from different test cases in a given suite will
land in different categories in the web UI, for example:
- APIServer_configmap-vol-per-node
- APIServer_downwardapi-vol-per-node

Alternatively, same effect could be achieved by generating separate test
descriptions in the perfdash config. It was not chosen, because it would
mean that perfdash config needs to be changed every time some test case
is added or removed.

This commit also includes other minor, but related changes like:
- increasing with of category field in the web UI
- removing 'pod-with-ephemeral-volume' configs

Testing:
- run perfdash locally, verify that data for both storage-suite and
separate storage jobs is displayed as expected

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -282,24 +282,14 @@ var (
 		"APIServer": {
 			"Responsiveness": []TestDescription{
 				{
-					Name:             "pod-with-ephemeral-volume-startup-latency",
-					OutputFilePrefix: "APIResponsiveness",
-					Parser:           parsePerfData,
-				},
-				{
-					Name:             "storage-",
+					Name:             "storage",
 					OutputFilePrefix: "APIResponsiveness",
 					Parser:           parsePerfData,
 				},
 			},
 			"RequestCount": []TestDescription{
 				{
-					Name:             "pod-with-ephemeral-volume-startup-latency",
-					OutputFilePrefix: "APIResponsiveness",
-					Parser:           parseRequestCountData,
-				},
-				{
-					Name:             "storage-",
+					Name:             "storage",
 					OutputFilePrefix: "APIResponsiveness",
 					Parser:           parseRequestCountData,
 				},
@@ -308,12 +298,7 @@ var (
 		"E2E": {
 			"PodStartup": []TestDescription{
 				{
-					Name:             "pod-with-ephemeral-volume-startup-latency",
-					OutputFilePrefix: "PodStartupLatency_PodWithMultiVolumeStartupLatency",
-					Parser:           parsePerfData,
-				},
-				{
-					Name:             "storage-",
+					Name:             "storage",
 					OutputFilePrefix: "PodStartupLatency_PodWithVolumesStartupLatency",
 					Parser:           parsePerfData,
 				},
--- a/perfdash/google-gcs-downloader.go
+++ b/perfdash/google-gcs-downloader.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -78,28 +79,22 @@ func (g *GoogleGCSDownloader) getData() (JobToCategoryData, error) {
 		if tests.Prefix == "" {
 			return nil, fmt.Errorf("Invalid empty Prefix for job %s", job)
 		}
-		for categoryLabel, categoryMap := range tests.Descriptions {
-			for testLabel := range categoryMap {
-				resultLock.Lock()
-				if _, found := result[tests.Prefix]; !found {
-					result[tests.Prefix] = make(CategoryToMetricData)
-				}
-				if _, found := result[tests.Prefix][categoryLabel]; !found {
-					result[tests.Prefix][categoryLabel] = make(MetricToBuildData)
-				}
-				if _, found := result[tests.Prefix][categoryLabel][testLabel]; found {
-					return result, fmt.Errorf("Duplicate name %s for %s", testLabel, tests.Prefix)
-				}
-				result[tests.Prefix][categoryLabel][testLabel] = &BuildData{Job: job, Version: "", Builds: map[string][]perftype.DataItem{}}
-				resultLock.Unlock()
-			}
-		}
 		go g.getJobData(&wg, result, &resultLock, job, tests)
 	}
 	wg.Wait()
 	return result, nil
 }
 
+/*
+getJobData fetches build numbers, reads metrics data from GCS and
+updates result with parsed metrics for a given prow job. Assumptions:
+- metric files are in /artifacts directory
+- metric file names have following prefix: {{OutputFilePrefix}}_{{Name}},
+  where OutputFilePrefix and Name are parts of test description (specified in prefdash config)
+- if there are multiple files with a given prefix, then expected format is
+  {{OutputFilePrefix}}_{{Name}}_{{SuiteId}}. SuiteId is appended to the category label,
+  which allows comparing metrics across several runs in a given suite
+*/
 func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategoryData, resultLock *sync.Mutex, job string, tests Tests) {
 	defer wg.Done()
 	buildNumbers, err := g.GoogleGCSBucketUtils.getBuildNumbersFromJenkinsGoogleBucket(job)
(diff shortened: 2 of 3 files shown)


perfdash/config.go
```diff
@@ -282,24 +282,14 @@ var (
 		"APIServer": {
 			"Responsiveness": []TestDescription{
 				{
-					Name:             "pod-with-ephemeral-volume-startup-latency",
-					OutputFilePrefix: "APIResponsiveness",
-					Parser:           parsePerfData,
-				},
-				{
-					Name:             "storage-",
+					Name:             "storage",
```
mborsz: Do the "old storage" test still work after this change?

pierewoj: It seems to work properly:
* I've started the perfdash locally and it has both the "old storage" metrics and "storage suite" metrics
* Checked files from old test run (https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability-max-downwardapi-vol-per-node/1178934805790724098/artifacts/) and each it looks like using "storage-" or "storage" prefixes should have same effect

perfdash/google-gcs-downloader.go
```diff
@@ -120,33 +114,57 @@ func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategor
...
 					if err != nil || len(artifacts) == 0 {
-						fmt.Printf("Error while looking for %s* in %s build %v: %v\n", fileStem, job, buildNumber, err)
+						fmt.Printf("Error while looking for %s* in %s build %v: %v\n", searchPrefix, job, buildNumber, err)
 						continue
 					}
-					metricsFilename := artifacts[0][strings.LastIndex(artifacts[0], "/")+1:]
-					if len(artifacts) > 1 {
-						fmt.Printf("WARNING: found multiple %s files with data, reading only one: %s\n", fileStem, metricsFilename)
+					for _, artifact := range artifacts {
+						metricsFileName := artifact[strings.LastIndex(artifact, "/")+1:]
```
mborsz: Please use https://golang.org/pkg/path/filepath/#Base -- it will be easier to read IMHO.

perfdash/google-gcs-downloader.go
```diff
@@ -120,33 +114,57 @@ func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategor
...
 	}
 }
 
+func getResultCategory(metricsFileName string, filePrefix string, category string, artifacts []string) string {
+	if len(artifacts) <= 1 {
+		return category
+	}
+	// if there are more artifacts, assume that this is a test suite run
+	trimmed := strings.TrimPrefix(metricsFileName, filePrefix+"_")
+	suiteId := strings.Split(trimmed, "_")[0]
```
mborsz: What if len(strings.Split(trimmed, "_")) == 0?

pierewoj: from split documentation:
```
// If s does not contain sep and sep is not empty, Split returns a
// slice of length 1 whose only element is s.
(...)
func Split(s, sep string) []string { return genSplit(s, sep, 0, -1) }
```

perfdash/google-gcs-downloader.go
```diff
@@ -120,33 +114,57 @@ func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategor
 		for categoryLabel, categoryMap := range tests.Descriptions {
 			for testLabel, testDescriptions := range categoryMap {
 				for _, testDescription := range testDescriptions {
-					fileStem := fmt.Sprintf("artifacts/%v_%v", testDescription.OutputFilePrefix, testDescription.Name)
-					artifacts, err := g.GoogleGCSBucketUtils.listFilesInBuild(job, buildNumber, fileStem)
+					filePrefix := fmt.Sprintf("%v_%v", testDescription.OutputFilePrefix, testDescription.Name)
+					searchPrefix := fmt.Sprintf("artifacts/%v", filePrefix)
```
mborsz: Nit: Why this change? I think I preferred the previous version.

perfdash/google-gcs-downloader.go
pierewoj: good tip, fixed

perfdash/google-gcs-downloader.go
pierewoj: mostly because `filePrefix` is used later on when constructing the `resultCategory`, I wanted to avoid building the prefix twice so that we cna change it in one place (hopefully) if the filename conventions change

perfdash/google-gcs-downloader.go
```diff
@@ -120,33 +115,57 @@ func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategor
...
+					for _, artifact := range artifacts {
+						metricsFileName := filepath.Base(artifact)
+						resultCategory := getResultCategory(metricsFileName, filePrefix, categoryLabel, artifacts)
+						testDataResponse, err := g.GoogleGCSBucketUtils.getFileFromJenkinsGoogleBucket(job, buildNumber,
+							fmt.Sprintf("artifacts/%v", metricsFileName))
+						if err != nil {
+							fmt.Fprintf(os.Stderr, "Error when reading response Body: %v\n", err)
+							continue
+						}
+						buildData := getBuildData(result, tests.Prefix, resultCategory, testLabel, job, resultLock)
```
mborsz: Don't you need to hold resultLock here?

perfdash/google-gcs-downloader.go
mborsz: Good point.

perfdash/google-gcs-downloader.go
mborsz: OK, makes sense.

perfdash/google-gcs-downloader.go
pierewoj: We actually hold the result lock when accessing the map, see the `getBuildData` impl

mborsz: You are right. It's OK.

perfdash/google-gcs-downloader.go
krzysied: Can you give example of a file name and how will it behave when there is no suiteId?
file name  = "metric_test_timestamp" => suiteId = timestamp?

perfdash/google-gcs-downloader.go
```diff
@@ -120,33 +115,57 @@ func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategor
...
 			}
 		}
 	}
 }
 
+func getResultCategory(metricsFileName string, filePrefix string, category string, artifacts []string) string {
+	if len(artifacts) <= 1 {
+		return category
+	}
+	// if there are more artifacts, assume that this is a test suite run
```
krzysied: nit: Capital letter and period.

perfdash/google-gcs-downloader.go
pierewoj: Yes, that's correct. But this would mean that we'll have two files with same "metric_test" prefix and a different "timestamp" for the same run.

We could add some validation around it. For example we could expect that 
`len(strings.Split(trimmed, "_")) ` is greater than 2. However, I suggest to keep it simple for now. What do you think?

perfdash/google-gcs-downloader.go
pierewoj: Fixed

perfdash/google-gcs-downloader.go
krzysied: > Yes, that's correct. But this would mean that we'll have two files with same "metric_test" prefix and a different "timestamp" for the same run.

What if there is only one file with this name format? Aren't all of our tests creating files like this unless we migrate them to test suite?
Example `ResourceUsageSummary_density_2019-10-02T10:08:21Z.json`

krzysied: For some reason I missed 
```
if len(artifacts) <= 1 {
	return category
}
```
It should behave correctly for old summaries.

PR #823, opened 2019-10-02
ClusterLoader - Latency pod memory update - 100 nodes

krzysied:
Using better memory estimation for latency pods in gce 100 test. This will save ~200MB for potential probe pods.

Scheduler uses allocatable memory to evaluate node score. Nodes have < 3500MB memory available. Therefore 350MB should be a good choice.

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -13,7 +13,7 @@
 # decreases the value of priority function in scheduler by one point.
 # This results in decreased probability of choosing the same node again.
 {{$LATENCY_POD_CPU := DefaultParam .LATENCY_POD_CPU 100}}
-{{$LATENCY_POD_MEMORY := DefaultParam .LATENCY_POD_MEMORY 400}}
+{{$LATENCY_POD_MEMORY := DefaultParam .LATENCY_POD_MEMORY 350}}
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/density/100_nodes/override.yaml
```diff
@@ -1 +1,2 @@
 DENSITY_RESOURCE_CONSTRAINTS_FILE: ./100_nodes/constraints.yaml
+LATENCY_POD_MEMORY: 350
```
wojtek-t: Why can't we change it everywhere? What's the point of introducing differences between tests?

krzysied: You're right. I'm changing density test default.

mm4tt:
Could you check why the tests failed?

krzysied:
Tests failed with `ERROR: (gcloud.compute.routers.create) Quota 'ROUTERS' exceeded. Limit: 10.0 globally.` during `up` phase.
I will check project quota.

krzysied:
Resources are available in the project, so it seems it was a random collision with other tests (I guess).

krzysied:
TL; DR Reason of test failure was unrelated to changes introduced in this PR.

PR #825, opened 2019-10-03
Increase perfdash mem limit and change format of category label

pierewoj:


perfdash/deployment.yaml
```diff
@@ -39,8 +39,8 @@ spec:
         resources:
           requests:
             cpu: 600m
-            memory: 4000Mi
+            memory: 8000Mi
```
mm4tt: NIT: 8Gi?

And below

PR #828, opened 2019-10-03
Extended Load: NetworkPolicies

jkaniuk:
Copy of: https://github.com/kubernetes/perf-tests/pull/719 with fixed paths

This is most likely no-op until we turn on some Network Policy Provider that will start enfocring these network policies.

It should be pretty straightforward to turn on Calico both in GKE and in GCE.
This should be done separately to isolate any potential performance
impact of tuning it just on.

/cc oxddr
/cc mm4tt


jkaniuk:
Comments from previous PR:

@oxddr on Jul 31 Contributor
You should add this knob in a separate PR, then enable it in perf-tests presubmit and the follow with the rest of this PR. This way, the presubmit will test the codepath/scenarios that's being added as part of this PR preventing breakages.

This has been documented in here

 @mm4tt
mm4tt on Jul 31 Author Contributor
This is an experimental suite, it's not part of any CI/CD tests. Therefore, I don't think that the rollout precedure applies here :)

 @oxddr
oxddr on Aug 1 Contributor
Make sense.

Issue #246, opened 2018-10-15
ClusterLoader: HA cluster support


krzysied:
Currently clusterloader supports only single-master cluster. This should be changed - clusterloader should handle multiple masters.

wojtek-t:
It's not about cluster-loader per-se.

It's about some of the measurements, that aren't really HA-friendly.

wojtek-t:
/assign @mborsz @mm4tt 

You already did some work around that - can one of you describe what is missing here?

mm4tt:
What is missing is basically going through all callers of [GetMasterIp and GetMasterInternalIp methods](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/config/cluster.go) and migrating them to use multiple IPs if feasible. 

I'll try to do it in a free cycles.

alejandrox1:
Hi all,
does anyone mind if i take this?
/assign @alejandrox1

mm4tt:
Hey @alejandrox1,

We don't mind at all, quite the opposite :) Let me know if you need any assistance with this, I'll be happy to help.

Thanks!

alejandrox1:
thank you Matt

And an update: i'am actively working on this and will get a pr out as soon as possible :)
/lifecycle active

alejandrox1:
holidays got me bad.
Back on this. Gonna start with the resource metric measurement part

alejandrox1:
lost the bandwidth for working on this. sorry for not being able to drive this to completion but i'll unassign to let others sign up.
/unassign @alejandrox1

PR #833, opened 2019-10-05
Enabled HA support for ETCD measurements

alejandrox1:
This patch enables clusterloaderv2 etcd measurements for multiple master
nodes.
ref: https://github.com/kubernetes/perf-tests/issues/246

### DIFF
--- a/clusterloader2/pkg/measurement/common/etcd_metrics.go
+++ b/clusterloader2/pkg/measurement/common/etcd_metrics.go
@@ -48,6 +48,7 @@ func createEtcdMetricsMeasurement() measurement.Measurement {
 }
 
 type etcdMetricsMeasurement struct {
+	sync.Mutex
 	isRunning bool
 	stopCh    chan struct{}
 	wg        *sync.WaitGroup
@@ -66,9 +67,10 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig) 
 	if err != nil {
 		return nil, err
 	}
-	host, err := util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp())
-	if err != nil {
-		return nil, err
+	hosts := config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(hosts) < 1 {
+		klog.Warningf("ETCD measurements will be disabled due to no MasterIps: %v", hosts)
+		return nil, nil
 	}
 
 	switch action {
@@ -78,11 +80,15 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig) 
 		if err != nil {
 			return nil, err
 		}
-		e.startCollecting(host, provider, waitTime)
+		for _, h := range hosts {
+			e.startCollecting(h, provider, waitTime)
+		}
 		return nil, nil
 	case "gather":
-		if err = e.stopAndSummarize(host, provider); err != nil {
-			return nil, err
+		for _, h := range hosts {
+			if err = e.stopAndSummarize(h, provider); err != nil {
+				return nil, err
+			}
 		}
 		content, err := util.PrettyPrintJSON(e.metrics)
 		if err != nil {
@@ -111,17 +117,29 @@ func (e *etcdMetricsMeasurement) String() string {
 func (e *etcdMetricsMeasurement) startCollecting(host, provider string, interval time.Duration) {
 	e.isRunning = true
 	e.wg.Add(1)
+
+	collectEtcdDatabaseSize := func() error {
+		dbSize, err := e.getEtcdDatabaseSize(host, provider)
+		if err != nil {
+			return err
+		}
+
+		e.Lock()
+		defer e.Unlock()
+		e.metrics.MaxDatabaseSize = math.Max(e.metrics.MaxDatabaseSize, dbSize)
+
+		return nil
+	}
 	go func() {
 		defer e.wg.Done()
 		for {
 			select {
 			case <-time.After(interval):
-				dbSize, err := e.getEtcdDatabaseSize(host, provider)
+				err := collectEtcdDatabaseSize()
 				if err != nil {
 					klog.Errorf("%s: failed to collect etcd database size", e)
 					continue
 				}
-				e.metrics.MaxDatabaseSize = math.Max(e.metrics.MaxDatabaseSize, dbSize)
 			case <-e.stopCh:
 				return
 			}
@@ -136,17 +154,28 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string) error {
 	if err != nil {
 		return err
 	}
-	for _, sample := range samples {
+
+	collectEtcdMetrics := func(sample *model.Sample) {
+		var hist *measurementutil.HistogramVec
 		switch sample.Metric[model.MetricNameLabel] {
 		case "etcd_disk_backend_commit_duration_seconds_bucket":
-			measurementutil.ConvertSampleToBucket(sample, &e.metrics.BackendCommitDuration)
+			hist = &e.metrics.BackendCommitDuration
 		case "etcd_debugging_snap_save_total_duration_seconds_bucket":
-			measurementutil.ConvertSampleToBucket(sample, &e.metrics.SnapshotSaveTotalDuration)
+			hist = &e.metrics.SnapshotSaveTotalDuration
 		case "etcd_disk_wal_fsync_duration_seconds_bucket":
-			measurementutil.ConvertSampleToBucket(sample, &e.metrics.WalFsyncDuration)
+			hist = &e.metrics.WalFsyncDuration
 		case "etcd_network_peer_round_trip_time_seconds_bucket":
-			measurementutil.ConvertSampleToBucket(sample, &e.metrics.PeerRoundTripTime)
+			hist = &e.metrics.PeerRoundTripTime
+		default:
+			return
 		}
+
+		e.Lock()
+		measurementutil.ConvertSampleToBucket(sample, hist)
+		e.Unlock()
+	}
+	for _, sample := range samples {
+		collectEtcdMetrics(sample)
 	}
 	return nil
 }
--- a/clusterloader2/pkg/measurement/util/histogram_test.go
+++ b/clusterloader2/pkg/measurement/util/histogram_test.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/prometheus/common/model"
+)
+
+func TestConvertSampleToBucket(t *testing.T) {
+	tests := map[string]struct {
+		samples []*model.Sample
+		want    *HistogramVec
+	}{
+		"convert single sample to bucket": {
+			samples: []*model.Sample{
+				{
+					Metric: model.Metric{"name": "value", "le": "le-value"},
+					Value:  1,
+				},
+			},
+			want: &HistogramVec{
+				{
+					Labels: map[string]string{
+						"name": "value",
+					},
+					Buckets: map[string]int{
+						"le-value": 1,
+					},
+				},
+			},
+		},
+		"convert multiple samples to bucket": {
+			samples: []*model.Sample{
+				{
+					Metric: model.Metric{"name": "value", "le": "le-value"},
+					Value:  1,
+				},
+				{
+					Metric: model.Metric{"name": "value", "le": "le-value"},
+					Value:  2,
+				},
+			},
+			want: &HistogramVec{
+				{
+					Labels: map[string]string{
+						"name": "value",
+					},
+					Buckets: map[string]int{
+						"le-value": 3,
+					},
+				},
+			},
+		},
+	}
+
+	var inputHistogramVec HistogramVec
+	for name, test := range tests {
+		inputHistogramVec = HistogramVec{
+			{
+				Labels: map[string]string{
+					"name": "value",
+				},
+				Buckets: map[string]int{
+					"le-value": 0,
+				},
+			},
+		}
+		for _, sample := range test.samples {
+			ConvertSampleToBucket(sample, &inputHistogramVec)
+		}
+		if !reflect.DeepEqual(&inputHistogramVec, test.want) {
+			t.Errorf("error %s: \n\tgot %#v \n\twanted %#v", name, &inputHistogramVec, test.want)
+		}
+	}
+}
--- a/clusterloader2/pkg/measurement/util/histogram.go
+++ b/clusterloader2/pkg/measurement/util/histogram.go
@@ -58,5 +58,5 @@ func ConvertSampleToBucket(sample *model.Sample, h *HistogramVec) {
 		hist = NewHistogram(labels)
 		*h = append(*h, *hist)
 	}
-	hist.Buckets[string(sample.Metric["le"])] = int(sample.Value)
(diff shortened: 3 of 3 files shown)


alejandrox1:
/title Migrating GetMasterIP callers to use multiple IPs
/milestone v1.17

alejandrox1:
/retest
thought i fixed those errors https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/833/pull-perf-tests-clusterloader2/1180612144207499264/

alejandrox1:
boskos err(?): https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/833/pull-perf-tests-clusterloader2-kubemark/1180660597503561728/#1:build-log.txt%3A401

alejandrox1:
/assign @mm4tt @wojtek-t 
ptal

clusterloader2/pkg/util/util.go
```diff
@@ -75,6 +75,15 @@ func GetStringOrDefault(dict map[string]interface{}, key string, defaultValue st
 	return value, err
 }
 
+// GetStringSliceOrDefault tries to return value from map cast to string slice type. If value doesn't exist default value is used.
```
mm4tt: Could you extract changes in this file to a separate PR?

alejandrox1: Should I separate the rewrite of `GetMasterIp()` to `GetMasterIps() then?
As it stands, all the callers depend on `GetStringSliceOrDefault()`, i.e., https://github.com/kubernetes/perf-tests/pull/833/files#diff-1501ea581b9304b1c1a8b79122c99a6eR69

alejandrox1: I guess, I could keep the usage of `GetStringOrDefault` and just call `GetMasterIps()` beforehand. WOuld that be prefered?

clusterloader2/pkg/config/cluster.go
```diff
@@ -50,13 +50,12 @@ type PrometheusConfig struct {
 	ScrapeKubeProxy    bool
 }
 
-// GetMasterIp returns the first master ip, added for backward compatibility.
-// TODO(mmatt): Remove this method once all the codebase is migrated to support multiple masters.
-func (c *ClusterConfig) GetMasterIp() string {
+// GetMasterIps returns all registered master IPs.
+func (c *ClusterConfig) GetMasterIps() []string {
 	if len(c.MasterIPs) > 0 {
```
mm4tt: Once we get rid of all the callers of this method we won't need it anymore. You can just delete it and get masterIps directly from the MasterIPs field in the downstream code.

clusterloader2/pkg/util/util.go
mm4tt: You can send a PR adding GetStringSLiceOrDefault method (basically all the changes in this file) separately. As it's very straighforward change it will get merged the same day. In the meantime we can continue working on this one, which is  a bit more tricky and will require more work.

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -61,9 +61,14 @@ func (p *profileMeasurement) populateProfileConfig(config *measurement.Measureme
...
 	}
-	if p.config.host, err = util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp()); err != nil {
+	masterIPs, err := util.GetStringSliceOrDefault(config.Params, "hosts", config.ClusterFramework.GetClusterConfig().GetMasterIps())
+	if err != nil {
 		return err
 	}
+	if masterIPs == nil || len(masterIPs) < 1 {
+		return fmt.Errorf("empty slice: there are no master IPs registered %v", masterIPs)
+	}
+	p.config.host = masterIPs[0]
```
mm4tt: Same here and in other files.

Also, it might be a good idea to tackle each of these files in a separate PR (depending on how much work needs to be done to rewrite a single one). If it's more than 50 lines per file then it will better to split them into separate PRs (it will be easier / faster to review).

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -66,10 +66,14 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig)
...
 	}
-	host, err := util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp())
+	masterIPs, err := util.GetStringSliceOrDefault(config.Params, "hosts", config.ClusterFramework.GetClusterConfig().GetMasterIps())
 	if err != nil {
 		return nil, err
 	}
+	if masterIPs == nil || len(masterIPs) < 1 {
+		return nil, fmt.Errorf("empty slice: there are no master IPs registered %v", masterIPs)
+	}
+	host := masterIPs[0]
```
mm4tt: Actually this is a bit more tricky, If we have more than one masterIp we should monitor each of them, not only the first one. So, we'll need to call e.startCollecting for each masterIp (host) and then rewrite the stopAndSummarize and other parts to make sure we collect and include data from each host in the summary returned by this measurement.

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -43,15 +43,15 @@ func createEtcdMetricsMeasurement() measurement.Measurement {
...
+		metrics: make(map[string]*etcdMetrics),
 	}
 }
 
 type etcdMetricsMeasurement struct {
 	isRunning bool
 	stopCh    chan struct{}
 	wg        *sync.WaitGroup
-	metrics   *etcdMetrics
+	metrics   map[string]*etcdMetrics
```
alejandrox1: The intent here is to organize the etcdMetrics per host (master IP) in order to have `e.startCollecting()` and `e.stopAndSummarize()` collect data on a single host.
The keys for the map are strings in order to conform with [`ClusterConfig.MasterIPs`](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/config/cluster.go#L37).

alejandrox1:
sorry for the delay on this. Will prioritize this going forward
ptal 
/cc @mm4tt

alejandrox1:
clusterloader2 kubemark failed in the up stage https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/833/pull-perf-tests-clusterloader2-kubemark/1185283924603965440/#1:build-log.txt%3A356

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -197,7 +205,7 @@ type etcdMetrics struct {
 	SnapshotSaveTotalDuration measurementutil.HistogramVec `json:"snapshotSaveTotalDuration"`
 	PeerRoundTripTime         measurementutil.HistogramVec `json:"peerRoundTripTime"`
 	WalFsyncDuration          measurementutil.HistogramVec `json:"walFsyncDuration"`
-	MaxDatabaseSize           float64                      `json:"maxDatabaseSize"`
+	MaxDatabaseSize           float64                      `json:"maxDatabaseSizes"`
```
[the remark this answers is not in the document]
alejandrox1: :sweat_smile: thank you for catching this!

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -66,9 +66,13 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig)
 	if err != nil {
 		return nil, err
 	}
-	host, err := util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp())
-	if err != nil {
-		return nil, err
+	hosts := config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(hosts) < 1 {
+		return nil, fmt.Errorf("empty slice: there are no master IPs registered %v", hosts)
```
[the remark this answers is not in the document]
alejandrox1: Ah, this makes sense. thank you

alejandrox1:
/cc @mm4tt 
fixed this PR based on your feedback (thank you!).
could you please take another look?

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -139,13 +147,21 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string) error {
 	for _, sample := range samples {
 		switch sample.Metric[model.MetricNameLabel] {
 		case "etcd_disk_backend_commit_duration_seconds_bucket":
+			e.Lock()
```
wojtek-t: please avoid locking in every case, just do lock before switch and unlock at the end

or even better (which you could do also above), is to opaque the critical section into func and use defer, i.e.
```
func () {
  e.Lock()
  defer e.Unlock()
  switch sample....
}()
```

What you get then is that even if you add return at some point or sth like that, it will still work.

alejandrox1: Thank you very much for your comment @wojtek-t !
Fixed it

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -111,17 +117,29 @@ func (e *etcdMetricsMeasurement) String() string {
 func (e *etcdMetricsMeasurement) startCollecting(host, provider string, interval time.Duration) {
 	e.isRunning = true
 	e.wg.Add(1)
+
+	collectEtcdDatabaseSize := func(host, provider string) error {
```
mm4tt: NIT: There is no need for this function to accept the provider string as it doesn't change within the method.

clusterloader2/pkg/measurement/util/histogram_test.go
```diff
@@ -0,0 +1,114 @@
...
+				},
+			},
+		},
+	}
+
+	for name, test := range tests {
+		for _, sample := range test.samples {
+			ConvertSampleToBucket(sample, test.inputHistogramVec)
+		}
+		if !Equal(test.inputHistogramVec, test.want) {
```
mm4tt: QQ, wouldn't using reflect.DeepEqual work here? If so, there is no need to create the Equal method.

clusterloader2/pkg/measurement/util/histogram_test.go
```diff
@@ -0,0 +1,114 @@
...
+						"le-value": 3,
+					},
+				},
+			},
+		},
+	}
+
+	for name, test := range tests {
+		for _, sample := range test.samples {
+			ConvertSampleToBucket(sample, test.inputHistogramVec)
```
mm4tt: Looks like HistogramVec is initialized exactly the same in all cases, is there a need to do that?
I think you could save a  few lines by removing the inputHistogramVec from the test struct and here do something like:
```
var inputHistogramVec HistogramVec
for _, sample := range test.samples {
  ConvertSampleToBucket(sample, &inputHistogramVec)
}
```

clusterloader2/pkg/measurement/util/histogram_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	for i, v := range *a {
+		if !reflect.DeepEqual(v, (*b)[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+func TestConvertSampleToBucket(t *testing.T) {
```
mm4tt: Nice, thanks for adding the test!

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -136,7 +154,11 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string) error {
 	if err != nil {
 		return err
 	}
-	for _, sample := range samples {
+
+	collectEtcdMetrics := func(sample *model.Sample) {
+		e.Lock()
+		defer e.Unlock()
+
```
mm4tt: Alternatively you could also refactor this to first in switch find address to the HistogramVec you should be modifying and then have a short Lock-Unlock block where you would call `measurementutil.ConvertSampleToBucket(sample, histogramVec)`

clusterloader2/pkg/measurement/util/histogram_test.go
wojtek-t: +1

clusterloader2/pkg/measurement/common/etcd_metrics.go
alejandrox1: i see. I wanted to be as explicit as possible but it makes sense.
I removed the parameters for this method (host as well as provider) since host doesn't change either within the method.

clusterloader2/pkg/measurement/util/histogram_test.go
alejandrox1: ha, it does indeed work. I created the Equal() method while I was working on the unit test to double check I was doing the correct thing. I replaced `Equal()` for `reflect.DeepEqual()` in the latest patch.

clusterloader2/pkg/measurement/util/histogram_test.go
alejandrox1: No need, I don't think. Just wanted to be as explicit as possible. I actually wanted to check if you all thought this was a good test case.

clusterloader2/pkg/measurement/common/etcd_metrics.go
alejandrox1: Done.

alejandrox1:
Thank you for the code reviews @mm4tt @wojtek-t 
I updated this PR, please take a look at your convenience :smile:

mm4tt:
Looks like the e2e tests are failing with:
```
panic: runtime error: invalid memory address or nil pointer dereference
W1030 03:20:51.144] [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x11e7b91]
W1030 03:20:51.145] 
W1030 03:20:51.145] goroutine 12449 [running]:
W1030 03:20:51.145] k8s.io/perf-tests/clusterloader2/pkg/measurement/util.ConvertSampleToBucket(0xc001037be0, 0x0)
W1030 03:20:51.145] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/util/histogram.go:51 +0x141
W1030 03:20:51.146] k8s.io/perf-tests/clusterloader2/pkg/measurement/common.(*etcdMetricsMeasurement).stopAndSummarize.func1(0xc001037be0)
W1030 03:20:51.146] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/etcd_metrics.go:172 +0xb7
W1030 03:20:51.146] k8s.io/perf-tests/clusterloader2/pkg/measurement/common.(*etcdMetricsMeasurement).stopAndSummarize(0xc000828930, 0xc00002c2e0, 0xb, 0x7ffc59eb5759, 0x3, 0x0, 0x0)
W1030 03:20:51.146] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/etcd_metrics.go:176 +0x102
W1030 03:20:51.146] k8s.io/perf-tests/clusterloader2/pkg/measurement/common.(*etcdMetricsMeasurement).Execute(0xc000828930, 0xc00256f900, 0xc005643268, 0x0, 0x0, 0x0, 0x8000102)
W1030 03:20:51.147] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/etcd_metrics.go:89 +0x1c9
W1030 03:20:51.147] k8s.io/perf-tests/clusterloader2/pkg/measurement/common/bundle.execute(0x18404c0, 0xc000828930, 0xc00256f900, 0xd, 0xc005643300, 0x0, 0x0, 0xc0004eaa80)
W1030 03:20:51.147] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/bundle/test_metrics.go:266 +0xf8
W1030 03:20:51.147] k8s.io/perf-tests/clusterloader2/pkg/measurement/common/bundle.(*testMetrics).Execute(0xc0001a6460, 0xc00256f810, 0xb, 0xc0001d9d10, 0xb, 0x1840800, 0xc0001a6460)
W1030 03:20:51.147] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/bundle/test_metrics.go:184 +0xeaa
W1030 03:20:51.148] k8s.io/perf-tests/clusterloader2/pkg/measurement.(*MeasurementManager).Execute(0xc0001120f0, 0xc0001d9d20, 0xb, 0xc0001d9d10, 0xb, 0xc0007f9530, 0x0, 0x0)
```

Could you take a look?

alejandrox1:
awesome, fixed the bug in the latest version of the patch.

mm4tt:
Nice, thanks!

/lgtm
/approve

PR #837, opened 2019-10-07
Enable simple Promethues API responsiveness measurement.

oxddr:
Measure APIResponsivenessPrometheus twice (gather action):

1. First call mimics old APIResponsiveness, but uses new metric instead of the old summary. 
2. Second call uses the new two-level query. 

/ref https://github.com/kubernetes/perf-tests/issues/498

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -723,16 +723,21 @@ steps:
     Method: APIResponsiveness
     Params:
       action: gather
-  - Identifier: APIResponsivenessPrometheus
+  - Identifier: APIResponsivenessPrometheusSimple
     Method: APIResponsivenessPrometheus
     Params:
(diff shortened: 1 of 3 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -718,7 +718,6 @@ steps:
       objectTemplatePath: service.yaml
 
 - name: Collecting measurements
-  measurements:
```
mm4tt: Why are you removing this?

wojtek-t: Seems like a bug - lgtm modulo this one

oxddr: I've erroneously deleted it. Fixed now.

oxddr:
@mm4tt, PTAL

PR #838, opened 2019-10-08
Remove resource usage constraints.

mborsz:
As discussed offline, maintaining resource usage constraints seems to be unnecessary cost (we can't remember any case when this has detected a real issue, while we quickly found many cases when resource consumption growth was caused by new feature/version of component).

I propose removing the limits at all.

/assign @wojtek-t
/assign @mm4tt 
/cc @krzysied


clusterloader2/testing/density/100_nodes/override.yaml
```diff
@@ -1 +1 @@
-DENSITY_RESOURCE_CONSTRAINTS_FILE: ./100_nodes/constraints.yaml
+# TODO(mborsz): Remove this file once we stop using this file in test-infra.
```
wojtek-t: Can we remove it from test-infra first?

mborsz: Removed PTAL

PR #839, opened 2019-10-08
Add system pod metrics measurement in clusterloader

pierewoj:
In this commit, "system pod metrics" measurement is added. Its purpose
is to gather data about system pod restarts. This measurement will be
invoked on every test run, but it's logic is no-op by default.

It will be enabled by setting "systemPodMetricsEnabled: true" in
TestMetrics measurement. This is to be done in follow-up commits.

Testing:
* ran storage suite against existing cluster
* verified that SystemPodMetrics artifact is being created
with proper content if flag is set to true
* verified that no such artifact is created when this flag is absent

### DIFF
--- a/clusterloader2/pkg/measurement/common/system_pod_metrics.go
+++ b/clusterloader2/pkg/measurement/common/system_pod_metrics.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	systemPodMetricsName            = "SystemPodMetrics"
+	systemNamespace                 = "kube-system"
+	systemPodMetricsEnabledFlagName = "systemPodMetricsEnabled"
+)
+
+func init() {
+	if err := measurement.Register(systemPodMetricsName, createSystemPodMetricsMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", systemPodMetricsName, err)
+	}
+}
+
+func createSystemPodMetricsMeasurement() measurement.Measurement {
+	return &systemPodMetricsMeasurement{}
+}
+
+// Gathers metrics for system pods, right now it only gathers container restart counts
+type systemPodMetricsMeasurement struct{}
+
+type containerMetrics struct {
+	Name         string `json:"name"`
+	RestartCount int32  `json:"restartCount"`
+}
+
+type podMetrics struct {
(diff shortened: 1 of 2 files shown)


oxddr:
Please fix boilerplate. 

It might be worth adding this check to pre-push hook in your git repo to detect such problem early. Below content of my pre-push hook. 

```
#!/usr/bin/env bash

set -o errexit
set -o nounset
set -o pipefail

GIT_ROOT=$(dirname "${BASH_SOURCE[0]}")/../../

cd "${GIT_ROOT}"

./verify/verify-boilerplate.sh
./verify/verify-dashboard-format.sh
./verify/verify-flags-underscore.py
./verify/verify-gofmt.sh
./verify/verify-golint.sh
./verify/test.sh
```

clusterloader2/pkg/measurement/common/system_pod_metris.go
```diff
@@ -0,0 +1,127 @@
...
+	if err := measurement.Register(systemPodMetricsName, createSystemPodMetricsMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", systemPodMetricsName, err)
+	}
+}
+
+func createSystemPodMetricsMeasurement() measurement.Measurement {
+	return &systemPodMetricsMeasurement{}
+}
+
+type systemPodMetricsMeasurement struct{}
```
oxddr: Please document what this measurement does.

clusterloader2/pkg/measurement/common/system_pod_metris.go
```diff
@@ -0,0 +1,127 @@
+/*
```
oxddr: Fix typo in the name of this file (metris -> metrics)

clusterloader2/pkg/measurement/common/system_pod_metris.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+type podMetrics struct {
+	Name       string             `json:"name"`
+	Containers []containerMetrics `json:"containers"`
+}
+
+type systemPodsMetrics struct {
+	Pods []podMetrics `json:"pods"`
+}
```
oxddr: Do we plan to extend this measurement with other metrics in the future? Unless the answer is "yes" I'd use more specific term than metrics (e.g. "restarts").

clusterloader2/pkg/measurement/common/system_pod_metris.go
```diff
@@ -0,0 +1,127 @@
...
+	summary := measurement.CreateSummary(systemPodMetricsName, "json", content)
+	return []measurement.Summary{summary}, nil
+}
+
+func systemPodMetricsEnabled(config *measurement.MeasurementConfig) bool {
+	systemPodMetricsEnabled, err := util.GetBoolOrDefault(config.Params, systemPodMetricsEnabledFlagName, false)
+	if err != nil {
+		klog.Errorf("failed to check %v flag: %v", systemPodMetricsEnabledFlagName, err)
+		return false
+	}
```
oxddr: I'd rather not swallow the error and would return it. Fail fast!

pierewoj: makes sense

clusterloader2/pkg/measurement/common/system_pod_metris.go
pierewoj: Discussed offline

clusterloader2/pkg/measurement/common/system_pod_metrics.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := measurement.Register(systemPodMetricsName, createSystemPodMetricsMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", systemPodMetricsName, err)
+	}
+}
+
+func createSystemPodMetricsMeasurement() measurement.Measurement {
+	return &systemPodMetricsMeasurement{}
+}
+
+// Gathers metrics for system pods, for instance number of container restarts
```
oxddr: Please be specific that we **only** gather restarts now.

pierewoj: Fixed

oxddr:
Please assign someone with approval bit, I don't have it.

PR #842, opened 2019-10-09
Subtract initial restart counts when calculating system pod metrics

pierewoj:
Before this change, system pod measurement gathered metrics only at the
end of each test. This means that the restart count in the present in
the summary was equal to the restart count since each given container
was created (this is usually during cluster creation). This means that
we could have high restart count metric for some test even though no
restart happened during that test, which could be misleading.

To fix this, in this commit we add listing system pods before the test
(in measurement "start" action). Then in "gather" action, initial
restart counts are subtracted from the actual restart counts. In effect,
restart count present in the summary is the number of restarts that
occured during execution of this test.

Testing:
* ran storage suite locally
* verified SystemPodMetrics summary is being created
* verified that test fails if we execute gather without executing start
action first
* unit tests for the restart count subtraction logic

### DIFF
--- a/clusterloader2/pkg/measurement/common/system_pod_metrics.go
+++ b/clusterloader2/pkg/measurement/common/system_pod_metrics.go
@@ -17,6 +17,8 @@ limitations under the License.
 package common
 
 import (
+	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


krzysied:
Good job!
The code looks good to me. 
Nit: The one thing that i would change is the way of passing `systemPodsMetrics`. Sometimes it is by value, sometimes by reference... Maybe we can mostly use `*systemPodsMetrics`?

pierewoj:
@krzysied 
Yeah, makes sense - I've changed it to be a little bit more consistent.

clusterloader2/pkg/measurement/common/system_pod_metrics.go
```diff
@@ -111,7 +168,17 @@ func extractMetrics(lst *v1.PodList) systemPodsMetrics {
 		}
 		metrics.Pods = append(metrics.Pods, podMetrics)
 	}
-	return metrics
+	return &metrics
+}
+
+func buildSummary(podMetrics *systemPodsMetrics) ([]measurement.Summary, error) {
+	content, err := util.PrettyPrintJSON(&podMetrics)
```
krzysied: &podMetrics == **systemPodsMetrics?

pierewoj: seems like json encoder will deference arbitrary number of pointers recirsively (ptrEncoder executes typeEncoder on the value that pointer points to...)

https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L432
https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L874-L885

Anyway it is not readable and looks like a bug, I'll fix it

PR #845, opened 2019-10-09
Refactor master-ip manifests

mborsz:
Refactoring the master-ip manifests to:
* share logic between kubemark and gce ('default')
* add support for multiple apiservers in gce (reusing the logic that already worked in kubemark)
/assign @mm4tt


mm4tt:

(To test perf-tests presubmits using boskos)

clusterloader2/go.mod
```diff
@@ -0,0 +1,69 @@
+module k8s.io/perf-tests/clusterloader2
```
mm4tt: Remove this file?

mm4tt:
As discussed personally, please make sure that no regression is introduced in GKE

mm4tt:
An unintended side effect of this change is that our kubemark tests (e.g. presubmit) stopped scrapping etcd. Please fix

mborsz:
See https://github.com/kubernetes/test-infra/pull/14911

clusterloader2/pkg/prometheus/manifests/default/kube-scheduler-service.yaml
```diff
@@ -1,15 +0,0 @@
-apiVersion: v1
```
mm4tt: I think that this approach has the benefit of working both in GCE and GKE while the one from this PR won't work in GKE when it comes to system pods like kube-scheduler or kube-controller-manager, right?

mm4tt:
/lgtm
/approve

Once this gets submitted could you check whether nothing got broken in our grafana dashboards? I guess some of the target names may have changed which may break some graphs.

PR #849, opened 2019-10-10
Enable collection of system pod metrics if flag is set

pierewoj:
Flag is already enabled by enable_system_pod_metrics.yaml override in
several tests (perf-tests presubmits for example, see test-infra).

In this commit, system pod metrics measurement is being enabled if
$ENABLE_SYSTEM_POD_METRICS flag is set.
This will result in all system pods being listed twice (first in start
action, then in gather action) for all tests that use this override.

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -18,6 +18,7 @@
 {{$MIN_SATURATION_PODS_TIMEOUT := 180}}
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/density/config.yaml
```diff
@@ -72,6 +73,9 @@ steps:
       action: start
       nodeMode: {{$NODE_MODE}}
       resourceConstraints: {{$DENSITY_RESOURCE_CONSTRAINTS_FILE}}
+      {{if $ENABLE_SYSTEM_POD_METRICS}}
+      systemPodMetricsEnabled: true
+      {{end}}
```
krzysied: Just out of the curiosity, why can't we have something like:
```
systemPodMetricsEnabled: {{$ENABLE_SYSTEM_POD_METRICS}}
```

pierewoj: The difference is how we will handle ENABLE_SYSTEM_POD_METRICS=false case.

In this PR we won't set systemPodMetricsEnabled at all, in your version we'll set systemPodMetricsEnabled=false.

They are handled in the same way so it should be equivalent. I actually like your version better, will change.

PR #850, opened 2019-10-10
Reverting to the state at 2bcba1 on release-1.13 branch

oxddr:
The nearest commit to when v1.13.0 tag was added in k/k repo. 

/assign @mm4tt @wojtek-t


wojtek-t:
/hold

I'm not sure it's what we want. It may not even have working ClusterLoader2...

I'm wondering if maybe tests is the only thing that we want to revert. @mm4tt - WDYT?

mm4tt:
+1 to reverting just configs for 1.13 and maybe 1.14

For 1.15 and forward we can most likely revert the whole branch.

mm4tt:
As discussed personally, given that the tests currently pass on gce 1.13 using the legacy testconfigs, I think it would be enough to just switch the configs with the legacy configs for this branch.

Issue #797, opened 2019-09-13
Figure out a mechanism to avoid breaking tests in k/k repo


wojtek-t:
Yesterday merging https://github.com/kubernetes/perf-tests/pull/788 broke tests in k/k repo in for 1.16 release branch, because it was depending on https://github.com/kubernetes/kubernetes/pull/82579 that was merged only at head.

Some time ago, we set up presubmits in perf-tests repo, to avoid breaking k/k repo. But that only works for master branch of k/k. We should figure out a way to protect release branches too.

I think there are two main options:
1. add more presubmits (i.e. for release branches)
 -though this is relatively expensive)

2. version perf-tests repo
- for every single release branch of Kubernetes, we create a corresponding branch in perf-tests repo and configure presubmits/postsubmits for k/k repo to use clustersloader from that brach
- that has also significant advantage of no need to explicitly version tests for different branches

I'm personally leaning towards the second option. Thoughts?
@kubernetes/sig-scalability-bugs

wojtek-t:
@mm4tt

mm4tt:
+1 to 2. as it would solve a few other issues in addition to presubmits/breaking k/k repo

Would be nice if we had some automation behind creating new branches in k/perf-tests, but we can start with doing that manually.
Once we have branches for each release it should be quite straightforward to use the `fork-per-release-replacements` feature to configure our CI/CD jobs to checkout proper branch of k8s.io/perf-tests instead of always checking out master. 

So, if I'm not missing anything, this should be actually pretty simple to implement.

oxddr:
Option 1 seems to be partial subset of the option 2 (modulo additional branches in k/perf-tests). I'd lean to the option 1. Maintaining a single branch is a simple mental model - one need to remember that we need to be compatible with least supported version. Incompatible changes seem to be fairly rare beasts. 

A problem I see for the option 2 is that not having an automation and processes from day 0 will put us in the same situation sooner or later - we're humans after all ;)

@mm4tt out of curiosity: what are those other things that having a branch parity between k/k and k/perf-tests solves?

krzysied:
+1 for option 2.
We can remove legacy test directories, when there is a versioning.

mm4tt:
/remove-lifecycle stale

This is in progress and very close to completion.

wojtek-t:
We have already versioned perf-tests repo. There is basically one more follow up on it: https://github.com/kubernetes/test-infra/issues/17097, but it can be closed.

PR #854, opened 2019-10-10
Reverting state at release1.14 branch

oxddr:
1. Revert the state to the same commit as 1.15
2. Revert {load,density}/config.yaml to cd2363 (around when v1.14.0 was added to k/k)

ref https://github.com/kubernetes/perf-tests/issues/797


oxddr:
/assign @wojtek-t 

I've just tested this branch against k/k release-1.14 and it worked fine.

PR #857, opened 2019-10-11
Add support for viewing system pod metrics in perfdash

pierewoj:
For each build, perfdash will display maximum restart count per each
container name (for example metadata-proxy, kube-apiserver,
etcd-container etc).

In addition a test was added and makefile was changed to run tests
before run.

Testing:
- started perfdash locally

### DIFF
--- a/perfdash/parser.go
+++ b/perfdash/parser.go
@@ -366,3 +366,47 @@ func parseHistogramMetric(metricName string) func(data []byte, buildNumber int, 
 		}
 	}
 }
+
+func parseSystemPodMetrics(data []byte, buildNumber int, testResult *BuildData) {
+	type containerMetrics struct {
+		Name         string `json:"name"`
+		RestartCount int32  `json:"restartCount"`
+	}
+
+	type podMetrics struct {
+		Name       string             `json:"name"`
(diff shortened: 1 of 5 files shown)


perfdash/parser.go
```diff
@@ -366,3 +366,46 @@ func parseHistogramMetric(metricName string) func(data []byte, buildNumber int,
...
+
+	restartCounts := make(map[string]int32)
+
+	for _, pod := range obj.Pods {
+		for _, container := range pod.Containers {
+			restartCounts[container.Name] = container.RestartCount
+		}
+	}
+
+	perfData := perftype.DataItem{Unit: "", Labels: map[string]string{
```
krzysied: This formatting looks weird...

perfdash/parser.go
```diff
@@ -366,3 +366,46 @@ func parseHistogramMetric(metricName string) func(data []byte, buildNumber int,
...
+		"RestartCount": "RestartCount",
+	}, Data: make(map[string]float64)}
+	for container, cnt := range restartCounts {
+		curMaxCnt := float64(0)
```
krzysied: IIUC we are copying restartCounts to perfData.Data. Is it possible to have container as a key more than once?

Issue #790, opened 2019-09-12
Calculating num replicas for DaemonSets should take into account affinity


mm4tt:
/area clusterloader
/label help-wanted

mozhuli:
I can help it, Can I assign it to me? @mm4tt

mm4tt:
Sure, would be great if you could take it :)

The purpose of this issue is to address these TODOs:
* https://github.com/kubernetes/perf-tests/blob/7d226a65336414a2d23141d6cc09f6ac4e4d09f3/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go#L294
* https://github.com/kubernetes/perf-tests/blob/7d226a65336414a2d23141d6cc09f6ac4e4d09f3/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go#L342

wojtek-t:
This is now fixed.

PR #861, opened 2019-10-15
Calculating num replicas for DaemonSets take into account affinity

mozhuli:
Take into account nodeAffinity when calculating num replicas for DaemonSets
Ref #790 

/assign @mm4tt

### DIFF
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	clientset "k8s.io/client-go/kubernetes"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
 )
 
@@ -290,8 +291,7 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
 		}
 		return 0, nil
 	case *appsv1.DaemonSet:
-		// TODO(#790): In addition to nodeSelector the affinity should be also taken into account
-		return getNumSchedulableNodesMatchingSelector(c, typed.Spec.Template.Spec.NodeSelector)
+		return getNumSchedulableNodesMatchingNodeSelectorAndNodeAffinity(c, typed.Spec.Template.Spec.NodeSelector, typed.Spec.Template.Spec.Affinity)
 	case *batch.Job:
 		if typed.Spec.Parallelism != nil {
 			return *typed.Spec.Parallelism, nil
@@ -302,20 +302,24 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
 	}
 }
 
-// getNumSchedulableNodesMatchingSelector returns the number of schedulable nodes matching the provided selector.
-func getNumSchedulableNodesMatchingSelector(c clientset.Interface, nodeSelector map[string]string) (int32, error) {
+// getNumSchedulableNodesMatchingNodeSelectorAndNodeAffinity returns the number of schedulable nodes matching both nodeSelector and NodeAffinity.
+func getNumSchedulableNodesMatchingNodeSelectorAndNodeAffinity(c clientset.Interface, nodeSelector map[string]string, affinity *corev1.Affinity) (int32, error) {
 	selector, err := metav1.LabelSelectorAsSelector(metav1.SetAsLabelSelector(nodeSelector))
 	if err != nil {
 		return 0, err
 	}
 	listOpts := metav1.ListOptions{LabelSelector: selector.String()}
-	list, err := client.ListNodesWithOptions(c, listOpts)
+	nodeList, err := client.ListNodesWithOptions(c, listOpts)
 	if err != nil {
 		return 0, err
 	}
 	var numSchedulableNodes int32
-	for _, node := range list {
-		if !node.Spec.Unschedulable {
+	for i := range nodeList {
+		matched, err := podMatchesNodeAffinity(affinity, &nodeList[i])
+		if err != nil {
+			return 0, err
+		}
+		if !nodeList[i].Spec.Unschedulable && matched {
 			numSchedulableNodes++
 		}
 	}
@@ -338,8 +342,11 @@ func tryAcquireReplicasFromUnstructuredSpec(c clientset.Interface, spec map[stri
 		if err != nil {
 			return 0, err
 		}
-		// TODO(#790): In addition to nodeSelector the affinity should be also taken into account
-		return getNumSchedulableNodesMatchingSelector(c, nodeSelector)
+		affinity, err := getDaemonSetAffinityFromUnstructuredSpec(spec)
+		if err != nil {
+			return 0, err
+		}
+		return getNumSchedulableNodesMatchingNodeSelectorAndNodeAffinity(c, nodeSelector, affinity)
 	case "Job":
 		replicas, found, err := unstructured.NestedInt64(spec, "parallelism")
 		if err != nil {
@@ -372,6 +379,24 @@ func getDaemonSetNodeSelectorFromUnstructuredSpec(spec map[string]interface{}) (
 	}
 	nodeSelector, found, err := unstructured.NestedStringMap(podSpec, "nodeSelector")
 	return nodeSelector, err
+}
+
+func getDaemonSetAffinityFromUnstructuredSpec(spec map[string]interface{}) (*corev1.Affinity, error) {
+	template, found, err := unstructured.NestedMap(spec, "template")
+	if err != nil || !found {
+		return nil, err
+	}
+	podSpec, found, err := unstructured.NestedMap(template, "spec")
+	if err != nil || !found {
+		return nil, err
+	}
+	unstructuredAffinity, found, err := unstructured.NestedMap(podSpec, "affinity")
+	if err != nil || !found {
+		return nil, err
+	}
+	var affinity *corev1.Affinity
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredAffinity, affinity)
+	return affinity, err
 }
 
 // IsEqualRuntimeObjectsSpec returns true if given runtime objects have identical specs.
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -412,3 +394,53 @@ func GetNumObjectsMatchingSelector(c dynamic.Interface, namespace string, resour
...
+		nodeAffinity := affinity.NodeAffinity
+		// if no required NodeAffinity requirements, will do no-op, means select all nodes.
+		if nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+			return true
+		}
+
+		// Match node selector for requiredDuringSchedulingIgnoredDuringExecution.
+		if nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+			nodeSelectorTerms := nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+			nodeAffinityMatches = nodeAffinityMatches && nodeMatchesNodeSelectorTerms(node, nodeSelectorTerms)
```
krzysied: Why not something like this?
```
if !nodeMatchesNodeSelectorTerms(node, nodeSelectorTerms) {
      return false
}
```

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -412,3 +394,53 @@ func GetNumObjectsMatchingSelector(c dynamic.Interface, namespace string, resour
...
+
+	}
+	return nodeAffinityMatches
+}
+
+// nodeMatchesNodeSelectorTerms checks if a node's labels satisfy a list of node selector terms,
+// terms are ORed, and an empty list of terms will match nothing.
+func nodeMatchesNodeSelectorTerms(node corev1.Node, nodeSelectorTerms []corev1.NodeSelectorTerm) bool {
+	for _, req := range nodeSelectorTerms {
+		nodeSelector, err := v1helper.NodeSelectorRequirementsAsSelector(req.MatchExpressions)
```
krzysied: Why are we returning `false` instead of propagating error?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -412,3 +394,53 @@ func GetNumObjectsMatchingSelector(c dynamic.Interface, namespace string, resour
...
+	affinity := podSpec.Affinity
+	if affinity != nil && affinity.NodeAffinity != nil {
+		nodeAffinity := affinity.NodeAffinity
+		// if no required NodeAffinity requirements, will do no-op, means select all nodes.
+		if nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+			return true
+		}
+
+		// Match node selector for requiredDuringSchedulingIgnoredDuringExecution.
+		if nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
```
krzysied: Isn't this check already done above?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -302,20 +302,15 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
...
+func getNumSchedulableNodesMatchingSelector(c clientset.Interface, podSpec corev1.PodSpec) (int32, error) {
+	nodes, err := client.ListNodes(c)
 	if err != nil {
 		return 0, err
 	}
 	var numSchedulableNodes int32
-	for _, node := range list {
-		if !node.Spec.Unschedulable {
+	for _, node := range nodes {
+		if podMatchesNodeLabels(podSpec, node) && !node.Spec.Unschedulable {
```
krzysied: nit: I would change the order of the conditions to utilize power of lazy evaluation :)

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -302,20 +302,15 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
...
-func getNumSchedulableNodesMatchingSelector(c clientset.Interface, nodeSelector map[string]string) (int32, error) {
-	selector, err := metav1.LabelSelectorAsSelector(metav1.SetAsLabelSelector(nodeSelector))
-	if err != nil {
-		return 0, err
-	}
-	listOpts := metav1.ListOptions{LabelSelector: selector.String()}
-	list, err := client.ListNodesWithOptions(c, listOpts)
+// getNumSchedulableNodesMatchingSelector returns the number of schedulable nodes matching the provided pod spec.
+func getNumSchedulableNodesMatchingSelector(c clientset.Interface, podSpec corev1.PodSpec) (int32, error) {
+	nodes, err := client.ListNodes(c)
```
krzysied: nit: Why to not use previous version of listing? We would get all nodes that satisfy nodeSelector and then we need to only check affinity.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
mozhuli: all have done except `RequiredDuringSchedulingRequiredDuringExecution`, since it has not implemented

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
mozhuli: here we just care about pod scheduled on the node or not,  we don't need to know the error info. and the error just is `not a valid node selector operator`.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -302,20 +302,20 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
 	}
 }
 
-// getNumSchedulableNodesMatchingSelector returns the number of schedulable nodes matching the provided selector.
-func getNumSchedulableNodesMatchingSelector(c clientset.Interface, nodeSelector map[string]string) (int32, error) {
-	selector, err := metav1.LabelSelectorAsSelector(metav1.SetAsLabelSelector(nodeSelector))
+// getNumSchedulableNodesMatchingSelector returns the number of schedulable nodes matching the provided pod spec.
```
mm4tt: In that case, shouldn't we also update the func name to match the description?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -334,12 +334,12 @@ func getReplicasFromUnstrutured(c clientset.Interface, obj *unstructured.Unstruc
 func tryAcquireReplicasFromUnstructuredSpec(c clientset.Interface, spec map[string]interface{}, kind string) (int32, error) {
 	switch kind {
 	case "DaemonSet":
-		nodeSelector, err := getDaemonSetNodeSelectorFromUnstructuredSpec(spec)
+		var podSpec corev1.PodSpec
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(spec, &podSpec)
 		if err != nil {
-			return 0, err
+			return 0, fmt.Errorf("try to pod spec failed, %v", err)
```
mm4tt: NIT: casting to PodSpec failed?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
mm4tt: I think that Krzysiek's point was that there is no need to check for != nil here. nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution has to bee not nil here as otherwise you would exit from the fun in line 416

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
mm4tt: This shouldn't really happen assuming the DaemonSet's podSpec is correct, right?
In that case, I believe returning an error is what we should do.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
krzysied: > I think that Krzysiek's point was that there is no need to check for != nil here.

That's right. :+1:

krzysied:
@mozhuli Thanks for resolving all my comments. Great job!
I see that @mm4tt has some additional comments, so I guess he will lgtm the PR when it's finished.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -290,8 +291,10 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
 		}
 		return 0, nil
 	case *appsv1.DaemonSet:
-		// TODO(#790): In addition to nodeSelector the affinity should be also taken into account
-		return getNumSchedulableNodesMatchingSelector(c, typed.Spec.Template.Spec.NodeSelector)
+		if typed.Spec.Template.Spec.Affinity != nil && typed.Spec.Template.Spec.Affinity.NodeAffinity != nil {
```
wojtek-t: nit - I would prefer passing types.Spec.Template.Spec.Affinity here and do the nil handling in the getNumSchedu... function

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -302,20 +305,24 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (in
...
 	if err != nil {
 		return 0, err
 	}
 	var numSchedulableNodes int32
-	for _, node := range list {
-		if !node.Spec.Unschedulable {
+	for _, node := range nodes {
+		matched, err := podMatchesNodeAffinity(nodeAffinity, node)
+		if err != nil {
+			return 0, nil
```
wojtek-t: This doesn't seem correct to me - shouldn't we return err here?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
wojtek-t: This still isn't resolved - please remove this check - after if from line 452, we already know it's not nil.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -412,3 +437,48 @@ func GetNumObjectsMatchingSelector(c dynamic.Interface, namespace string, resour
 	err := client.RetryWithExponentialBackOff(client.RetryFunction(listFunc))
 	return numObjects, err
 }
+
+// The pod can only schedule onto nodes that satisfy requirements in NodeAffinity.
+func podMatchesNodeAffinity(affinity *corev1.Affinity, node corev1.Node) (bool, error) {
```
wojtek-t: For performance reasons, we don't want to pass corev1.Node (because that is resulting in making a copy of the object).
Instead you want to pass *corev1.Node

wojtek-t: Same everywhere below

PR #862, opened 2019-10-15
Add an option to CL2 to preload given list of images.

oxddr:
/assign @mm4tt 

Please let me know if more documentation should be added, either here or in the code.

### DIFF
--- a/clusterloader2/pkg/imagepreload/imagepreload.go
+++ b/clusterloader2/pkg/imagepreload/imagepreload.go
@@ -1,0 +1,181 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package imagepreload
+
+import (
+	"strings"
+	"sync"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/flags"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
+)
+
+const (
+	informerTimeout = time.Minute
+	manifest        = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml"
+	namespace       = "preload"
+	daemonsetName   = "preload"
+	pollingInterval = 5 * time.Second
+	// TODO(oxddr): verify whether 5 minutes is a sufficient timeout
+	pollingTimeout = 5 * time.Minute
+)
+
+var images []string
+
+func init() {
+	flags.StringSliceEnvVar(&images, "node-preload-images", "NODE_PRELOAD_IMAGES", []string{}, "List of images to preload on each node in the test cluster before executing tests")
+}
+
+type controller struct {
+	// lock for controlling access to doneNodes in PreloadImages
+	lock sync.Mutex
+
+	config          *config.ClusterLoaderConfig
+	framework       *framework.Framework
+	templateMapping map[string]interface{}
+	images          []string
+}
+
+// Setup ensures every node in cluster preloads given list of images before starting tests.
+// It does it by creating a Daemonset that call "docker pull" and awaits for Node object to be updated.
+// As a side-effect of the image preloading, size of Node objects is increased.
+//
+// Preloading is skipped in kubemark or if no images have been specified.
+func Setup(conf *config.ClusterLoaderConfig, f *framework.Framework) error {
+	mapping, err := config.GetMapping(conf)
+	if err != nil {
+		return err
+	}
+
+	ctl := &controller{
+		config:          conf,
+		framework:       f,
+		templateMapping: mapping,
+		images:          images,
+	}
+	return ctl.PreloadImages()
+}
+
+func (c *controller) PreloadImages() error {
+	if len(images) == 0 {
+		klog.Warning("No images specified. Skipping image preloading")
+		return nil
+	}
+	if c.config.ClusterConfig.Provider == "kubemark" {
+		klog.Warning("Image preloading is disabled in kubemark")
+		return nil
+	}
+
+	kclient := c.framework.GetClientSets().GetClient()
+
+	doneNodes := make(map[string]bool)
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	nodeInformer := informer.NewInformer(
+		kclient,
+		"nodes",
+		util.NewObjectSelector(),
+		func(old, new interface{}) { c.checkNode(new, doneNodes) })
+	if err := informer.StartAndSync(nodeInformer, stopCh, informerTimeout); err != nil {
+		return err
+	}
+
+	klog.Infof("Creating namespace %s...", namespace)
+	if err := client.CreateNamespace(kclient, namespace); err != nil {
+		return err
+	}
+
+	klog.Info("Creating daemonset to preload images...")
+	c.templateMapping["Images"] = c.images
+	if err := c.framework.ApplyTemplatedManifests(manifest, c.templateMapping); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/imagepreload/manifests/example.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: apps/v1
```
mm4tt: Why do we need this example?

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,161 @@
...
+	return ctl.PreloadImages()
+}
+
+func (i *controller) PreloadImages() error {
+	if len(*images) == 0 {
+		klog.Warning("No images specified. Skipping image preloading")
+		return nil
+	}
+	if i.config.ClusterConfig.Provider == "kubemark" {
+		klog.Warning("Image preloading is disabled in kubemark")
```
mm4tt: Oh, this is unfortunate as it will make kubemark tests to differ even more than they do today.
I think we haven't thought of that...
@wojtek-t, any thoughts?

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	informerTimeout = time.Minute
+	manifest        = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml"
+	namespace       = "preload"
+	pollingInterval = 5 * time.Second
+	pollingTimeout  = 5 * time.Minute
+)
+
+var images = pflag.StringSlice("node-preload-images", []string{}, "List of images to preload on each node in the test cluster before executing tests")
```
mm4tt: Can we make it an env var so it can be easily set for all tests via presets?

Can be addressed in a follow up PR.

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func checkNode(obj interface{}, done map[string]bool) {
+	if obj == nil {
+		return
+	}
+
+	node := obj.(*v1.Node)
+	if hasPreloadedImages(node) {
+		done[node.Name] = true
```
mm4tt: Is it a documented and guaranteed feature of an informer that event handlers are called from a single goroutine and hence don't have to be safe thread? If not, and it's only an implementation detail, then I think adding a mutex here wouldn't hurt.

clusterloader2/pkg/imagepreload/imagepreload.go
wojtek-t: +1
at least a TODO here

clusterloader2/pkg/imagepreload/imagepreload.go
wojtek-t: I think we should solve it directly in kubemark.
Basically, how I would imagine this is to:
- on startup of hollow-kubelet, it list images from the node it's running on (it requires passing additional flags to it so that it can do that)
- it creates fakeDocker client with with that list of images
- and uses it for its whole lifetime

but that requires handling separate in kubemark - shouldn't be part of this PR.

wojtek-t: I think we should file an issue for it anyway - this will be useful independently from this PR.

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if i.config.ClusterConfig.Provider == "kubemark" {
+		klog.Warning("Image preloading is disabled in kubemark")
+		return nil
+	}
+
+	clusterSize := i.config.ClusterConfig.Nodes
+	kclient := i.framework.GetClientSets().GetClient()
+
+	// 1. setup node watcher
```
wojtek-t: Remove - we shouldn't add comment saying what the code is doing when the code is obvious - which is the case here.

Same below.

clusterloader2/pkg/imagepreload/imagepreload.go
wojtek-t: We should handle it with mutex. Please add a thread-safe structure to keep this info.

clusterloader2/pkg/imagepreload/manifests/example.yaml
wojtek-t: Agree - please remove.

clusterloader2/pkg/imagepreload/imagepreload.go
oxddr: Done.

clusterloader2/pkg/imagepreload/manifests/example.yaml
oxddr: Leftover from testing. Removed.

mm4tt:

(to test boskos presubmits)

oxddr:
PTAL.

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,166 @@
...
+	flags.StringSliceEnvVar(&images, "node-preload-images", "NODE_PRELOAD_IMAGES", []string{}, "List of images to preload on each node in the test cluster before executing tests")
+}
+
+type controller struct {
+	config          *config.ClusterLoaderConfig
+	framework       *framework.Framework
+	templateMapping map[string]interface{}
+	images          []string
+
+	lock sync.Mutex
```
wojtek-t: nit: move lock to the top? Or above the variables it's protecting.

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,166 @@
...
+	if len(images) == 0 {
+		klog.Warning("No images specified. Skipping image preloading")
+		return nil
+	}
+	if c.config.ClusterConfig.Provider == "kubemark" {
+		klog.Warning("Image preloading is disabled in kubemark")
+		return nil
+	}
+
+	clusterSize := c.config.ClusterConfig.Nodes
```
wojtek-t: Instead of relying on the config, can we use the function that we already have:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go#L266

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,166 @@
...
+	kclient := c.framework.GetClientSets().GetClient()
+
+	doneNodes := make(map[string]bool)
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	ni := informer.NewInformer(
```
wojtek-t: nit: Call it nodeInformer - short names are good, but readable are better :)

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	klog.Info("Creating daemonset to preload images...")
+	c.templateMapping["Images"] = c.images
+	if err := c.framework.ApplyTemplatedManifests(manifest, c.templateMapping); err != nil {
+		return err
+	}
+
+	klog.Info("Waiting for Node objects to be updated...")
+	if err := wait.Poll(pollingInterval, pollingTimeout, func() (bool, error) {
```
wojtek-t: Are we sure pollingtimeout = 5m will be enough?

oxddr: We use this value in other places. Hard to tell to be honest, but I'd suspect this to be fine. We can always bump it if needed.

clusterloader2/pkg/imagepreload/imagepreload.go
oxddr: I am not sure how this is supposed to work for nodes.

wojtek-t: I was thinking about passing the DaemonSet you are creating to this function.

clusterloader2/pkg/imagepreload/imagepreload.go
wojtek-t: What places? Here we wait until N pods will download M images - I don't think we have precedent for that anywhere.
At least add a TODO to validate this value.

clusterloader2/pkg/imagepreload/imagepreload.go
oxddr: I am not sure what the benefit of doing it this way.

clusterloader2/pkg/imagepreload/imagepreload.go
oxddr: For that specifically defined task no - we don't have. But in other places (e.g. Prometheus). I'll add a note on that.

mm4tt:
LGTM - /lgtm

I see that there are two open comments created by Wojtek, so leaving the `approve` for him

clusterloader2/pkg/imagepreload/imagepreload.go
wojtek-t: The benefit is that this way it may be flake.
The first example that comes to my mind is that we are accepting the fact that some small percentage of nodes won't come up (and the cluster is still treated as healthy).
In this case, this will timeout, because if 1 node is unhealthy (or even not registered), it will never preload anything. But the number from Config still contains it.

oxddr: That sounds fine to me, that we timeout here if some nodes won't become healthy - in the name of the principle, that we should fail fast.

wojtek-t: My point is that if we would use what I'm suggesting, we wouldn't have to timeout, because we would know for how many pods we should wait.

wojtek-t: So if you don't want to do that, at the very least it requires a TODO.

PR #870, opened 2019-10-23
Introduce flags for restart count checks

pierewoj:
It is no longer used anywhere, so it should be safe (see
clusterloader2/docs/experiments.md). The list of containes was
formulated based on recent gce-5000 and kubemark-5000 test runs.


clusterloader2/testing/experiments/enable_restart_count_check.yaml
```diff
@@ -0,0 +1,2 @@
+ENABLE_RESTART_COUNT_CHECK: true
+IGNORE_RESTARTS_OF: prometheus-to-sd-exporter,metadata-proxy,kube-proxy,fluentd-gcp,kubernetes-dashboard,kube-scheduler,kube-controller-manager
```
mborsz: I think this list should be per test type, like:
* gce 5000: with disabled fluentd-gcp
* kubemark: kube-scheduler, kube-controller-manager

This way we will be able to detect if kube-scheduler is restarting in gce 5000 (which would be a regression) even before we debug this in kubemark

clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
```diff
@@ -0,0 +1 @@
+IGNORE_RESTARTS_OF: prometheus-to-sd-exporter,metadata-proxy,kube-proxy,fluentd-gcp
```
mm4tt: I would like to avoid situation where we add something here and it stays as ignored forever.

Could you add (in a comment) a sentence about each container (why it's being ignored) and/or link to github issue?

PR #876, opened 2019-10-25
Enable restart count validation in clusterloader

pierewoj:
In this change, restart count validation is implemented. Note that it
will only affect tests, which have this feature enabled (by using
enable_restart_count_check.yaml override).

Testing:
* ran kubemark test locally using clusterloader, cluster running on GCE

/hold 
need to finish testing locally


pierewoj:
comments fixed, tested load/test 

/hold cancel

PR #880, opened 2019-10-30
Fix some apiserver graphs.

mm4tt:
There has been some recent refactoring that changed Prometheus target names and accidentally broke some graphs.


pierewoj:
Did you test this somehow?

Regarding commit message, s/THere/There/

Also, I see that there are many different commit message styles in this repo, what about trying to stick to some? It makes going through the git log much easier, I personally like https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53 

Otherwise lgtm

mm4tt:
> Did you test this somehow?
> 
> Regarding commit message, s/THere/There/
> 
> Also, I see that there are many different commit message styles in this repo, what about trying to stick to some? It makes going through the git log much easier, I personally like https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
> 
> apart from that lgtm

Tested manually.
Updated the commit message.
That's an interesting idea but let's discuss it offline :)

PR #882, opened 2019-11-04
Experiment guard for image preloanding

oxddr:
/assign @mm4tt


oxddr:
/close

This is actually not needed.

PR #884, opened 2019-11-05
Allow for 2 restarts of CoreDNS in tests.

oxddr:
/assign @pierewoj 
/assign @mm4tt

### DIFF
--- a/clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
+++ b/clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
@@ -2,8 +2,10 @@ RESTART_COUNT_THRESHOLD_OVERRIDES: |
   # To be investigated in https://github.com/kubernetes/perf-tests/issues/872.
   fluentd-gcp: 999
 
-  # These are ran on each node, so it might happen that we'll get a few restarts
-  # due to hardware failure for example.
+  # Main purpose of this check is detection crashlooping pods.
+  # It was extended to check whether master pods were restarted even once.
+  # For components that we run multiple instances of we should be less aggressive and tolerate restarts e.g. due to node restarts.
   kube-proxy: 2
   metadata-proxy: 2
   prometheus-to-sd-exporter: 2
+  coredns: 2


clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
```diff
@@ -7,3 +7,7 @@ RESTART_COUNT_THRESHOLD_OVERRIDES: |
   kube-proxy: 2
   metadata-proxy: 2
   prometheus-to-sd-exporter: 2
+
+  # CoreDNS runs multiple copies. We might be unlucky with hardware here too
+  # Example: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1190313018366365696
+  coredns: 2
```
mm4tt: Shouldn't we set it to 1 in such case?

@pierewoj, why do we use 2 in the pods above (kube-proxy and others)?

oxddr: No strong preferences on an actual value, except it shouldn't be 0. We can always tweak it if needed. 

To me 2 for components running on each node sounds like a reasonable value.

pierewoj: Main purpose of this check was AFAIK detection crashlooping pods, we extended it a bit to be notified if master pods are restarted even once. I think we might be better off staying on the less aggressive side with the thresholds for now to reduce number of false positives.

See http://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=SystemPodMetrics&metricname=Load_SystemPodMetrics&RestartCount=RestartCount

mm4tt: Thanks, that makes sense. I agree that most likely the cost of chasing each 2 restarts it's not worth the potential gain. 

Can we explain this (why we use 2) in this file for posterity?

pierewoj: Yes

oxddr:
PTAL

Issue #887, opened 2019-11-06
CL2 density and load test shouldn't require # nodes to be multiple of 100


mm4tt:
If you try running load or density test with #nodes not being multiple of 100 (e.g. with less than 100 nodes) you'll get some strange error caused by attempt to divide by 0:

```
error: decoding failed: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal number -9223372036854775808 into Go struct field Phase.Steps.Phases.ReplicasPerNamespace of type int32
```

We should fix that to unblock people who want to start their adventure with clusterloader2 but don't necessary have enough resources to run 100 node cluster.

PR #888, opened 2019-11-06
Check division by zero in template func

hypnoglow:
Adds an explicit check on division by zero for template functions. This way we get more precise errors like:

```
config reading error: replacing placeholders error: template: :34:23: executing "" at <DivideInt $totalPods $namespaces>: error calling DivideInt: division by zero
```

instead of weird ones like:

```
config reading error: decoding failed: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal number -4611686018427388 into Go struct field Phase.ReplicasPerNamespace of type int32
```

Part of #887


mm4tt:
Nice, thanks @hypnoglow!

/ok-to-test
/lgtm
/approve

PR #891, opened 2019-11-07
Enabled HA support for profile measurements

alejandrox1:
This patch enables clusterloaderv2 profile measurements for multiple master
nodes.

ref: #246

### DIFF
--- a/clusterloader2/pkg/measurement/common/profile.go
+++ b/clusterloader2/pkg/measurement/common/profile.go
@@ -17,6 +17,7 @@ limitations under the License.
 package common
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -49,7 +50,7 @@ func init() {
 type profileConfig struct {
 	componentName string
 	provider      string
-	host          string
+	hosts         []string
 	kind          string
 }
 
@@ -59,9 +60,6 @@ func (p *profileMeasurement) populateProfileConfig(config *measurement.Measureme
 		return err
 	}
 	if p.config.provider, err = util.GetStringOrDefault(config.Params, "provider", config.ClusterFramework.GetClusterConfig().Provider); err != nil {
-		return err
-	}
-	if p.config.host, err = util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp()); err != nil {
 		return err
 	}
 	return nil
@@ -86,6 +84,11 @@ func createProfileMeasurementFactory(name, kind string) func() measurement.Measu
 }
 
 func (p *profileMeasurement) start(config *measurement.MeasurementConfig) error {
+	p.config.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(p.config.hosts) < 1 {
+		return errors.New("Profile measurements will be disabled due to no MasterIps")
+	}
+
 	if err := p.populateProfileConfig(config); err != nil {
 		return err
 	}
@@ -107,13 +110,13 @@ func (p *profileMeasurement) start(config *measurement.MeasurementConfig) error 
 			case <-p.stopCh:
 				return
 			case <-time.After(profileFrequency):
-				profileSummary, err := p.gatherProfile(config.ClusterFramework.GetClientSets().GetClient())
+				profileSummaries, err := p.gatherProfile(config.ClusterFramework.GetClientSets().GetClient())
 				if err != nil {
 					klog.Errorf("failed to gather profile for %#v: %v", *p.config, err)
 					continue
 				}
-				if profileSummary != nil {
-					p.summaries = append(p.summaries, profileSummary)
+				if profileSummaries != nil {
+					p.summaries = append(p.summaries, profileSummaries...)
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


alejandrox1:
/assign @mm4tt @wojtek-t 
please take a look

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -61,8 +61,9 @@ func (p *profileMeasurement) populateProfileConfig(config *measurement.Measureme
 	if p.config.provider, err = util.GetStringOrDefault(config.Params, "provider", config.ClusterFramework.GetClusterConfig().Provider); err != nil {
 		return err
 	}
-	if p.config.host, err = util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp()); err != nil {
-		return err
+	p.config.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(p.config.hosts) < 1 {
+		klog.Warningf("Profile measurements will be disabled due to no MasterIps: %v", p.config.hosts)
```
mm4tt: NIT: If `len(p.config.hosts) < 1` then p.config.hosts = [] and the message will be always the same.
`Profile measurements will be disabled due to no MasterIps: []`

Maybe get rid of the %v part ?

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -159,33 +160,41 @@ func (p *profileMeasurement) String() string {
...
-		return measurement.CreateSummary(profilePrefix, "pprof", string(body)), err
+		summary := measurement.CreateSummary(profilePrefix, "pprof", string(body))
+		return []measurement.Summary{summary}, err
+	}
+
+	// Get the profile data over SSH.
+	// Start by checking that the provider allows us to do so.
+	if p.config.provider == "gke" {
+		// Only logging error for gke. SSHing to gke master is not supported.
+		klog.Errorf("%s: failed to execute curl command on master through SSH", p.name)
```
mm4tt: Can we change it to Warning? This shouldn't be Error

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -159,33 +160,41 @@ func (p *profileMeasurement) String() string {
 	return p.name
 }
 
-func (p *profileMeasurement) gatherProfile(c clientset.Interface) (measurement.Summary, error) {
+func (p *profileMeasurement) gatherProfile(c clientset.Interface) ([]measurement.Summary, error) {
 	profilePrefix := fmt.Sprintf("%s_%s", p.config.componentName, p.name)
```
mm4tt: Can we add host to the profilePrefix? Otherwise it would be impossible to tell from which host (master) the profile comes.

wojtek-t:
@alejandrox1 - will you have time to apply the comments?

clusterloader2/pkg/measurement/common/profile.go
alejandrox1: Thank you for the catch!

clusterloader2/pkg/measurement/common/profile.go
alejandrox1: Done!
thank you for your comment.

alejandrox1:
Sorry for the delay @wojtek-t @mm4tt - i got caught up on the us holiday season and stuff :sweat_smile: 
Just updated this PR and will resume working on #246

alejandrox1:
not sure if prev error was a flake or not so...

prev run: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/891/pull-perf-tests-clusterloader2/1206403270419943424/

mm4tt:
> not sure if prev error was a flake or not so...
> /test pull-perf-tests-clusterloader2
> 
> prev run: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/891/pull-perf-tests-clusterloader2/1206403270419943424/

The test fails due to https://github.com/kubernetes/kubernetes/issues/86302

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -61,8 +61,9 @@ func (p *profileMeasurement) populateProfileConfig(config *measurement.Measureme
 	if p.config.provider, err = util.GetStringOrDefault(config.Params, "provider", config.ClusterFramework.GetClusterConfig().Provider); err != nil {
 		return err
 	}
-	if p.config.host, err = util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp()); err != nil {
-		return err
+	p.config.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(p.config.hosts) < 1 {
```
mm4tt: Can we move this to the start() method and exit in that case? 
There is no point in executing any other code if there are no hosts to scrape the profiles from :)

alejandrox1: Done. thank you for pointing this out :sweat_smile:

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -86,6 +84,11 @@ func createProfileMeasurementFactory(name, kind string) func() measurement.Measu
 }
 
 func (p *profileMeasurement) start(config *measurement.MeasurementConfig) error {
+	p.config.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
```
mm4tt: Ideally, we'd allow the `hosts` to be specified via config.Params. So in  populateProfileConfig() we'd have
```
	if p.config.host, err = util.GetStringsOrDefault(config.Params, "hosts", config.ClusterFramework.GetClusterConfig().MasterIps); err != nil {
    return err
}
```

And here, after the populateProfileConfig, we'd have this check whether hosts are not empty.

But, this is fine, we currently don't set host anywhere via measurement params, so this can be improved in a separate PR

mm4tt:
Thanks, @alejandrox1. This is really great!

PR #893, opened 2019-11-12
update clusterloader2/vendor for scheduler metrics

RainbowMango:
## Why we need this
The vendor for scheduler metrics is too old.

Please refer to https://github.com/kubernetes/kubernetes/pull/83838#issuecomment-552704198

## How To Do
By command :
`govendor fetch k8s.io/kubernetes/pkg/scheduler/metrics`

### DIFF
--- a/clusterloader2/vendor/github.com/blang/semver/LICENSE
+++ b/clusterloader2/vendor/github.com/blang/semver/LICENSE
@@ -1,0 +1,22 @@
+The MIT License
+
+Copyright (c) 2014 Benedikt Lang <github at benediktlang.de>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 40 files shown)


RainbowMango:
@oxddr 
Thanks for your quick response.

I have two questions about the stuff, can you help to clarify?
- Does this will take effect immediately?
- Is there any side effect if we update all packages in the vendor?

oxddr:
1) All scalability test use perf-test repo at HEAD (subject to be change in the nearest future)
2) It depends what you mean specifically as "side-effect". AFAIK there is no other effect of updating vendor than on building scalability tests.

RainbowMango:
Got it. Thanks for your clarification.

Maybe I will send another update later. For PR: https://github.com/kubernetes/kubernetes/pull/83837

PR #896, opened 2019-11-12
Add missing start for APIResponsivenessPrometheusSimple

mm4tt:
Without it the `startTime` is set to `time.Zero` which results in querying metrics for the whole lifetime of a cluster while we should query only for the time of the test.


pierewoj:
Why do you only change load config and not density as well?

mm4tt:
Good point, I caught this while looking at load test but looks like density has the same error. Done

PR #900, opened 2019-11-14
Add Sleep measurement

mborsz:
/assign @mm4tt

### DIFF
--- a/clusterloader2/pkg/measurement/common/sleep.go
+++ b/clusterloader2/pkg/measurement/common/sleep.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/sleep.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func createSleepMeasurement() measurement.Measurement {
+	return &sleepMeasurement{}
+}
+
+type sleepMeasurement struct{}
+
+func (s *sleepMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	action, err := util.GetString(config.Params, "action")
```
wojtek-t: Do we even need to introduce "action" parameter for this measurement?

mm4tt: +1, let's keep it simple

mm4tt:
Thanks, @mborsz !

mm4tt:
Also travis seems to be failing with:
```
Boilerplate header is wrong for: /home/travis/gopath/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/sleep.go
The command "./verify/verify-boilerplate.sh" exited with 1.
```

Issue #895, opened 2019-11-12
clusterloader2 testing result missing ControllerManagerMetrics


sonyafenge:
Run clusterloader2 with default config under ./testing and get all result for controller manager is null.
cmd:
```
./run-e2e.sh cluster-loader2 --provider=kubemark --report-dir=/tmp/results --kubeconfig=[PWD]/kubeconfig.yaml --testconfig=testing/density/config.yaml --testconfig=testing/load/config.yaml
```

test result for controller manager:
```
"ControllerManagerMetrics": {
    "garbage_collector_attempt_to_delete_queue_latency": null,
    "garbage_collector_attempt_to_delete_work_duration": null,
    "garbage_collector_attempt_to_orphan_queue_latency": null,
    "garbage_collector_attempt_to_orphan_work_duration": null,
    "garbage_collector_dirty_processing_latency_microseconds": null,
    "garbage_collector_event_processing_latency_microseconds": null,
    "garbage_collector_graph_changes_queue_latency": null,
    "garbage_collector_graph_changes_work_duration": null,
    "garbage_collector_orphan_processing_latency_microseconds": null,
    "namespace_queue_latency": null,
    "namespace_queue_latency_count": null,
    "namespace_queue_latency_sum": null,
    "namespace_retries": null,
    "namespace_work_duration": null,
    "namespace_work_duration_count": null,
    "namespace_work_duration_sum": null
  },
```

mm4tt:
It seems to be also the case for our CI/CD tests, e.g. https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1194165744330346499/artifacts/MetricsForE2E_load_2019-11-12T08:37:41Z.json

mm4tt:
This file seems to be generated by this measurement - https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/metrics_for_e2e.go
This is where I would start looking.

mm4tt:
Looks like this measurement is based on the metric_grabber.go from k8s.io/kubernetes. Maybe we're using too old version of it and something has changed recently which results in these metrics being null. Just a theory but would be worth checking. If it's true then updating the vendor dir should fix the problem (we can try this nevertheless).

sonyafenge:
[...]

PR #901, opened 2019-11-15
[issuefix]:clusterloader2 testing result missing ControllerManagerMet…

sonyafenge:
…rics #895

- remove all outdated metrics labels for interestingControllerManagerMetricsLabels and replace with new labels
 Changes to be committed:
	modified:   clusterloader2/pkg/measurement/common/metrics_for_e2e.go

**Review tips:**
I listed all the available metrics labels in case you want to add anyone during review.


clusterloader2/pkg/measurement/common/metrics_for_e2e.go
```diff
@@ -38,23 +38,106 @@ var interestingApiServerMetricsLabels = []string{
 }
 
 var interestingControllerManagerMetricsLabels = []string{
```
mm4tt: Thanks Sonya!

Looking at your PR it got me thinking whether we need this "interesting" metric logic.
The cons of it are that we need to keep this list up2date which has non-zero maintanance costs (e.g. someone adds new metric or metric gets replaced by other and we need to update it here).
And when it comes to pros  the only one I see is less storage used for a test run results but these savings are in KB (maybe small MB) and comparing to GBs of logs a single run can generate it doesn't really matter.

Given that, could you remove this "interesting...MetricsLables and simply make this code dump all metrics? Let's start with KubeControllerManager, we'll see how it goes and if I haven't missed anything we can later change it for other components.

Thanks!

sonyafenge: I agree with you and will submit new update soon.

PR #903, opened 2019-11-20
Add last restart reason to system pod metrics

pierewoj:
This is a change in clusterloader system pod metrics measurement. To
speed up rootcausing, in this change restart reason field is added to
the measurement output.

Testing:
- set up cluster with 1 node
- ran density test with enabled system_pod_metrics
- restarted one container manually
- verified that SystemPodMetrics artifact is generated properly


mm4tt:
Very nice!

/lgtm
/approve

PR #906, opened 2019-11-21
Start enforcing scheduler throughput in density test

mm4tt:
This is to catch regressions like https://github.com/kubernetes/kubernetes/pull/85030.

This will affect only tests for 500+ node clusters as in smaller clusters the scheduler QPS setting may be less than 100 and hence to 100 throughput won't be achievable.

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -8,6 +8,7 @@
 {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 {{$DENSITY_TEST_THROUGHPUT := DefaultParam .DENSITY_TEST_THROUGHPUT 20}}
+{{$SCHEDULER_THROUGHPUT_THRESHOLD := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 0}}
 # LATENCY_POD_MEMORY and LATENCY_POD_CPU are calculated for 1-core 4GB node.
 # Increasing allocation of both memory and cpu by 10%
 # decreases the value of priority function in scheduler by one point.
@@ -137,6 +138,7 @@ steps:
     Method: SchedulingThroughput
     Params:
       action: gather
+      threshold: {{$SCHEDULER_THROUGHPUT_THRESHOLD}}
 
 - name: Starting latency pod measurements
   measurements:


clusterloader2/testing/density/config.yaml
```diff
@@ -137,6 +137,13 @@ steps:
     Method: SchedulingThroughput
     Params:
       action: gather
+      {{if ge .Nodes 500}}
+      # Setting to 90 should allow us to catch regressions like
+      # https://github.com/kubernetes/kubernetes/pull/85030 while not making the tests flaky.
+      # We do that only for 500+ node cluster as in smaller clusters the scheduler QPS may be
+      # less than 100 and hence the 100 throughput won't be achievable.
+      threshold: 90
+      {{end}}
```
wojtek-t: I'm not sure I like it - this is super hacky...
And we actually bump it for 100-node clusters too on GCE.

mm4tt: Why it's hacky? IMHO, it's reasonable to assume that certain throughput will be achievable only in large enough clusters
I agree that we won't be alerted on regressions in  100 node gce clusters, but this is still way better than what we currently have  = no alerts at all.

mm4tt: The alternative is to gate this by provider which imho is more hacky. I don't see anything hacky in saying that we enforce scheduler throughput in clusters having at least X nodes. If you don't like  X=500 then I'm open for discussion :)

wojtek-t: I don't like baking assumptions about providers (it is kind-of provider-specific logic) into our generic tests.
Note that, we actually set qps=100 in e2e-common preset:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L87

Rather than that, I would prefer making a threshold here some kind of parameter, and baking explicit thesholds into test definitions (overrides) used explicitly in tests we want to test against it.

clusterloader2/testing/density/config.yaml
```diff
@@ -8,6 +8,7 @@
 {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 {{$DENSITY_TEST_THROUGHPUT := DefaultParam .DENSITY_TEST_THROUGHPUT 20}}
+{{$SCHEDULER_THROUGHPUT_THRESHOLD := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD -1}}
```
wojtek-t: Why not set it to 0 as default?
Throughput will always be >= 0, so then you won't need if below.

clusterloader2/testing/density/config.yaml
wojtek-t: I like the new approach - thanks.

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/922

clusterloader2/testing/density/config.yaml
mm4tt: Done.

PR #908, opened 2019-11-27
Allow a single restart of master components 

oxddr:
fixes: https://github.com/kubernetes/kubernetes/issues/85682

/assign @pierewoj

### DIFF
--- a/clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
+++ b/clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
@@ -9,3 +9,9 @@ RESTART_COUNT_THRESHOLD_OVERRIDES: |
   metadata-proxy: 2
   prometheus-to-sd-exporter: 2
   coredns: 2
+
+  # Allow for a single restart of master components.
+  # As long as tests are passing, a single restart shouldn't be a problem
+  kube-scheduler: 1
+  kube-controller-manager: 1
+  l7-lb-controller: 1


clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
```diff
@@ -9,3 +9,9 @@ RESTART_COUNT_THRESHOLD_OVERRIDES: |
   metadata-proxy: 2
   prometheus-to-sd-exporter: 2
   coredns: 2
+
+  # Allow for a single restart of master components.
+  # As long as tests are passing, a single restart shouldn't be a problem
+  kube-scheduler: 1
+  kube-controller-manager: 1
+  l7-lb-controller: 1
```
pierewoj: If we consider single restart not a problem for any component, i am a bit worried that we might end up listing all pods here eventually. What do you think about adding "default" allowed restart count?

This could be a separate system_pod_metrics measurement param for "DEFAULT_ALLOWED_RESTART_COUNT" or we could just specify it in already existing override map, for example as "deafault: 2" (second approach should be slightly simpler, it's just a change in clusterloader + specifying defaults in ignore_lnown_restarts yaml configs).

clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
```diff
@@ -9,3 +9,9 @@ RESTART_COUNT_THRESHOLD_OVERRIDES: |
   metadata-proxy: 2
   prometheus-to-sd-exporter: 2
   coredns: 2
+
```
pierewoj: Do we want to specify it in ignore_known_kubemark_container_restarts.yaml as well?

pierewoj:
This should work so adding LGTM, but please take a look at comments

/lgtm

clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
wojtek-t: We should allow for specifying a default.

Regarding this change in general - do we understand why those restarts are happening? In general, I agree we may not necessary need to fail the tests on them (at least now), but I think we should at least have issues to investigate them.

oxddr: I'd go for a default that's set to at least 1. Let's merge this though first and then we can clean this up, once the change you mentioned is merged. 

@wojtek-t: I don't plan to debug this further: single restarts can happen for myriads of reason. From my experience debugging them is a waste of time, especially tests passed just fine.

clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
oxddr: Actually - why do we have a separate list for kubemark?

pierewoj: Because there are some containers that restart just in kubemark tests (for example at the time of implementing crashlooping pod detection. scheduler/controller-manager restarts only happened in kubemark). We wanted to temporarily allow it in kubemark tests and still be able to notice regressions in non-kubemark runs.

clusterloader2/testing/experiments/ignore_known_gce_container_restarts.yaml
wojtek-t: > @wojtek-t: I don't plan to debug this further: single restarts can happen for myriads of reason. From my experience debugging them is a waste of time, especially tests passed just fine.

I think I disagree with the latter - we already had a bunch of cases, where restarts were actually caused by real bugs.

PR #912, opened 2019-11-28
Add PriorityClass with a high priority value to Prometheus.

oxddr:
Running Promtheus is critical to even start tests, it should evict other
non-monitoring pod if needed.

/assign @wojtek-t 

ref https://github.com/kubernetes/kubernetes/issues/85683

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/00monitoring-priorityclass.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/00monitoring-priorityclass.yaml
@@ -1,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: monitoring
+value: 1000000
+globalDefault: false
+description: "This priority class should be used for monitoring pods only."
--- a/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
@@ -16,6 +16,7 @@ spec:
   baseImage: gcr.io/k8s-testimages/quay.io/prometheus/prometheus
   nodeSelector:
     beta.kubernetes.io/os: linux
+  priorityClassName: monitoring
   replicas: 1
   resources:
     requests:


clusterloader2/pkg/prometheus/manifests/00monitoring-priorityclass.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: monitoring
+value: 1000000
```
wojtek-t: Won't it collide with: https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/daemonset-priorityclass.yaml
?

clusterloader2/pkg/prometheus/manifests/00monitoring-priorityclass.yaml
```diff
@@ -0,0 +1,7 @@
...
+globalDefault: false
+description: "This priority class should be used for monitoring pods only."
```
wojtek-t: While in principle I agree with this approach, I would like to understand the consequences for Kubemark better.
- What will happen if we preempt a bunch of hollow-nodes to make place for prometheus?
- Don't we need to add some additional logic to ensure that they will really be up when we will be starting tests?
- Don't we need to add some logic to ensure that the preempted ones will be cleaned-up (and nothing will get scheduled later on them?)

There are couple questions that I would like to see clear answers for.

oxddr: Discussed in person.

Prometheus having a higher priority will evict some hollow-nodes, which can

1. New hollow-nodes won't be registered before an actual test starts. This is not a high risk. 
2. Old hollow-nodes won't be remove and be used for scheduling for some time (it'd be worth to find out what "some time" is in this case), which can cause unwanted behavior in tests (e.g. failing pod startup latency). 

We decided not to proceed with this PR because of 2. and instead add another, tainted node, on which hollow-nodes won't be scheduled.

PR #913, opened 2019-12-02
Allow Prometheus to run on a tainted heapster node.

oxddr:
See https://github.com/kubernetes/kubernetes/issues/85683 and https://github.com/kubernetes/kubernetes/pull/85797 for context. 

/assign @wojtek-t

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
@@ -34,6 +34,10 @@ spec:
     fsGroup: 2000
     runAsNonRoot: true
     runAsUser: 1000
+  tolerations:
+  - key: "monitoring"
+    operator: "Exists"
+    effect: "NoSchedule"
   serviceAccountName: prometheus-k8s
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -34,6 +34,10 @@ spec:
     fsGroup: 2000
     runAsNonRoot: true
     runAsUser: 1000
+  tolerations:
+  - key: "monitoring"
+    operator: "Exists"
+    effect: "PreferNoSchedule"
```
wojtek-t: Prefer won't force anything - is that really what you want?
Shouldn't we use `NoSchedule`

oxddr: Yes, I don't think it's going to be a problem if we schedule one or two hollow nodes or other infrastructure pods on this node. I don't have a strong preference on this though.

oxddr: Done

PR #914, opened 2019-12-02
Use APIResponsivenessPrometheusSimple instead of APIResponsiveness.

oxddr:
The latter is based on a deprecated metric. The former mimics the behavior using Prometheus and has been already enable in all tests. 

ref https://github.com/kubernetes/kubernetes/issues/73519

/assign @wojtek-t

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -74,10 +74,6 @@ chaosMonkey:
 steps:
 - name: Starting measurements
   measurements:
-  - Identifier: APIResponsiveness
-    Method: APIResponsiveness
-    Params:
-      action: reset
   - Identifier: APIResponsivenessPrometheus
     Method: APIResponsivenessPrometheus
     Params:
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -729,17 +725,11 @@ steps:
...
   measurements:
-  - Identifier: APIResponsiveness
-    Method: APIResponsiveness
-    Params:
-      action: gather
   - Identifier: APIResponsivenessPrometheusSimple
     Method: APIResponsivenessPrometheus
     Params:
       action: gather
-      {{if $ENABLE_PROMETHEUS_API_RESPONSIVENESS}}
```
wojtek-t: This is currently not enabled for violations in our largest tests (both gce and kubemarks).
I would prefer explicitly enabling it first, and only then removing the rest.

Do we even know if we won't start failing significantly on it?

oxddr: It's enable in Kubemark test (https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml), but you're right - it's not enable in gce5000. I looked through my local repo, where I had unmerged commit. Please see: https://github.com/kubernetes/test-infra/pull/15462

PR #921, opened 2019-12-10
Cl2: Make CL2_* envs available in templates

mm4tt:
This will make all env variables starting with CL2_ prefix available in the CL2 template files.

It will allow us to leverage the presets mechanism in cl2 templates and (in some cases) use it instead of relying on cumbersome test override files.

The problem it solves is that sometimes we want to set something for almost all (but not all tests) tests.
Currently, the only way to do that is the create an override file and pass it to many test definitions.
This is arduous for many reasons, e.g. it requires at least 2 PRs (one to perf-test with the override file, second to test-infra) and it requires keeping many places in sync.

### DIFF
--- a/clusterloader2/pkg/config/template_provider.go
+++ b/clusterloader2/pkg/config/template_provider.go
@@ -21,12 +21,15 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"sync"
 	"text/template"
 
+	goerrors "github.com/go-errors/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/perf-tests/clusterloader2/api"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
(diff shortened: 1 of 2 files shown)


mm4tt:
This will be needed for https://github.com/kubernetes/perf-tests/pull/906

clusterloader2/pkg/config/template_provider.go
```diff
@@ -21,12 +21,15 @@ import (
...
 	"io/ioutil"
+	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"sync"
 	"text/template"
 
+	goErrors "github.com/go-errors/errors"
```
wojtek-t: convention is to use small letters only for aliases

mm4tt: Done. 

In the ideal world we'd have a lint check for such things. 
@oxddr, you were looking at some golint alternatives, right? Would it be possible to configure them to detect such things?

oxddr: Enforcing naming convention could be done using the current linter. However we have many violations in our codebase, so I've put the change on hold. I will have a look at it again in spare time.

PR #923, opened 2019-12-10
Use Measurement's indentifier in logging.

oxddr:
This allows to clearly identify logs if we have two, distinct instances of the same measurement.

/assign @mm4tt


mm4tt:
Can we have both measurementName and identifier? I think there is a value in ability to search by either of them

oxddr:
> Can we have both measurementName and identifier? I think there is a value in ability to search by either of them

We could, but in all cases but one they are same values. And in the case where they are different (APIResponsivenessPrometheues) Identifier contains name of the measurement so grepping still works. 

Plus you always get name of the file, from which logging was invoked.

PR #927, opened 2019-12-10
Implement fetching prow config from directories

pierewoj:
At this point, we need to specify each prow config file explicitly in
deployment yaml. This will be a problem for adding sig-release jobs to
pefdash, since these configs are autogenerated by config forker.

With this change, we'd be able to just specify "githubConfigDir" to the
sig-release dir and perfdash will ingest all the configs from that dir.

Testing:
* started perfdash locally with
githubConfigDir=path-to-sig-release-configs, verified configs are
fetched
* tested that perfadash still works without providing githubConfigDir
flag

### DIFF
--- a/perfdash/github-configs-fetcher.go
+++ b/perfdash/github-configs-fetcher.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+type githubDirContent struct {
+	Name        string `yaml:"name"`
+	Path        string `yaml:"path"`
+	DownloadUrl string `yaml:"download_url"`
+	Type        string `yaml:"type"`
+	Url         string `yaml:"url"`
+}
+
+// GetConfigsFromGithub gets config paths from github directory. It uses github API,
+// which is documented here: https://developer.github.com/v3/repos/contents/
+//
+// Example url: https://api.github.com/repos/kubernetes/test-infra/contents/config/jobs/kubernetes/sig-release/release-branch-jobs
+//
+// Different branch can be specified by appending "?ref=branch-name" parameter at the end
+// of the url.
+func GetConfigsFromGithub(url string) ([]string, error) {
+	var result []string
+	contents, err := getGithubDirContents(url)
+	if err != nil {
+		return nil, err
+	}
+	for _, c := range contents {
+		// Dirs and non-yaml files are ignored; this means that there is no
+		// recursive search, it should be good enough for now.
+		if c.Type == "file" && strings.HasSuffix(c.Name, ".yaml") {
+			result = append(result, c.DownloadUrl)
+		}
+	}
+	return result, nil
+}
+
+func getGithubDirContents(url string) ([]githubDirContent, error) {
+	fmt.Printf("Downloading github spec from %v\n", url)
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("error calling github API %s: %v", url, err)
+	}
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("error reading github response %s: %v", url, err)
+	}
+	var decoded []githubDirContent
+	err = yaml.Unmarshal(b, &decoded)
+	if err != nil {
+		return nil, err
+	}
+	return decoded, nil
+}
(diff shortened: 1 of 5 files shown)


perfdash/github-configs-fetcher.go
```diff
@@ -0,0 +1,77 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
mm4tt: 2019?

perfdash/github-configs-fetcher.go
```diff
@@ -0,0 +1,77 @@
...
+/*
+Gets config paths from github directory. It uses github API, which is documented
+here: https://developer.github.com/v3/repos/contents/
+
+Example url: https://api.github.com/repos/pierewoj/test-infra/contents/config/jobs/kubernetes/sig-release/release-branch-jobs
+
+Different branch can be specified by appending "?ref=branch-name" parameter at the end
+of the url.
+*/
+func GetConfigsFromGithub(url string) ([]string, error) {
```
mm4tt: Comment for this method needs to start with `GetConfigsFromGithub ...`
golint complains about that

perfdash/github-configs-fetcher.go
```diff
@@ -0,0 +1,77 @@
...
+		// recursive search, it should be good enough for now.
+		if c.Type == "file" && strings.HasSuffix(c.Name, ".yaml") {
+			result = append(result, c.DownloadUrl)
+		}
+	}
+	return result, nil
+}
+
+func getGithubDirContents(url string) ([]githubDirContent, error) {
+	fmt.Printf("Downloading github spec from %v\n", url)
```
mm4tt: Use klog/log?

Here and other places

perfdash/google-gcs-downloader.go
```diff
@@ -37,6 +37,7 @@ import (
 // GoogleGCSDownloaderOptions is an options for GoogleGCSDownloader.
 type GoogleGCSDownloaderOptions struct {
 	ConfigPaths        []string
+	GithubConfigDir    []string
```
mm4tt: This is a list of dirs so GithubConfigDirs

perfdash/google-gcs-downloader.go
```diff
@@ -66,6 +67,19 @@ func NewGoogleGCSDownloader(opt *GoogleGCSDownloaderOptions) (*GoogleGCSDownload
 
 // TODO(random-liu): Only download and update new data each time.
 func (g *GoogleGCSDownloader) getData() (JobToCategoryData, error) {
+	configPaths := g.Options.ConfigPaths
```
mm4tt: It's strange to have a get*() function modify something and this will modify the g.Options.ConfigPaths.
Moreover, if you call it twice you'll have the github config paths listed twice in the  g.Options.ConfigPaths.

Copy the slice instead?

perfdash/perfdash.go
```diff
@@ -48,6 +48,7 @@ func initGoogleDownloaderOptions() {
 	pflag.BoolVar(&options.OverrideBuildCount, "force-builds", false, "Whether to enforce number of builds to process as passed via --builds flag. This would override values defined by \"perfDashBuildsCount\" label on prow job")
 	pflag.IntVar(&options.DefaultBuildsCount, "builds", maxBuilds, "Total builds number")
 	pflag.StringArrayVar(&options.ConfigPaths, "configPath", []string{}, "Paths/urls to the prow config")
+	pflag.StringArrayVar(&options.GithubConfigDir, "githubConfigDir", []string{}, "Github API url to the prow config directory, all configs from this dir will be used")
```
mm4tt: urls, dirs

perfdash/github-configs-fetcher.go
```diff
@@ -0,0 +1,77 @@
...
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("error calling github API %s: %v", url, err)
```
mm4tt: Nit: Consider using github.com/go-errors/errors instead (they have stacktraces!) :)

perfdash/github-configs-fetcher.go
pierewoj: That's a good idea but we use fmt.Printf in other places. If we wan't to use klog, I'd migrate all logging places at once (if we don't do that, some lines will have timestamp, some won't etc).

Let's do this in a separate PR.

perfdash/google-gcs-downloader.go
pierewoj: done

perfdash/github-configs-fetcher.go
pierewoj: done

perfdash/github-configs-fetcher.go
pierewoj: done

perfdash/google-gcs-downloader.go
pierewoj: Good idea, thanks, done

perfdash/perfdash.go
pierewoj: done

perfdash/github-configs-fetcher.go
pierewoj: I'll create a separate issue to use it in perfdash, are you fine with this?

mm4tt: Sgtm, thanks

pierewoj: https://github.com/kubernetes/perf-tests/issues/929

perfdash/github-configs-fetcher.go
pierewoj: https://github.com/kubernetes/perf-tests/issues/929

perfdash/github-configs-fetcher.go
pierewoj: https://github.com/kubernetes/perf-tests/issues/929

perfdash/github-configs-fetcher.go
wojtek-t: We generally try to not have comments in /* .. */

Can you please change to simply start each line with //

pierewoj: done

mm4tt:
/lgtm
/approve

Thanks!

PR #930, opened 2019-12-12
Support for windows nodes

YangLu1031:
Add support for windows nodes to do node-throughput test.
1. the pod startup latency threshold on windows node is not clear yet, so set 20m for now.

cmd to run on k8s windows cluster:

$GOPATH/src/k8s.io/perf-tests/run-e2e.sh cluster-loader2 --nodes=1 --provider=gce --report-dir=/workspace/_artifacts --testconfig=testing/node-throughput/config.yaml --testoverrides=./testing/node-throughput/windows_override.yaml

### DIFF
--- a/clusterloader2/testing/node-throughput/windows_override.yaml
+++ b/clusterloader2/testing/node-throughput/windows_override.yaml
@@ -1,0 +1,3 @@
+POD_COUNT: 80
+CONTAINER_IMAGE: gcr.io/gke-release/pause-win:1.0.0
+POD_STARTUP_LATENCY_THRESHOLD: 20m
--- a/clusterloader2/testing/node-throughput/config.yaml
+++ b/clusterloader2/testing/node-throughput/config.yaml
@@ -2,8 +2,10 @@
 # - This test is designed for 1-node cluster.
 
 #Constants
-{{$POD_COUNT := 100}}
+{{$POD_COUNT := DefaultParam .POD_COUNT 100}}
 {{$POD_THROUGHPUT := DefaultParam .POD_THROUGHPUT 2}}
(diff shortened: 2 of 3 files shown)


clusterloader2/testing/node-throughput/windows_override.yaml
```diff
@@ -0,0 +1,3 @@
+POD_COUNT: 80
```
oxddr: out of curiosity: why do you decrease number of pods? It was 100 before.

clusterloader2/testing/node-throughput/windows_override.yaml
```diff
@@ -0,0 +1,3 @@
...
+CONTAINER_IMAGE: gcr.io/gke-release/pause-win:1.0.0
+POD_STARTUP_LATENCY_THRESHOLD: 20m
```
oxddr: This seems to be a magnitude higher than the default (5s) - why?

clusterloader2/testing/node-throughput/windows_override.yaml
wojtek-t: We were never testing windows nodes - so it's hard to say what it was before :)

clusterloader2/testing/node-throughput/windows_override.yaml
wojtek-t: For experimenting with widnwos nodes it makes sense to me.

mm4tt: Yeah, it was explained in the PR description. Looks good to me as well, although wouldn't hurt to add a TODO / comment here.

mm4tt:
Nice!

/lgtm

clusterloader2/testing/node-throughput/windows_override.yaml
oxddr: I must have missed that. Make sense.

PR #933, opened 2019-12-13
add scheduling cycle e2e latency

yiyang5055:
add e2e_scheduling_duration_seconds to perf_test
ref  [#86216](https://github.com/kubernetes/kubernetes/issues/86126)


liu-cong:
cc @ahg-g 
cc

Issue #929, opened 2019-12-11
Improve perfdash logging


pierewoj:
Right now, perfdash uses fmt.Printf to log errors. It is not ideal because logs do not contain timestamps, log source etc.

Proposed solution:
* change all ocurrances of fmt.Printf to klog in perfdash
* consider using https://github.com/go-errors/errors for handling errors (so that they contain stack traces)
* you can use `make run` for testing

Sakuralbj:
I'd like to help.Can I assign it to me? @pierewoj

pierewoj:
Sure, let me know if you need any help, either here or on k8s slack (I'm under Jakub Pierewoj there). Thanks!

PR #934, opened 2019-12-15
Perfdash: Improve perfdash logging

Sakuralbj:
Fix [Issue929](https://github.com/kubernetes/perf-tests/issues/929).
Change all ocurrances of fmt.Printf to klog in perfdash.
Using go-errors for handling errors will be a separate PR

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -19,8 +19,8 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"k8s.io/klog"
 	"net/http"
-	"os"
 	"strconv"
 	"strings"
 
@@ -384,10 +384,10 @@ func getProwConfig(configPaths []string) (Jobs, error) {
 	jobs := Jobs{}
 
 	for _, configPath := range configPaths {
-		fmt.Fprintf(os.Stderr, "Fetching config %s\n", configPath)
+		klog.Infof("Fetching config %s", configPath)
 		// Perfdash supports only yamls.
 		if !strings.HasSuffix(configPath, ".yaml") {
-			fmt.Fprintf(os.Stderr, "%s is not an yaml file!\n", configPath)
+			klog.Warningf("%s is not an yaml file!", configPath)
 			continue
 		}
 		var content []byte
@@ -408,13 +408,13 @@ func getProwConfig(configPaths []string) (Jobs, error) {
 		for _, periodic := range conf.Periodics {
 			config, err := parsePeriodicConfig(periodic)
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "warning: failed to parse config of %q due to: %v\n",
+				klog.Errorf("warning: failed to parse config of %q due to: %v",
 					periodic.Name, err)
 				continue
 			}
 			shouldUse, err := checkIfConfigShouldBeUsed(config)
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "warning: failed to validate config of %q due to: %v\n",
+				klog.Errorf("warning: failed to validate config of %q due to: %v",
 					periodic.Name, err)
 				continue
 			}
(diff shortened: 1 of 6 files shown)


perfdash/config.go
```diff
@@ -384,10 +384,10 @@ func getProwConfig(configPaths []string) (Jobs, error) {
 	jobs := Jobs{}
 
 	for _, configPath := range configPaths {
-		fmt.Fprintf(os.Stderr, "Fetching config %s\n", configPath)
+		klog.Errorf("Fetching config %s\n", configPath)
```
pierewoj: This does not seem like an error level (even though it was written to os.Stderr). It will be logged for every config path provided, so I'd keep it at INFO level, this is klog.Infof

Also - consider removing the newline char from all logging invocations - according to https://godoc.org/github.com/kubernetes/klog, klog appends newlines if they are not present.

perfdash/config.go
```diff
@@ -384,10 +384,10 @@ func getProwConfig(configPaths []string) (Jobs, error) {
...
 		// Perfdash supports only yamls.
 		if !strings.HasSuffix(configPath, ".yaml") {
-			fmt.Fprintf(os.Stderr, "%s is not an yaml file!\n", configPath)
+			klog.Errorf("%s is not an yaml file!\n", configPath)
```
pierewoj: Also, this would be better as INFO or WARNING in my opinion (can happen for directories with OWNERS file and it is totally fine).

perfdash/downloader.go
```diff
@@ -106,23 +107,23 @@ func (j *JobToCategoryData) ServeCategoryNames(res http.ResponseWriter, req *htt
 func (j *JobToCategoryData) ServeMetricNames(res http.ResponseWriter, req *http.Request) {
 	jobname, ok := getURLParam(req, "jobname")
 	if !ok {
-		fmt.Printf("Url Param 'jobname' is missing\n")
+		klog.Infof("Url Param 'jobname' is missing\n")
```
pierewoj: Don't these logs mean that the request returns nothing? This is something like 404 Not Found case, maybe we should log such issues in this method (and also in `ServeMetricNames` below) as Warning?

perfdash/google-gcs-downloader.go
```diff
@@ -122,20 +122,20 @@ func (g *GoogleGCSDownloader) getJobData(wg *sync.WaitGroup, result JobToCategor
...
+		klog.Infof("Fetching %s build %v...\n", job, buildNumber)
 		for categoryLabel, categoryMap := range tests.Descriptions {
 			for testLabel, testDescriptions := range categoryMap {
 				for _, testDescription := range testDescriptions {
 					filePrefix := fmt.Sprintf("%v_%v", testDescription.OutputFilePrefix, testDescription.Name)
 					searchPrefix := fmt.Sprintf("artifacts/%v", filePrefix)
 					artifacts, err := g.GoogleGCSBucketUtils.listFilesInBuild(job, buildNumber, searchPrefix)
 					if err != nil || len(artifacts) == 0 {
-						fmt.Printf("Error while looking for %s* in %s build %v: %v\n", searchPrefix, job, buildNumber, err)
+						klog.Infof("Error while looking for %s* in %s build %v: %v\n", searchPrefix, job, buildNumber, err)
```
pierewoj: Errorf?

perfdash/perfdash.go
```diff
@@ -57,9 +58,10 @@ func initGoogleDownloaderOptions() {
 }
 
 func main() {
-	fmt.Println("Starting perfdash...")
+	klog.InitFlags(nil)
+	klog.Infoln("Starting perfdash...")
 	if err := run(); err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		klog.Errorln(err)
```
pierewoj: Do we need the `Errorln`? From the docs https://godoc.org/github.com/kubernetes/klog#Errorf a newline is appended when missing.

Won't klog.Error be good enough here?

pierewoj:
I'll be away for a few weeks, assigning to Matt to check next revisions
/assign @mm4tt

perfdash/perfdash.go
Sakuralbj: Thank you. I have changed it.

mm4tt:
/lgtm
/approve

Thanks, @Sakuralbj for helping us with this!

PR #915, opened 2019-12-02
Remove APIResponsiveness

oxddr:
ref https://github.com/kubernetes/kubernetes/issues/73519

PR #935, opened 2019-12-16
Use APIResponsivenessPrometheusSimple in legacy test configs.

jkaniuk:
Some scalability tests are failing after https://github.com/kubernetes/perf-tests/pull/915 was merged

https://k8s-testgrid.appspot.com/sig-scalability-experiments#storage
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-storage-scalability/1205162963401445380

https://k8s-testgrid.appspot.com/sig-scalability-node#node-throughput
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-node-throughput/1205121688543432710

```
[measurement call APIResponsiveness - APIResponsiveness error: unknown measurement method APIResponsiveness
measurement call APIResponsiveness - APIResponsiveness error: unknown measurement method APIResponsiveness]
```

/cc oxddr

### DIFF
--- a/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
@@ -37,11 +37,11 @@ tuningSets:
 steps:
 # Start measurements
 - measurements:
-  - Identifier: APIResponsiveness
-    Method: APIResponsiveness
-    Params:
-      action: reset
   - Identifier: APIResponsivenessPrometheus
+    Method: APIResponsivenessPrometheus
+    Params:
+      action: start
+  - Identifier: APIResponsivenessPrometheusSimple
     Method: APIResponsivenessPrometheus
     Params:
       action: start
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/density/legacy/config.yaml
```diff
@@ -17,6 +17,7 @@
 {{$MIN_LATENCY_PODS := 500}}
 {{$MIN_SATURATION_PODS_TIMEOUT := 180}}
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
+{{$USE_SIMPLE_LATENCY_QUERY := DefaultParam .USE_SIMPLE_LATENCY_QUERY false}}
```
wojtek-t: legacy/ files shouldn't be use

Instead - we should fix the configs to use appropriate branches from perf-tests repo

@mm4tt @oxddr

mm4tt: +1, @oxddr, can we switch 1.13 tests to use 1.13 branch of perf-test? Is there anything left to do before we do that?

jkaniuk: release-1.13 was reverted to same state as release-1.14, so that should solve the issue with older test

independently, other test are failing due to unknown metric:
https://k8s-testgrid.appspot.com/sig-scalability-node#node-throughput
https://k8s-testgrid.appspot.com/sig-scalability-experiments#storage

I have updated the PR

PR #950, opened 2019-12-19
Fix GlobalQPSLoad to really share rate.Limiter

mborsz:
The previous attempt doesn't work. Cluster loader creates new instance of GlobalQPSLoad for all phases, so fact that they use inner *rate.Limiter doesn't change anything.

/assign @wojtek-t


wojtek-t:
LGTM - please fix gofmt though

mborsz:
gofmt seems to be happy now

PR #966, opened 2020-01-02
Implement WaitForNodes measurement

mm4tt:
One of the use-cases for this is to allow running simple ClusterAutoscaler tests.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_nodes.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_nodes.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"time"
+
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/wait_for_nodes.go
```diff
@@ -0,0 +1,95 @@
...
+
+	options := &measurementutil.WaitForNodeOptions{
+		Selector:             selector,
+		MinDesiredNodeCount:  minNodeCount,
+		MaxDesiredNodeCount:  maxNodeCount,
+		EnableLogging:        true,
+		CallerName:           w.String(),
+		WaitForNodesInterval: defaultWaitForNodesInterval,
+	}
+	return nil, measurementutil.WaitForNodes(config.ClusterFramework.GetClientSets().GetClient(), stopCh, options)
```
wojtek-t: Don't we want to allow for configuring it somehow? Ensuring that those nodes are Ready is the first things that comes to my mind.
But I can imagine other conditions too (e.g. no NetworkUnavailable condition, no DiskPressure, etc.)

mm4tt: I think that eventually we want :) It's just a first version of this measurement, we'll build on that. 

It already supports selectors, so if a condition can be expressed via selector then we already support it.

wojtek-t: > It already supports selectors, so if a condition can be expressed via selector then we already support it.

It can't...

I'm wondering if the existence of node is actually enough - whether node being Ready is actually what we want.

mm4tt: Done, as discussed personally.

mm4tt:
PTAL

clusterloader2/pkg/measurement/util/wait_for_nodes.go
```diff
@@ -0,0 +1,74 @@
...
+	defer ps.Stop()
+
+	nodeCount := getNumReadyNodes(ps.List())
+	for {
+		select {
+		case <-stopCh:
+			return fmt.Errorf("timeout while waiting for [%d-%d] Nodes with selector '%v' to be ready - currently there is %d Nodes",
+				options.MinDesiredNodeCount, options.MaxDesiredNodeCount, options.Selector.String(), nodeCount)
+		case <-time.After(options.WaitForNodesInterval):
+			nodeCount = len(ps.List())
```
wojtek-t: this should compute ready nodes

mm4tt: Good catch, done.

PR #967, opened 2020-01-03
Reduce the pod count to pass the windows load test

YangLu1031:
The windows load test times out from [testgrid](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-node-throughput/1212907172682272773). And after SSH  to windows vm, the pause image is loaded, and around 30 pods are up running. Maybe 80 is too big, so change to 10 to see if it's the cause.


YangLu1031:
cc @yliaog

PR #973, opened 2020-01-07
scale up windows load test

YangLu1031:
1. Run load test locally, for 50 pods it takes 10mins to startup, looks acceptable
2. Also in case of timeout, extends operation timeout to 30mins.


clusterloader2/testing/node-throughput/config.yaml
```diff
@@ -33,7 +34,7 @@ steps:
       apiVersion: v1
       kind: ReplicationController
       labelSelector: group = latency
-      operationTimeout: 15m
+      operationTimeout: {{$OPTS_TIMEOUT}}
```
pjh: Maybe name the variable "REPLICATION_TIMEOUT" or "OPERATION_TIMEOUT" ?

OPTS is not clear / specific enough.

PR #977, opened 2020-01-09
Fix scraping kube-apiserver pprofs

mm4tt:


clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -172,17 +172,6 @@ func (p *profileMeasurement) gatherProfile(c clientset.Interface) ([]measurement
 	var summaries []measurement.Summary
 	for _, host := range p.config.hosts {
 		profilePrefix := fmt.Sprintf("%s_%s_%s", host, p.config.componentName, p.name)
-
-		if p.config.componentName == "kube-apiserver" {
```
wojtek-t: Discussed offline - this will break setups where we can't ssh to master.

PR #978, opened 2020-01-09
add e2e_scheduling_duration_seconds in the scheduler perf dashboard

yiyang5055:
add e2e_scheduling_duration_seconds in the scheduler perf dashboard.

ref[ #86126](https://github.com/kubernetes/kubernetes/issues/86126)
/cc @liu-cong

### DIFF
--- a/clusterloader2/pkg/measurement/common/scheduler_latency.go
+++ b/clusterloader2/pkg/measurement/common/scheduler_latency.go
@@ -38,7 +38,7 @@ import (
 const (
 	schedulerLatencyMetricName = "SchedulingMetrics"
 
-	e2eScheduling = "e2eScheduling"
+	schedulerE2eSchedulingMetricName = "e2e_scheduling_duration_seconds_bucket"
 
 	schedulingLatencyMetricName = model.LabelValue(schedulermetric.SchedulerSubsystem + "_" + schedulermetric.SchedulingLatencyName)
 	singleRestCallTimeout       = 5 * time.Minute
@@ -118,7 +118,12 @@ func (s *schedulerLatencyMeasurement) getSchedulingLatency(c clientset.Interface
 		return nil, err
 	}
 
+	hist := measurementutil.NewHistogram(nil)
 	for _, sample := range samples {
+		if sample.Metric[model.MetricNameLabel] == schedulerE2eSchedulingMetricName {
+			measurementutil.ConvertSampleToHistogram(sample, hist)
+			continue
+		}
 		if sample.Metric[model.MetricNameLabel] != schedulingLatencyMetricName {
 			continue
 		}
@@ -133,8 +138,6 @@ func (s *schedulerLatencyMeasurement) getSchedulingLatency(c clientset.Interface
 			metric = &result.PreemptionEvaluationLatency
 		case schedulermetric.Binding:
 			metric = &result.BindingLatency
-		case e2eScheduling:
-			metric = &result.E2eSchedulingLatency
 		}
 
 		if metric == nil {
@@ -148,12 +151,30 @@ func (s *schedulerLatencyMeasurement) getSchedulingLatency(c clientset.Interface
 		metric.SetQuantile(quantile, time.Duration(int64(float64(sample.Value)*float64(time.Second))))
 	}
 
+	if err := s.setQuantileFromHistogram(&result.E2eSchedulingLatency, hist); err != nil {
+		return nil, err
+	}
+
 	content, err := util.PrettyPrintJSON(result)
 	if err != nil {
 		return nil, err
 	}
 	summary := measurement.CreateSummary(schedulerLatencyMetricName, "json", content)
 	return []measurement.Summary{summary}, nil
+}
+
+// Set quantile of LatencyMetric from Histogram
+func (s *schedulerLatencyMeasurement) setQuantileFromHistogram(metric *measurementutil.LatencyMetric, hist *measurementutil.Histogram) error {
+	quantiles := []float64{0.5, 0.9, 0.99}
+	for _, quantile := range quantiles {
+		histQuantile, err := hist.Quantile(quantile)
+		if err != nil {
+			return err
+		}
+		metric.SetQuantile(quantile, time.Duration(int64(histQuantile*float64(time.Second))))
+	}
+
+	return nil
 }
 
 // Sends request to kube scheduler metrics
--- a/clusterloader2/pkg/measurement/util/histogram.go
+++ b/clusterloader2/pkg/measurement/util/histogram.go
@@ -17,15 +17,77 @@ limitations under the License.
 package util
 
 import (
+	"math"
 	"reflect"
+	"sort"
+	"strconv"
 
 	"github.com/prometheus/common/model"
 )
+
+// Bucket of a histogram
+type bucket struct {
+	upperBound float64
+	count      float64
+}
+
+type buckets []bucket
+
+func (b buckets) Len() int           { return len(b) }
+func (b buckets) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
+func (b buckets) Less(i, j int) bool { return b[i].upperBound < b[j].upperBound }
+
+// bucketQuantile return Quantile of calculate from buckets
+// copy from https://github.com/kubernetes/kubernetes/pull/85861/files#diff-8fc16044937fc03c39b585a657b20150R179-R209
+func bucketQuantile(q float64, buckets buckets) float64 {
+	if q < 0 {
+		return math.Inf(-1)
+	}
+	if q > 1 {
+		return math.Inf(+1)
+	}
+
+	if len(buckets) < 2 {
+		return math.NaN()
+	}
+
+	rank := q * buckets[len(buckets)-1].count
+	b := sort.Search(len(buckets)-1, func(i int) bool { return buckets[i].count >= rank })
+
+	if b == 0 {
+		return buckets[0].upperBound * (rank / buckets[0].count)
+	}
+
+	// linear approximation of b-th bucket
+	brank := rank - buckets[b-1].count
+	bSize := buckets[b].upperBound - buckets[b-1].upperBound
+	bCount := buckets[b].count - buckets[b-1].count
+
+	return buckets[b-1].upperBound + bSize*(brank/bCount)
+}
 
 // Histogram is a structure that represents distribution of data.
 type Histogram struct {
 	Labels  map[string]string `json:"labels"`
 	Buckets map[string]int    `json:"buckets"`
+}
+
+// Quantile calculates the quantile 'q' based on the given buckets of Histogram.
+func (h *Histogram) Quantile(q float64) (float64, error) {
+	var b []bucket
+	for k, v := range h.Buckets {
+		upper, err := strconv.ParseFloat(k, 64)
+		if err != nil {
+			return math.MaxFloat64, err
+		}
+		b = append(b, bucket{
+			upperBound: upper,
+			count:      float64(v),
+		})
+	}
+
+	sort.Sort(buckets(b))
+	return bucketQuantile(q, b), nil
 }
 
 // HistogramVec is an array of Histograms.
@@ -43,7 +105,7 @@ func NewHistogram(labels map[string]string) *Histogram {
 func ConvertSampleToBucket(sample *model.Sample, h *HistogramVec) {
 	labels := make(map[string]string)
 	for k, v := range sample.Metric {
-		if k != "le" {
+		if k != model.BucketLabel {
 			labels[string(k)] = string(v)
 		}
 	}
@@ -58,5 +120,18 @@ func ConvertSampleToBucket(sample *model.Sample, h *HistogramVec) {
 		hist = NewHistogram(labels)
 		*h = append(*h, *hist)
 	}
-	hist.Buckets[string(sample.Metric["le"])] += int(sample.Value)
+	hist.Buckets[string(sample.Metric[model.BucketLabel])] += int(sample.Value)
+}
+
+// ConvertSampleToHistogram converts prometheus sample into Histogram.
+func ConvertSampleToHistogram(sample *model.Sample, h *Histogram) {
+	labels := make(map[string]string)
+	for k, v := range sample.Metric {
+		if k != model.BucketLabel {
+			labels[string(k)] = string(v)
+		}
+	}
+
+	h.Labels = labels
+	h.Buckets[string(sample.Metric[model.BucketLabel])] += int(sample.Value)
 }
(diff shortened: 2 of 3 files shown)


perfdash/parser.go
```diff
@@ -264,6 +265,29 @@ func parseOperationLatency(latency latencyMetric, operationName string) perftype
...
+			klog.Errorf("err in build %d: no +Inf bucket: %v", buildNumber, hist)
+			continue
+		}
+		for kBucket, vBucket := range hist[i].Buckets {
+			if kBucket != "+Inf" {
+				if count == 0 {
+					perfData.Data["<= "+kBucket+"s"] = 0
+					continue
+				}
+				perfData.Data["<= "+kBucket+"s"] = float64(vBucket) / float64(count) * 100
```
liu-cong: This is calculating percentages in buckets? I don't think this is useful. We are interested in average latency and percentiles.

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -216,5 +219,11 @@ type schedulingMetrics struct {
 	PriorityEvaluationLatency   measurementutil.LatencyMetric `json:"priorityEvaluationLatency"`
 	PreemptionEvaluationLatency measurementutil.LatencyMetric `json:"preemptionEvaluationLatency"`
 	BindingLatency              measurementutil.LatencyMetric `json:"bindingLatency"`
-	E2eSchedulingLatency        measurementutil.LatencyMetric `json:"e2eSchedulingLatency"`
+	E2eSchedulingLatency        measurementutil.HistogramVec  `json:"e2eSchedulingLatency"`
```
liu-cong: I think a cleaner design is to make the E2eSchedulingLatency conform to the LatencyMetric struct (you may want to add Average to the existing LatencyMetric since average is important too). And you shouldn't need any changes on the perfdash side. 

BTW, this [PR](https://github.com/kubernetes/kubernetes/pull/85861) has logic on how to approximately calculate percentiles from Histograms. 
If it's too complex, feel free to just implement average latency without percentiles.

perfdash/parser.go
yiyang5055: should calculate percentiles from Histograms, rather than change E2eSchedulingLatency from  LatencyMetric to HistogramVec? @liu-cong

clusterloader2/pkg/measurement/common/scheduler_latency.go
yiyang5055: Got it,  I misunderstood it.

perfdash/parser.go
liu-cong: Yeah end goal is to have latency percentile and average. Changing E2eSchedulingLatency type from LatencyMetric to HistogramVec is implementation detail, which I have an alternative suggestion.

I think my other comment clarifies this. Let me know if you still have questions.

yiyang5055: thank you, no questions

yiyang5055:
@liu-cong  I revised this pr to calculate percentiles from Histograms, please rereview

liu-cong:
I discussed this with @ahg-g and right now it's not clear if it's "legal" to use the code to calculate percentiles from histograms because that's from Prometheus. 
What we can do now is:
* Revert back to use the buckets, there is a  [parseHistogramMetric](https://github.com/kubernetes/perf-tests/blob/982668b933540721dfc713627697fbdc6e9af5c2/perfdash/parser.go#L326) function to use. I think you initial PR uses that or similar.
* For the bucket distribution, in order not to confuse with percentiles and averages, we can name it `<metric name> Bucket Distribution` in the dashboard.
* (Future work) we can add a new dashboard page with analytics such as  percentiles and averages. We will have analytics and distribution pages at the same time since distributions are accurate, while analytics are more intuitive to reason about.

Sorry for the churn. But hopefully we will still be able to use your work once we figure out the legal stuff. Please save your work :)

yiyang5055:
ok, i start a new pr and add a new metric name for use histograms, this pr could keep it. @liu-cong

ahg-g:
> (Future work) we can add a new dashboard page with analytics such as percentiles and averages. We will have analytics and distribution pages at the same time since distributions are accurate, while analytics are more intuitive to reason about.

Yes, it turns out we can do something about that prometheus code, see suggestion here: https://github.com/kubernetes/kubernetes/pull/85861/files#r367096687

yiyang5055:
> > (Future work) we can add a new dashboard page with analytics such as percentiles and averages. We will have analytics and distribution pages at the same time since distributions are accurate, while analytics are more intuitive to reason about.
> 
> Yes, it turns out we can do something about that prometheus code, see suggestion here: https://github.com/kubernetes/kubernetes/pull/85861/files#r367096687

ok, thank you,  i got it.

yiyang5055:
take a look, I fork the quantile.go into a new subdir of `third_party/forked/`, thanks @ahg-g  @liu-cong

ahg-g:
> take a look, I fork the quantile.go into a new subdir of `third_party/forked/`, thanks @ahg-g @liu-cong

Thanks, but we should fork it in ```https://github.com/kubernetes/kubernetes/tree/master/third_party/forked``` and the clusterloader imports the forked package from kubernetes.

ahg-g:
@yiyang5055 thanks a lot for all the effort, so we reached a decision on the other PR to implement our own function to compute quantiles, [here is the implementation](https://github.com/kubernetes/kubernetes/pull/85861/files#diff-8fc16044937fc03c39b585a657b20150R179-R209). 

Can you please remove all the files under forked and just copy that function to ```clusterloader2/pkg/measurement/util/histogram.go```

ahg-g:
@ingvagabund would you like to pick this up if @yiyang5055 can't continue to work on it?

We need to do something similar to what you did in the other metrics PR: calculate percentiles of e2e scheduling latency to show on perf dashboard.

ingvagabund:
> @ingvagabund would you like to pick this up if @yiyang5055 can't continue to work on it?

Yes, count me in

yiyang5055:
@ahg-g sorry, It was my vacation recently, and i will continue to work on it as soon as possible.

yiyang5055:
> @yiyang5055 thanks a lot for all the effort, so we reached a decision on the other PR to implement our own function to compute quantiles, [here is the implementation](https://github.com/kubernetes/kubernetes/pull/85861/files#diff-8fc16044937fc03c39b585a657b20150R179-R209).
> 
> Can you please remove all the files under forked and just copy that function to `clusterloader2/pkg/measurement/util/histogram.go`

@ahg-g done

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -156,6 +163,20 @@ func (s *schedulerLatencyMeasurement) getSchedulingLatency(c clientset.Interface
...
 
+// Set quantile of LatencyMetric from Histogram
+func (s *schedulerLatencyMeasurement) setQuantileFromHistogram(metric *measurementutil.LatencyMetric, hist *measurementutil.Histogram) error {
+	quantiles := []float64{0.5, 0.9, 0.99}
+	for _, quantile := range quantiles {
+		q, err := hist.Quantile(quantile)
+		if err != nil {
+			return err
+		}
+		metric.SetQuantile(quantile, time.Duration(int64(q*float64(time.Second))))
```
ahg-g: shouldn't this be ```q``` instead of ```quantile```?

clusterloader2/pkg/measurement/util/histogram.go
```diff
@@ -60,3 +113,16 @@ func ConvertSampleToBucket(sample *model.Sample, h *HistogramVec) {
 	}
 	hist.Buckets[string(sample.Metric["le"])] += int(sample.Value)
 }
+
+// ConvertSampleToHistogram converts prometheus sample into Histogram.
+func ConvertSampleToHistogram(sample *model.Sample, h *Histogram) {
+	labels := make(map[string]string)
+	for k, v := range sample.Metric {
+		if k != "le" {
```
ahg-g: what is "le"? define a constant inside the function for it to better understand what it represents.

yiyang5055:
@ahg-g @ingvagabund i found this function have a problem, [here is the implement](https://github.com/kubernetes/kubernetes/pull/85861/files#diff-8fc16044937fc03c39b585a657b20150R179-R209), and It's unstable. the results of each calculation are different even though the same input.

ahg-g:
> @ahg-g @ingvagabund i found this function have a problem, [here is the implement](https://github.com/kubernetes/kubernetes/pull/85861/files#diff-8fc16044937fc03c39b585a657b20150R179-R209), and It's unstable. the results of each calculation are different even though the same input.

Thanks, @ingvagabund we should add unit tests to your PR, sorry didn't notice this earlier.

ingvagabund:
> @ahg-g @ingvagabund i found this function have a problem, here is the implement, and It's unstable. the results of each calculation are different even though the same input.

@yiyang5055 can you provide the input which makes the function unstable?

clusterloader2/pkg/measurement/util/histogram.go
```diff
@@ -17,17 +17,70 @@ limitations under the License.
...
 // Histogram is a structure that represents distribution of data.
 type Histogram struct {
 	Labels  map[string]string `json:"labels"`
 	Buckets map[string]int    `json:"buckets"`
 }
 
+// Quantile calculates the quantile 'q' based on the given buckets of Histogram.
+func (h *Histogram) Quantile(q float64) (float64, error) {
+	var b []bucket
+	for k, v := range h.Buckets {
```
ingvagabund: I see. `h.Buckets` is a map so the individual buckets are processed randomly. You need to make sure the list of  buckets that goes into `bucketQuantile` is sorted by bucket's upper bound. That's why I created `Validate` function to make sure the buckets are properly sorted.

yiyang5055: yeah, the buckets parameter passed to the bucketQuantile function is not sorted by bucket's upperbound, but i think we should sort list of buckts in bucketQuantile to ensure the correctness of result.

ingvagabund: Implementation of `bucketQuantile` you copy-pasted is private. In my case all the buckets are always properly sorted. So it's safe to call it as it is. In your case where the source histogram buckets are represented as map, it's your implementation detail to make sure the buckets are properly sorted.

ahg-g: @ingvagabund if we create a ```NewHistogram``` function to create a [```Histogram```](https://github.com/kubernetes/kubernetes/pull/87923/files#diff-be7d56a59338e31acbaf032075732116R188) instance, then this PR can import your implementation instead of copying it.

ingvagabund: +1 for it, just finishing the unit tests

ingvagabund: This is the best I can offer right now: https://github.com/kubernetes/kubernetes/pull/87923/files#diff-be7d56a59338e31acbaf032075732116R234

`Histogram` data type under `k8s.io/component-base/metrics/testutil` is a wrapper around prometheus's Histogram. It needs to be a wrapper so I can clear all histogram samples and reset it to its initial state. Thus, `GetHistogramFromBuckets` will accept a list of buckets (which is generic enough to be used for various histogram source formats). However, computing average will give nonsensical values since we don't have the samples to get overall sample sum.

ahg-g:
The other PR merged, can you please rebase and import implementation? You will need to create a storage.Hiistogram from the samples, ordered by upper bound, can you do that, or should @ingvagabund take over?

ingvagabund:
@yiyang5055 feel free to lemme know if you need any help

clusterloader2/pkg/measurement/util/histogram.go
yiyang5055: "le" is a label that  defines the upper bound of a bucket of a histogram ("le" -> "less or equal").  it define in  `github.com/prometheus/common/model` , i will change to use `model.BucketLabel`.

yiyang5055:
@ahg-g I sort buckets before calling the bucketQuantile function, and make sure the list of buckets that goes into bucketQuantile is sorted by bucket's upper bound.  is ok for you?  If you don't ok, let @ingvagabund  take over it.

ahg-g:
> @ahg-g I sort buckets before calling the bucketQuantile function, and make sure the list of buckets that goes into bucketQuantile is sorted by bucket's upper bound. is ok for you? otherwise let @ingvagabund take over it.

Thanks for the effort. @ingvagabund can you try and rework this PR to import your implementation of the quantiles instead of copying it.

ingvagabund:
> Thanks for the effort. @ingvagabund can you try and rework this PR to import your implementation of the quantiles instead of copying it.

Sure thing. @yiyang5055 I can either give you diff which you can apply or can open another PR and include yours in it. Which one do you prefer?

ahg-g:
> > Thanks for the effort. @ingvagabund can you try and rework this PR to import your implementation of the quantiles instead of copying it.
> 
> Sure thing. @yiyang5055 I can either give you diff which you can apply or can open another PR and include yours in it. Which one do you prefer?

@ingvagabund can you please send your PR for review? I don't think we have enough time to go back and forth on this because the freeze is very close.

ingvagabund:
> @ingvagabund can you please send your PR for review? I don't think we have enough time to go back and forth on this because the freeze is very close.

https://github.com/kubernetes/perf-tests/pull/1061

ingvagabund:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/978/pull-perf-tests-clusterloader2/1230833030562582528/artifacts/SchedulingMetrics_load_2020-02-21T12:59:01Z.json

@ahg-g even here you get:
```json
{
  "predicateEvaluationLatency": {
    "Perc50": 0,
    "Perc90": 0,
    "Perc99": 0
  },
  "priorityEvaluationLatency": {
    "Perc50": 0,
    "Perc90": 0,
    "Perc99": 0
  },
  "preemptionEvaluationLatency": {
    "Perc50": 0,
    "Perc90": 0,
    "Perc99": 0
  },
  "bindingLatency": {
    "Perc50": 0,
    "Perc90": 0,
    "Perc99": 0
  },
  "e2eSchedulingLatency": {
    "Perc50": -9223372036854775808,
    "Perc90": -9223372036854775808,
    "Perc99": -9223372036854775808
  }
}
```

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -38,7 +38,7 @@ import (
 const (
 	schedulerLatencyMetricName = "SchedulingMetrics"
 
-	e2eScheduling = "e2eScheduling"
+	e2eSchedulingMetricName = "e2e_scheduling_duration_seconds_bucket"
```
ahg-g: you need to prefix this with "scheduler_"

PR #979, opened 2020-01-14
Simplify logging in the isNodeConditionSetAsExpected

mm4tt:
It's an anti-pattern to control level of logging via method arguments, a logging framework features should be used for that.
Changing the log level to V(4) so it can be turned on/off by setting the --v flag appropriately.

A justification for making this change is the CA tests where nodes appear and disappear, in the current for this logging spams the CL2 output.

### DIFF
--- a/clusterloader2/pkg/util/cluster.go
+++ b/clusterloader2/pkg/util/cluster.go
@@ -80,9 +80,9 @@ func IsNodeSchedulableAndUntainted(node *corev1.Node) bool {
 // 2) it's Ready condition is set to true
 // 3) doesn't have NetworkUnavailable condition set to true
 func isNodeSchedulable(node *corev1.Node) bool {
-	nodeReady := isNodeConditionSetAsExpected(node, corev1.NodeReady, true, false)
+	nodeReady := isNodeConditionSetAsExpected(node, corev1.NodeReady, true)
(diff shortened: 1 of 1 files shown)


oxddr:
Thanks for the cleanup. Similar logic is used in some other places in CL2 - we should remove it too (not in this PR though). 

/lgtm
/approve

clusterloader2/pkg/util/cluster.go
```diff
@@ -96,24 +96,20 @@ func isNodeUntainted(node *corev1.Node) bool {
...
 		// Ensure that the condition type and the status matches as desired.
 		if cond.Type == conditionType {
 			if wantTrue == (cond.Status == corev1.ConditionTrue) {
 				return true
 			}
-			if !silent {
-				klog.Infof("Condition %s of node %s is %v instead of %t. Reason: %v, message: %v",
-					conditionType, node.Name, cond.Status == corev1.ConditionTrue, wantTrue, cond.Reason, cond.Message)
-			}
+			klog.V(4).Infof("Condition %s of node %s is %v instead of %t. Reason: %v, message: %v",
```
wojtek-t: There are couple more places in perf-tests repo that do that - can you please also fix them (or open an issue)?

oxddr: https://github.com/kubernetes/perf-tests/issues/980

PR #984, opened 2020-01-16
update scheduler metrics

xiaoanyunfei:
What type of PR is this?

/kind cleanup

Fixes #https://github.com/kubernetes/kubernetes/issues/86969


perfdash/parser.go
```diff
@@ -245,23 +245,36 @@ type latencyMetric struct {
 }
 
 type schedulingMetrics struct {
-	PredicateEvaluationLatency  latencyMetric `json:"predicateEvaluationLatency"`
-	PriorityEvaluationLatency   latencyMetric `json:"priorityEvaluationLatency"`
-	PreemptionEvaluationLatency latencyMetric `json:"preemptionEvaluationLatency"`
-	BindingLatency              latencyMetric `json:"bindingLatency"`
-	E2eSchedulingLatency        latencyMetric `json:"e2eSchedulingLatency"`
-	ThroughputAverage           float64       `json:"throughputAverage"`
-	ThroughputPerc50            float64       `json:"throughputPerc50"`
```
mm4tt: Were these fields not used anywhere? :)

perfdash/parser.go
```diff
@@ -245,23 +245,36 @@ type latencyMetric struct {
...
+	E2eSchedulingLatency        histogramVec `json:"e2eSchedulingLatency"`
 }
 
-func parseOperationLatency(latency latencyMetric, operationName string) perftype.DataItem {
-	perfData := perftype.DataItem{Unit: "ms", Labels: map[string]string{"Operation": operationName}, Data: make(map[string]float64)}
-	perfData.Data["Perc50"] = float64(latency.Perc50) / float64(time.Millisecond)
-	perfData.Data["Perc90"] = float64(latency.Perc90) / float64(time.Millisecond)
-	perfData.Data["Perc99"] = float64(latency.Perc99) / float64(time.Millisecond)
-	return perfData
+func parseOperationLatency(histogramVecMetric histogramVec, operationName string) perftype.DataItem {
```
mm4tt: IIUC, this changes are backward compatible, your're just adding a new metric (SchedulingAlgorithmLatency) and refactoring this method to make it more generic.

If that's the case, could you lease send the perf-dash changes in a separate PR? We'll submit and deploy a new version of it and verify that the change doesn't break anything (unfortunately, we don't have any e2e tests for perf-dash). Once it's done we can submit the CL2 changes

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -119,35 +114,37 @@ func (s *schedulerLatencyMeasurement) getSchedulingLatency(c clientset.Interface
...
+			hist = &result.PreemptionEvaluationLatency
+		case "scheduler_binding_duration_seconds_bucket":
+			hist = &result.BindingLatency
+		case "scheduler_scheduling_algorithm_duration_seconds_bucket":
+			hist = &result.SchedulingAlgorithmLatency
+		case "scheduler_e2e_scheduling_duration_seconds_bucket":
+			hist = &result.E2eSchedulingLatency
 		}
 
-		var metric *measurementutil.LatencyMetric
```
liu-cong: Let's keep the existing summary metrics to not lose history, and add the new ones separately. We can remove the summary metrics once they are removed from scheduler.

perfdash/parser.go
liu-cong: +1

Can we refactor the existing `parseHistogramMetric` function to make it more generic?

https://github.com/kubernetes/perf-tests/blob/982668b933540721dfc713627697fbdc6e9af5c2/perfdash/parser.go#L326

PR #988, opened 2020-01-16
Slow down the speed of pods creation to 1 pod / 30s

YangLu1031:
Slow down the pods creation to 1 pod / 30s to avoid the PLEG not healthy issue.


YangLu1031:
@mm4tt @oxddr @mborsz Was working with Wojciech before and thanks to his fast reviews, but he seems OOO for the next week, could you help to review this PR, need to see the test results. Thanks!

clusterloader2/testing/node-throughput/windows_override.yaml
```diff
@@ -1,5 +1,5 @@
-POD_COUNT: 20
-POD_THROUGHPUT: 2
+POD_COUNT: 80
+POD_THROUGHPUT: 0.03
```
mm4tt: Creating 80 pods with 0.03 pods/s throughput would take 80/0.03/60 = ~44 hours. 
Why do you need to lower the pod throughput that much?

mm4tt:
/approve

/hold
Could you update the PR description to explain better the justification behind making these changes? E.g. answering the question I asked in the other comment. Feel free to void the hold once you do it, thanks!

clusterloader2/testing/node-throughput/windows_override.yaml
pjh: Lowering the throughput by this much basically eliminates a variable. Yang's testing so far found that creating Windows pods too rapidly consistently leads the Windows node to become Unhealthy with a status like "PLEG is not healthy: pleg was last seen active 14m57.9413932s ago; threshold is 3m0s".

Dropping the rate significantly lets us crank up the variable that we really care about for now: the maximum number of pods we can run on the Windows node. We'll come back and evaluate + investigate the pod creation rate more thoroughly later. This has been done for Linux [in the past](https://github.com/kubernetes/kubernetes/issues/27388) but not Windows.

PR #991, opened 2020-01-17
Enable HA support for resource metric measurements

alejandrox1:
xref: https://github.com/kubernetes/perf-tests/issues/246

### DIFF
--- a/clusterloader2/pkg/measurement/common/resource_usage.go
+++ b/clusterloader2/pkg/measurement/common/resource_usage.go
@@ -47,7 +47,7 @@ func createResourceUsageMetricMeasurement() measurement.Measurement {
 }
 
 type resourceUsageMetricMeasurement struct {
-	gatherer            *gatherers.ContainerResourceGatherer
+	gatherers           []*gatherers.ContainerResourceGatherer
 	resourceConstraints map[string]*measurementutil.ResourceConstraint
 }
 
@@ -66,10 +66,9 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Measurement
 		if err != nil {
 			return nil, err
 		}
-		host, err := util.GetStringOrDefault(config.Params, "host", config.ClusterFramework.GetClusterConfig().GetMasterIp())
-		if err != nil {
-			return nil, err
-		}
+
+		hosts := config.ClusterFramework.GetClusterConfig().MasterIPs
+
 		nodeMode, err := util.GetStringOrDefault(config.Params, "nodeMode", "")
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/resource_usage.go
```diff
@@ -104,34 +103,48 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Measurement
...
+		}
+
+		for i := range e.gatherers {
+			go e.gatherers[i].StartGatheringData()
 		}
-		go e.gatherer.StartGatheringData()
 		return nil, nil
 	case "gather":
-		if e.gatherer == nil {
+		if e.gatherers == nil || len(e.gatherers) < 1 {
```
mm4tt: NIT: You can do simply
```
if len(e.gatherers) == 0 {
...
}
```

as `len(nil)` returns 0

clusterloader2/pkg/measurement/common/resource_usage.go
```diff
@@ -104,34 +103,48 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Measurement
...
+		for i := range e.gatherers {
+			summary, err := e.gatherers[i].StopAndSummarize([]int{50, 90, 99, 100})
+			if err != nil {
+				return nil, err
+			}
+			content, err := util.PrettyPrintJSON(summary)
+			if err != nil {
+				return nil, err
+			}
+			resourceSummary := measurement.CreateSummary(resourceUsageMetricName, "json", content)
```
mm4tt: Can we add the "host" prefix to the summary name so we can distinguish them?

clusterloader2/pkg/measurement/common/resource_usage.go
```diff
@@ -104,34 +103,48 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Measurement
...
-			InKubemark:                        strings.ToLower(provider) == "kubemark",
-			Nodes:                             nodesSet,
-			ResourceDataGatheringPeriod:       60 * time.Second,
-			MasterResourceDataGatheringPeriod: 10 * time.Second,
-			PrintVerboseLogs:                  false,
-		}, nil)
-		if err != nil {
-			return nil, err
+		for _, host := range hosts {
+			gatherer, err := gatherers.NewResourceUsageGatherer(config.ClusterFramework.GetClientSets().GetClient(), host, provider, gatherers.ResourceGathererOptions{
```
mm4tt: I checked the code of the ContainerResourceGatherer and it's a bit messy (sorry about that!) It looks like `host` field is used only when the `InKubemark` option is true. Otherwise the host field is ignored and depending on the `Nodes` strategy different set of nodes is being monitored in terms of resource usage.

Given that, could you first refactor the `container_resource_gatherer.go` file to make the `host` field optional (part of the `ResourceGathererOptions`). Actually we can name it KubemarkMasterHost (and remove `InKubemark` field) to clean it up a bit. Once we do it it should be easier to refactor this file. 

Thanks!

alejandrox1: Absolutely, will start working on that! and thank you for your review.

mm4tt: Thanks @alejandrox1 !

wojtek-t:
@alejandrox1 - will you have time to get back to it?

PR #994, opened 2020-01-23
[ClusterLoader2] change slice related Go template functions

jprzychodzen:


mm4tt:
Could you provide some examples of how the new template functions should be used? And what is the problem with Seq you're trying to address?

clusterloader2/pkg/config/template_functions.go
```diff
@@ -55,14 +57,18 @@ func GetFuncs() template.FuncMap {
...
 		"Seq":           seq,
+		"SliceOfZeros":  sliceOfZeros,
 		"SubtractFloat": subtractFloat,
 		"SubtractInt":   subtractInt,
 		"YamlQuote":     yamlQuote,
 	}
 }
 
+// seq returns a slice of size 'size' filled with zeros.
+// Deprecated: Naming generates confusion. Please use 'SliceOfZeros' for explicit zero values or 'Loop' for.
```
mm4tt: ... or 'Loop' for. 

Isn't there missing anything?

mm4tt:
/approve
/lgtm

/hold

Feel free to cancel the hold once you fix the comment.

jprzychodzen:
Just for a record:
`Seq` requires Go syntax to provide incremental integers - it's an error prone approach. With a single value rage, a number of `0` values is generated, and this is counter intuitive.

Example from clusterloader2/testing/experimental/storage/pod-startup/volume-types/downwardapi/deployment_with_downwardapi.yaml
Old syntax:
```yaml
{{ range $volumeIndex, $vol := Seq .VolumesPerPod }}
- name: vol-{{$volumeIndex}}
  mountPath: /usr/share/{{$volumeIndex}}
{{ end }}
```
`$vol` is discarded as not used

New syntax:
```yaml
{{ range $volumeIndex := Loop .VolumesPerPod }}
- name: vol-{{$volumeIndex}}
  mountPath: /usr/share/{{$volumeIndex}}
{{ end }}
```

clusterloader2/pkg/config/template_functions.go
jprzychodzen: It was missing.

Issue #985, opened 2020-01-16
CL2 kube-apiserver profiles are not collected


mm4tt:
/kind bug

This is a result of recent refactoring. It's not that simple to get the profiles using curl as we need to authenticate (or allow unauthenticated users to call /debug/profile path).
I believe a solution to that would be to switch back to collecting profiles via k8s client (as we used to). This wouldn't work in HA setups but there is already a bug for that (https://github.com/kubernetes/perf-tests/issues/246)

alejandrox1:
There is this note from the profile code https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/profile.go#L182-L184
```go
// TODO(#246): This will connect to a random master in HA (multi-master) clusters, fix it.
if p.config.componentName == "kube-apiserver" {
    body, err := c.CoreV1().RESTClient().Get().AbsPath("/debug/pprof/" + p.config.kind).DoRaw()				
```
When you say 
> I believe a solution to that would be to switch back to collecting profiles via k8s client (as we used to).

Does that mean that we should use that method to do so?

Currently that part of the logic is executed if `p.config.provider == "gke"`, could we also not add some other piece of configuration that will allow the user to use the client to collectprofile data?

mm4tt:
Yeah, the problem is that we cannot access api-server debug endpoint without authenticating. It means that we cannot easily use curl to do that. 
A partial solution would be to switch the code back to using k8s client, but as I documented this wouldn't work in HA env.

mm4tt:
And to clarify, we should go with this partial solution, at least for now. This is what this issue is about.

PR #995, opened 2020-01-24
Fix gathering apiserver profiles

mm4tt:
Fixes https://github.com/kubernetes/perf-tests/issues/985

### DIFF
--- a/clusterloader2/pkg/measurement/common/profile.go
+++ b/clusterloader2/pkg/measurement/common/profile.go
@@ -21,8 +21,12 @@ import (
 	"sync"
 	"time"
 
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


mm4tt:
/hold 

Let's verify the presubmits

mm4tt:
/hold cancel
Profiles look good

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -91,6 +95,12 @@ func (p *profileMeasurement) start(config *measurement.MeasurementConfig) error
 		klog.Warning("Profile measurements will be disabled due to no MasterIps")
 		return nil
 	}
+	c := config.ClusterFramework.GetClientSets().GetClient()
```
jkaniuk: Any reason to put it so early in execution?
It is hard to look for "c" so maybe I am missing some use of this client. Could you use more descriptive name?

mm4tt: Reason is that it's used a few lines below.

I renamed to k8sClient to make it more descriptive, thanks!

jkaniuk: Thx. Now I see it :)

PR #999, opened 2020-01-29
Log response' body from a call to Prometheus (proxied by apiserver).

oxddr:
/assign @mm4tt


mborsz:
I was debugging a case today when this would be useful.

Thanks, good idea!

PR #1000, opened 2020-01-29
Move probes to util-images dir

mm4tt:
Thanks to @jprzychodzen changes we can leverage the common util-images build presubmit and we don't need to maintain the vendor dir anymore.


wojtek-t:
Nice

/lgtm

PR #1001, opened 2020-01-29
Retry 409 resource-quotas errors in CL2

mm4tt:
We noticed runs where some calls failed with:
```
measurement call InClusterNetworkLatency - InClusterNetworkLatency error: error while applying (/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/ping-client-service.yaml): Operation cannot be fulfilled on resourcequotas "gke-resource-quotas": the object has been modified; please apply your changes to the latest version and try again
```

This seems to be the exactly the situation described in https://github.com/kubernetes/kubernetes/issues/67761

### DIFF
--- a/clusterloader2/pkg/framework/client/objects.go
+++ b/clusterloader2/pkg/framework/client/objects.go
@@ -64,6 +64,8 @@ func IsRetryableAPIError(err error) bool {
 	// These errors may indicate a transient error that we can retry in tests.
 	if apierrs.IsInternalError(err) || apierrs.IsTimeout(err) || apierrs.IsServerTimeout(err) ||
 		apierrs.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) ||
+		// Retryable resource-quotas conflict errors may be returned in some cases, e.g. https://github.com/kubernetes/kubernetes/issues/67761
+		isResourceQuotaConflictError(err) ||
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/framework/client/objects.go
```diff
@@ -64,6 +64,8 @@ func IsRetryableAPIError(err error) bool {
 	// These errors may indicate a transient error that we can retry in tests.
 	if apierrs.IsInternalError(err) || apierrs.IsTimeout(err) || apierrs.IsServerTimeout(err) ||
 		apierrs.IsTooManyRequests(err) || utilnet.IsProbableEOF(err) || utilnet.IsConnectionReset(err) ||
+		// 409 error may be returned in some cases, e.g. https://github.com/kubernetes/kubernetes/issues/67761
+		apierrs.IsConflict(err) ||
```
wojtek-t: This one is fairly tricky - not every conflict will be automatically retryable.
I'm slightly afraid of doing that automatically for every case.

Instead of that, we may allow that explicitly in specific calls...

wojtek-t:
Thanks

/lgtm
/hold cancel

TBBle:
If the logs from the original failures are still around, are they only replicating when deploying Services? Or was that just a lucky example in the original report?

mm4tt:
We don't have logs anymore, but Services were just a "lucky example",  the error could have happened for other resource types as well.

TBBle:
Drat. That widens the scope of causes a bit. >_<

Thank you.

Issue #1002, opened 2020-01-30
Remove apiserver_request_total's client label in master-dashboard


mm4tt:
The client label of the apiserver_request_total metric is being removed in https://github.com/kubernetes/kubernetes/pull/87669.
This will break some of the graphs in the master-dashboard.json. We should update the prometheus queries to not use the client label.

mm4tt:
Thanks @hypnoglow for looking into this!

mm4tt:
This is done, IIRC.

/close

Thanks, @hypnoglow!

PR #1010, opened 2020-02-03
dashboards: remove apiserver_request_total client label

hypnoglow:
Related issue: https://github.com/kubernetes/perf-tests/issues/1002

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
+++ b/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
@@ -34,10 +34,6 @@ def api_call_latency(title, verb, scope, threshold):
 
 
 CLUSTERLOADER_PANELS = [
-    d.simple_graph(
-        "Requests",
-        'sum(irate(apiserver_request_total{client="clusterloader/v0.0.0 (linux/amd64) kubernetes/$Format", verb!="CONNECT"}[1m])) by (verb, resource)',
-    ),
     api_call_latency(
         title="Read-only API call latency (percentaile=99, scope=resource, threshold=1s)",
         verb="GET",
@@ -72,12 +68,12 @@ HEALTH_PANELS = [
     ),
     d.simple_graph(
         "Pod creations",
-        'sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource=""}[1m])) by (client)',
+        'sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource=""}[1m]))',
         yAxes=g.single_y_axis(format=g.OPS_FORMAT),
     ),
     d.simple_graph(
         "Pod bindings",
-        'sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource="binding"}[1m])) by (client)',
+        'sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource="binding"}[1m]))',
         yAxes=g.single_y_axis(format=g.OPS_FORMAT),
     ),
     # It's not clear which "Component restarts" shows more accurate results.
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -36,7 +36,7 @@ def api_call_latency(title, verb, scope, threshold):
 CLUSTERLOADER_PANELS = [
```
mm4tt: Actually, let's get rid of this whole panel. Without being able to plot requests coming from clusterloader there is no point in having it.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -72,12 +72,12 @@ def api_call_latency(title, verb, scope, threshold):
     ),
     d.simple_graph(
         "Pod creations",
-        'sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource=""}[1m])) by (client)',
```
mm4tt: Let's not get rid of the sum aggregation. We want to see an overall pod creation rate in the cluster, without sum you'll get a separate graph for every unique tuple of prometheus labels, e.g. we can have multiple api-servers with different labels, etc.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -72,12 +72,12 @@ def api_call_latency(title, verb, scope, threshold):
...
+        'irate(apiserver_request_total{verb="POST", resource="pods", subresource=""}[1m])',
         yAxes=g.single_y_axis(format=g.OPS_FORMAT),
     ),
     d.simple_graph(
         "Pod bindings",
-        'sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource="binding"}[1m])) by (client)',
```
mm4tt: Same here

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
hypnoglow: Removed and `CLUSTERLOADER_PANELS ` and `api_call_latency` definition that is not used anymore.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
hypnoglow: Fixed, now only `by (client)` is dropped.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -18,52 +18,6 @@
...
-                % {"verb": verb, "scope": scope},
-                legendFormat="{{verb}} {{scope}}/{{resource}}",
-            ),
-        ],
-        yAxes=g.single_y_axis(format=g.SECONDS_FORMAT),
-    )
-
-
-CLUSTERLOADER_PANELS = [
-    d.simple_graph(
```
mm4tt: Sorry, my bad. I meant to delete only this graph (requests), the other graphs (api-call-latency graphs) are actually useful and don't depend on the client label.

hypnoglow: No problem, `CLUSTERLOADER_PANELS` is back with only first graph deleted.

mm4tt:
/lgtm
/approve

Thanks, @hypnoglow !

wojtek-t:
Closed and reopened to trigger travis.

PR #1006, opened 2020-02-03
Independence from 'verify/verify-flags/known-flags.txt'

ChakshuGupta13:
Remove dependence from 'verify/verify-flags/known-flags.txt' of 'verify/verify-flags-underscore.py'. The 'verify/verify-flags-underscore.py' script will have to just go over all *.go files in the repo (except certain directories and files) and check that all flags are in the proper format.

Fix #997

PR #1011, opened 2020-02-04
[Access-tokens] Update access-tokens to version v0.0.6

jprzychodzen:
Changelog:
* Add layer of parallelization per token
* Make requests cheaper with  ResoruceVersion : 0

/sig scalability

### DIFF
--- a/util-images/access-tokens/cmd/main.go
+++ b/util-images/access-tokens/cmd/main.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	certutil "k8s.io/client-go/util/cert"
+	flowcontrol "k8s.io/client-go/util/flowcontrol"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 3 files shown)


mm4tt:
Please revert your changes in known-flags.txt, it's not needed to update it anymore, see https://github.com/kubernetes/perf-tests/pull/1006

jkaniuk:
> Please revert your changes in known-flags.txt, it's not needed to update it anymore, see #1006

Why do we have it then?

mm4tt:
@jkaniuk, we don't have it anymore, it was removed in https://github.com/kubernetes/perf-tests/pull/1006

util-images/access-tokens/cmd/main.go
```diff
@@ -103,5 +109,6 @@ func newConfig(tokenFile, rootCAFile string) (*rest.Config, error) {
 		TLSClientConfig: tlsClientConfig,
 		BearerToken:     string(token),
 		BearerTokenFile: tokenFile,
+		RateLimiter:     flowcontrol.NewFakeAlwaysRateLimiter(),
```
mm4tt: Cannot you just set QPS: *qps here?

util-images/access-tokens/cmd/main.go
```diff
@@ -29,18 +29,21 @@ import (
...
 )
 
 var (
-	namespace = flag.String("namespace", "", "Namespace that is targeted with request flow.")
-	tokenDirs = flag.StringSlice("access-token-dirs", nil, "Directories with access-token. A worker will be started per each dir/token")
-	qps       = flag.Float64("qps-per-worker", 0.5, "QPS per worker")
+	namespace       = flag.String("namespace", "", "Namespace that is targeted with request flow.")
+	tokenDirs       = flag.StringSlice("access-token-dirs", nil, "Directories with access-token. A worker will be started per each dir/token")
+	qps             = flag.Float64("qps-per-worker", 0.5, "QPS per worker")
+	workersPerToken = flag.Int("workers-per-token", 1, "Number of workers using the same token")
```
mm4tt: Why do we need this additional layer here of parallelism here? Is there any use-case for it?

The downside of it is that it's getting more complicated, total qps is now: #tokens * qps * worksersPerToken

util-images/access-tokens/cmd/main.go
jprzychodzen: It's more complicated than that:
  1. We are already rate limiting in `func worker` - we do not need another limiter here
  2. Setting a QPS would require to set Burst parameter. It raises a question what would be sane value for Burst.

mm4tt: Setting burst to 1 should work, we need 1 initial request and then the the rate limiter's bucket should be refilled with the exact qps we use.

But I see your point and I agree. There is no point in having two rate-limiting mechanism.

Could you just add a comment here explaining that, basically what you wrote in point 1. ?

jprzychodzen: Added, done.

util-images/access-tokens/cmd/main.go
jprzychodzen: It was mostly related to broken implementation of `func worker` that could be throttled in a case if server in not responding fast enough. Changed implementation to asynchronous and removed `workersPerToken` flag

PR #1012, opened 2020-02-04
Change formater flag to one that actually rewrites

jprzychodzen:


verify/verify-gofmt.sh
```diff
@@ -39,7 +39,7 @@ find_files() {
     \) -name '*.go'
 }
 
-GOFMT="gofmt -s"
+GOFMT="gofmt -w"
```
wojtek-t: This is verify script - if you want to rewrite one, please create update-gofmt.sh

[BTW, the -s option is actually useful even with -w]

PR #1021, opened 2020-02-05
[Access-tokens][ClusterLoader2] Add standalone test

jprzychodzen:
Add standalone test for ClusterLoader2 that allows benchmarking and
testing access tokens thresholds.

This test is based on Mateusz's code from github.com/mm4tt/k8s-util.

### DIFF
--- a/clusterloader2/testing/access-tokens/config.yaml
+++ b/clusterloader2/testing/access-tokens/config.yaml
@@ -1,0 +1,179 @@
+# Stress testing access token validation
+#
+# Targeting 2 000 tokens with 5 000 total QPS for 5k node cluster, so it's 2.5
+# QPS per token.
+#
+# For this test number of tokens is not changed with number of nodes.
+# By default, those 2 000 tokens are are assigned to 80 service accounts, with
+# 25 tokens each. There is 1:1 mapping between deployments and service
+# accounts, so 80 deployments is generated, each with one pod.
+#
+# For smaller cluster, we scale down lineary QPS per token to
+# 2.5 * (Number of nodes)/(5 000). This results in 1 QPS per node, if there is
+# 2 000 tokens.
+#
+# Structure and mapping:
+# * For each namespace (by default 1), we are generating service accounts and
+#   deployments (by default 80).
+# * For each service account we are generating tokens (by default 25).
+# * For each deployment we are creating pods (by default 1) and for those pods
+#   we are mounting all tokens generated from linked service account.
+# * Each pod is running a number of clients equal to number of assigned tokens.
+#
+# When defining your own parameters:
+# Number of tokens = ${namespaces} * ${serviceAccounts} * ${tokensPerServiceAccount}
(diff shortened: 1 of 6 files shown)


clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -0,0 +1,138 @@
+# Stress testing access token validation
+#
+# Targeting 2 000 tokens with 20 000 total QPS
```
mm4tt: Ideally, the test should scale with the cluster size (as load test does). In such case we'll be able to run it along our CI/CD tests at 100, 500 and 5k node scale.

jprzychodzen: Added linear scaling of QPS. Actually, we could get more QPS for smaller clusters, but this sets reasonable requirements and expectations what can be delivered.

jprzychodzen:
Decreased QPS target number to match real-world expectations

clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -0,0 +1,137 @@
...
+name: access-tokens
+automanagedNamespaces: {{$namespaces}}
+tuningSets:
+  - name: Sequence
+    parallelismLimitedLoad:
+      parallelismLimit: 1
+steps:
+- name: Starting measurements
+  measurements:
+    - Identifier: APIResponsivenessPrometheus
```
mm4tt: Please also start the TestMetrics measurement (you can see how it's done in density or load test).
It's a bundle measurement producing a lot of useful data (e.g. profiles, restart checks, resource usages, etc.)

clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -0,0 +1,137 @@
+# Stress testing access token validation
+#
+# Targeting 2 000 tokens with 5 000 total QPS for 5k node cluster, so it's 2.5 QPS per token.
+# For smaller cluster, we want to scale down lineary QPS per token to 2.5 * (Number of nodes)/(5 000).
```
mm4tt: Could you make it more explicit that this test always creates 2K tokens and 80 pods and only the QPS depends on the cluster size?

jprzychodzen: Added comment and calculations.

clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -0,0 +1,167 @@
+# Stress testing access token validation
+#
+# Targeting 2 000 tokens with 5 000 total QPS for 5k node cluster, so it's 2.5 QPS per token.
+#
+# Number of tokens does not scale with number of nodes.
+# By default, those 2 000 tokens are are assigned to 80 service accounts, with 25 tokens each.
```
mm4tt: Sorry for nitpicking, but I'd also add a line about number of pods.
IIUC there will be always 80 pods, which means that this test can be run even on a 1 node cluster (assuming the node is big enough). If we had more pods then we'd run into 110pods/node limitation.

PR #1022, opened 2020-02-05
port number of etcd insecure access should be configurable

chendave:
The port number should be configurable instead of hard-coded,
actually, for the cluster which is setup by kubeadm, the insecure
access number is "2381" by default.

It's hard-coded as "2382", this will make end-users face issues
when etcd metrics is needed, change source code to gather data for
end-user does not make sense, we'd better to provide a option so that
the port number could be specified before the job is started.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -61,6 +61,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
 	// TODO(#595): Change the name of the MASTER_IP and MASTER_INTERNAL_IP flags and vars to plural
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/config/cluster.go
```diff
@@ -34,6 +34,7 @@ type ClusterConfig struct {
 	KubeConfigPath             string
 	Nodes                      int
 	Provider                   string
+	EtcdInsecurePort           string
 	MasterIPs                  []string
```
oxddr: Use IntEnvVar flag - ports can be represented as integers. Change type of port to int throughout your PR.

clusterloader2/cmd/clusterloader.go
```diff
@@ -61,6 +61,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcdinsecureport", "ETCD_INSECURE_PORT", "2381", "Inscure http port")
```
oxddr: nit: make it `etcd-insecure-port` so it reads better

chendave: done

chendave:
Hi @oxddr , thanks for your review and comments! could you pls take another look. 
btw, I don't know why some testcases failed with Jenkins, maybe infrastructure issues?

clusterloader2/pkg/config/cluster.go
chendave: done

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -180,21 +182,24 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string) error {
...
 	}
 
 	// In https://github.com/kubernetes/kubernetes/pull/74690, mTLS is enabled for etcd server
-	// http://localhost:2382 is specified to bypass TLS credential requirement when checking
-	// etcd /metrics and /health.
-	if samples, err := e.sshEtcdMetrics("curl http://localhost:2382/metrics", host, provider); err == nil {
+	// in order to bypass TLS credential requirement when checking etc /metrics and /health, you
+	// need to provide the insecure http port number to access etcd, http://localhost:2381 for
+	// example.
+	cmd := "curl http://localhost:" + strconv.Itoa(port) + "/metrics"
```
oxddr: With port being int, I'd use sprintf to do the conversion for us: `cmd := fmt.Sprinf("curl http://localhost:%d/metrics", port)`

oxddr:
I've left a single nit. Otherwise LGTM.

clusterloader2/pkg/measurement/common/etcd_metrics.go
chendave: done

PR #1023, opened 2020-02-05
Modify ListRuntimeObjectsForKind to work for CRDs as well

mborsz:
With this change WaitForControlledPodsRunning should work for CRDs wrapping groups of pods with format like:

```
metadata:
  namespace: namespace-1
spec:
  replicas: 200
  selector:
     # any labelSelector, e.g.
     matchLabels:
       name: xx
```
i.e.:
* metadata.namespace must point to the namespace that the object lives in
* spec.replicas must describe desired number of pods
* spec.selector must describe a selector that can be used to find all pods associated with given CRD

WaitForControlledPodsRunning will wait until exactly `spec.replica` pods in given namespace match `spec.selector` and is in running state.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -360,8 +360,10 @@ func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(runtim
 func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion() (int, uint64, error) {
 	var desiredCount int
 	var maxResourceVersion uint64
-	objects, err := runtimeobjects.ListRuntimeObjectsForKind(w.clusterFramework.GetClientSets().GetClient(),
-		w.kind, w.selector.Namespace, w.selector.LabelSelector, w.selector.FieldSelector)
+	objects, err := runtimeobjects.ListRuntimeObjectsForKind(
+		w.clusterFramework.GetClientSets().GetClient(),
+		w.clusterFramework.GetDynamicClients().GetClient(),
+		w.gvr, w.kind, w.selector.Namespace, w.selector.LabelSelector, w.selector.FieldSelector)
 	if err != nil {
 		return desiredCount, maxResourceVersion, fmt.Errorf("listing objects error: %v", err)
 	}
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
@@ -37,8 +37,8 @@ import (
 )
 
 // ListRuntimeObjectsForKind returns objects of given kind that satisfy given namespace, labelSelector and fieldSelector.
-// TODO: using dynamic interface rather than clientset interface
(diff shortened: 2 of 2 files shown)


mborsz:
/hold

I still plan to run some test to validate whether this really works.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -360,8 +360,10 @@ func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(runtim
 func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion() (int, uint64, error) {
 	var desiredCount int
 	var maxResourceVersion uint64
-	objects, err := runtimeobjects.ListRuntimeObjectsForKind(w.clusterFramework.GetClientSets().GetClient(),
-		w.kind, w.selector.Namespace, w.selector.LabelSelector, w.selector.FieldSelector)
+	objects, err := runtimeobjects.ListRuntimeObjectsForKind(
+		w.clusterFramework.GetClientSets().GetClient(),
+		w.clusterFramework.GetDynamicClients().GetClient(),
```
wojtek-t: just pass MultiClientSet instead of passing both

mborsz: As discussed offline, I would need to pass both MultiClientSet and MutliDynamicClient, so it's better to keep it this way.

mborsz:
I tested this manually and found one bug. It's fixed now.
/hold cancel

FTR, files I used:
```yaml
# config.yaml
name test:
automanagedNamespaces: 1
tuningSets:
- name: 100qps
  GlobalQPSLoad:
    qps: 100
    burst: 10
- name: RandomizedUpdateTimeLimited
  RandomizedTimeLimitedLoad:
    timeLimit: 2h
steps:
- name: Starting measurement for waiting for pods
  measurements:
  - Identifier: WaitForRunningMicroservices
    Params:
      action: start
      apiVersion: samplecontroller.k8s.io/v1alpha1
      kind: Foo
      labelSelector: l1 = v1
      operationTimeout: 1h
- name: Create Foos
  phases:
  - namespaceRange:
      min: 1
      max: 1
    replicasPerNamespace: 1
    tuningSet: 100qps
    objectBundle:
    - basename: foo
      objectTemplatePath: foo.yaml
- name: Waiting for objects become updated
  measurements:
  - Identifier: WaitForRunningMicroservices
    Params:
      action: gather

# foo.yaml
apiVersion: samplecontroller.k8s.io/v1alpha1
kind: Foo
metadata:
  name: example-foo
  labels:
    l1: v1
spec:
  selector:
    matchLabels:
      l1: v1
  replicas: 3
```

Testing procedure:
```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/sample-controller/master/artifacts/examples/crd-status-subresource.yaml
./clusterloader2 --config conifg.yaml

(in background)
kubectl run -l l1=v1 -n $namespace --image=nginx pod-1
kubectl run -l l1=v1 -n $namespace --image=nginx pod-2
kubectl run -l l1=v1 -n $namespace --image=nginx pod-3
```

The expected result is that clusterloader finds new pods and finishes with success.

PR #1025, opened 2020-02-05
Add support for podLabelSelector to WaitForControlledPodsRunning

mborsz:
It tries to solve problem that there is a race between rolling update of pods and WaitForControlledPodsRunning waiting for updated pods.

Sample usage:

First, create deployment label `pod-revision: old`.
```yaml
- name: Starting measurement for waiting for pods
  measurements:
  - Identifier: WaitForRunningPods
    Params:
      action: start
      apiVersion: apps/v1
      kind: Deployment
      labelSelector: l1 = v1
      operationTimeout: 1h
- name: Create deployments with pod-revision=old label
  phases:
  - namespaceRange:
      min: 1
      max: 1
    replicasPerNamespace: 1
    tuningSet: 100qps
    objectBundle:
    - basename: foo
      templateFillMap:
         labelToSet: old
- name: Waiting for objects become created
  measurements:
  - Identifier: WaitForRunningPods
    Params:
      action: gather
```

Then, before we trigger update of deployments, we "limit" WaitForControllerPodsRunning's window to only updated pods by setting pod labels selector:

```yaml
- name: Set pod selector
  measurements:
  - Identifier: WaitForRunningPods
    Params:
       action: setPodLabelSelector
       selector:
          pod-revision: new
- name: Update Foos
  phases:
  - namespaceRange:
      min: 1
      max: 1
    replicasPerNamespace: 1
    tuningSet: 100qps
    objectBundle:
    - basename: foo
      templateFillMap:
         labelToSet: new
```
At this point, each deployment object update will trigger new watch inside of WaitForControllerPodsRunning with additional selector that will limit "watch window" to new pods only. This way, we in the next step we will wait only for updated pods (not seeing old pods at all).

```yaml
- name: Waiting for objects become updated
  measurements:
  - Identifier: WaitForRunningPods
    Params:
      action: gather
 ```

The main drawback of such approach is that we do not check if old pods were deleted at all. Potentially due to some bug, they may leak and we won't catch this.

/assign @mm4tt 
/assign @wojtek-t

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -23,6 +23,9 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


mborsz:
/hold
I still need to test if this works and we still need to discuss whether this is a right approach to take.

PR #1031, opened 2020-02-07
Differentiate different kinds of errors

chendave:
If the ssh connection to the host is successful but the cmd run failed,
the error message should be more helpful instead of misleading us with
ssh connection error.

`runSSHCommand` method provides us a way to judge this based on the return
`code`.


wojtek-t:
/lgtm
/approve

Thanks!

PR #1033, opened 2020-02-07
Add params for mem/cpu requests in load test deployments

mm4tt:


mm4tt:
Needed to add `"` to fix the `reading template (deployment.yaml) error: parsing error: template: :3: bad number syntax: "10m"` error

PR #1035, opened 2020-02-07
Add pod update checking to WaitForControlledPodsRunning

mborsz:
This implements an experimental support for waiting for pod updates in WaitForControlledPodsRunning.

```
- name: Starting measurement for waiting for pods
  measurements:
  - Identifier: WaitForRunningPods
    Method: WaitForControlledPodsRunning
    Params:
      action: start
      apiVersion: apps/v1
      kind: Deployment
      labelSelector: l1 = v1
      checkIfPodsAreUpdated: true
      operationTimeout: 1h
- name: Create deployments with envVar set to old
  phases:
  - namespaceRange:
      min: 1
      max: 1
    replicasPerNamespace: 1
    tuningSet: 100qps
    objectBundle:
    - basename: foo
      objectTemplatePath: deployment.yaml
      templateFillMap:
         envToSet: old
- name: Waiting for objects become created
  measurements:
  - Identifier: WaitForRunningPods
    Method: WaitForControlledPodsRunning
    Params:
      action: gather
```

The new logic is enabled by setting `checkIfPodsAreUpdated: true` flag (name TBD).
At this point, nothing special happens.

Let's trigger some update:
```
- name: Update Foos
  phases:
  - namespaceRange:
      min: 1
      max: 1
    replicasPerNamespace: 1
    tuningSet: 100qps
    objectBundle:
    - basename: foo
      objectTemplatePath: deployment.yaml
      templateFillMap:
         envToSet: new
- name: Waiting for objects become updated
  measurements:
  - Identifier: WaitForRunningPods
    Method: WaitForControlledPodsRunning
    Params:
      action: gather
```
The `gather` call will block here until `replicas` pods exist in given deployment **that match deployment's spec.template.spec**.
Without this change, we would wait only for any `replicas` pod exist in given deployment (regardless if they are updated or not) which is not correct.

/assign @wojtekt-t
/assign @mm4tt

PR #1037, opened 2020-02-07
Remove tolerations from daemonset.yaml

mborsz:
This is unnecessary (daemoset controller adds this toleration anyway) and breaks https://github.com/kubernetes/perf-tests/pull/1035


clusterloader2/testing/load/daemonset.yaml
```diff
@@ -28,14 +28,10 @@ spec:
             memory: "10M"
       priorityClassName: daemonset-priorityclass-0 # Name is autogenerated, hence the -0 prefix.
       terminationGracePeriodSeconds: 1
-      # Add not-ready/unreachable tolerations for 15 minutes so that node
```
mm4tt: We're still adding tolerations (just not for 15minutes) so I don't see a reason for deleting the comment.

If what you wrote in the PR description is true (and daemonset controller adds these tolerations) then we can remove the whole tolerations section, right?

mborsz: Right

PR #1048, opened 2020-02-10
Add "etcd_operation" and "etcd_type" variables to Grafana dashboard

oxddr:
This should ease querying and allows to remove some dedicated graphs as they can be now easily substituted.

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/dashboards/defaults.py
+++ b/clusterloader2/pkg/prometheus/manifests/dashboards/defaults.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 import attr
 from grafanalib import core as g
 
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/manifests/dashboards/defaults.py
```diff
@@ -29,7 +29,24 @@ class Dashboard(g.Dashboard):
         default=g.Templating(
             list=[
                 # Make it possible to use $source as a source.
-                g.Template(name="source", type="datasource", query="prometheus")
+                g.Template(name="source", type="datasource", query="prometheus"),
+                g.Template(
```
mborsz: I think this adds variables to all dashboards including the ones which has nothing to do with etcd (e.g. dns, network). Is this correct?

Can we somehow avoid that?

oxddr: Done.

clusterloader2/pkg/prometheus/manifests/dashboards/defaults.py
```diff
@@ -21,18 +21,14 @@
...
     time = attr.ib(default=g.Time("now-30d", "now"))
-    templating = attr.ib(
-        default=g.Templating(
-            list=[
-                # Make it possible to use $source as a source.
-                g.Template(name="source", type="datasource", query="prometheus")
-            ]
-        )
-    )
+    # Make it possible to use $source as a source.
```
mborsz: IIUC you dropped new variables completely.

I didn't mean to remove etcd_operation and etcd_type variables completely -- rather to somehow move them to master-dashboard only.

oxddr: They were moved not deleted, please see bottom of master-dashboard.dashboard.py

mborsz:
Cool! Thanks!

/lgtm

PR #1049, opened 2020-02-11
scale down to 80 pods

YangLu1031:
Test failed for 100 pods with timeout, so scale down back to 80 pods.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-windows-node-throughput/1225796067589623808


pjh:
Thanks Yang, also LGTM.

@wojtek-t do you have any ideas about why we can get up to 100 pods just fine in manual test runs but we can't seem to reach 100 pods yet in continuous e2e runs? We'll continue investigating to see if we can find any differences in the manual vs. e2e setup/environment, but if you have any ideas or past experience please let us know.

wojtek-t:
> @wojtek-t do you have any ideas about why we can get up to 100 pods just fine in manual test runs but we can't seem to reach 100 pods yet in continuous e2e runs? We'll continue investigating to see if we can find any differences in the manual vs. e2e setup/environment, but if you have any ideas or past experience please let us know.

No idea - we generally get pretty consistent results across manual and continuous runs.
How do you run your manual tests? I guess there has to be some difference between how you setup your cluster there.

PR #1051, opened 2020-02-11
Change checkIfPodsAreUpdated default to true

mborsz:
We should switch default WaitForControllerPodsRunning behavior to the right one: where we correctly distinguish updated pods from state ones.

The original change (https://github.com/kubernetes/perf-tests/pull/1035) has been merged 4 hours ago so maybe let's wait one more day before submitting this.
/hold


mborsz:
PTAL
/hold cancel

PR #1054, opened 2020-02-13
clean up stale namespaces before execution

chendave:
It's not likely execution is finished normally everytime, the execution
might be aborted abnormally, this will make the auto managed namespaces
become orphans and no chance to be clean up on the test machine, thus,
impact the data we got from execution.

The change cleans up all those namespaces before the execution, give
the execution a fairly clean environment.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -62,6 +62,7 @@ func initClusterFlags() {
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "Whether to delete all stale namespaces before the test execution.")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
 	// TODO(#595): Change the name of the MASTER_IP and MASTER_INTERNAL_IP flags and vars to plural
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
--- a/clusterloader2/pkg/framework/framework.go
+++ b/clusterloader2/pkg/framework/framework.go
@@ -34,6 +34,8 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
+var namespaceID = regexp.MustCompile(`^test-[a-z0-9]+-[0-9]+$`)
+
 // Framework allows for interacting with Kubernetes cluster via
 // official Kubernetes client.
 type Framework struct {
@@ -115,22 +117,30 @@ func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
 }
 
 // ListAutomanagedNamespaces returns all existing automanged namespace names.
-func (f *Framework) ListAutomanagedNamespaces() ([]string, error) {
-	var automanagedNamespacesList []string
+func (f *Framework) ListAutomanagedNamespaces() ([]string, []string, error) {
+	var automanagedNamespacesList, staleNamespaces []string
 	namespacesList, err := client.ListNamespaces(f.clientSets.GetClient())
 	if err != nil {
-		return automanagedNamespacesList, err
+		return automanagedNamespacesList, staleNamespaces, err
 	}
 	for _, namespace := range namespacesList {
 		matched, err := f.isAutomanagedNamespace(namespace.Name)
 		if err != nil {
-			return automanagedNamespacesList, err
+			return automanagedNamespacesList, staleNamespaces, err
 		}
 		if matched {
 			automanagedNamespacesList = append(automanagedNamespacesList, namespace.Name)
+		} else {
+			// check further whether the namespace is a automanaged namespace created in previous test execution.
+			// this could happen when the execution is aborted abornamlly, and the resource is not able to be
+			// clean up.
+			matched := f.isStaleAutomanagedNamespace(namespace.Name)
+			if matched {
+				staleNamespaces = append(staleNamespaces, namespace.Name)
+			}
 		}
 	}
-	return automanagedNamespacesList, nil
+	return automanagedNamespacesList, staleNamespaces, nil
 }
 
 // DeleteAutomanagedNamespaces deletes all automanged namespaces.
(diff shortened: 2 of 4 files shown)


oxddr:
And they are lost :(

clusterloader2/pkg/framework/framework.go
```diff
@@ -115,22 +115,30 @@ func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
 }
 
 // ListAutomanagedNamespaces returns all existing automanged namespace names.
-func (f *Framework) ListAutomanagedNamespaces() ([]string, error) {
-	var automanagedNamespacesList []string
+func (f *Framework) ListAutomanagedNamespaces() ([]string, []string, error) {
+	var automanagedNamespacesList, staleAutomanagedNamespacesList []string
```
oxddr: What about just `staleNamespaces`. This is a function-level variable, so I don't think we need to be extra verbose in here.

clusterloader2/pkg/framework/framework.go
```diff
@@ -155,6 +163,26 @@ func (f *Framework) DeleteAutomanagedNamespaces() *errors.ErrorList {
 	return errList
 }
 
+// DeleteStaleAutomanagedNamespaces deletes all stale automanged namespaces.
+func (f *Framework) DeleteStaleAutomanagedNamespaces(namespaces []string) *errors.ErrorList {
```
oxddr: Rename to `DeleteNamespace` - this function is not limited to automated and stale namespaces and could be used in other places.

clusterloader2/pkg/framework/framework.go
```diff
@@ -216,3 +244,8 @@ func (f *Framework) ApplyTemplatedManifests(manifestGlob string, templateMapping
 func (f *Framework) isAutomanagedNamespace(name string) (bool, error) {
 	return regexp.MatchString(f.automanagedNamespacePrefix+"-[1-9][0-9]*", name)
 }
+
+func (f *Framework) isStaleAutomanagedNamespace(name string) bool {
+	var namespaceID = regexp.MustCompile(`^test-[a-z0-9]+-[0-9]+$`)
```
oxddr: Make it package-level variable. No need to recompute regex each time we call this function.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -60,13 +60,20 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
 	if err := ctx.GetChaosMonkey().Init(conf.ChaosMonkey, stopCh); err != nil {
 		return errors.NewErrorList(fmt.Errorf("error while creating chaos monkey: %v", err))
 	}
-	automanagedNamespacesList, err := ctx.GetClusterFramework().ListAutomanagedNamespaces()
+	automanagedNamespacesList, staleAutomanagedNamespacesList, err := ctx.GetClusterFramework().ListAutomanagedNamespaces()
```
oxddr: Drop `List` suffix. 

I know the convention in this file is different, but this naming seems to be against Go convention of using short and concise names. Type of a variable can be easily inferred from its type ;)

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -60,13 +60,20 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
...
 	if len(automanagedNamespacesList) > 0 {
 		return errors.NewErrorList(fmt.Errorf("pre-existing automanaged namespaces found"))
 	}
+	if len(staleAutomanagedNamespacesList) > 0 {
+		klog.Warningf("stale automanaged namespaces found")
+		if errList := ctx.GetClusterFramework().DeleteStaleAutomanagedNamespaces(staleAutomanagedNamespacesList); !errList.IsEmpty() {
+			klog.Warningf("stale automanaged namespace cleanup error: %v", errList.String())
+		}
+	}
+
```
oxddr: Guard this logic behind a flag. I can imagine a situation, when we don't want clusterloader to delete any namespaces.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -60,13 +60,20 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
...
 	if err != nil {
 		return errors.NewErrorList(fmt.Errorf("automanaged namespaces listing failed: %v", err))
 	}
 	if len(automanagedNamespacesList) > 0 {
 		return errors.NewErrorList(fmt.Errorf("pre-existing automanaged namespaces found"))
 	}
+	if len(staleAutomanagedNamespacesList) > 0 {
+		klog.Warningf("stale automanaged namespaces found")
+		if errList := ctx.GetClusterFramework().DeleteStaleAutomanagedNamespaces(staleAutomanagedNamespacesList); !errList.IsEmpty() {
+			klog.Warningf("stale automanaged namespace cleanup error: %v", errList.String())
```
oxddr: This should be an error not a warning.

chendave:
@oxddr thanks you for all those valuable suggestion, I will address them later.

clusterloader2/pkg/test/simple_test_executor.go
chendave: I was thinking that failed to delete those namespaces should not block the whole process, I am good with change this to error after the new flag is introduced.

clusterloader2/cmd/clusterloader.go
```diff
@@ -62,6 +62,7 @@ func initClusterFlags() {
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespace, "delete-stale-ns", "DELETE_STALE_NAMESPACE", false, "Whether to delete all stale namespaces before the test execution.")
```
wojtek-t: nit: delete-state-namespace

let's try to be consistent with naming between flag and env-var (and let's not shortcut those)

clusterloader2/pkg/framework/framework.go
```diff
@@ -115,22 +117,30 @@ func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
...
 	for _, namespace := range namespacesList {
 		matched, err := f.isAutomanagedNamespace(namespace.Name)
 		if err != nil {
-			return automanagedNamespacesList, err
+			return automanagedNamespacesList, staleNamespaces, err
 		}
 		if matched {
 			automanagedNamespacesList = append(automanagedNamespacesList, namespace.Name)
+		} else {
+			//check further whether the namespace is a automanaged namespace created in previous test execution.
```
wojtek-t: nit: "// check..." (whitespace after "//")

here and below

chendave:
@wojtek-t thanks for the comments!

clusterloader2/cmd/clusterloader.go
```diff
@@ -62,6 +62,7 @@ func initClusterFlags() {
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespace, "delete-stale-namespace", "DELETE_STALE_NAMESPACE", false, "Whether to delete all stale namespaces before the test execution.")
```
wojtek-t: Sorry one more nit:
probably should be "DeleteStaleNamespaces" (and delete-stale-namespaces) [i.e. plural namespaces]

chendave: never mind, good catch!

PR #1055, opened 2020-02-13
Refresh templates on every time range change

mborsz:
Without this change, the templates are calculated only once, for default time range (30d) so for older snapshot they are empty and there is no simple way to refresh them.

Refreshing the set of values seems to be quite fast so I hope this won't be a big performance issue, but we will see.

/assign @oxddr


oxddr:
That sounds reasonable to me, I don't expect big performance hit because of that (but we will see). Thanks!

/lgtm

PR #1058, opened 2020-02-17
Using nodeSelector & Tolerations to schedule windows test pods

YangLu1031:
1. For simplicity, we taint linux node and untaint windows node to schedule pause image on windows node for pod startup latency test before. But now need to schedule prometheus server pods on linux node as well, so add nodeSelector & Tolerations to schedule pause image on windows node. 
2. Bumped to latest pause image version.


clusterloader2/testing/node-throughput/rc.yaml
```diff
@@ -24,6 +28,12 @@ spec:
       # Add not-ready/unreachable tolerations for 15 minutes so that node
       # failure doesn't trigger pod deletion.
       tolerations:
+      {{if .WindowsNodeTest}}
+      - key: "node.kubernetes.io/os"
+        operator: "Equal"
+        value: "win1809"
```
pjh: 1809 will be going away soon, we should switch to 2019 instead. Could be in this PR or a follow-up.

YangLu1031: I see, seems need to change to more relax operator? like  "Exists"

Issue #1057, opened 2020-02-14
Clusterloader2 use existing prometheus?


aakarshg:
Hello Some of the distros like OpenShift come with a prometheus installation on the cluster, is there anyway we can use this prometheus server's metrics instead of trying to stand up another one, and then have node exporter running ?

wojtek-t:
@mm4tt @mborsz @oxddr

oxddr:
At the moment I don't this is possible. But the reason for that is we have some assumptions backed in the clusterloader code (@mm4tt correct me if I am wrong). This should be matter of a single PR to change how the current setup Prometheus works.

mm4tt:
Depending on what you mean by "prometheus installation".

Our setup is based on [prometheus-operator](https://github.com/coreos/prometheus-operator), it Prometheus-operator is avaiable in the cluster then it should be doable to have cl2 to use it instead of spinning up a new one. But as @oxddr wrote it would require some code changes and it's not possible as of now.

But taking a step back, what is the problem you're trying to solve? It shouldn't be an issue to have 2 (or more) Prometheus servers running in cluster simultaneously.

aakarshg:
If a small change will let us use existing prometheus-operator on cluster, we would at least like to see what the change looks like so that we can apply the change locally to test if it'll work..

We're currently evaluating clusterloader2 for our needs. We're hitting all sorts of problems with the clusterloader2 trying to deploy its own prometheus-operator pods, some of them are privilege related, some of them oddly are due to prometheus-operator launched through clusterloader2 trying to use alertmanager pods in another namespace ( because we've an existing prom stack deployment ). I can try to look into creating another set of manifests specifically for Openshift but all of this seems like doubling the effort given that we already have a prometheus stack on the cluster.

mm4tt:
I see, deploying two prometheus-operators is definitely not a good idea. 

What we could try to do is to extract the logic we have for setting up prometheus operator and add an ability in cl2 to call it (or not call it) conditionally. If prometheus operator is already present in the cluster then it should be doable for cl2 to re-use it. I'm not sure if it will work out of the box as we might have some baked in assumptions about prometheus in our code. But it's definitively worth trying.

Unfortunately, we don't have any free cycles currently to do it. I'd appreciate if you could take a shot at it.
[...]

PR #1064, opened 2020-02-19
Remove alerting config from prometheus-prometheus

mm4tt:
We don't use prometheus alerting anywhere in our prometheus monitoring.

Ref. https://github.com/kubernetes/perf-tests/issues/1057


mm4tt:

This is another time when I see a test failing with this error:
```
[measurement call NetworkProgrammingLatency - NetworkProgrammingLatency error: got unexpected number of samples: 0]
```

Will open an issue for it

mm4tt:
The issue for NPL failures - https://github.com/kubernetes/perf-tests/issues/1068

mm4tt:
Presubmit currently fails due to https://github.com/kubernetes/test-infra/issues/16391

mm4tt:

The last failure looks like https://github.com/kubernetes/perf-tests/issues/1068

mm4tt:
@oxddr, it looks like the cl2 presubmit for this PR is failing **consistently** with 
```
[measurement call NetworkProgrammingLatency - NetworkProgrammingLatency error: got unexpected number of samples: 0]
```

Any ideas?

oxddr:
I'll have a closer look later today or tomorrow. I suspect this maybe related to alert-rules. In NetworkingProgrammingLatency we use precomputes. But that's weak hypothesis, because you're removing alert config after all.

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -26,10 +21,6 @@ spec:
       # Start with 2Gi and add 2Gi for each 2K nodes.
       memory: {{MultiplyInt 2 (AddInt 1 (DivideInt .Nodes 2000))}}Gi
       {{end}}
-  ruleSelector:
```
oxddr: `ruleSelector` is not only used for alerting, but also for recording rules. We rely on recording rules to compute Network Programming SLO, hence the failed presubmit. Please restore this section. I'd suggest maybe to remove `role: alert-rules` label and replace it with something more meaningful (here and in prometheus-rule.yaml file).

mm4tt: Thanks, good catch. Done.

PR #1065, opened 2020-02-19
Clean up useless information.

Sakuralbj:
KubemarkServiceMonitors are not used now. I remove code that uses it.


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -142,15 +141,10 @@ func (pc *PrometheusController) SetUpPrometheusStack() error {
...
 	}
-	if pc.isKubemark() {
-		if err := pc.applyManifests(kubemarkServiceMonitors); err != nil {
-			return err
-		}
-	} else {
-		if err := pc.applyManifests(defaultServiceMonitors); err != nil {
-			return err
-		}
+	if err := pc.applyManifests(defaultServiceMonitors); err != nil {
```
mm4tt: This is causing the kubemark tests to fail.

it should be
```
if !pc.isKubemark() {
  if err := pc.applyManifests(defaultServiceMonitors); err != nil {
    return err
  }
}
```

mm4tt:
Thanks @Sakuralbj!

clusterloader2/pkg/prometheus/prometheus.go
Sakuralbj: Thank you.  Is it means if pc.isKubemark()=true,we don't need to apply defaultServiceMonitors?

mm4tt: Yes, if `pc.isKubemark()=true` then we **should not** apply defaultServiceMonitors, otherwise the test will fail.

Sakuralbj: Thank you.I fix it.

Issue #1072, opened 2020-02-22
useless namespaces should be cleanup all at once


chendave:
Here is the patch on how to reproduce the issue, 

```patch
diff --git a/clusterloader2/pkg/framework/framework.go b/clusterloader2/pkg/framework/framework.go
index 05df6bf5..db3b9f29 100644
--- a/clusterloader2/pkg/framework/framework.go
+++ b/clusterloader2/pkg/framework/framework.go
@@ -170,8 +170,10 @@ func (f *Framework) DeleteNamespaces(namespaces []string) *errors.ErrorList {
        var wg wait.Group
        errList := errors.NewErrorList()
        for _, namespace := range namespaces {
+               fmt.Println("namespace found: ", namespace)
                clientSet := f.clientSets.GetClient()
                wg.Start(func() {
+                       fmt.Println("namespace going to be deleted: ", namespace)
                        if err := client.DeleteNamespace(clientSet, namespace); err != nil {
                                errList.Append(err)
                                return
```

```bash
go run cmd/clusterloader.go ...
```

and here is what I found,

...
namespace found:  test-tp7qhi-93
namespace found:  test-tp7qhi-94
namespace found:  test-tp7qhi-95
namespace found:  test-tp7qhi-96
namespace found:  test-tp7qhi-97
namespace found:  test-tp7qhi-98
namespace found:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-84
namespace going to be deleted:  test-tp7qhi-82
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-5
namespace going to be deleted:  test-tp7qhi-76
**namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99
namespace going to be deleted:  test-tp7qhi-99**
...

and eventually, only partial namespaces could be deleted, others still go stale.

PR #1073, opened 2020-02-22
Stale namespaces should be cleaned all at once

chendave:
This change fixs the race condition that's happened around
the `wait.Group` and some code refactoring as well.

fix: https://github.com/kubernetes/perf-tests/issues/1072

### DIFF
--- a/clusterloader2/pkg/framework/framework.go
+++ b/clusterloader2/pkg/framework/framework.go
@@ -143,20 +143,27 @@ func (f *Framework) ListAutomanagedNamespaces() ([]string, []string, error) {
 	return automanagedNamespacesList, staleNamespaces, nil
 }
 
+func (f *Framework) deleteNamespace(namespace string) error {
+	clientSet := f.clientSets.GetClient()
+	if err := client.DeleteNamespace(clientSet, namespace); err != nil {
+		return err
+	}
+	if err := client.WaitForDeleteNamespace(clientSet, namespace); err != nil {
+		return err
+	}
+	return nil
+}
+
 // DeleteAutomanagedNamespaces deletes all automanged namespaces.
 func (f *Framework) DeleteAutomanagedNamespaces() *errors.ErrorList {
 	var wg wait.Group
 	errList := errors.NewErrorList()
 	for i := 1; i <= f.automanagedNamespaceCount; i++ {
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/framework/framework.go
```diff
@@ -167,19 +171,11 @@ func (f *Framework) DeleteAutomanagedNamespaces() *errors.ErrorList {
 
 // DeleteNamespaces deletes the list of namespaces.
 func (f *Framework) DeleteNamespaces(namespaces []string) *errors.ErrorList {
-	var wg wait.Group
+	var wg sync.WaitGroup
 	errList := errors.NewErrorList()
 	for _, namespace := range namespaces {
-		clientSet := f.clientSets.GetClient()
```
mm4tt: This method is indeed erroneous, the problem is that you are capturing the loop variable (namespace) in the inline func you pass to the wg.Start. In the next iteration the value of the namespace variable will change and this is causing the race conditions.

This can be fixed with a one liner:
```
for _, namespace := range namespaces {
  namespace := namespace // One line fix, namespace is now not a loop variable anymore
  clientSet  = ...
```

clusterloader2/pkg/framework/framework.go
```diff
@@ -143,22 +143,26 @@ func (f *Framework) ListAutomanagedNamespaces() ([]string, []string, error) {
 	return automanagedNamespacesList, staleNamespaces, nil
 }
 
+func (f *Framework) deleteNS(namespace string, wg *sync.WaitGroup, errList *errors.ErrorList) {
```
mm4tt: Please change the signature to

```
func (f *Framework) deleteNamespace(namespace string) error {
```

This is actually a good refactoring as it will get rid of duplicated code.

clusterloader2/pkg/framework/framework.go
chendave: cool!

mm4tt:
Thanks, @chendave!

/lgtm
/approve

/hold
This looks good to me, but could you please update the PR message?

chendave:
hey @mm4tt , thanks for your great suggestion here!

mm4tt:
Thanks for fixing the bug and refactoring the code!

/hold cancel

Issue #1007, opened 2020-02-03
Combine density and load test into a single test


mm4tt:
It basically means getting rid of the density test and adding the missing functionalities to the load test (SchedulerThroughput and PodStartupLatency)

wojtek-t:
@mborsz @marseel

wojtek-t:
This has been implemented in the meantime.

/close

PR #1078, opened 2020-02-25
[Remove density] Fix metrics in perf dash

jprzychodzen:
Add experiment specific metric to Perf-Dash.
Rename experiment specific metric in ClusterLoader2 experiment config.

Ref. #1007

/assign @mm4tt

### DIFF
--- a/clusterloader2/testing/load/experimental-config.yaml
+++ b/clusterloader2/testing/load/experimental-config.yaml
@@ -113,7 +113,7 @@ steps:
     Method: APIResponsivenessPrometheus
     Params:
       action: start
-  - Identifier: LoadPodStartupLatency
+  - Identifier: CreatePhasePodStartupLatency
     Method: PodStartupLatency
     Params:
       action: start
@@ -403,7 +403,7 @@ steps:
 # BEGIN scheduler throughput
 - name: Creating scheduler throughput measurements
   measurements:
-  - Identifier: HighThroughputStartupLatency
+  - Identifier: HighThroughputPodStartupLatency
     Method: PodStartupLatency
     Params:
       action: start
@@ -447,7 +447,7 @@ steps:
       action: gather
 - name: Collecting scheduler throughput measurements
   measurements:
-  - Identifier: HighThroughputStartupLatency
+  - Identifier: HighThroughputPodStartupLatency
     Method: PodStartupLatency
     Params:
       action: gather
@@ -916,7 +916,7 @@ steps:
     Params:
       action: gather
   {{end}}
-  - Identifier: LoadPodStartupLatency
+  - Identifier: CreatePhasePodStartupLatency
     Method: PodStartupLatency
     Params:
       action: gather
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -19,10 +19,11 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
-	"k8s.io/klog"
 	"net/http"
 	"strconv"
 	"strings"
+
+	"k8s.io/klog"
 
 	"github.com/ghodss/yaml"
 )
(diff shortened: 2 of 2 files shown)


clusterloader2/testing/load/experimental-config.yaml
```diff
@@ -113,7 +113,7 @@ steps:
     Method: APIResponsivenessPrometheus
     Params:
       action: start
-  - Identifier: LoadPodStartupLatency
+  - Identifier: PodStartupLatency
```
mm4tt: There is another PodStartupLatency measurement already defined in this config, line 481.
I consciously named it differently to avoid conflicts and to be able to differentiate them.

perfdash/config.go
```diff
@@ -80,6 +81,21 @@ var (
...
+				Name:             "load",
+				OutputFilePrefix: "PodStartupLatency_HighThroughputStartupLatency",
+				Parser:           parsePerfData,
+			}},
+			"LoadPodStartup": []TestDescription{{
+				Name:             "load",
+				OutputFilePrefix: "PodStartupLatency_PodStartupLatency",
+				Parser:           parsePerfData,
+			}},
+			"LoadSaturationPodStartup": []TestDescription{{
```
mm4tt: We don't have such measurement anywhere (saturation changed name to high-throughput)

clusterloader2/testing/load/experimental-config.yaml
jprzychodzen: Changed name to `Load...` to `Second...` so i's not confusing from perf-dash perspective how metric name is generated.

perfdash/config.go
jprzychodzen: This measurement exists in density test - I've copied the name in case if someone will enable the same measurement in load test.

Changed `Saturation` to `Second` to match Identifier from `experimental-config.yaml`.

clusterloader2/testing/load/experimental-config.yaml
```diff
@@ -113,7 +113,7 @@ steps:
     Method: APIResponsivenessPrometheus
     Params:
       action: start
-  - Identifier: LoadPodStartupLatency
+  - Identifier: SecondPodStartupLatency
```
mm4tt: I don't like "second", let's make the name more descriptive for people who are not familiar how load test is structured. 
It will be better even for us, I'm quite sure that we'll forget what "second" refers to after some time.

What about LoadPhasePodStartupLatency or CreatePhasePodStartupLatency or something like that?

perfdash/config.go
```diff
@@ -80,6 +81,21 @@ var (
 				OutputFilePrefix: "ResourceUsageSummary",
 				Parser:           parseResourceUsageData,
 			}},
+			"LoadHighThroughputStartup": []TestDescription{{
```
mm4tt: Let's name it with PodStartup prefix for consistency

jprzychodzen: Done. Also, removed duplicated 'Startup'

clusterloader2/testing/load/experimental-config.yaml
jprzychodzen: Changed to `CreatePhasePodStartupLatency`. `Load...` creates a confusion with a name of a whole test.

clusterloader2/testing/load/experimental-config.yaml
```diff
@@ -447,7 +447,7 @@ steps:
       action: gather
 - name: Collecting scheduler throughput measurements
   measurements:
-  - Identifier: HighThroughputStartupLatency
+  - Identifier: PodStartupHighThroughputLatency
```
jprzychodzen: Should be `HighThroughputPodStartupLatency`

clusterloader2/testing/load/experimental-config.yaml
```diff
@@ -403,7 +403,7 @@ steps:
 # BEGIN scheduler throughput
 - name: Creating scheduler throughput measurements
   measurements:
-  - Identifier: HighThroughputStartupLatency
+  - Identifier: PodStartupHighThroughputLatency
```
mm4tt: HighThroughputPodStartupLatency? :)

HighThroughput refers to pods not to latency

perfdash/config.go
```diff
@@ -80,6 +81,21 @@ var (
...
 				Parser:           parseResourceUsageData,
 			}},
+			"LoadCreatePhasePodStartup": []TestDescription{{
+				Name:             "load",
+				OutputFilePrefix: "PodStartupLatency_CreatePhasePodStartupLatency",
+				Parser:           parsePerfData,
+			}},
+			"LoadHighThroughputPodStartup": []TestDescription{{
+				Name:             "load",
+				OutputFilePrefix: "PodStartupLatency_HighThroughputPodStartupLatency",
```
mm4tt: Here it's good :)

perfdash/config.go
jprzychodzen: Actually, for density we are using `SaturationPodStartupLatency`. I will change new metrics to match this pattern

clusterloader2/testing/load/experimental-config.yaml
jprzychodzen: I've send inconsistent PR. It's already HighThroughputPodStartupLatency.

mm4tt:
/lgtm
/approve

Thanks, @jprzychodzen. You'll need to build and deploy a new version of perf-dash to apply this change. Please ping me if you don't know how to do it

PR #1087, opened 2020-02-28
[Access-tokens] Backport test and dependencies from 'master'

jprzychodzen:
Backport:
* Access-tokens test
* Template function 'Loop'
* 'Sleep' measurement


jprzychodzen:
Tested this change locally with 100 nodes cluster for k8s 1.16.

PR #1088, opened 2020-02-28
perfdash: display metrics from ci-benchmark-scheduler-perf-master

ingvagabund:
Time to display collected metrics from test/integration/scheduler_perf.

Jenkins Job: https://github.com/kubernetes/test-infra/pull/16421

Fixes: https://github.com/kubernetes/kubernetes/issues/86972

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -298,6 +298,15 @@ var (
 				OutputFilePrefix: "BenchmarkResults",
 				Parser:           parsePerfData,
 			}},
+			"BenchmarkPerfResults": []TestDescription{{
+				// Expected file prefix string is constructed as
+				// OutputFilePrefix+"_"+Name. Given we currently generate
+				// file prefixed with "BenchmarkPerfScheduling_" followed by a date,
+				// set the Name to an empty string.
+				Name:             "",
+				OutputFilePrefix: "BenchmarkPerfScheduling",
+				Parser:           parsePerfData,
+			}},
 		},
 	}
 


ingvagabund:
FYI @ahg-g @liu-cong

perfdash/config.go
```diff
@@ -298,6 +298,11 @@ var (
 				OutputFilePrefix: "BenchmarkResults",
 				Parser:           parsePerfData,
 			}},
+			"BenchmarkPerfResults": []TestDescription{{
+				Name:             "",
```
liu-cong: why empty name?

ingvagabund: Perfdash is looking for files prefixed with `$OutputFilePrefix_$Name` string. In our case we generated files prefixed with `BenchmarkPerfScheduling_` followed by a date. I can update https://github.com/kubernetes/kubernetes/blob/master/test/integration/scheduler_perf/util.go#L115 to add `benchmark` string into the file name so we can set the Name to `benchmark` as is done in `BenchmarkResults` case. Though, I don't think it makes any difference.

liu-cong: Either do that or simply add a comment here.

ingvagabund: Done

ingvagabund:
@wojtek-t PTAL

ingvagabund:
@mm4tt PTAL

mm4tt:
/approve

+ @oxddr, current oncall, to assist you with deploying the new version of perf-dash.

Ideally, we should have a better process for that. We could make it work similarly to kubekins where there is a post-submit job configured to automatically build and push a new image every time the relevant code changes. I'll file an issue for that

oxddr:
/lgtm 
Once this PR is merged I am going to release new perf-dash version. 

> Ideally, we should have a better process for that. We could make it work similarly to kubekins where there is a post-submit job configured to automatically build and push a new image every time the relevant code changes. I'll file an issue for that

We've talked about it in the past and decided not to do anything, as perf-dash (in the current form) at some point is going to be replaced. On the other hand, building (and even rolling out) an image should be easy to do with Cloud Build.

mm4tt:
> We've talked about it in the past and decided not to do anything, as perf-dash (in the current form) at some point is going to be replaced. On the other hand, building (and even rolling out) an image should be easy to do with Cloud Build.

I think it shouldn't be more than ~2h of work (given that there is an existing post-submit we can just copy and tune to our needs). Given that it takes ~10-15min to  build/push new perf-dash and we're pushing it 3rd time this month, I believe such investment would quickly pay-off.

ahg-g:
Thanks Jan, here is what the dashboard is reporting: http://perf-dash.k8s.io/#/?jobname=scheduler-perf-benchmark&metriccategoryname=Scheduler&metricname=BenchmarkPerfResults&Metric=scheduler_scheduling_algorithm_priority_evaluation_seconds&Name=BenchmarkPerfScheduling%2FSchedulingBasic%2F5000Nodes%2F5000InitPods%2F1000PodsToSchedule

Not as consistent as we hoped for, I guess the type of machine the test is run on impacts the results. I wonder if we can tune the test run such that it is allocated a specific amount of CPU/memory to get more consistent lines.

oxddr:
I don't have an access to Prow, but last time I checked it (months ago) it used uniform machines to run tests. Specifying CPU and mem is definitely worth checking IMO.

ingvagabund:
> Not as consistent as we hoped for, I guess the type of machine the test is run on impacts the results. I wonder if we can tune the test run such that it is allocated a specific amount of CPU/memory to get more consistent lines.

> I don't have an access to Prow, but last time I checked it (months ago) it used uniform machines to run tests. Specifying CPU and mem is definitely worth checking IMO.

Checking the artifacts every test container is created with:
```
"resources":{"requests":{"memory":"1Gi"}}
```

https://github.com/kubernetes/test-infra/pull/16687 for constraining cpu as well to 4 cores for now.

Issue #1083, opened 2020-02-27
wrong schedulingThroughput in density test


xial-thu:
/kind bug
/area clusterloader

When calculating scheduling throughput in density test, I found the calculation is incorrect when running a small test of 300 saturation pods on 16 nodes. The log shows

> I0227 17:38:29.247294 2069840 scheduling_throughput.go:103] SchedulingThroughput: starting collecting throughput data
I0227 17:38:34.247614 2069840 scheduling_throughput.go:118] SchedulingThroughput: labelSelector(group = saturation): 15 pods scheduled
I0227 17:38:39.247969 2069840 scheduling_throughput.go:118] SchedulingThroughput: labelSelector(group = saturation): 127 pods scheduled
I0227 17:38:44.248357 2069840 scheduling_throughput.go:118] SchedulingThroughput: labelSelector(group = saturation): 300 pods scheduled
I0227 17:38:49.248779 2069840 scheduling_throughput.go:118] SchedulingThroughput: labelSelector(group = saturation): 300 pods scheduled
I0227 17:38:54.249241 2069840 scheduling_throughput.go:118] SchedulingThroughput: labelSelector(group = saturation): 300 pods scheduled
I0227 17:38:55.132166 2069840 scheduling_throughput.go:131] SchedulingThroughput: gathering data

300 pods are scheduled within 15 seconds but `gather` stage happens after 25 seconds. In .json output, the average throughput is 12, which is exactly **300/25**. But wait, we should exclude data when all pods are scheduled.

```
{
  "average": 12,
  "perc50": 3,
  "perc90": 34.6,
  "perc99": 34.6
}
```

Below is the part of configuration

```yaml
- name: Collecting saturation pod measurements
  measurements:
  - Identifier: WaitForRunningSaturationDeployments
    Method: WaitForControlledPodsRunning
    Params:
      action: gather
- measurements:
  - Identifier: SaturationPodStartupLatency
    Method: PodStartupLatency
    Params:
      action: gather
- measurements:
  - Identifier: SchedulingThroughput
    Method: SchedulingThroughput
    Params:
      action: gather
      threshold: {{$SCHEDULER_THROUGHPUT_THRESHOLD}}
```

wojtek-t:
This is a know issue - we're going to rewrite this measurement to be based on prometheus metrics.

@mm4tt - FYI

mm4tt:
[...]

PR #1093, opened 2020-03-02
[Remove density] Rework SchedulingThroughput

jprzychodzen:
* Add measurmentInterval parameter to SchedulingThroughput
* SchedulingThroughput threshold compares to max value, instead of 99th
  percentile

### DIFF
--- a/clusterloader2/pkg/measurement/common/scheduling_throughput.go
+++ b/clusterloader2/pkg/measurement/common/scheduling_throughput.go
@@ -32,6 +32,7 @@ import (
 
 const (
 	schedulingThroughputMeasurementName = "SchedulingThroughput"
+	defaultSchedulingThroughputInterval = 5 * time.Second
 )
 
 func init() {
@@ -70,9 +71,12 @@ func (s *schedulingThroughputMeasurement) Execute(config *measurement.Measuremen
 		if err := selector.Parse(config.Params); err != nil {
 			return nil, err
 		}
-
+		measurmentInterval, err := util.GetDurationOrDefault(config.Params, "measurmentInterval", defaultSchedulingThroughputInterval)
+		if err != nil {
+			return nil, err
+		}
 		s.stopCh = make(chan struct{})
-		return nil, s.start(config.ClusterFramework.GetClientSets().GetClient(), selector)
+		return nil, s.start(config.ClusterFramework.GetClientSets().GetClient(), selector, measurmentInterval)
 	case "gather":
 		threshold, err := util.GetFloat64OrDefault(config.Params, "threshold", 0)
 		if err != nil {
@@ -94,7 +98,7 @@ func (*schedulingThroughputMeasurement) String() string {
 	return schedulingThroughputMeasurementName
 }
 
-func (s *schedulingThroughputMeasurement) start(clientSet clientset.Interface, selector *measurementutil.ObjectSelector) error {
+func (s *schedulingThroughputMeasurement) start(clientSet clientset.Interface, selector *measurementutil.ObjectSelector, measurmentInterval time.Duration) error {
 	ps, err := measurementutil.NewPodStore(clientSet, selector)
 	if err != nil {
 		return fmt.Errorf("pod store creation error: %v", err)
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -70,9 +70,12 @@ func (s *schedulingThroughputMeasurement) Execute(config *measurement.Measuremen
 		if err := selector.Parse(config.Params); err != nil {
 			return nil, err
 		}
-
+		measurmentInterval, err := util.GetDurationOrDefault(config.Params, "measurmentInterval", defaultWaitForPodsInterval)
```
mm4tt: Would you mind creating a new const to replace `defaultWaitForPodsInterval`? 
The `defaultWaitForPodsInterval` is defined in other file, with main purpose of being the interval for "waitForPods" measurement. We shouldn't have a dependency like that - imagine someone changing wait for pods interval and having a side effect of changing how scheduler_throughput works, not really a best code :)

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -167,4 +171,5 @@ type schedulingThroughput struct {
 	Perc50  float64 `json:"perc50"`
 	Perc90  float64 `json:"perc90"`
 	Perc99  float64 `json:"perc99"`
+	Max     float64 `json:"max"`
```
mm4tt: Are we going to delete the Average field as well?

clusterloader2/testing/load/experimental-config.yaml
```diff
@@ -424,6 +424,7 @@ steps:
     Params:
       action: start
       labelSelector: group = scheduler-throughput
+      measurmentInterval: 1s
```
mm4tt: This is great, if we make this change then we can also reduce the number of the "scheduler-throughput" pods. We don't need 1000 anymore, something like 300 should be enough.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
jprzychodzen: I'm not sure about it - I know that there is #1083 about broken metric.

Let's leave it as it for now without touching this issue - I will either fix avg or remove it, however it's not a critical part of solving Scheduler Throughput meassurment.

clusterloader2/testing/load/experimental-config.yaml
jprzychodzen: This is only experimental config - if this will work, I will work on lowering those numbers. Right now it's only a matter of few seconds. Please keep it as is.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
jprzychodzen: I'm aware of relationship between `defaultWaitForPodsInterval` and `waitForPods`, however I'm not sure that we really want to remove this side-effect.

Aligning waitForPods and schedulingThroughput has a benefit that we have a shared precision, and that looks reasonable. If you feel bad about it, then I guess it should be refactored to some shared common default precision variable. WDYT?

mm4tt: I don't agree that shared precision looks reasonable. It's not even a precision in case of the waitForPods. 
In the waitForPods this parameter only controls how often we check whether a certain condition has occurred. One could easily argue, that we could increase this parameter to 30s to reduce CPU usage of clusterloader2 and it would have no-impact on our tests.
On the other hand, increasing this param to 30s would completely break the scheduler throughput for smaller tests.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
mm4tt: Sound good as long as we have in on the radar

clusterloader2/testing/load/experimental-config.yaml
mm4tt: Personally, I'd just do these two steps at once, but you're the owner of this issue, it's your call :)

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/1007

clusterloader2/pkg/measurement/common/scheduling_throughput.go
jprzychodzen: After second though - removed Avg as we are no using it. It there will be a FR to keep it, then we will need to figure out how should it behave.

Resolving #1083

clusterloader2/pkg/measurement/common/scheduling_throughput.go
jprzychodzen: Added independent default equal to 5s to be backward compatible.

PR #1084, opened 2020-02-27
clusterloader2: collect scheduling_algorithm_duration_seconds to see algorithm latency without binding

ingvagabund:
Also, update vendored `k8s.io/kubernetes/pkg/scheduler/metrics`.

PR #1094, opened 2020-03-02
Update perfdash to 2.20.

oxddr:
/assign @mm4tt


oxddr:
ref https://github.com/kubernetes/perf-tests/pull/1084

PR #1097, opened 2020-03-03
[Access-tokens] Backport test and dependencies from 'master'

jprzychodzen:
Backport:
* Access-tokens test
* Template function 'Loop'
* Template function 'ifThenElse'
* 'Sleep' measurement
* backport simple_exectuor template parameters

Ref. kubernetes/kubernetes#83375

### DIFF
--- a/verify/verify-flags/known-flags.txt
+++ b/verify/verify-flags/known-flags.txt
@@ -20,6 +20,10 @@ n-runs-count
 ping-server-address
 ping-server-bind-address
 ping-sleep-duration
+prometheus-scrape-etcd
+prometheus-scrape-kube-proxy
+prometheus-scrape-kubelets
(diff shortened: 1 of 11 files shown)


verify/verify-flags/known-flags.txt
```diff
@@ -20,6 +20,10 @@ n-runs-count
 ping-server-address
 ping-server-bind-address
 ping-sleep-duration
+prometheus-scrape-etcd
+prometheus-scrape-kube-proxy
```
mborsz: Is this needed? You don't backport any Prometheus component?

jprzychodzen: It's not necessary, but if someone uses `https://github.com/kubernetes/perf-tests#repository-setup` from master branch, then it breaks pushing.

PR #1098, opened 2020-03-03
perfdash: add deploy target to Makefile.

oxddr:
Also make deploy.sh to use temp directory, which is removed by the script fully.

/assign @mm4tt

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -27,5 +27,8 @@ container: perfdash
 push: container
 	gcloud docker -s $(REPO) -- push $(REPO)/perfdash:$(TAG)
 
+deploy: container
+	./deploy.sh
+
 clean:
 	rm -f perfdash
--- a/perfdash/deploy.sh
+++ b/perfdash/deploy.sh
@@ -14,14 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -euxo pipefail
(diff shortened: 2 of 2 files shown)


perfdash/Makefile
```diff
@@ -27,5 +27,8 @@ container: perfdash
 push: container
 	gcloud docker -s $(REPO) -- push $(REPO)/perfdash:$(TAG)
 
+deploy: containter
```
mm4tt: typo: container

perfdash/deploy.sh
```diff
@@ -14,14 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -euxo pipefail
+set -euo pipefail
+set -x
```
mm4tt: Could you explain why you're making this change (I'm curious what's the difference)

oxddr: `set -euo pipefail` is the usual clause put at the top of all bash scripts, to have sane behavior and I as is is easily recognizable. `-x` is to make the script verbose and you can see it much less often that the former. I like to have it on a separate line, so I am not surprised, when the script starts to spit many lines. 

This might be personal I'm fine with reverting it.

mm4tt: This is fine. Thanks for explaining

/lgtm
/approve

PR #1102, opened 2020-03-03
Disable Go Modules in scripts used by travis (and pre-push hook)

oxddr:
/assign @mm4tt

### DIFF
--- a/verify/test.sh
+++ b/verify/test.sh
@@ -39,4 +39,4 @@ find_test_dirs() {
   )
 }
 
-go test $(find_test_dirs)
+GO111MODULE=off go test $(find_test_dirs)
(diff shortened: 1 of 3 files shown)


verify/test.sh
```diff
@@ -39,4 +39,4 @@ find_test_dirs() {
   )
 }
 
-go test $(find_test_dirs)
+GO111MODULE=off go test $(find_test_dirs)
```
mm4tt: This isn't really necessary, right? We use golang1.11 and all the code is in the gopath inside the travis container, hence everything works as if GO111MODULE was set to off

Would you mind explaining more why we need this change? I don't see anything in the PR description

oxddr: This script is used also outside Travis, e.g. in the pre-push hook (it was also changed in this PR). I personally run those scripts on my machine, and with go1.13+ they stopped to work as modules are on by default. 

From Travis perspective, this shouldn't change anything.

PR #1103, opened 2020-03-03
Update to dnsperf 2.3.0 in dns tests.

prameshj:
This version has the option to send queries over tcp and using a
specific source port.

This PR adds the change to the Makefile and Dockerfile. The tests will be modified to use this image in a followup PR.


prameshj:
/hold

There is more work involved to get the histogram generation patch into the new dnsperf code and build the image. This is needed for reporting latency after the test runs.

PR #1106, opened 2020-03-04
Fix confusing log line in exposeAPIServerMetrics

mm4tt:
This method used to be called only for kubemark clusters, but it's not anymore. When it's called for non-kubemark cluster the log statement is confusing.


oxddr:
/lgtm

Thanks!

Issue #1005, opened 2020-02-03
NodeKiller seems to be not working in 100 node 1.17 / master performance tests


mm4tt:
Original debugging done by @jkaniuk:

> In 100 nodes OSS performance tests of 1.16:  
> https://k8s-testgrid.appspot.com/sig-scalability-gce#gce-cos-1.16-scalability-100

> NodeKiller is consistently failing:
> https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-stable1/1219228169567997957

> ```
> W0120 12:25:21.234] I0120 12:25:21.234558   12979 nodes.go:105] NodeKiller: Rebooting "e2e-big-minion-group-tt6r" to repair the node
> W0120 12:25:24.556] I0120 12:25:24.555774   12979 ssh.go:38] ssh to "e2e-big-minion-group-tt6r" finished with "External IP address was not found; defaulting to using IAP > tunneling.\npacket_write_wait: Connection to UNKNOWN port 65535: Broken pipe\r\nERROR: (gcloud.compute.ssh) [/usr/bin/ssh] exited with return code [255].\n": exit status 255
> W0120 12:25:24.556] E0120 12:25:24.555839   12979 nodes.go:108] NodeKiller: Error while rebooting node "e2e-big-minion-group-tt6r": exit status 255
> ```

mm4tt:
Looks like it transiently fails in 1.16, meaning that some of the ssh calls succeed and some not (within a single run), e.g.

OK - `W0129 12:36:05.537] I0129 12:36:05.536771   10910 ssh.go:38] ssh to "e2e-big-minion-group-5l39" finished with "External IP address was not found; defaulting to using IAP tunneling.\nWarning: Permanently added 'compute.691072012589517573' (ED25519) to the list of known hosts.\r\nWarning: Stopping docker.service, but it can still be activated by:\n  docker.socket\n": <nil>`

BAD - `W0129 12:46:08.636] I0129 12:46:08.636013   10910 ssh.go:38] ssh to "e2e-big-minion-group-5l39" finished with "External IP address was not found; defaulting to using IAP tunneling.\npacket_write_wait: Connection to UNKNOWN port 65535: Broken pipe\r\nERROR: (gcloud.compute.ssh) [/usr/bin/ssh] exited with return code [255].\n": exit status 255`

mm4tt:
[...]

PR #1107, opened 2020-03-04
[NodeKiller] Change rounding of a number of selected nodes

jprzychodzen:
Ref. #1005

### DIFF
--- a/clusterloader2/pkg/chaos/nodes.go
+++ b/clusterloader2/pkg/chaos/nodes.go
@@ -18,6 +18,7 @@ package chaos
 
 import (
 	"fmt"
+	"math"
 	"math/rand"
 	"sync"
 	"time"
@@ -86,6 +87,7 @@ func (k *NodeKiller) pickNodes() ([]v1.Node, error) {
 	for i := range prometheusPods {
 		if prometheusPods[i].Spec.NodeName != "" {
 			nodesHasPrometheusPod.Insert(prometheusPods[i].Spec.NodeName)
+			klog.Infof("%s: Node %s removed from killing. Runs pod %s", k, prometheusPods[i].Spec.NodeName, prometheusPods[i].Name)
 		}
 	}
 
@@ -98,9 +100,13 @@ func (k *NodeKiller) pickNodes() ([]v1.Node, error) {
 	rand.Shuffle(len(nodes), func(i, j int) {
 		nodes[i], nodes[j] = nodes[j], nodes[i]
 	})
-	numNodes := int(k.config.FailureRate * float64(len(nodes)))
+	numNodes := int(math.Ceil(k.config.FailureRate * float64(len(nodes))))
+	klog.Infof("%s: %d nodes available, wants to fail %d nodes", k, len(nodes), numNodes)
 	if len(nodes) > numNodes {
-		return nodes[:numNodes], nil
+		nodes = nodes[:numNodes]
+	}
+	for _, node := range nodes {
+		klog.Infof("%s: Node %q schedule for failure", k, node.Name)
 	}
 	return nodes, nil
 }
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -67,8 +67,8 @@ tuningSets:
 chaosMonkey:
   nodeFailure:
     failureRate: 0.01
-    interval: 1m
-    jitterFactor: 10.0
+    interval: 5m
+    jitterFactor: 2.0
     simulatedDowntime: 10m
 {{end}}
 steps:


clusterloader2/pkg/chaos/nodes.go
```diff
@@ -98,11 +99,15 @@ func (k *NodeKiller) pickNodes() ([]v1.Node, error) {
...
 		nodes[i], nodes[j] = nodes[j], nodes[i]
 	})
-	numNodes := int(k.config.FailureRate * float64(len(nodes)))
-	if len(nodes) > numNodes {
-		return nodes[:numNodes], nil
+	expectedNodesToFail := k.config.FailureRate * float64(len(nodes))
+	nodesToFail := nodes[:0]
+	klog.Infof("%s: %d nodes available, expecting to fail %f nodes", k, len(nodes), expectedNodesToFail)
+	for _, node := range nodes {
+		if rand.Float64() < k.config.FailureRate {
```
mm4tt: This is already randomized, we call rand.Shuffle in line 99

clusterloader2/pkg/chaos/nodes.go
```diff
@@ -124,7 +129,7 @@ func (k *NodeKiller) kill(nodes []v1.Node) {
 			time.Sleep(time.Duration(k.config.SimulatedDowntime))
 
 			klog.Infof("%s: Rebooting %q to repair the node", k, node.Name)
-			err = util.SSH("sudo reboot", &node, nil)
+			err = util.SSH("nohup sudo reboot +1s > /dev/null 2> /dev/null < /dev/null &", &node, nil)
```
mm4tt: It could be a different PR (or at least commit).
Also, could you explain how it works? Ideally, in a comment. Personally, I don't understand the +1s part

clusterloader2/pkg/chaos/nodes.go
```diff
@@ -98,11 +99,15 @@ func (k *NodeKiller) pickNodes() ([]v1.Node, error) {
 	rand.Shuffle(len(nodes), func(i, j int) {
 		nodes[i], nodes[j] = nodes[j], nodes[i]
 	})
-	numNodes := int(k.config.FailureRate * float64(len(nodes)))
-	if len(nodes) > numNodes {
-		return nodes[:numNodes], nil
+	expectedNodesToFail := k.config.FailureRate * float64(len(nodes))
+	nodesToFail := nodes[:0]
+	klog.Infof("%s: %d nodes available, expecting to fail %f nodes", k, len(nodes), expectedNodesToFail)
```
mm4tt: +1 for adding more logs (I'd even add more, e.g. listing exactly which nodes were returned here)

clusterloader2/pkg/chaos/nodes.go
jprzychodzen: Sure, moved to  PR #1109  with extensive comment.

clusterloader2/pkg/chaos/nodes.go
jprzychodzen: They are randomized, but the size is constant.

And if we are expecting that less than a single node will fail (as for load test, with failure rate 0.01, and less than 100 eligible nodes, as some of them are running Prometheus), no node is scheduled.

clusterloader2/pkg/chaos/nodes.go
jprzychodzen: This information is available from 'kill' function in line 117, but I can duplicate that.

clusterloader2/pkg/chaos/nodes.go
mm4tt: I see, good debugging!
But, wouldn't be better to simply set numNodes to 1 if it's 0?
What I don't like about your solution is that, while it'll kill the `FailureRate` of nodes on average, we'll have test-runs where no nodes were killed and we'll have runs where 1+ nodes were killed. It's quite likely that it will make the test flaky - most likely it won't be failing but it can result in big variance of many metrics.

jprzychodzen:
[Persubmit job does not have chaos monkey override enabled](https://github.com/kubernetes/test-infra/blob/308dab8f7f91eb1afccb5f174ca7c1f73f4d28b7/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L305) , so this code is never invoked at this job.

clusterloader2/pkg/chaos/nodes.go
jprzychodzen: Sure, changed that to get more stable tests. Also, increased minimal interval in load test to prevent overkilling.

jprzychodzen:
Fixes #1005

PR #1109, opened 2020-03-05
[NodeKiller] Schedule reboot and disconnect from node

jprzychodzen:


jprzychodzen:
/assign @mm4tt 
Ref. #1005
Ref. #1107

clusterloader2/pkg/chaos/nodes.go
```diff
@@ -124,7 +124,16 @@ func (k *NodeKiller) kill(nodes []v1.Node) {
...
 
 			klog.Infof("%s: Rebooting %q to repair the node", k, node.Name)
-			err = util.SSH("sudo reboot", &node, nil)
+			// Scheduling a reboot in one second, then disconnecting.
+			//
+			// Bash command explanation:
+			// 'nohup' - Making sure that end of SSH connection signal will not break sudo
+			// 'sudo' - Elevated priviliages, required by 'reboot'
+			// 'reboot' - Rebooting node
+			// '+1s' - Parameter to 'reboot', to wait 1 second before rebooting.
```
mm4tt: I might be wrong but I'm not sure whether reboot accepts such parameter. I checked `sudo reboot --help` on COS system and such feature wasn't documented anywhere.

clusterloader2/pkg/chaos/nodes.go
```diff
@@ -124,7 +124,16 @@ func (k *NodeKiller) kill(nodes []v1.Node) {
...
+			// '> /dev/null 2> /dev/null < /dev/null' - File descriptor redirect, as we are closing terminal
```
mm4tt: The explanation is not very clear. Wouldn't be better to write something like:
```
// redirect all three I/O to avoid ssh hanging, see https://web.archive.org/web/20090429074212/http://www.openssh.com/faq.html#3.10
```
?

clusterloader2/pkg/chaos/nodes.go
jprzychodzen: You are right, this should use `shutdown` command.

clusterloader2/pkg/chaos/nodes.go
jprzychodzen: Added this comment

Issue #1101, opened 2020-03-03
Service Creation Latency Measurement fails when PodService is Nil


vamossagar12:
Service Creation Latency measurement throws a Segmentation Violation Error at the following line:

https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/service_creation_latency.go#L316

This happens when exec_service returns PodService as nil here:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/execservice/exec_service.go#L137

The problem is because the error returned by above method is not being handled:

https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/service_creation_latency.go#L306

We will need to handle this.

Also, we shouldn't run service_creation_latency measurement if execservice is disabled.

PR #1111, opened 2020-03-05
Fix to handle exec_service error + not starting service creation latency measurment when --enable-exec-service flag is disabled

vamossagar12:
This PR fixes https://github.com/kubernetes/perf-tests/issues/1101. It adds checks to handle errors thrown by exec_service and also, to not start service_creation_latency when --enable-exec-service flag is disabled.

### DIFF
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -86,6 +86,9 @@ func (s *serviceCreationLatencyMeasurement) Execute(config *measurement.Measurem
 	if err != nil {
 		return nil, err
 	}
+	if !config.ClusterLoaderConfig.EnableExecService {
+		return nil, fmt.Errorf("enable-exec-service flag not enabled")
+	}
 
 	switch action {
 	case "start":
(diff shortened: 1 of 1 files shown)


vamossagar12:
@mm4tt , as discussed separate PR to handle service_creation_latency measurement related scenarios.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -304,6 +307,11 @@ func (p *pingChecker) run() {
 			}
 			// TODO(#685): Make ping checks less communication heavy.
 			pod, err := execservice.GetPod()
+			if err != nil {
+				success = 0
```
mm4tt: I'm not very familiar with this code, but as a general rule, if you're swallowing the error it would be good to at least log it. Otherwise it would be really hard to debug if it stops working.

Could you add a klog.Warning somewhere here?

vamossagar12: Yeah that was a miss. Added it.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -304,6 +307,12 @@ func (p *pingChecker) run() {
 			}
 			// TODO(#685): Make ping checks less communication heavy.
 			pod, err := execservice.GetPod()
+			if err != nil {
+				klog.Warningf("No exec service pod found in cluster.")
```
mm4tt: Please include the original error in the message, e.g.
```
klog.Warningf("call to execservice.GetPod() ended with error: %v", err)
```

vamossagar12: done.

mm4tt:
/ok-to-test
/lgtm
/approve

Thanks @vamossagar12 !

PR #1123, opened 2020-03-13
add arm64 support to netperf image

lubinsz:
I have tested it with Arm64 k8s cluster, the result seems OK.

Signed-off-by: Bin Lu <bin.lu@arm.com>

### DIFF
--- a/hack/build/init-buildx.sh
+++ b/hack/build/init-buildx.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


lubinsz:
@girishkalele 
Can you check it?
/assign @girishkalele

lubinsz:
cc @BenTheElder

hack/build/init-buildx.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
```
BenTheElder: this script is a hacky experiment coped from the kind repo, I'm not ready to encourage replicating it throughout the project yet.

BenTheElder:
I don't own this, however I am noting that we should probably not copy those scripts from kind yet (also normally when copying code it is good to note [where it came from](https://github.com/kubernetes-sigs/kind/blob/49eba13ba76d76da867e4949486a3b3c315dbda1/hack/build/init-buildx.sh)).
/uncc

PR #1124, opened 2020-03-16
Add measurement to collect windows node cpu usage per process

YangLu1031:
1. As windows server container doesn't support privileged containers, so to collect the cpu usage per process, have to install the [wmi_exporter](https://github.com/martinlindhe/wmi_exporter#installation) onto windows node using `gcloud ssh` command. As a result, this feature is only supported for GCE cluster.
2. Use the "[Additional Scrape configs](https://github.com/coreos/prometheus-operator/blob/master/Documentation/additional-scrape-config.md)" in prometheus-operator to scrape windows node.
3. Add a new measurement to collect the metrics once when all pods are running, as we want to find out if there's any noticeable high cpu usage process. Later on, we may change to calculate the percentiles during the test run and put it on perf-dash board.

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
+++ b/clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package slos
+
+import (
+	"math"
+	"time"
+
+	"github.com/prometheus/common/model"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	windowsResourceUsagePrometheusMeasurementName = "WindowsResourceUsagePrometheus"
+	// get top 10 non-system processes with highest cpu usage within 1min query window size
+	cpuUsageQueryTop10                        = `topk(10, sum by (process) (rate(wmi_process_cpu_time_total{process!~"Idle|Total|System"}[1m]) / on(job) group_left wmi_cs_logical_processors) * 100)`
+	currentWindowsResourceUsageMetricsVersion = "v1"
+)
+
+type windowsResourceUsageGatherer struct{}
+
+func (w *windowsResourceUsageGatherer) IsEnabled(config *measurement.MeasurementConfig) bool {
+	return true
+}
+
+func (w *windowsResourceUsageGatherer) String() string {
+	return windowsResourceUsagePrometheusMeasurementName
+}
+
+func init() {
+	create := func() measurement.Measurement { return createPrometheusMeasurement(&windowsResourceUsageGatherer{}) }
+	if err := measurement.Register(windowsResourceUsagePrometheusMeasurementName, create); err != nil {
+		klog.Fatalf("Cannot register %s: %v", windowsResourceUsagePrometheusMeasurementName, err)
+	}
+}
+
+func convertToPerfData(samples []*model.Sample) *measurementutil.PerfData {
+	perfData := &measurementutil.PerfData{Version: currentWindowsResourceUsageMetricsVersion}
+	for _, sample := range samples {
+		item := measurementutil.DataItem{
+			Data: map[string]float64{
+				"CPU_Usage": math.Round(float64(sample.Value)*100) / 100,
+			},
+			Unit: "%",
+			Labels: map[string]string{
+				"Process": string(sample.Metric["process"]),
+			},
+		}
+		perfData.DataItems = append(perfData.DataItems, item)
+	}
+	return perfData
+}
+
+// Gather gathers the metrics and convert to json summary
+func (w *windowsResourceUsageGatherer) Gather(executor QueryExecutor, startTime time.Time, config *measurement.MeasurementConfig) (measurement.Summary, error) {
+	samples, err := executor.Query(cpuUsageQueryTop10, time.Now())
+	if err != nil {
+		return nil, err
+	}
+	content, err := util.PrettyPrintJSON(convertToPerfData(samples))
+	if err != nil {
+		return nil, err
+	}
+	summaryName, err := util.GetStringOrDefault(config.Params, "summaryName", windowsResourceUsagePrometheusMeasurementName)
+	if err != nil {
+		return nil, err
+	}
+	summary := measurement.CreateSummary(summaryName, "json", content)
+	return summary, nil
+}
--- a/clusterloader2/pkg/prometheus/gce_windows_util.go
+++ b/clusterloader2/pkg/prometheus/gce_windows_util.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prometheus
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
+)
+
+// gceNode is a struct storing gce node info
+type gceNode struct {
+	projectID string
+	zone      string
+	nodeName  string
+}
+
+func setUpWindowsNodeAndTemplate(k8sClient kubernetes.Interface, mapping map[string]interface{}) error {
+	// Get the gce windows node info
+	windowsNode, err := getGceWindowsNodeFromKubernetesService(k8sClient)
+	if err != nil {
+		return err
+	}
+	// Install the wmi exporter onto windows node
+	if err := installWmiExporter(k8sClient, windowsNode, mapping); err != nil {
+		return err
+	}
+	mapping["WINDOWS_NODE_NAME"] = windowsNode.nodeName
+	mapping["PROMETHEUS_SCRAPE_WINDOWS_NODES"] = true
+	return nil
+}
+
+func isWindowsNodeScrapingEnabled(mapping map[string]interface{}, clusterLoaderConfig *config.ClusterLoaderConfig) bool {
+	if windowsNodeTest, exists := mapping["WINDOWS_NODE_TEST"]; exists && windowsNodeTest.(bool) && clusterLoaderConfig.ClusterConfig.Provider == "gce" {
+		return true
+	}
+	return false
+}
+
+func installWmiExporter(k8sClient kubernetes.Interface, windowsNode *gceNode, mapping map[string]interface{}) error {
+	wmiExporterURL, ok := mapping["WMI_EXPORTER_URL"]
+	if !ok {
+		return fmt.Errorf("missing setting up wmi exporter download url")
+	}
+	wmiExporterCollectors, ok := mapping["WMI_EXPORTER_ENABLED_COLLECTORS"]
+	if !ok {
+		return fmt.Errorf("missing setting up wmi exporter enabled collectors")
+	}
+	// Tried executing the invoke-webrequest cmd directly in gcloud ssh bash, but get this error:
+	// 		invoke-webrequest : Win32 internal error "Access is denied" 0x5 occurred while reading the console output buffer.
+	// 		Contact Microsoft Customer Support Services.
+	// After wrap it up in script block, it works well, also support timeout setting.
+	installWmiCmdTemplate := `Start-Job -Name InstallWmiExporter -ScriptBlock {invoke-webrequest -uri %s -outfile C:\wmi_exporter.msi; Start-Process msiexec.exe -Wait -ArgumentList '/I C:\wmi_exporter.msi ENABLED_COLLECTORS=%s LISTEN_PORT=5000 /quiet'}; Wait-Job -Timeout 300 -Name InstallWmiExporter`
+	installWmiExporterCmd := fmt.Sprintf(installWmiCmdTemplate, wmiExporterURL, wmiExporterCollectors)
+	powershellCmd := fmt.Sprintf(`powershell.exe -Command "%s"`, installWmiExporterCmd)
+	if windowsNode == nil {
+		return fmt.Errorf("no windows nodes available to install wmi exporter")
+	}
+	klog.Infof("Installing wmi exporter onto projectId: %s, zone: %s, nodeName: %s with cmd: %s", windowsNode.projectID, windowsNode.zone, windowsNode.nodeName, installWmiExporterCmd)
+	cmd := exec.Command("gcloud", "compute", "ssh", "--project", windowsNode.projectID, "--zone", windowsNode.zone, windowsNode.nodeName, "--command", powershellCmd)
+	err := cmd.Run()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func getGceWindowsNodeFromKubernetesService(k8sClient kubernetes.Interface) (*gceNode, error) {
+	var nodeList *corev1.NodeList
+	f := func() error {
+		var err error
+		nodeList, err = k8sClient.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: "kubernetes.io/os=windows"})
+		return err
+	}
+
+	if err := client.RetryWithExponentialBackOff(client.RetryFunction(f)); err != nil {
+		return nil, err
+	}
+
+	if len(nodeList.Items) == 0 {
+		return nil, fmt.Errorf("no windows nodes available in kubernetes service")
+	}
+	// TODO: Modify to support multiple windows nodes soon.
+	providerStrs := strings.Split(nodeList.Items[0].Spec.ProviderID, "/")
+	if len(providerStrs) < 5 {
+		return nil, fmt.Errorf("no valid gce provider ID available")
+	}
+
+	// Gce providerID format: gce://{projectID}/{zone}/{nodeName}
+	return &gceNode{
+		projectID: providerStrs[2],
+		zone:      providerStrs[3],
+		nodeName:  providerStrs[4],
+	}, nil
+}
--- a/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
@@ -1,4 +1,5 @@
 {{$PROMETHEUS_SCRAPE_KUBELETS := DefaultParam .PROMETHEUS_SCRAPE_KUBELETS false}}
+{{$PROMETHEUS_SCRAPE_WINDOWS_NODES := DefaultParam .PROMETHEUS_SCRAPE_WINDOWS_NODES false}}
 
 apiVersion: monitoring.coreos.com/v1
 kind: Prometheus
@@ -47,3 +48,8 @@ spec:
             storage: {{MultiplyInt 10 (AddInt 1 (DivideInt .Nodes 1000))}}Gi
   query:
     maxSamples: 100000000
+  {{if $PROMETHEUS_SCRAPE_WINDOWS_NODES}}
+  additionalScrapeConfigs:
+    name: windows-scrape-configs
+    key: windows-scrape-configs.yaml
+  {{end}}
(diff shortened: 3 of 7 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -335,6 +392,35 @@ func dumpAdditionalLogsOnPrometheusSetupFailure(k8sClient kubernetes.Interface)
...
+	}
+
+	if err := client.RetryWithExponentialBackOff(client.RetryFunction(f)); err != nil {
+		return nil, err
+	}
+
+	if len(nodeList.Items) == 0 {
+		return nil, fmt.Errorf("no windows nodes available in kubernetes service")
+	}
+	// As we could scrape only one windows node as of now, so pick the first
```
yliaog: why only one node now?

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
pjh: 2020

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	windowsResourceUsagePrometheusMeasurementName = "WindowsResourceUsagePrometheus"
+	// get top 10 non-system processes with highest cpu usage within 1min query window size
+	cpuUsageQueryTop10 = `topk(10, sum by (process) (rate(wmi_process_cpu_time_total{process!~"Idle|Total|System"}[1m]) / on(job) group_left wmi_cs_logical_processors) * 100)`
+)
+
+type windowsResourceUsageMetricMeasurement struct {
```
pjh: Is this type actually used anywhere?

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -0,0 +1,90 @@
...
+			Labels: map[string]string{
+				"Process": string(sample.Metric["process"]),
+			},
+		}
+		perfData.DataItems = append(perfData.DataItems, item)
+	}
+	return perfData
+}
+
+// Gather the metrics and convert to json summary
```
pjh: This should be "Gather gathers the metrics and converts..."

clusterloader2/pkg/prometheus/manifests/prometheus-additional-scrape-configs.yaml
```diff
@@ -0,0 +1,12 @@
+{{$WINDOWS_NODE_NAME := DefaultParam .WINDOWS_NODE_NAME ""}}
```
pjh: Maybe name this file "windows-additional-scrape-configs.yaml" instead?

clusterloader2/pkg/prometheus/manifests/prometheus-additional-scrape-configs.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+apiVersion: v1
+kind: Secret
+metadata:
+  namespace: monitoring
+  name: windows-scrape-configs
+stringData:
+  prometheus-additional.yaml: |-
```
pjh: windows-additional.yaml? windows-scrape-configs.yaml?

To future-proof this in case someone else wants to add additional scrape configs in the future that might not apply to Windows nodes.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -48,6 +50,12 @@ const (
 	checkPrometheusReadyTimeout  = 15 * time.Minute
 	numK8sClients                = 1
 	nodeExporterPod              = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/exporters/node-exporter.yaml"
+	wmiExporterURL               = "https://github.com/martinlindhe/wmi_exporter/releases/download/v0.10.1/wmi_exporter-0.10.1-amd64.msi"
+	wmiExporterCollectors        = "process,cs"
```
pjh: It would be nice if there was a way to specify this as part of the yaml config somewhere, instead of hard-coding it here. Not sure if there's an easy way to do that though.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -118,18 +133,60 @@ func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (p
...
 
+func enableScrapingWindowsNode(mapping map[string]interface{}, clusterLoaderConfig *config.ClusterLoaderConfig) bool {
+	if windowsNodeTest, exists := mapping["WINDOWS_NODE_TEST"]; exists && windowsNodeTest.(bool) && clusterLoaderConfig.ClusterConfig.Provider == "gce" {
+		return true
+	}
+	return false
+}
+
+func installWmiExporter(k8sClient kubernetes.Interface) error {
+	InstallWmiCmdTemplate := `Start-Job -Name InstallWmiExporter -ScriptBlock {invoke-webrequest -uri %s -outfile C:\wmi_exporter.msi; Start-Process msiexec.exe -Wait -ArgumentList '/I C:\wmi_exporter.msi ENABLED_COLLECTORS=%s LISTEN_PORT=5000 /quiet'}; Wait-Job -Name InstallWmiExporter`
```
pjh: Is there a way to give the `Wait-Job` a timeout, so that this won't hang forever in case it fails?

pjh: InstallWmiCmdTemplate -> installWmiCmdTemplate?

same for other variables below.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -118,18 +133,60 @@ func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (p
...
+	InstallWmiExporterCmd := fmt.Sprintf(InstallWmiCmdTemplate, wmiExporterURL, wmiExporterCollectors)
+	powershellCmd := fmt.Sprintf(`powershell.exe -Command "%s"`, InstallWmiExporterCmd)
+	if windowsNode == nil {
```
pjh: It's weird that windowsNode is a global variable. Is there a way to pass it as an arg here?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -118,18 +133,60 @@ func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (p
...
+		return fmt.Errorf("no windows nodes available to install wmi exporter")
+	}
+	klog.Infof("Installing wmi exporter onto projectId: %s, zone: %s, nodeName: %s with cmd: %s", windowsNode.projectID, windowsNode.zone, windowsNode.nodeName, InstallWmiExporterCmd)
+	cmd := exec.Command("gcloud", "compute", "ssh", "--project", windowsNode.projectID, "--zone", windowsNode.zone, windowsNode.nodeName, "--command", powershellCmd)
+	err := cmd.Run()
+	if err != nil {
+		return err
```
pjh: Write a log message here?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -335,6 +392,35 @@ func dumpAdditionalLogsOnPrometheusSetupFailure(k8sClient kubernetes.Interface)
 	klog.Info(string(s))
 }
 
+func getGceWindowsNodeNameFromKubernetesService(k8sClient kubernetes.Interface) (*gceNode, error) {
```
pjh: Should this function and `type gceNode` go in a separate `gce.go` file?

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: 1. The windows node name would be passed into `targets: ["{{$WINDOWS_NODE_NAME}}:5000"]` in the additional-scrape-config.yaml file, and I don't know how to dynamically create slice of targets in yaml file yet. Will change it if I figured out later. 
2. And load test is for 1-node cluster, so scrape one node is sufficient as of now.

clusterloader2/pkg/prometheus/manifests/prometheus-additional-scrape-configs.yaml
mm4tt: +1

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -0,0 +1,90 @@
...
+
+func init() {
+	create := func() measurement.Measurement { return createPrometheusMeasurement(&windowsResourceUsageGatherer{}) }
+	if err := measurement.Register(windowsResourceUsagePrometheusMeasurementName, create); err != nil {
+		klog.Fatalf("Cannot register %s: %v", windowsResourceUsagePrometheusMeasurementName, err)
+	}
+}
+
+func convertToPerfData(samples []*model.Sample) *measurementutil.PerfData {
+	perfData := &measurementutil.PerfData{Version: currentAPICallMetricsVersion}
```
mm4tt: Let's not mix constants across different measurements, this measurement has nothing to do with API Call Latency one.

clusterloader2/pkg/prometheus/manifests/prometheus-additional-scrape-configs.yaml
YangLu1031: Seems all the prometheus yaml files have `prometheus-` prefix, so change it to "prometheus-windows-scrape-configs.yaml"

clusterloader2/pkg/prometheus/manifests/prometheus-additional-scrape-configs.yaml
YangLu1031: make sense. changed to windows-scrape-configs.yaml

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
YangLu1031: Oh, forgot to clean it up. Removed

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	windowsResourceUsagePrometheusMeasurementName = "WindowsResourceUsagePrometheus"
+	// get top 10 non-system processes with highest cpu usage within 1min query window size
+	cpuUsageQueryTop10 = `topk(10, sum by (process) (rate(wmi_process_cpu_time_total{process!~"Idle|Total|System"}[1m]) / on(job) group_left wmi_cs_logical_processors) * 100)`
```
mm4tt: IIUC, this query has unbounded timeframe meaning that it will be always taking into account all metrics since the prometheus server was started. Usually we try to limit the window to the actual timeframe of the test. If you run your test multiple times (either run CL2 multiple times without tearing down prometheus server or run a suite of multiple windows tests  in single CL2 invocation) this measurements will not work as expected. Could you fix it? I'm not familiar with the promQL syntax you used here (is it something new?), but maybe you can get some inspiration comparing how it's done in api-call-latency measurement - https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -47,3 +47,8 @@ spec:
             storage: {{MultiplyInt 10 (AddInt 1 (DivideInt .Nodes 1000))}}Gi
   query:
     maxSamples: 100000000
+  {{if and .WINDOWS_NODE_TEST (eq .PROVIDER "gce")}}
```
mm4tt: Let's keep the yamls as simple as possible. It would be better to have a dedicated variable controlling whether scraping windows nodes should be enabled or not. Something like:
```
{{if  .PROMETHEUS_SCRAPE_WINDOWS_NODES}}
...
{{end}}
```

This is a common pattern, e.g. see https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/manifests/master-ip/master-endpoints.yaml

clusterloader2/pkg/prometheus/prometheus.go
mm4tt: Yeah, would be great. Also this file is quite large already (close to 400 lines). Can we extract all the windows features to a separate file and just call something from here?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -118,18 +133,60 @@ func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (p
 		clusterLoaderConfig.PrometheusConfig.ScrapeKubeProxy = mapping["PROMETHEUS_SCRAPE_KUBE_PROXY"].(bool)
 	}
 	mapping["PROMETHEUS_SCRAPE_KUBELETS"] = clusterLoaderConfig.PrometheusConfig.ScrapeKubelets
+	// Add PROVIDER to template mapping for conditions control in prometheus-prometheus.yaml
+	if _, exists := mapping["PROVIDER"]; !exists {
```
mm4tt: If we simplify the yamls we won't need this

clusterloader2/pkg/prometheus/prometheus.go
mm4tt: Here is an example of slice variable in yaml - https://github.com/kubernetes/perf-tests/blob/3b2fdc539fb14ae7bb64e71370868fb1eb421e0b/clusterloader2/pkg/prometheus/manifests/master-ip/master-endpoints.yaml#L14

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: Seems not necessary, as the err will be logged in the caller method `SetUpPrometheusStack()` in clusterloader.go

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
YangLu1031: 1. I see, and CMIIW, I think I set it as 1min static bounded timeframe, as I wanted to collect the peak usage when all pods are running.
2. As I need to divide by logical processors which is a constant. If divided directly there was syntax error as it didn't work with rate() function, so I spent some time and found that using the many-to-one left join works fine. Reference: https://prometheus.io/docs/prometheus/latest/querying/operators/#many-to-one-and-one-to-many-vector-matches

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: As I separated the steps of 1) fetching windowsNode object to fill the windows scrape configs yaml, 2) install the wmi exporter from windowsNode object, so need a global object to share windows node info. Now combined them together, doesn't need it anymore.

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: Yeah, added 5m timeout as of now `-Timeout 300`

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
YangLu1031: I see, fixed

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: Ok, created `gce_windows_util.go` and put it there

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
YangLu1031: fixed

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: Not difficult, moved to `windows_override.yaml`

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: Thanks, will support multiple nodes in following PR.

clusterloader2/pkg/prometheus/gce_windows_util.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// gceNode is a struct storing gce node info
+type gceNode struct {
+	projectID string
+	zone      string
+	nodeName  string
+}
+
+func setUpWindowsNodeAndTemplate(k8sClient kubernetes.Interface, mapping map[string]interface{}) error {
```
pjh: This has to be (or at least should be ) capitalized in order to be exported outside the package, right? https://golang.org/doc/effective_go.html#commentary. It should also have a comment if it's exported.

Please check throughout the file.

clusterloader2/pkg/prometheus/gce_windows_util.go
```diff
@@ -0,0 +1,112 @@
...
+func installWmiExporter(k8sClient kubernetes.Interface, windowsNode *gceNode, mapping map[string]interface{}) error {
+	wmiExporterURL, ok := mapping["WMI_EXPORTER_URL"]
+	if !ok {
+		return fmt.Errorf("missing setting up wmi exporter download url")
+	}
+	wmiExporterCollectors, ok := mapping["WMI_EXPORTER_ENABLED_COLLECTORS"]
+	if !ok {
+		return fmt.Errorf("missing setting up wmi exporter enabled collectors")
+	}
+	installWmiCmdTemplate := `Start-Job -Name InstallWmiExporter -ScriptBlock {invoke-webrequest -uri %s -outfile C:\wmi_exporter.msi; Start-Process msiexec.exe -Wait -ArgumentList '/I C:\wmi_exporter.msi ENABLED_COLLECTORS=%s LISTEN_PORT=5000 /quiet'}; Wait-Job -Timeout 300 -Name InstallWmiExporter`
```
pjh: You could add a comment here explaining what you found: that the only way you could get this to work was using -ScriptBlock, executing it directly was failing for some reason, etc.

pjh:
This looks great, thanks Yang!

LGTM, will leave the rest of the review to the perf-tests folks :)

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -47,3 +48,8 @@ spec:
             storage: {{MultiplyInt 10 (AddInt 1 (DivideInt .Nodes 1000))}}Gi
   query:
     maxSamples: 100000000
+  {{if $PROMETHEUS_SCRAPE_WINDOWS_NODES}}
+  additionalScrapeConfigs:
```
mm4tt: I think you could move additionalScrapeConfigs outside the if block

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -118,6 +118,15 @@ func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (p
 		clusterLoaderConfig.PrometheusConfig.ScrapeKubeProxy = mapping["PROMETHEUS_SCRAPE_KUBE_PROXY"].(bool)
 	}
 	mapping["PROMETHEUS_SCRAPE_KUBELETS"] = clusterLoaderConfig.PrometheusConfig.ScrapeKubelets
+
+	// If enabled prometheus scrape windows node, need to setup windows node and template mapping
+	if enableScrapingWindowsNode(mapping, clusterLoaderConfig) {
```
mm4tt: Can we move it outside the constructor? I believe a better place for that code would be in the SetUpPrometheusStack method

clusterloader2/pkg/prometheus/gce_windows_util.go
mm4tt: I think it's going to be called from the same package (prometheus)

clusterloader2/pkg/prometheus/gce_windows_util.go
```diff
@@ -0,0 +1,112 @@
...
+	// Install the wmi exporter onto windows node
+	if err := installWmiExporter(k8sClient, windowsNode, mapping); err != nil {
+		return err
+	}
+	mapping["WINDOWS_NODE_NAME"] = windowsNode.nodeName
+	mapping["PROMETHEUS_SCRAPE_WINDOWS_NODES"] = true
+	return nil
+}
+
+func enableScrapingWindowsNode(mapping map[string]interface{}, clusterLoaderConfig *config.ClusterLoaderConfig) bool {
```
mm4tt: I'd prefer changing the name, `enableScrapingWindowsNode` suggests it's actually enabling something while it only checks whether something should be enabled or not. What about `shouldEnableScrapingWindowsNodes` or `isWindowsNodeScrapingEnabled` ?

clusterloader2/pkg/prometheus/gce_windows_util.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	if err := client.RetryWithExponentialBackOff(client.RetryFunction(f)); err != nil {
+		return nil, err
+	}
+
+	if len(nodeList.Items) == 0 {
+		return nil, fmt.Errorf("no windows nodes available in kubernetes service")
+	}
+	// As we could scrape only one windows node as of now, so pick the first
```
mm4tt: Can we add a TODO here to support more nodes in the future?

clusterloader2/pkg/prometheus/gce_windows_util.go
YangLu1031: Yeah, should be only called within prometheus package, so no need to expose it.

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
YangLu1031: Do you mean moving the additionalScrapeConfigs to the top of if block? I tried but got this error if `$PROMETHEUS_SCRAPE_WINDOWS_NODES` is false.
```
spec.additionalScrapeConfigs: Invalid value: "null": spec.additionalScrapeConfigs in body must be of type object: "null"
```

clusterloader2/pkg/prometheus/gce_windows_util.go
YangLu1031: Right, added

clusterloader2/pkg/prometheus/prometheus.go
YangLu1031: Agree, moved

clusterloader2/pkg/prometheus/gce_windows_util.go
YangLu1031: Agree, changed to isWindowsNodeScrapingEnabled

clusterloader2/pkg/prometheus/gce_windows_util.go
YangLu1031: Updated

YangLu1031:
@mm4tt Addressed your comments, can you take a look again?

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
mm4tt: Thanks for checking! Let's leave it as is in this case.

PR #1125, opened 2020-03-17
Set the extended load test override defaults to true

maniSbindra:
This PR is to resolve issue #1036 . As described in #1036 to resolve the issue changes also need to be made in k8s.io/test-infra, for which PR https://github.com/kubernetes/test-infra/pull/16818 has been created

Issue #803, opened 2019-09-18
Make PV test work on kubemark clusters


mm4tt:
Enabling PVs in the load test failed the kubemark presubmit because scheduler failed to schedule pods with PVs
```
Unable to schedule test-y6u59n-1/small-statefulset-0-0: no fit: 0/101 nodes are available: 1 node(s) were unschedulable, 100 node(s) had volume node affinity conflict.; waiting
```

In order to make it work on kubemark we'll most likely have to change (or ensure they already work this way) a few places in code so they're faking operations related to attaching and mounting PDs in kubemark
* Attach Detach Controller
* HollowKubelet
* Scheduler
* ?

Ref. https://github.com/kubernetes/perf-tests/issues/704

Jukie:
I'd like to pick this up, where would I begin?

mm4tt:
Hey, @Jukie, that's great to hear!

This one is a bit more complicated, I'll need to think about it more and come up with a more concrete list of steps. Unfortunately, I won't have time to do it until ~mid next week.
In the meantime, feel free to pick another issue from the [help wanted list](https://github.com/kubernetes/perf-tests/labels/help%20wanted). I think https://github.com/kubernetes/perf-tests/issues/595 is relatively simple and will get you through all the prep work required to start contributing to kubernetes (forking repos, signing CLA, etc.).

mm4tt:
/remove-lifecycle stale

We still want to do it...

PR #1126, opened 2020-03-18
Add CL2 prefix to ENABLE_PVS variable, to allow value to be overriden using env variable

maniSbindra:
This PR is based on the conversation https://github.com/kubernetes/perf-tests/pull/1125#discussion_r393675913. After this change we will be able to override the value of ENABLE_PVS variable. The default value of this variable will be true. 

In tests with kubemark cluster we will override this value to false due to issue https://github.com/kubernetes/perf-tests/issues/803.

Once this is merged we will PR into the test-infra repo to override this value to false for kubemark cluster tests.


clusterloader2/testing/load/config.yaml
```diff
@@ -20,7 +20,7 @@
 {{$ENABLE_CONFIGMAPS := DefaultParam .ENABLE_CONFIGMAPS false}}
 {{$ENABLE_DAEMONSETS := DefaultParam .ENABLE_DAEMONSETS false}}
 {{$ENABLE_JOBS := DefaultParam .ENABLE_JOBS false}}
-{{$ENABLE_PVS := DefaultParam .ENABLE_PVS false}}
+{{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
```
mm4tt: Let's just change the name, don't change the default yet. If you change the default it will break the kubemark tests. 
Adding
/ok-to-test
to let you see that the kubemark presubmit will fail (unless you change the default back to false)

Also, would it be a problem for you to update all variables (add prefix) for consistency? It'd be really great if you could do it

maniSbindra: Thanks, I will add the prefix to all variables defined under https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides

mm4tt: SG, thanks!

maniSbindra: Done. Request you to please review if anything needs to be modified. Thanks Mani.

maniSbindra:
Hi @mm4tt, I am not sure why the clusterloader2 test is failing. Could you suggest what I need to do?

mm4tt:
It might be a flake, let's try again

mm4tt:
Looks like the density test timed out in presubmit, strange given that this PR is only touching load test.
Let's try one more time

maniSbindra:
Thanks @mm4tt, this time it succeeded.

mm4tt:
/lgtm
/approve

Thanks!

PR #1127, opened 2020-03-19
clusterloader2: improve experience on aks

alexeldeib:
AKS, much like GKE, does not support SSH to master nodes or similar methods of contacting them. This PR avoids trying to do so as part of cluster loader tests for AKS. Without this, the TestMetrics gather step fails to SSH without a signer for the provider.

I imagine the same patterns would be broadly applicable for managed clusters. Perhaps we could generalize this?

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -144,10 +144,10 @@ func completeConfig(m *framework.MultiClientSet) error {
 		}
 	}
 	if clusterLoaderConfig.ClusterConfig.Provider != "aks" || clusterLoaderConfig.ClusterConfig.Provider != "gke" {
-		clusterLoaderConfig.ClusterConfig.IsSSHSupported = true
+		clusterLoaderConfig.ClusterConfig.IsSSHToMasterSupported = true
 	}
 	if clusterLoaderConfig.ClusterConfig.Provider != "aks" {
-		clusterLoaderConfig.ClusterConfig.IsPprofSupported = true
+		clusterLoaderConfig.ClusterConfig.IsAPIServerPprofExposed = true
 	}
 	return nil
 }
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -147,7 +147,7 @@ func completeConfig(m *framework.MultiClientSet) error {
 		clusterLoaderConfig.ClusterConfig.IsSSHToMasterSupported = true
 	}
 	if clusterLoaderConfig.ClusterConfig.Provider != "aks" {
-		clusterLoaderConfig.ClusterConfig.IsAPIServerPprofExposed = true
+		clusterLoaderConfig.ClusterConfig.IsAPIServerPprofEnabled = true
 	}
 	return nil
 }
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -143,8 +143,11 @@ func completeConfig(m *framework.MultiClientSet) error {
 			klog.Errorf("Getting master internal ip error: %v", err)
 		}
 	}
-	if clusterLoaderConfig.ClusterConfig.Provider == "aks" || clusterLoaderConfig.ClusterConfig.Provider == "gke" {
-		clusterLoaderConfig.ClusterConfig.IsSSHSupported = false
+	if clusterLoaderConfig.ClusterConfig.Provider != "aks" || clusterLoaderConfig.ClusterConfig.Provider != "gke" {
+		clusterLoaderConfig.ClusterConfig.IsSSHSupported = true
+	}
+	if clusterLoaderConfig.ClusterConfig.Provider != "aks" {
+		clusterLoaderConfig.ClusterConfig.IsPprofSupported = true
 	}
 	return nil
 }
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -143,6 +143,9 @@ func completeConfig(m *framework.MultiClientSet) error {
 			klog.Errorf("Getting master internal ip error: %v", err)
 		}
 	}
+	if clusterLoaderConfig.ClusterConfig.Provider == "aks" || clusterLoaderConfig.ClusterConfig.Provider == "gke" {
+		clusterLoaderConfig.ClusterConfig.IsSSHSupported = false
+	}
 	return nil
 }
 
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -43,9 +43,9 @@ type ClusterConfig struct {
 	// IsSSHToMasterSupported is false on some managed providers.
 	// Clusterloader will not attempt operations requiring SSH when this value is false, such as some metrics collection routines.
 	IsSSHToMasterSupported bool
-	// IsAPIServerPprofExposed is false for some managed providers.
-	// When IsAPIServerPprofExposed is false, clusterloader will avoid using using pprof to collect metrics.
-	IsAPIServerPprofExposed bool
+	// IsAPIServerPprofEnabled is false for some managed providers.
+	// When IsAPIServerPprofEnabled is false, clusterloader will avoid collecting kube-apiserver pprofs.
+	IsAPIServerPprofEnabled bool
 }
 
 // PrometheusConfig represents all flags used by prometheus.
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -40,8 +40,12 @@ type ClusterConfig struct {
 	MasterName                 string
 	KubemarkRootKubeConfigPath string
 	DeleteStaleNamespaces      bool
-	IsSSHSupported             bool
-	IsPprofSupported           bool
+	// IsSSHToMasterSupported is false on some managed providers.
+	// Clusterloader will not attempt operations requiring SSH when this value is false, such as some metrics collection routines.
+	IsSSHToMasterSupported bool
+	// IsAPIServerPprofExposed is false for some managed providers.
+	// When IsAPIServerPprofExposed is false, clusterloader will avoid using using pprof to collect metrics.
+	IsAPIServerPprofExposed bool
 }
 
 // PrometheusConfig represents all flags used by prometheus.
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -40,6 +40,7 @@ type ClusterConfig struct {
 	MasterName                 string
 	KubemarkRootKubeConfigPath string
 	DeleteStaleNamespaces      bool
+	IsSSHSupported             bool
 }
 
 // PrometheusConfig represents all flags used by prometheus.
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -41,6 +41,7 @@ type ClusterConfig struct {
 	KubemarkRootKubeConfigPath string
 	DeleteStaleNamespaces      bool
 	IsSSHSupported             bool
+	IsPprofSupported           bool
 }
 
 // PrometheusConfig represents all flags used by prometheus.
--- a/clusterloader2/pkg/measurement/common/etcd_metrics.go
+++ b/clusterloader2/pkg/measurement/common/etcd_metrics.go
@@ -60,6 +60,13 @@ type etcdMetricsMeasurement struct {
 // - start - Starts collecting etcd metrics.
 // - gather - Gathers and prints etcd metrics summary.
 func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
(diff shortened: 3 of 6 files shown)


clusterloader2/pkg/chaos/nodes.go
```diff
@@ -82,7 +82,7 @@ func (r *eventRecorder) record(a nodeAction, nodeName string) {
 
 // NewNodeKiller creates new NodeKiller.
 func NewNodeKiller(config api.NodeFailureConfig, client clientset.Interface, provider string) (*NodeKiller, error) {
-	if provider != "gce" && provider != "gke" {
+	if provider != "gce" && provider != "aks" {
```
marwanad: why remove gke?

alexeldeib: whoops

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -184,7 +184,8 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string, port in
 
 func (e *etcdMetricsMeasurement) getEtcdMetrics(host, provider string, port int) ([]*model.Sample, error) {
 	// Etcd is only exposed on localhost level. We are using ssh method
-	if provider == "gke" {
+	// TODO(ace): --managed flag instead of provider switch?
+	if provider == "gke" || provider == "aks" {
 		klog.Infof("%s: not grabbing etcd metrics through master SSH: unsupported for gke", e)
```
mm4tt: Please change to ...unsupported for %s", e, provider)

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -187,6 +187,11 @@ func (p *profileMeasurement) gatherProfile(c clientset.Interface) ([]measurement
 	for _, host := range p.config.hosts {
 		profilePrefix := fmt.Sprintf("%s_%s_%s", host, p.config.componentName, p.name)
 
+		// ssh to collect profile metrics not supported on aks masters
+		if p.config.provider == "aks" {
+			return nil, nil
```
mm4tt: It's possible to do it without SSH, see line 202. Would it be feasible for aks?

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -184,7 +184,8 @@ func (e *etcdMetricsMeasurement) stopAndSummarize(host, provider string, port in
 
 func (e *etcdMetricsMeasurement) getEtcdMetrics(host, provider string, port int) ([]*model.Sample, error) {
 	// Etcd is only exposed on localhost level. We are using ssh method
-	if provider == "gke" {
+	// TODO(ace): --managed flag instead of provider switch?
```
mm4tt: I'd rather create a new field in ClusterLoaderConfig: `isSshSupported` that would be initialized depending on the provider. Then we could have a generic condition for disabling things that require ssh

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -226,7 +226,7 @@ func (s *schedulerLatencyMeasurement) sendRequestToScheduler(c clientset.Interfa
 		responseText = string(body)
 	} else {
 		// If master is not registered fall back to old method of using SSH.
-		if provider == "gke" {
+		if provider == "gke" || provider == "aks" {
```
mm4tt: Same here

clusterloader2/pkg/measurement/common/etcd_metrics.go
mm4tt: Actually ClusterConfig would be a better place

alexeldeib: moved to cluster config and used the value throughout. I left the prometheus one as individually checking the values since it's not technically the same condition -- let me know if you think some more cleanup is appropriate there.

clusterloader2/pkg/measurement/common/profile.go
alexeldeib: testing this out, I think it should work...will report back

clusterloader2/cmd/clusterloader.go
```diff
@@ -143,6 +143,9 @@ func completeConfig(m *framework.MultiClientSet) error {
 			klog.Errorf("Getting master internal ip error: %v", err)
 		}
 	}
+	if clusterLoaderConfig.ClusterConfig.Provider == "aks" || clusterLoaderConfig.ClusterConfig.Provider == "gke" {
+		clusterLoaderConfig.ClusterConfig.IsSSHSupported = false
```
mm4tt: Do we set it to true anywhere? :)

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -75,6 +75,7 @@ func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig)
 	}
 
 	etcdInsecurePort := config.ClusterFramework.GetClusterConfig().EtcdInsecurePort
+	isSSHSupported := config.ClusterFramework.GetClusterConfig().IsSSHSupported
```
mm4tt: Hmm, I think that if the ssh is not supported then this measurement has nothing to do. In such cases we should just log and info and return from the Execute method. Does it make sense?

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -189,8 +189,8 @@ func (p *profileMeasurement) gatherProfile(c clientset.Interface) ([]measurement
 
 		// Get the profile data over SSH.
 		// Start by checking that the provider allows us to do so.
-		if p.config.provider == "gke" || shouldGetAPIServerByK8sClient(p.config.componentName) {
-			// SSH to master is not possible in gke, but if the component is
+		if !isSSHSupported || shouldGetAPIServerByK8sClient(p.config.componentName) {
+			// SSH to master is not possible in gke/ or ks, but if the component is
```
mm4tt: s/ks/eks/?

But, given that we have this isSSHSupported now, maybe let's remove the specific providers from the comments. Such comments will get quickly outdated once we disabled ssh for another provider. So maybe something like:
```
If we're here it means that ssh to master is not supported in the cluster. But if the component is kube-apiserver then we can get ...
```

WDYT?

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -226,7 +226,7 @@ func (s *schedulerLatencyMeasurement) sendRequestToScheduler(c clientset.Interfa
 		responseText = string(body)
 	} else {
 		// If master is not registered fall back to old method of using SSH.
-		if provider == "gke" {
+		if !isSSHSupported {
```
mm4tt: Similarly to the etcd measurement, this measurement will have nothing to do if `!(masterRegistered || isSshSupported)`. Can we move this check to the top of the execute method?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -96,7 +96,7 @@ func NewPrometheusController(clusterLoaderConfig *config.ClusterLoaderConfig) (p
 		delete(mapping, "MasterIps")
 	}
 	if _, exists := mapping["PROMETHEUS_SCRAPE_APISERVER_ONLY"]; !exists {
-		mapping["PROMETHEUS_SCRAPE_APISERVER_ONLY"] = clusterLoaderConfig.ClusterConfig.Provider == "gke"
+		mapping["PROMETHEUS_SCRAPE_APISERVER_ONLY"] = clusterLoaderConfig.ClusterConfig.Provider == "gke" || clusterLoaderConfig.ClusterConfig.Provider == "aks"
```
mm4tt: Yeah, this is good. It shouldn't rely on the isSSHSupported

alexeldeib:
I really appreciate the quick reviews 🙂 thank you!

I've gone ahead and hoisted moist of the checks earlier. It doesn't seem like the pprof approach worked on AKS, I need to figure out why though because I would expect it to. For now I've added a flag similar to ssh. Let me know what you think of this? I also corrected the initial conditions to take advantage of the bool zero value.

sorry about random json -_-

clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -60,6 +60,13 @@ type etcdMetricsMeasurement struct {
 // - start - Starts collecting etcd metrics.
 // - gather - Gathers and prints etcd metrics summary.
 func (e *etcdMetricsMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+	isSSHSupported := config.ClusterFramework.GetClusterConfig().IsSSHSupported
+	// Etcd is only exposed on localhost level. We are using ssh method
+	if isSSHSupported {
```
mm4tt: `if !isSSHSupported` ?

Also, do we need to introduce this local variable? I don't think it improves readability a lot. What about
```
if !config.ClusterFramework.GetClusterConfig().IsSSHSupported {
  ...
}
```

clusterloader2/pkg/config/cluster.go
```diff
@@ -40,6 +40,8 @@ type ClusterConfig struct {
 	MasterName                 string
 	KubemarkRootKubeConfigPath string
 	DeleteStaleNamespaces      bool
+	IsSSHSupported             bool
```
mm4tt: We should make it clear that it's the SSH to master that is not supported (e.g. in GKE you can still ssh into nodes). 
`IsSSHToMasterSupported`?

clusterloader2/pkg/config/cluster.go
```diff
@@ -40,6 +40,8 @@ type ClusterConfig struct {
...
+	IsPprofSupported           bool
```
mm4tt: Similarly to the ssh we should make it clear what it means. Could you either change the name to make it more self documenting or add a comment (or both)?

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -61,14 +61,35 @@ type schedulerLatencyMeasurement struct{}
 // - reset - Resets latency data on api scheduler side.
 // - gather - Gathers and prints current scheduler latency data.
 func (s *schedulerLatencyMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
-	action, err := util.GetString(config.Params, "action")
+	isSSHSupported := config.ClusterFramework.GetClusterConfig().IsSSHSupported
+
+	c := config.ClusterFramework.GetClientSets().GetClient()
+	nodes, err := c.CoreV1().Nodes().List(metav1.ListOptions{})
```
mm4tt: Hmm, this is actually a quite heavy call (especially in large clusters) we shouldn't be making it for every Execute call. I'm aware that it was already like this, but let's make it better. Would you mind moving this logic to clusterloader.go's `completeConfig` method and adding a field to the ClusterConfig? Ideally, it could be done in a separate PR. Thanks!

alexeldeib: I can tackle this in a separate PR, sure. I assume it's done here because we need to check if the master is registered. I don't love that i'm adding a bunch of new fields to this struct, but I could add something like `IsMasterNodeRegistered` to cover this and set it using this logic in `completeConfig`? WDYT?

clusterloader2/pkg/config/cluster.go
alexeldeib: how's that?

clusterloader2/pkg/config/cluster.go
```diff
@@ -40,8 +40,12 @@ type ClusterConfig struct {
...
 	KubemarkRootKubeConfigPath string
 	DeleteStaleNamespaces      bool
-	IsSSHSupported             bool
-	IsPprofSupported           bool
+	// IsSSHToMasterSupported is false on some managed providers.
+	// Clusterloader will not attempt operations requiring SSH when this value is false, such as some metrics collection routines.
+	IsSSHToMasterSupported bool
+	// IsAPIServerPprofExposed is false for some managed providers.
+	// When IsAPIServerPprofExposed is false, clusterloader will avoid using using pprof to collect metrics.
+	IsAPIServerPprofExposed bool
```
mm4tt: Looking at the code, I thin the question is more whether the pprof is enabled in apiserver (we're exposing in on our own in profile.go). Can we change the name to IsAPIServerPprofEnabled and change the doc to say something like
`When IsAPIServerPprofEnabled is false, clusterloader2 will avoid collecting kube-apiserver pprofs`

clusterloader2/pkg/measurement/common/scheduler_latency.go
mm4tt: Yeah, that sounds great, thanks!

clusterloader2/pkg/config/cluster.go
alexeldeib: fixed 👍

alexeldeib:
Would you like me to squash commits before merge? I kept everything as-is until it looked good.

clusterloader2/pkg/measurement/common/scheduler_latency.go
wojtek-t: Yes - let's do in a separate PR please.

PR #1128, opened 2020-03-20
prefixing ENABLE_NETWORKPOLICIES with CL2_ for consistency 

maniSbindra:
This PR is to prefixing ENABLE_NETWORKPOLICIES variable with CL2_ so that variable can be overriden. This change was applied to other [override variables](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides) as a part of #1126. This will make the ENABLE_NETWORKPOLICIES variable consistent will all other [override variables](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides)


mm4tt:
/lgtm
/approve
/ok-to-test

Thanks!

PR #1131, opened 2020-03-23
Add ipv6 support for network performance test

thomasgaozx:
Worker nodes need to add square brackets ([...]) to IPv6 addresses
when connecting to the orchestrator otherwise the call to `rpc.DialHTTP`
fails.

This change has been tested on an ipv6 cluster with two worker nodes.

Co-authored-by: Allain Legacy <allain.legacy@windriver.com>
Signed-off-by: Thomas Gao <thomas.gao@windriver.com>

### DIFF
--- a/network/benchmarks/netperf/nptest/nptest.go
+++ b/network/benchmarks/netperf/nptest/nptest.go
@@ -542,17 +542,7 @@ func handleClientWorkItem(client *rpc.Client, workItem *WorkItem) {
 // isIPv6: Determines if an address is an IPv6 address
 func isIPv6(address string) bool {
 	x := net.ParseIP(address)
-	if x != nil {
-		// The net package does not have a good way to determine if an address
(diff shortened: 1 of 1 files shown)


network/benchmarks/netperf/nptest/nptest.go
```diff
@@ -539,17 +539,38 @@ func handleClientWorkItem(client *rpc.Client, workItem *WorkItem) {
 	time.Sleep(10 * time.Second)
 }
 
+// isIPv6: Determines if an address is an IPv6 address
+func isIPv6(address string) bool {
+	x := net.ParseIP(address)
+	if x != nil {
```
bowei: err cases first:

```
if x == nil {
  return false
}
```

why not 

```
return x != nil && !x.To4() == nil && x.To16() != nil
```

thomasgaozx:
/assign @mm4tt 

Sorry Matt, could you take a look at this as well?

PR #1132, opened 2020-03-23
Expand the scraping timeout

YangLu1031:
Found we couldn't get the cpu usage per process data during windows [load test](https://testgrid.k8s.io/google-windows#gce-windows-node-throughput).
So figured out that the scraping is pretty slow for `process` collector in wmi_exporter.
Tried locally, the prometheus server took around 1m30s to scrape windows target data when all 80 pods are running, so expand the scrape_timeout and interval.

Also created an [issue](https://github.com/martinlindhe/wmi_exporter/issues/487) for tracking

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/prometheus-windows-scrape-configs.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/prometheus-windows-scrape-configs.yaml
@@ -8,5 +8,7 @@ metadata:
 stringData:
   windows-scrape-configs.yaml: |-
     - job_name: "monitoring/windows_static"
+      scrape_interval: {{DefaultParam .CL2_WINDOWS_SCRAPE_INTERVAL "130s"}}
+      scrape_timeout: {{DefaultParam .CL2_WINDOWS_SCRAPE_TIMEOUT "120s"}}
       static_configs:
       - targets: ["{{$WINDOWS_NODE_NAME}}:5000"]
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/manifests/prometheus-windows-scrape-configs.yaml
```diff
@@ -8,5 +8,7 @@ metadata:
 stringData:
   windows-scrape-configs.yaml: |-
     - job_name: "monitoring/windows_static"
+      scrape_interval: 130s
```
mm4tt: I'm fine with this, but having a 130s scrape interval may make these metrics useless. Usually we scrape every 5-15s, if you scrape every ~2min and the slowness period you want to observer takes ~30s there is a big chance that you'll miss it. 
That given said, I understand that this is better than nothing, i.e. better to have working metrics with 130s scrape interval than not having the metrics at all.

/lgtm
/approve
/hold
On the other hand, it'd be better if it was a configurable so you could have a short interval for low-pod-density scenarios and increase it for high-pod-density. I think the easiest way to achieve that would be to introduce a param here, with CL2_[1] prefix, e.g. 
```
scrape_interval: {{DefaultParam .CL2_PROMETHEUS_WINDOWS_SCRAPE_INTERVAL 10s}}
```

[1] If a parameter starts with CL2_ you can override it by overriding the corresponding env variable. For manual runs you'd just do `export CL2_MY_PARAM=...`, for prow jobs you can use the ``--env`` flag, see example - https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L65

YangLu1031: Agree, and thanks! Made changes using CL2_ prefix. 
Also changed to use `irate()`, as I found `irate()` seems more stable than `rate()`

Issue #1130, opened 2020-03-23
[ClusterLoader2][NodeKiller] NodeKiller can leave broken nodes


jprzychodzen:
When a testing run uses NodeKiller, some nodes can be left in a broken state - with disabled services. This can influence results while running multiple testing jobs on the same cluster.

It happens when testing finishes before NodeKiller reboot is triggered.

Sample log:
```
...
I0323 08:03:00.196712   16174 resource_gather_worker.go:89] Closing worker for e2e-test-jprzychodzen-minion-group-2zfc
I0323 08:03:00.197101   16174 container_resource_gatherer.go:182] Waitgroup finished.
I0323 08:03:00.237245   16174 system_pod_metrics.go:124] collecting system pod metrics...
I0323 08:03:00.394737   16174 system_pod_metrics.go:228] Loaded restart count threshold overrides: map[fluentd-gcp:1 kube-proxy:1 metadata-p
roxy:1 prometheus-to-sd-exporter:1]
I0323 08:03:00.396502   16174 simple_test_executor.go:150] Step "Collecting measurements" ended
I0323 08:03:00.411423   16174 simple_test_executor.go:114] Summary of Chaos Monkey execution
NodeKiller: Recorded following events
NodeKiller: At Mon Mar 23 07:56:55 UTC 2020 stopService happend for node e2e-test-jprzychodzen-minion-group-49jb
I0323 08:03:00.411453   16174 probes.go:133] Stopping DnsLookupLatency probe...
I0323 08:03:20.715459   16174 probes.go:133] Stopping InClusterNetworkLatency probe...
I0323 08:03:46.222799   16174 simple_test_executor.go:370] Resources cleanup time: 45.811343198s
I0323 08:03:46.222842   16174 clusterloader.go:189] --------------------------------------------------------------------------------
I0323 08:03:46.222848   16174 clusterloader.go:190] Test Finished
I0323 08:03:46.222853   16174 clusterloader.go:191]   Test: testing/load/config.yaml
I0323 08:03:46.222859   16174 clusterloader.go:192]   Status: Success
I0323 08:03:46.222864   16174 clusterloader.go:196] --------------------------------------------------------------------------------
I0323 08:03:46.223239   16174 prometheus.go:181] Tearing down prometheus stack
```
In this sample NodeKiller did not schedule reboot an ClusterLoader2 has ended working, leaving e2e-test-jprzychodzen-minion-group-49jb without reboot.

[...]

PR #1135, opened 2020-03-25
[NodeKiller] Rebooting broken nodes before exiting the test

mcastellin:
With this change I tried to introduce some sort of wait mechanism for the NodeKiller subroutine.
The rationale behind is it avoiding that cluster nodes are left in a broken state by NodeKiller if the test finishes while downtime is being simulated for some nodes.

Now the chaos monkey can return back to its caller a WaitGroup that can be used to wait for the subroutine termination before proceeding to test summary.

/assign @jprzychodzen 
Ref: #1130

### DIFF
--- a/clusterloader2/pkg/test/simple_test_executor.go
+++ b/clusterloader2/pkg/test/simple_test_executor.go
@@ -55,11 +55,46 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
 	klog.Infof("AutomanagedNamespacePrefix: %s", ctx.GetClusterFramework().GetAutomanagedNamespacePrefix())
 	defer cleanupResources(ctx)
 	ctx.GetTuningSetFactory().Init(conf.TuningSets)
+
 	stopCh := make(chan struct{})
-	defer close(stopCh)
-	if err := ctx.GetChaosMonkey().Init(conf.ChaosMonkey, stopCh); err != nil {
+	chaosMonkeyWaitGroup, err := ctx.GetChaosMonkey().Init(conf.ChaosMonkey, stopCh)
+	if err != nil {
+		close(stopCh)
 		return errors.NewErrorList(fmt.Errorf("error while creating chaos monkey: %v", err))
 	}
+	errList := ste.ExecuteTestSteps(ctx, conf)
+	close(stopCh)
+
+	if chaosMonkeyWaitGroup != nil {
+		// Wait for the Chaos Monkey subroutine to end
+		klog.Info("Waiting for the chaos monkey subroutine to end...")
+		chaosMonkeyWaitGroup.Wait()
+		klog.Info("Chaos monkey ended.")
+	}
+
+	for _, summary := range ctx.GetMeasurementManager().GetSummaries() {
+		if ctx.GetClusterLoaderConfig().ReportDir == "" {
+			klog.Infof("%v: %v", summary.SummaryName(), summary.SummaryContent())
+		} else {
+			testDistinctor := ""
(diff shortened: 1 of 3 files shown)


mcastellin:
As discussed with @jprzychodzen in #1130 it would also be nice to have a way to mock the NodeKiller so we can implement unit testing for this part of the code. I have submitted the code change for now to get some feedback on the idea. I will try implement some unit testing later to complement this one.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -55,17 +56,50 @@ func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *error
...
+				continue
+			}
+		}
+	}
+	klog.Infof(ctx.GetChaosMonkey().Summary())
+	return errList
+}
+
+// ExecuteTestSteps executes all test steps provided in configuration
+func (ste *simpleTestExecutor) ExecuteTestSteps(ctx Context, conf *api.Config) (*sync.WaitGroup, *errors.ErrorList) {
```
wojtek-t: I like the overall approach, but that it is introducing unnecessary complication.

Let's:
- create stopCh and ChaosMonkey outside of this function (in the ExecuteTest one)
- this one then doesn't need to know (and return) about WaitGroup

mcastellin: No problem, I refactored it to only return the errorList

mcastellin:
@wojtek-t Thanks for reviewing my work. 
Can you help me out understanding what is wrong with the build? I really can't point my finger on it..

jprzychodzen:
> Can you help me out understanding what is wrong with the build? I really can't point my finger on it..

@mcastellin, what do you mean by 'what is wrong'? It seems that commit passed tests, so it builds. Are you referring to @wojtek-t comments?

mcastellin:
No @jprzychodzen , I was referring to the errors in the first build. The second one passed though.. so I believe it was just a flake.

wojtek-t:
Looks reasonable to me - I will let @jprzychodzen to make a deeper review though.

/approve

mcastellin:
@jprzychodzen 
In addition to this change, I have refactored the NodeKiller to provide two providers:
- A "CloudKill" provider when running in GCE
- a "DummyKill" provider that for now just simulates the chaos monkey with a sleep command

I believe this is what you asked in #1130, right?

If you're interested I can make this part of this PR or follow up with another PR once we decide what to do with this one. What do you think?

wojtek-t:
> If you're interested I can make this part of this PR or follow up with another PR once we decide what to do with this one. What do you think?

Separate PR please. Whenever it's possible, split things into multiple PRs - it makes reviewing easier and faster. Review time is exponential to the size of the PR usually...

mcastellin:
> Separate PR please. Whenever it's possible, split things into multiple PRs - it makes reviewing easier and faster. Review time is exponential to the size of the PR usually...

Agreed, I also though to hold this off. I'll wait until this one is resolved then. Thanks for your feedback @wojtek-t

jprzychodzen:
@mcastellin, I've checked the code, thanks for this contribution!

I need to confirm that this work with few edge cases - there are no unit tests on Node Killer/Chaos Monkey level and right now Chaos Monkey is not enabled in presubmit. I will finish that tomorrow. Some (probably) unrelated problem happened (resource cleanup) and I need to double check that.

Please don't add those provides right now. It would be even better if you could add each provider by a separate PR - smaller PRs are easier to review.

Also, could you squash those three commits into single commit? Thanks!

jprzychodzen:
It seems that the issue I've seen is related to some other problem. I will prepare a replication script and create a separate issue for that.

@mcastellin thanks for contribution.

Issue #948, opened 2019-12-19
CL2 stopped gathering etcd profiles


mm4tt:
![8Y2WptGNdE4](https://user-images.githubusercontent.com/2604887/71157982-4e4c5500-2243-11ea-8842-3023322eb04b.png)

This is most likely due to https://github.com/kubernetes/kubernetes/pull/77561 and the port change.

mm4tt:
I tried changing etcd port to 2382 in profile.go in https://github.com/kubernetes/perf-tests/pull/949, but unfortunately it didn't help

This suggests that now we need to use https to get the etcd profiles, which makes the things a tad more complicated

mm4tt:
/good-first-issue

A bit complicated for the first issue, but I think it should be doable. If there is anyone who would like to take this one I can help with this one. I have some ideas that we could try. It'd be great to have the etcd profiles back

skurtzemann:
Hi @mm4tt,
I'll like to try this issue as my first issue ;) How can I start ?

mm4tt:
Hey, @skurtzemann  
Thanks for reaching out. 

Could you try preparing a PR where you'd change the curl command we use to gather etcd profiles, https://github.com/kubernetes/perf-tests/blob/9b73b5f9cc8ee2680474bc08a624833242bd9da8/clusterloader2/pkg/measurement/common/profile.go#L184, to pass there --cert and --key flags, similarly to https://github.com/kubernetes/perf-tests/blob/39b98208736db08b243b23ff53f1215c022cfb34/clusterloader2/pkg/measurement/common/etcd_metrics.go#L208

Please start with hardcoding the key path to `/etc/srv/kubernetes/pki/etcd-apiserver-server.key` and the cert path to `/etc/srv/kubernetes/pki/etcd-apiserver-server.crt`. I want to see whether it would work. If so, we'll figure out how to do it in a cleaner way.

Thanks!

PR #1136, opened 2020-03-25
CL2: fix etcd profiles gathering

skurtzemann:
This PR is to resolve issue #948 as seen with @mm4tt

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -61,6 +61,8 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdCertificatePath, "etcd-certificate", "ETCD_CERTIFICATE", "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt", "Path to the etcd certificate on the master machine")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdKeyPath, "etcd-key", "ETCD_KEY", "/etc/srv/kubernetes/pki/etcd-apiserver-server.key", "Path to the etcd key on the master machine")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "Whether to delete all stale namespaces before the test execution.")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
--- a/clusterloader2/pkg/measurement/common/profile.go
+++ b/clusterloader2/pkg/measurement/common/profile.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 	"time"
 
+	goerrors "github.com/go-errors/errors"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -122,7 +123,7 @@ func (p *profileMeasurement) start(config *measurement.MeasurementConfig, IsSSHT
 			case <-p.stopCh:
 				return
 			case <-time.After(profileFrequency):
-				profileSummaries, err := p.gatherProfile(k8sClient, IsSSHToMasterSupported)
+				profileSummaries, err := p.gatherProfile(k8sClient, IsSSHToMasterSupported, config)
 				if err != nil {
 					klog.Errorf("failed to gather profile for %#v: %v", *p.config, err)
 					continue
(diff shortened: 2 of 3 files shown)


mm4tt:
Thanks, @skurtzemann!

Let's see whether it works
/ok-to-test

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -181,7 +181,15 @@ func (p *profileMeasurement) gatherProfile(c clientset.Interface) ([]measurement
 		return nil, fmt.Errorf("profile gathering failed finding component port: %v", err)
 	}
 	profileProtocol := getProtocolForComponent(p.config.componentName)
-	getCommand := fmt.Sprintf("curl -s -k %slocalhost:%v/debug/pprof/%s", profileProtocol, profilePort, p.config.kind)
+
+	var getCommand string
```
mm4tt: Can we extract this logic (constructing getCommand) to a separate method?

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -181,7 +181,15 @@ func (p *profileMeasurement) gatherProfile(c clientset.Interface) ([]measurement
...
+	if p.config.componentName == "etcd" {
+		etcdCert := "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt"
```
mm4tt: Now we confirmed that it works we should create a flags for these two variables and default them to these values. The logic then should be: if flags are set, pass them and use https, if not don't pass and use http

mm4tt:
Works like a charm! Screenshot from the presubmit artifacts:
gJDD9EB3cGM

Ok, now let's clean this up a bit

clusterloader2/pkg/measurement/common/profile.go
skurtzemann: Ok, I'll do that

clusterloader2/pkg/measurement/common/profile.go
skurtzemann: Use flags or env variables (`ETCD_CERTIFICATE` and `ETCD_KEY`) like https://github.com/kubernetes/perf-tests/blob/39b98208736db08b243b23ff53f1215c022cfb34/clusterloader2/pkg/measurement/common/etcd_metrics.go#L203 ?

mm4tt: Let's use flags. We can use the "env-var" flag for convenience (can be set both via flag and env variable). The advantage is the basically a flag is clearly defined and documented (e.g. I can run --help and check what it does)

skurtzemann: > Now we confirmed that it works we should create a flags for these two variables and default them to these values. The logic then should be: if flags are set, pass them and use https, if not don't pass and use http

I create 2 flags but how can I get theses values into `clusterloader2/pkg/measurement/common/profile.go ` ?
https://github.com/kubernetes/perf-tests/pull/1136/commits/8e492ada85e2038ccebdb88b160ea306da0268da

clusterloader2/pkg/measurement/common/profile.go
skurtzemann: Done in https://github.com/kubernetes/perf-tests/pull/1136/commits/b6e71483ad60e548c065d1ed905ba87e101e5809

clusterloader2/cmd/clusterloader.go
```diff
@@ -61,6 +61,8 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.Nodes, "nodes", "NUM_NODES", 0, "number of nodes")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.Provider, "provider", "PROVIDER", "", "Cluster provider")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdCertificatePath, "etcd-certificate", "ETCD_CERTIFICATE", "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt", "Path to ectd certificate")
```
mm4tt: Let's change the description to be clear about where the key/certificate is, e.g. `Path to the etcd certificate on the master machine". Here and below.

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -221,6 +219,26 @@ func (p *profileMeasurement) shouldExposeApiServerDebugEndpoint() bool {
...
+	profilePort, err := getPortForComponent(p.config.componentName)
+	if err != nil {
+		return "", fmt.Errorf("get profile command failed finding component port: %v", err)
+	}
+	profileProtocol := getProtocolForComponent(p.config.componentName)
+
+	//TODO: use flags for etcdCert and etcdKey
+	var command string
+	if p.config.componentName == "etcd" {
+		etcdCert := "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt"
```
wojtek-t: Let's do this - you already set defaults to those values...

clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -221,6 +219,26 @@ func (p *profileMeasurement) shouldExposeApiServerDebugEndpoint() bool {
 	return p.config.componentName == "kube-apiserver"
 }
 
+func (p *profileMeasurement) getProfileCommand() (string, error) {
+	profilePort, err := getPortForComponent(p.config.componentName)
+	if err != nil {
+		return "", fmt.Errorf("get profile command failed finding component port: %v", err)
```
mm4tt: NIT: Could you use go-errors for all added / updated lines? Here is an example https://github.com/kubernetes/perf-tests/blob/75dfad9672dfabe11b292c246f8f99c79db84655/clusterloader2/pkg/config/template_provider.go#L240

clusterloader2/pkg/measurement/common/profile.go
mm4tt: The `Execute` method has the MeasurementConfig argument. You can access the ClusterConfig from there (via ClusterLoaderConfig, IIRC). Then you can either pipe the values you need here, or add them to the profileMeasurement struct.

wojtek-t: @skurtzemann - `start()` method already has access to the config, and that is calling `gatherProfile` - let's just pipe them here

skurtzemann: Ok, thanks for your help. I submit the updated PR with the config

skurtzemann:
If tests succeded I will clean all my commits (rebase from master make my very dirty commits :( )

mm4tt:
Looks good! 
One more thing before I approve, could you please squash all commits into one? 
Thanks!

mm4tt:
If I'm not available, you can also ping @wojtek-t for the approval

skurtzemann:
> Looks good!
> One more thing before I approve, could you please squash all commits into one?
> Thanks!

Commits squashed, tests in progress :)

mm4tt:
Nice!

/lgtm
/approve

PR #1138, opened 2020-03-27
Decouple SSH command execution in util.SSH to enable unit testing

### LINKED ISSUE #1137: [ClusterLoader2][NodeKiller] Refactor NodeKiller by adding 'provider' abstraction and move env specific calls over there
Two provides should be implemented to consider this issue resolved
 *  one for GCP environment (currenly inlined in [clusterloader2/pkg/chaos](../tree/master/clusterloader2/pkg/chaos) code)
 *  one for fake (testing) environment

Resolving this will unlock adding Kubemark provider and unit testing Chaos Monkey/Node Killer.

/good-first-issue

mcastellin:
Following up with #1137 we are starting to decouple code that depends on real cloud components to allow unit testing, starting with util.SSH. This utility executes commands against a GCloud cluster so it should be abstracted with an interface and replaceable in a go test.
- create interface for command execution in cluster nodes through SSH
- refactor code that is currently using util.SSH to use the new executor interface

### DIFF
--- a/clusterloader2/pkg/chaos/nodes.go
+++ b/clusterloader2/pkg/chaos/nodes.go
@@ -49,6 +49,7 @@ type NodeKiller struct {
 	// killedNodes stores names of the nodes that have been killed by NodeKiller.
 	killedNodes sets.String
 	recorder    *eventRecorder
+	ssh         util.SSHExecutor
 }
 
 type nodeAction string
@@ -85,7 +86,8 @@ func NewNodeKiller(config api.NodeFailureConfig, client clientset.Interface, pro
 	if provider != "gce" && provider != "gke" {
 		return nil, fmt.Errorf("provider %q is not supported by NodeKiller", provider)
 	}
-	return &NodeKiller{config, client, provider, sets.NewString(), newEventRecorder()}, nil
+	sshExecutor := &util.GCloudSSHExecutor{}
+	return &NodeKiller{config, client, provider, sets.NewString(), newEventRecorder(), sshExecutor}, nil
 }
 
 // Run starts NodeKiller until stopCh is closed.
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/chaos/nodes.go
```diff
@@ -94,111 +51,20 @@ func (k *NodeKiller) Run(stopCh <-chan struct{}, wg *sync.WaitGroup) {
...
-	wg.Add(len(nodes))
-	for _, node := range nodes {
-		k.killedNodes.Insert(node.Name)
-		node := node
-		go func() {
-			defer wg.Done()
-
-			klog.Infof("%s: Stopping docker and kubelet on %q to simulate failure", k, node.Name)
-			k.addStopServicesEvent(node.Name)
-			err := util.SSH("sudo systemctl stop docker kubelet", &node, nil)
```
jprzychodzen: Here we should only extract `err := util.SSH("sudo systemctl stop docker kubelet", &node, nil)` and `err = util.SSH("nohup sudo shutdown -r +1s > /dev/null 2> /dev/null < /dev/null &", &node, nil)` as provider specific code.

Shared code can be unit tested, provider specific code needs integration testing and it's harder to do locally.

mcastellin:
@jprzychodzen if the only problem you want to address is the execution of SSH commands I would suggest to abstract the SSH command execution at its core by providing an interface. I check and it's only used by the nodekiller and the PrometheusController (that I guess it's having the same issue and is not unit tested). I will update the code to do just that, let me know what you think

jprzychodzen:
@mcastellin Nice insight with SSH and Prometheus. This change should be merged as is to decouple SSH execution and allow testing on that level.

Still it does not allow testing of Node Killer - in unit test environment there is no K8s cluster that we can use, so all calls using client should also be abstracted. Please check my previous comments for guideline on that.

Could you change the title of this change and description (first comment) to match what it really does? It will not fix #1137, however it still should be merged.

mcastellin:
@jprzychodzen thanks for your feedback. Name and rescription is now changed.

/assign @jprzychodzen

mm4tt:
/approve

Hint for the future, it might be better to move changes not related to the problem you're solving (the .gitignore file) to a separate PR. It will make sure you're not blocked on approvals.

PR #1140, opened 2020-03-30
Revert "[NodeKiller] Schedule reboot and disconnect from node"

jprzychodzen:
This reverts commit f1647f4e3c1f05c29c5d04596a876371893420b7.

Required to make Node Killer work again - with shutdown it silently fails.

/hold


jprzychodzen:
@mcastellin Please be aware about this change.

mm4tt:
FTR, we know that something is wrong with the "shutdown" command, here is an example where 6 nodes were killed, but none of them properly rebooted - https://gke-scalability-prow.corp.goog/log?job=ci-kubernetes-e2e-scalability-experimental-load-containerd-jprzychodzen&id=1244575398428151814
We don't know what exactly is wrong, though. We manually verified that reverting this change helps, nodes started getting rebooted properly.

/lgtm
/approve

I'm re-opening the https://github.com/kubernetes/perf-tests/pull/1109 in that case.

PR #1145, opened 2020-03-30
Unpack the string values from the CL2_.* envs

mm4tt:
Without doing it, it's impossible to easily override boolean param via env. E.g. if I do
```
--env=ENABLE_MY_FEATURE=false
```

The 'false' will be treated as string, and then in the config file the if statement like this:
```
{{if .ENABLE_MY_FEATURE}}
...
{{end}}
```

Will evaluate to true and will be executed, because "false" is non-empty string that evaluates to true :)

### DIFF
--- a/clusterloader2/pkg/config/template_provider.go
+++ b/clusterloader2/pkg/config/template_provider.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"text/template"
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/config/template_provider.go
```diff
@@ -240,11 +241,24 @@ func LoadCL2Envs() (map[string]interface{}, error) {
 			return nil, goerrors.Errorf("unparsable string in os.Eviron(): %v", keyValue)
 		}
 		key, value := split[0], split[1]
-		mapping[key] = value
+		mapping[key] = unpackStringValue(value)
```
wojtek-t: Are you sure it won't break any existing test?

mm4tt: I'm 99% sure as of now.
Currently, we only use this feature in two places: scheduler_throughput and access_tokens. Both are enabled in the presubmit so we'll know for 100% in ~30min

PR #1146, opened 2020-03-30
Extend load test to immutable secrets/configmaps

wojtek-t:
Each deployment in load test has now both secret and configmap mounted. 90% of them are marked as immutable, only 10% is mutable (as it was before).

Scalability test coverage is a requirement for graduating "Immutable Secrets/ConfigMaps" feature to Beta.

This has been scale-tested already and the features was enabled by default in https://github.com/kubernetes/kubernetes/pull/89594

ref https://github.com/kubernetes/enhancements/issues/1412

### DIFF
--- a/clusterloader2/testing/load/configmap.yaml
+++ b/clusterloader2/testing/load/configmap.yaml
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{.Name}}
+{{if not (eq (Mod .Index 20) 0 19) }} # .Index % 20 in {0,19} - only 10% deployments will have non-immutable ConfigMap.
+immutable: true
+{{end}}
 data:
   data.yaml: |-
     a: 1
--- a/clusterloader2/testing/load/deployment.yaml
+++ b/clusterloader2/testing/load/deployment.yaml
@@ -28,14 +28,10 @@ spec:
             cpu: {{$CpuRequest}}
             memory: {{$MemoryRequest}}
         volumeMounts:
-          {{if (eq (Mod .Index 20) 0 19) }} # .Index % 20 in {0,19} - 10% deployments will have ConfigMap
           - name: configmap
             mountPath: /var/configmap
-          {{end}}
-          {{if (eq (Mod .Index 20) 10 19) }} # .Index % 20 in {10,19} - 10% deployments will have Secret
           - name: secret
             mountPath: /var/secret
-          {{end}}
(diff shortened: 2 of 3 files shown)


clusterloader2/testing/load/configmap.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{.Name}}
+{{if not (eq (Mod .Index 20) 0 19) }} # .Index % 20 in {0,19} - only 10% deployments will have non-immutable ConfigMap.
```
mm4tt, line 5: The if condition seems to be: `.Index % 20 not in {0,19}` which translates to 90% of deployments will have immutable Config map. 

It's basically what you wrote, but with less negations - easier to parse. 

NIT: Consider rewriting

clusterloader2/testing/load/deployment.yaml
```diff
@@ -28,14 +28,10 @@ spec:
             cpu: {{$CpuRequest}}
             memory: {{$MemoryRequest}}
         volumeMounts:
-          {{if (eq (Mod .Index 20) 0 19) }} # .Index % 20 in {0,19} - 10% deployments will have ConfigMap
```
mm4tt, line 31: So if I understand the change correctly, now every deployment will have both configmap and secret, 90% will be immutable, 10% mutable? If so, can we make it clear in the PR description / commit message?

wojtek-t: Yes - that's exactly the intention. Updating.

clusterloader2/testing/load/configmap.yaml
wojtek-t: Let's merge as is for now and will clean up later - I'm slightly afraid of breaking something now and I would like that to land asap to have as much soaking time as possible.

clusterloader2/testing/load/deployment.yaml
wojtek-t: done

wojtek-t:
@mm4tt - PTAL

clusterloader2/testing/load/configmap.yaml
mm4tt: I proposed to change only the comment, but I'm fine with doing that later. Thanks!

mm4tt:
FYI, @jprzychodzen current oncall

PR #1148, opened 2020-04-01
[Density][ChaosMonkey] Make config more predictable

jprzychodzen:
Density test flakes a lot with NodeKiller enabled. It's related to
very unstable and unpredictable when nodes will be removed from cluster

Ref. kubernetes/kubernetes#89051

/sig scalability
/assign Matt Matejczyk

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -44,9 +44,9 @@ tuningSets:
 chaosMonkey:
   nodeFailure:
     failureRate: 0.01
-    interval: 1m
-    jitterFactor: 10.0
-    simulatedDowntime: 10m
+    interval: 5m
+    jitterFactor: 1.0
+    simulatedDowntime: 5m
 {{end}}
 steps:
 - name: Starting measurements


mm4tt:
I think that no matter what we do the node-killer will have problems with density test. I'm worried about the "measuring the pod startup latency" step, if a node is killer during that time, it's very likely that the pod-startup latency will fail (we'll have ~1% of the pods with Xmin pod startup latency). 
Given that, I think we should simply disable the node-killer in the density test. 

On the other hand, we will run into this problems again when we'll be merging the density test with the load test. We'll need a way to stop node-killer for the pod-startup latency measurement step.  Such thing is currently not supported by the node-killer API.

But, given that the we keep discovering new issues with the node killer, I started wondering whether we shouldn't just disable it for now. It'd allow us to focus on more important efforts like containerd and merging the density with load test. Once we do that, we can get back to the node killer and think more holistically about it. Instead of patching hole over hole in it, maybe we should come up with a new, improved design that will solve all this problems at once.
@wojtek-t, WDYT?

wojtek-t:
> But, given that the we keep discovering new issues with the node killer, I started wondering whether we shouldn't just disable it for now. It'd allow us to focus on more important efforts like containerd and merging the density with load test. Once we do that, we can get back to the node killer and think more holistically about it. Instead of patching hole over hole in it, maybe we should come up with a new, improved design that will solve all this problems at once.
@wojtek-t, WDYT?

+1
I think that mergeing density and load is much more important. Once we also migrate to containerd, I really hope we will be able to start enforcing startup time across the whole test, not just for latency phase (which is super hacky). And once both happen, the impact of few pods affected by node killer will be negligible.
So I fully agree that we should address the issues in that order.

jprzychodzen:
Closing this for now. Will return to this problem after a merge of load and containerd.

wojtek-t:
Let's close as discussed above.

Issue #1036, opened 2020-02-07
Cleanup the "extended load test" overrides


mm4tt:
The extended load tests has been implemented and enabled everywhere as a part of https://github.com/kubernetes/perf-tests/issues/704.
We've been running these tests in many places (also at 5k node scales) for a few months. I believe we have enough confidence to make all the changes the defaults and cleanup all the gating logic we have.

We should get rid of all these overrides: https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides
And clean-up all places where they are used (e.g. 10+ occurrences in k8s.io/test-infra)

wojtek-t:
To clarify - the things to do are:
- update default for things that are enabled everywhere (PVs are still disabled in couple tests)
- then remove the overwrides
- cleanup ability to override those things

maniSbindra:
Hi @wojtek-t  I would like to work on this issue. 

I found 126 occurrences of "testing/load/experimental/override" in 7 files within  k8s.io/test-infra. All these occurrences use ```--test-cmd-args=--testoverrides=./testing/load/experimental/overrides/``` with one of the files listed in https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides . 

I wanted a bit more clarification regarding changes and tests I can use to verify the changes. So as apart of PR to resolve this issue should I 
*  Remove the ```--test-cmd-args=--testoverrides=./testing/load/experimental/overrides/``` switches from k8s.io/test-infra ?
* Remove the [override files](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides) ?
* Modify the ./clusterloader2/pkg/config/template_provider.go to cleanup ability to  override
?
* Would it be possible to share more details related to "update default for things that are enabled everywhere"


Thanks,
Mani

mm4tt:
Hey @maniSbindra, thanks for taking a look at this!

To give you an example of what "update defaults" means, let's take a look at the 
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/experimental/overrides/enable_configmaps.yaml
It overrides the `ENABLE_CONFIGMAPS` variable to `true`.
[...]

PR #1151, opened 2020-04-01
Delete Redundant "extended load test" override files

maniSbindra:
This PR deletes the redundant [override files](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides) , and fixes the issue #1036 .

After PR https://github.com/kubernetes/test-infra/pull/17034 is merged, the [override files](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load/experimental/overrides) are redundant and can be deleted.


wojtek-t:
@maniSbindra - thanks for this PR; it seems fine from OSS perspective, but we also need to update our internal tests in Google to stop referencing those files. once they are fixed, i will approve this PR.

/assign

maniSbindra:
Thanks @wojtek-t

Issue #604, opened 2019-07-04
Migrate perf-dash off debian


oxddr:
Perf-dash uses debian at the moment. We could try to migrate it to scratch or busybox. 

/area perf-dash

wojtek-t:
We should migrate to distroless.

/remove-lifecycle stale

skurtzemann:
Hi @oxddr, 
I'll like to get this issue and I had some questions ;)

- can we perharps include the `go build` process into the `Dockerfile` as a build stage instead of copying the binary into the `Dockerfile`
- what version of golang do you use to build `perf-dash`
- how can i plugs data into `perf-dash` to test it ?

Thanks

oxddr:
Awesome!

> * can we perharps include the `go build` process into the `Dockerfile` as a build stage instead of copying the binary into the `Dockerfile` ?

Sure, that's a good idea to make builds more reproducible. 

> * what version of golang do you use to build `perf-dash` ?

I've used go1.14 last time. I'd use whichever is the newest and works with the codebase. We don't have specific requirements around this as long as perf-dash works.

> * how can i plugs data into `perf-dash` to test it ?

Please have a look at `Makefile` and `run` target. It loads actual test data for few recent builds (you may want to decrease this number even more to speed things up). 

Please let me know if you have any other questions.

skurtzemann:
Thanks for your response.

Inside the `golang:1.14.1` I've got the following error after launch `make run` :

```
E0331 15:44:41.418705   12067 google-gcs-downloader.go:138] Error while looking for artifacts/in_cluster_network_latency_load* in ci-kubernetes-kubemark-500-gce-stable2 build 1244640772502851585: <nil>
```

oxddr:
AFAIK this is expected. perf-dash lookups cartesian product of {jobs, log files}, so it won't be able to find some of them. Error messages around it can be misleading it seems.

skurtzemann:
> AFAIK this is expected. perf-dash lookups cartesian product of {jobs, log files}, so it won't be able to find some of them. Error messages around it can be misleading it seems.

With these errors the dashboard will still display some data ?

oxddr:
Most of the time, this is expected. If you want to verify, data was loaded correctly, check some of the graphs (e.g. API responsivess or compare to perf-dash.k8s.io) . As mentioned earlier, those log statements are misleading.

wojtek-t:
[...]

PR #1154, opened 2020-04-01
Improve perfdash container build

skurtzemann:
This PR is to resolve issue #604 as seen with @oxddr

### DIFF
--- a/perfdash/Dockerfile
+++ b/perfdash/Dockerfile
@@ -12,8 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:jessie
-RUN apt-get update
-RUN apt-get install -y -qq ca-certificates
-ADD perfdash /perfdash
-ADD www /www
+# Build the application
+FROM golang:1.14-buster as build
+
+RUN go get -u github.com/tools/godep
+WORKDIR /go/src/app
+COPY . /go/src/app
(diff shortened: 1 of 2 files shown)


oxddr:
Looks good, thanks!

Can you please squash commits into a single one? Otherwise LGTM.

skurtzemann:
> Looks good, thanks!
> 
> Can you please squash commits into a single one? Otherwise LGTM.

Ok I will squash commits.

Another question, do I also need to
- bump `TAG` info `Makefile` and by the way update `deployment.yaml` 
- update `container` target into `Makefile` which will no more depend of `perfdash` target

oxddr:
> * bump `TAG` info `Makefile` and by the way update `deployment.yaml`

I'll take care of this at the same time I'll upgrade perfdash, since you don't have an access to GCP project running perf-dash. 

> * update `container` target into `Makefile` which will no more depend of `perfdash` target

Please do, in this PR ideally.

skurtzemann:
> > * bump `TAG` info `Makefile` and by the way update `deployment.yaml`
> 
> I'll take care of this at the same time I'll upgrade perfdash, since you don't have an access to GCP project running perf-dash.
> 
> > * update `container` target into `Makefile` which will no more depend of `perfdash` target
> 
> Please do, in this PR ideally.

PR is updated :)

oxddr:
/ok-to-test
/lgtm
/approve 

Thanks a lot for the PR!

PR #1155, opened 2020-04-02
Fixing comment with clusterloader2/pkg/util/ssh.go Exec method to make it compliant with golint

vamossagar12:


wojtek-t:
/ok-to-test
/lgtm
/approve

Thanks!

PR #1156, opened 2020-04-03
cleanup the clusterloader2 load testing config file and deployment yaml

maniSbindra:
As a part of #1125, default value of the variables
ENABLE_CONFIGMAPS, ENABLE_DAEMONSETS, ENABLE_JOBS, ENABLE_SECRETS, and ENABLE_STATEFULSETS are set to true, and these values are not overriden in any of the tests. The config file and deployment yaml can now be be simplified as discussed in https://github.com/kubernetes/perf-tests/issues/1036#issuecomment-607631768 .


wojtek-t:
Thanks!

/lgtm
/approve

PR #1157, opened 2020-04-06
Run dnsclient as a deployment in DNS performance tests

prameshj:
This change supports running dnsclient as a deployment. The replicas can be configured by editing the cluster/dnsperf-deploy.py, it will be made configurable in a followup change. This will help measure ClusterDNS and NodeLocal DNSCache performance at scale.

### DIFF
--- a/dns/README.md
+++ b/dns/README.md
@@ -221,6 +221,7 @@ query_file: ["nx-domain.txt", "outside.txt", "pod-ip.txt", "service.txt"]
 CREATE TABLE runs (
   run_id,
   run_subid,
+  pod_name,
   run_length_seconds,
   dnsmasq_cpu,
   dnsmasq_cache,
@@ -233,6 +234,7 @@ CREATE TABLE runs (
 CREATE TABLE results (
   run_id,
   run_subid,
+  pod_name,
   queries_sent,
   queries_completed,
   queries_lost,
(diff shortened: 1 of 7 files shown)


prameshj:
cc @mm4tt @bowei

mm4tt:
Forgive my ignorance,  I know nothing about the DNS performance test, but it sounds interesting.
Is the test run continuously somewhere? If so, could you link to a test grid? What exactly the test does?

Thanks!

/lgtm
/hold

Feel free to void whenever you want (e.g. if you want to get @bowei's approval)

prameshj:
> Forgive my ignorance, I know nothing about the DNS performance test, but it sounds interesting.
> Is the test run continuously somewhere? If so, could you link to a test grid? What exactly the test does?
> 
> Thanks!
> 
> /lgtm
> /hold
> 
> Feel free to void whenever you want (e.g. if you want to get Bowei Du's approval)

Thanks Matt! The test is run continuously here:
http://perf-dash.k8s.io/#/?jobname=kube-dns%20benchmark
http://perf-dash.k8s.io/#/?jobname=node-local-dns%20benchmark 

I have added more description in https://github.com/kubernetes/perf-tests/blob/master/dns/README.md

The test brings up one dnsclient pod on a GCE k8s cluster and sends DNS requests using dnsperf. We gather latency and QPS metrics. This PR allows us to run a deployment rather than a single pod so we can run this test on larger clusters and generate more DNS load.

mm4tt:
Nice, thanks for explaining!

PR #1158, opened 2020-04-06
Add windows memory usage measurement

YangLu1031:
1. Add memory usage measurement, query is from discussion: https://github.com/martinlindhe/wmi_exporter/issues/496
2. Rename cpu usage measurement methods
3. Bumped the default pause image version

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
+++ b/clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
@@ -30,10 +30,17 @@ import (
 const (
 	windowsResourceUsagePrometheusMeasurementName = "WindowsResourceUsagePrometheus"
 	// get top 10 non-system processes with highest cpu usage within 1min query window size
-	cpuUsageQueryTop10                        = `topk(10, sum by (process) (irate(wmi_process_cpu_time_total{process!~"Idle|Total|System"}[5m]) / on(job) group_left wmi_cs_logical_processors) * 100)`
+	cpuUsageQueryTop10 = `topk(10, sum by (process) (irate(wmi_process_cpu_time_total{process!~"Idle|Total|System"}[5m]) / on(job) group_left wmi_cs_logical_processors) * 100)`
+	// cpu usage metrics file name prefix
+	cpuUsageMetricsName = "WindowsCPUUsagePrometheus"
+	// get top 10 non-system processes with highest memory usage
+	memoryUsageQueryTop10 = `topk(10, sum(wmi_process_working_set{process!~"Idle|Total|System"}) by (process))`
+	// memory usage metrics file name prefix
+	memoryUsageMetricsName                    = "WindowsMemoryUsagePrometheus"
 	currentWindowsResourceUsageMetricsVersion = "v1"
 )
 
+type convertFunc func([]*model.Sample) *measurementutil.PerfData
 type windowsResourceUsageGatherer struct{}
 
 func (w *windowsResourceUsageGatherer) IsEnabled(config *measurement.MeasurementConfig) bool {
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/slos/windows_node_memory_usage.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+
+	"github.com/prometheus/common/model"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
```
mm4tt, line 30: There is a lot of boilerplate and copy-pasting here. Can we either:
1. Keep the windows_node_resource_usage measurement, that will do both cpu and memory
2. Extract the common parts somewhere and keep the different instance as short and simple as possible? E.g. you could have an abstract "WindowsResourceUsage" measurement that would take care of all the boilerplate. Then the particular instances of this measurement (e.g. cpu, memory, etc) would only have to provide the differentiating parts, e.g. the query to execute, name of fields in the summary, etc. 

WDYT?

wojtek-t: +1

YangLu1031: Thanks, make sense. 
For #1, as all the prometheus measurements need to implement the Gatherer interface in prometheus_measurement.go, and the `Execute()` method in prometheus_measurement.go could return multiple summary, but the Gather() method return only one summary, which seems not reasonable. Could we change the Gather() method to return list of summary? That will change the existing APIResponsiveness & NetworkingProgramming prometheus measurements also.
I'd prefer this modification, as it enables us to return multiple promethues measurement summary in one file to avoid the abstract layer. But if you think it might break the existing measurements, I will go for #2

mm4tt: Yeah, that's a great observation, the Gather in prometheus_measurement should return []measurement.Summary.
Could you send a separate PR to fix that?

YangLu1031: Sure

YangLu1031: Created a PR: https://github.com/kubernetes/perf-tests/pull/1159

wojtek-t: I already approved that one. Once that merges, please rebase this PR.

YangLu1031: rebased the PR, please take a look again

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -68,20 +74,48 @@ func convertToPerfData(samples []*model.Sample) *measurementutil.PerfData {
...
+			Labels: map[string]string{
+				"Process": string(sample.Metric["process"]),
+			},
+		}
+		perfData.DataItems = append(perfData.DataItems, item)
+	}
+	return perfData
+}
+
+func getSummary(query string, converter func(samples []*model.Sample) *measurementutil.PerfData, metricsName string, executor QueryExecutor, startTime time.Time, config *measurement.MeasurementConfig) (measurement.Summary, error) {
```
wojtek-t, line 94: Two nits:
1. please define a type for the function type, e.g.:
```
type convertFunc func([]*model.Sample) *measurementutil.PerfData
```
2. startTime seems unused - please remove

YangLu1031: Thanks, fixed

PR #1161, opened 2020-04-06
Revert "Improve perfdash container build #1154"

mm4tt:
The change broke deploying new versions of perf-dash. There are at least two things wrong:
1. The app was moved in the container from `/perfdash` to `/app/perfdash` but the `deployment.yaml` wasn't updated to reflect this change
2. Even with the fixed binary path the perf-dash is still not operational, it seems to be working, but it's impossible to connect to it via https. I assume it's because of the missing ca-certificates

I spent over 30 minutes debugging it. It looks like the change wasn't tested, please remember to test your changes in future. 
/assign @oxddr 
/cc @skurtzemann

### DIFF
--- a/perfdash/Dockerfile
+++ b/perfdash/Dockerfile
@@ -12,18 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Build the application
-FROM golang:1.14-buster as build
-
-RUN go get -u github.com/tools/godep
-WORKDIR /go/src/app
-COPY . /go/src/app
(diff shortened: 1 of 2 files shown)


skurtzemann:
@mm4tt : I'm sorry I only test with a local container :(
I'll retake the PR and deploy it on a k8s cluster before !

mm4tt:
No worries, it happens.

Also, you don't have to deploy it on your own. @oxddr should be able to do that once you prepare the PR. You can try bumping perfdash version artificially and try deploying it using new scripts to confirm whether it works. 

Thanks!

skurtzemann:
@mm4tt , @oxddr : new PR done

Issue #1096, opened 2020-03-03
Create api-availability measurement


mm4tt:
### Justification

See this comment - https://github.com/kubernetes/perf-tests/issues/1086#issuecomment-593918574

#### V0

1. Create a new "ApiAvailability" measurement
1. This measurement should periodically probe the apiserver's /healthz endpoint and record whether api was available or not
1. The measurement should output the following stats
    1. availability percentage (e.g. % of OK responses over all responses)
    1. Longest consecutively unavailability period
    1. ...

#### V1
1. Make the measurement fail if apiserver is consecutively  not available in the last XX min (e.g. 30min). 
1. Make the error critical to make sure the test execution is stopped in such case

#### V2

1. Come up with exact SLI/SLO definition, make it available in perf-dash for further analysis
1. Make it a WIP [Scalability SLO](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md)
1. Evaluate and promote to official SLO

/good-first-issue

vamossagar12:
/assign
I would like to take this up. Would these be more details furnished?

mm4tt:
Great, thanks @vamossagar12 

Let's start with the V0. Could you let me know which part of the description is not clear or requires further details?

vamossagar12:
thanks @mm4tt Couple of things:
1) how frequently should the probe happen? Would it be configurable or we can define it to be fixed?
2) The availability percentage should be wrt the entire perf test run and we have to keep track of percentage and longest unavailable period?

As an outside question, typically if the health end point is down for a configured time, response would be to scale up or something like that. Just wanted to know the rationale behind adding this particular measurement. Thanks!

wojtek-t:
> how frequently should the probe happen? Would it be configurable or we can define it to be fixed?

Seems like implementation detail. We can make it configurable, but I doubt we will be using different values once we agree on sth.

> The availability percentage should be wrt the entire perf test run and we have to keep track of percentage and longest unavailable period?

Correct

> As an outside question, typically if the health end point is down for a configured time, response would be to scale up or something like that. Just wanted to know the rationale behind adding this particular measurement. Thanks!

There are couple points:
- we want to ensure that we understand how availability looks like (I can imgine all other SLOs being fine, but cluster being periodicall unavailable)
- we can use it as optimization - if cluster is down for X minutes, the it probably won't get up ever, and we can shut down the test and save money

vamossagar12:
[...]

PR #1162, opened 2020-04-07
Api availability measurement

vamossagar12:
This PR is for https://github.com/kubernetes/perf-tests/issues/1096 which created a new measurement to track the availability of api-server by probing the /healthz end point

### DIFF
--- a/clusterloader2/pkg/measurement/common/api_availability_measurement.go
+++ b/clusterloader2/pkg/measurement/common/api_availability_measurement.go
@@ -1,0 +1,173 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/provider"
+	"sync"
+	"time"
+
+	goerrors "github.com/go-errors/errors"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	apiAvailabilityName = "APIAvailability"
+)
+
+func init() {
+	if err := measurement.Register(apiAvailabilityName, createAPIAvailabilityMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", apiAvailabilityName, err)
+	}
+}
+
+func createAPIAvailabilityMeasurement() measurement.Measurement {
+	return &apiAvailabilityMeasurement{}
+}
+
+type apiAvailabilityMeasurement struct {
+	isRunning           bool
+	stopCh              chan struct{}
+	hosts               []string
+	summaries           []measurement.Summary
+	clusterLevelMetrics *apiAvailabilityMetrics
+	hostLevelMetrics    map[string]*apiAvailabilityMetrics
+	wg                  sync.WaitGroup
+}
+
+func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.Config, provider provider.Provider) {
+	for _, host := range a.hosts {
+		// SSH and check the health of the host
+		command := fmt.Sprintf("curl -f -s -k %slocalhost:%v/healthz", "https://", 443)
+		sshResult, err := measurementutil.SSH(command, host+":22", provider)
+		availability := err != nil || sshResult.Code != 0
+		a.updateAvailabilityMetrics(availability, a.hostLevelMetrics[host])
+	}
+}
+
+func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
+	// Check the availability of the cluster by issuing a REST call to /healthz end point
+	result := c.CoreV1().RESTClient().Get().AbsPath("/healthz").Do(context.TODO())
+	status := 0
+	result.StatusCode(&status)
+	a.updateAvailabilityMetrics(status == 200, a.clusterLevelMetrics)
+}
+
+func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMasterSupported bool, probeDuration int) error {
+	a.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(a.hosts) < 1 {
+		return goerrors.Errorf("APIAvailability measurement can't start due to lack of master IPs")
+	}
+
+	k8sClient := config.ClusterFramework.GetClientSets().GetClient()
+	provider := config.ClusterFramework.GetClusterConfig().Provider
+
+	a.isRunning = true
+	a.stopCh = make(chan struct{})
+	a.wg.Add(1)
+
+	go func() {
+		defer a.wg.Done()
+		for {
+			select {
+			case <-a.stopCh:
+				return
+			case <-time.After(time.Duration(probeDuration)):
+				a.updateClusterAvailabilityMetrics(k8sClient)
+				if SSHToMasterSupported {
+					a.updateMasterAvailabilityMetrics(k8sClient, config, provider)
+				}
+			}
+		}
+	}()
+	return nil
+}
+
+// Execute starts the api-server healthz probe end point from start action and
+// collects availability metrics in gather.
+func (a *apiAvailabilityMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	SSHToMasterSupported := !config.ClusterFramework.GetClusterConfig().Provider.Features().SupportSSHToMaster
+
+	if !SSHToMasterSupported {
+		klog.Infof("ssh to master nodes not supported. Measurement would have only Cluster Level Metrics")
+	}
+
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	probeFrequency, err := util.GetIntOrDefault(config.Params, "frequency", 1)
+	if err != nil {
+		return nil, err
+	}
+
+	switch action {
+	case "start":
+		if a.isRunning {
+			klog.Infof("%s: measurement already running", a)
+			return nil, nil
+		}
+		return nil, a.start(config, SSHToMasterSupported, probeFrequency)
+	case "gather":
+		err := a.stopAndSummarize(SSHToMasterSupported, probeFrequency)
+		if err != nil {
+			return nil, err
+		}
+		return a.summaries, nil
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) stopAndSummarize(SSHToMasterSupported bool, probeFrequency int) error {
+	if !a.isRunning {
+		return nil
+	}
+	close(a.stopCh)
+	a.wg.Wait()
+
+	output := apiAvailabilityOutput{}
+	output.createClusterSummary(a.clusterLevelMetrics, probeFrequency)
+	if SSHToMasterSupported {
+		output.createMastersSummary(a.hostLevelMetrics, a.hosts, probeFrequency)
+	}
+
+	errList := errors.NewErrorList()
+	content, err := util.PrettyPrintJSON(output)
+	if err != nil {
+		errList.Append(errList, err)
+	} else {
+		summary := measurement.CreateSummary(apiAvailabilityName, "json", content)
+		a.summaries = append(a.summaries, summary)
+	}
+	return errList
+}
+
+// Dispose cleans up after the measurement.
+func (a apiAvailabilityMeasurement) Dispose() {}
+
+// String returns string representation of this measurement.
+func (a apiAvailabilityMeasurement) String() string {
+	return apiAvailabilityName
+}
--- a/clusterloader2/pkg/measurement/common/api_availability_metrics.go
+++ b/clusterloader2/pkg/measurement/common/api_availability_metrics.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"time"
+)
+
+type apiAvailabilityMetrics struct {
+	numSuccesses               int
+	numFailures                int
+	maxConsecutiveFailedProbes int
+	consecutiveFailedProbes    int
+}
+
+type apiAvailabilitySummary struct {
+	IP                         string        `json:"IP"`
+	AvailabilityPercentage     float32       `json:"availabilityPercentage"`
+	LongestUnavailableDuration time.Duration `json:"longestUnavailableDuration"`
+}
+
+type apiAvailabilityOutput struct {
+	ClusterMetrics *apiAvailabilitySummary   `json:"clusterMetrics"`
+	MasterMetrics  []*apiAvailabilitySummary `json:"masterMetrics"`
+}
+
+func (output apiAvailabilityOutput) createClusterSummary(metrics *apiAvailabilityMetrics, probeFrequency int) {
+	metrics.buildAPIAvailabilityMetricsSummary(probeFrequency, true, "")
+}
+
+func (output apiAvailabilityOutput) createMastersSummary(metrics map[string]*apiAvailabilityMetrics, hosts []string, probeFrequency int) {
+	for _, host := range hosts {
+		masterHostAvailabilitySummary := metrics[host].buildAPIAvailabilityMetricsSummary(probeFrequency, false, host)
+		output.MasterMetrics = append(output.MasterMetrics, masterHostAvailabilitySummary)
+	}
+
+}
+
+// updateMaxConsecutiveFailuresIfNeeded checks if the recently concluded consecutive failed number of probes is
+// higher than the max consecutive failed number of probes so far
+// if yes, then Update max consecutive failed probes
+func (a *apiAvailabilityMetrics) updateMaxConsecutiveFailuresIfNeeded() {
+	if a.consecutiveFailedProbes > a.maxConsecutiveFailedProbes {
+		a.maxConsecutiveFailedProbes = a.consecutiveFailedProbes
+	}
+}
+
+func (a *apiAvailabilityMetrics) updateFailureMetrics() {
+	a.numFailures++
+	a.consecutiveFailedProbes++
+	a.updateMaxConsecutiveFailuresIfNeeded()
+}
+
+func (a *apiAvailabilityMetrics) updateSuccessMetrics() {
+	a.numSuccesses++
+	a.consecutiveFailedProbes = 0
+}
+
+func (a *apiAvailabilityMeasurement) updateAvailabilityMetrics(apiServerAvailable bool, metrics *apiAvailabilityMetrics) {
+	if apiServerAvailable {
+		metrics.updateSuccessMetrics()
+	} else {
+		metrics.updateFailureMetrics()
+	}
+}
+
+func (a *apiAvailabilityMetrics) buildAPIAvailabilityMetricsSummary(probeFrequency int, clusterMetrics bool, hostIP string) *apiAvailabilitySummary {
+	// Gather availability a
+	availabilityPercentage := (float32(a.numSuccesses) / float32(a.numSuccesses+a.numFailures)) * 100
+	longestUnavailableDuration := time.Duration(a.maxConsecutiveFailedProbes * probeFrequency)
+
+	apiAvailabilitySummary := &apiAvailabilitySummary{}
+	apiAvailabilitySummary.AvailabilityPercentage = availabilityPercentage
+	apiAvailabilitySummary.LongestUnavailableDuration = longestUnavailableDuration
+
+	return apiAvailabilitySummary
+}


mm4tt:
/ok-to-test

Thanks, @vamossagar12!

I'll try to take a look tomorrow. Adding Wojtek in case he has capacity to take a look today
/assign @wojtek-t

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+	probeDuration, err := util.GetInt(config.Params, "duration")
+	if err != nil {
+		return nil, err
+	}
+	ticker := time.NewTicker(500 * time.Millisecond)
+	stopAvailabilityCheck := make(chan bool)
+	apiServerClient := config.ClusterFramework.GetClientSets().GetClient().CoreV1().RESTClient()
+	numProbes := 0
```
wojtek-t, line 56: Instead of having one huge Execute method, I would much rather prefer having a more organized approach with dedicate start() and gather() methods, those variables being measurement fields, etc.

As an example, PTAL at profile measurement, which is pretty similar conceptually to what you want to achieve:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/profile.go

vamossagar12: thanks @wojtek-t this was just an initial stab. That's why I created a draft PR to get comments and the direction i should be moving in.

vamossagar12: @wojtek-t thanks for pointing me to this example. As you said, it's pretty similar to what I want to achieve.
Couple of questions about interacting with api server. I see that there's explicit configs to interact with https and also add rbac roles to expose debug end point. Do you think we need all those features in this measurement?

wojtek-t: > I  see that there's explicit configs to interact with https and also add rbac roles to expose debug end point. Do you think we need all those features in this measurement?

https is definitely needed. Re rbac - I can't remember from the top of my head how /health is handled - I think it's not needed, but it requires checking once you implement the measurement itself.

vamossagar12: ok. Another question that I have is, the profile measurement creates summaries during every run and appends them to a list. Do we want to have something similar? What I understood from the OP was that we want to print the statistics only at the end of the test?

wojtek-t: We want to generate a single summary at the end. That said you need to keep track of the information you gather over time - so although slightly different, this will be fairly similar.

vamossagar12: yeah, i am already keeping track of some of those metrics. just need to surface them out at the end of the tests.

vamossagar12: hey i have made further changes to the measurements which are inline to the profile measurements and yet generating only a single summary at the end of gather phase.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+	hosts                  []string
+	apiAvailabilityMetrics *apiAvailabilityMetrics
+	wg                     sync.WaitGroup
+}
+
+func (a *apiAvailabilityMeasurement) apiServerAvailable(c clientset.Interface, SSHToMasterSupported bool, config *measurement.MeasurementConfig) bool {
+	for _, host := range a.hosts {
+		// Get the api-server measurement data over SSH.
+		// Start by checking that the provider allows us to do so.
+		if !SSHToMasterSupported {
```
wojtek-t, line 71: I don't think we should even consider SSH for this purpose.
Simply calling /healthz using client is what we want.

The problem is that using client you currently contact a random master. That said, maybe it's actually fine - if there is a single master down but load-balancer is not forwarding any requests to it, that sounds like a healthy cluster?

Thinking more about it, what we may consider is:
- healthiness at the cluster level
- healthiness of individual masters

wojtek-t: [That said, I'm not sure if in CL2 we support passing "cluster-level IP" and "master IPs". Maybe we do and the first is via kubeconfig?]

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func updateFailureMetrics(apiAvailabilityMetrics *apiAvailabilityMetrics) {
+	// increment total number of failed probes count, current number of failed probes by 1
+	apiAvailabilityMetrics.numFailures++
+	apiAvailabilityMetrics.consecutiveFailedProbes++
+	// if the previous probe was not a failure, then set the previous probe failure flag to true.
+	// This would be used by next probe
+	if !apiAvailabilityMetrics.prevProbeFailure {
+		apiAvailabilityMetrics.prevProbeFailure = true
```
wojtek-t, line 128: Why do we need it? if consecutiveFailedProbes == 0 (or 1 at this poin in the code) that means that prevProbe wasn't a failure, otherwise it was).

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func updateSuccessMetrics(apiAvailabilityMetrics *apiAvailabilityMetrics) {
+	// if the probe succeeded, then increase the number of successful probes count by 1
+	apiAvailabilityMetrics.numSuccess++
+	// if this successful probe was received after failing probe(s)
+	if apiAvailabilityMetrics.prevProbeFailure {
```
wojtek-t, line 136: This can be simplified a lot - see my comment above.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+// collect metrics in gather.
+func (a *apiAvailabilityMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+
+	SSHToMasterSupported := config.ClusterFramework.GetClusterConfig().SSHToMasterSupported
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+
+	probeDuration, err := util.GetInt(config.Params, "duration")
```
wojtek-t, line 160: duration isn't the best name - it's rather frequency maybe?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+	switch action {
+	case "start":
+		if a.isRunning {
+			klog.Infof("%s: measurement already running", a)
+			return nil, nil
+		}
+		return a.start(config, SSHToMasterSupported, probeDuration)
+	case "gather":
+		a.stop()
+		availabilityPercentage := (a.apiAvailabilityMetrics.numSuccess / a.apiAvailabilityMetrics.numSuccess) * 100
```
wojtek-t, line 174: Won't this always be 100?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: We should also default to something sane, maybe 1s?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) stop() {
+	if !a.isRunning {
+		return
+	}
+	close(a.stopCh)
+	a.wg.Wait()
```
wojtek-t, line 193: That doesn't fully solve the problem - if the last period was the longest one, it won't be reported.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+		klog.Fatalf("Cannot register %s: %v", apiAvailabilityName, err)
+	}
+}
+
+func createApiAvailabilityMeasurement() measurement.Measurement {
+	return &apiAvailabilityMeasurement{}
+}
+
+type apiAvailabilityOutput struct {
+	AvailabilityPercentage     int `json:"availabilityPercentage"`
```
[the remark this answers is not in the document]
vamossagar12: Makes sense. I made the change.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,202 @@
...
+	LongestUnavailableDuration int `json:"longestUnavailableDuration"`
```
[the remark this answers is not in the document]
vamossagar12: I wasn't aware of the duration type in golang. Thay sounds more appropriate. I modified it.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Well, I hadn't thought of doing it via the consecutiveFailedProbes counter. On second thoughts, when I re-looked at the code, the checks with a boolean flag seemed a bit more intuitive to understand that the previous probe was a failure than a counter based check. That's my personal opinion though and can change it if it seems complicated.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Same as above.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: True. Did both.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Can you plz help me understand this scenario? Are you saying that when gather was called, there could have been another probe which would have made it the longest failing one and because I am closing the channel, it will miss that probe?
I have limited understanding of go routines but if the above scenario is correct, then does it mean that the channel stopping would override any currently executing probe? Should I wait before closing to check if any probes are ongoing?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: That was a bad miss, sorry. Updated.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: > I don't think we should even consider SSH for this purpose.
> Simply calling /healthz using client is what we want.
> 
> The problem is that using client you currently contact a random master. That said, maybe it's actually fine - if there is a single master down but load-balancer is not forwarding any requests to it, that sounds like a healthy cluster?
> 
> Thinking more about it, what we may consider is:
> 
> * healthiness at the cluster level
> * healthiness of individual masters

As per my understanding, If a single master is down and LB isn't forwarding any requests to it, then it can't be qualified as a healthy cluster. 

I think with the current implementation, for non-ssh mode, what we are considering is cluster level health. So, if a request returns, we count it as a success. 
It's only in the ssh case, where we consider hosts but in this case, if any one of the hosts reports an error, it is counted as a failure. So, that's an individual master level probe. 

So, you are saying we report both cluster/individual level modes? If we add individual level, then we will also need to track which masters failed how many times?

wojtek-t: > As per my understanding, If a single master is down and LB isn't forwarding any requests to it, then it can't be qualified as a healthy cluster.

If there are multiple masters and one is down but not receiving any traffic, I claim that from user-perspective the cluster is healthy. And we care about user-perspective here.

> So, you are saying we report both cluster/individual level modes? If we add individual level, then we will also need to track which masters failed how many times?

yes - we need the same data for 1. cluster as a whole, 2. each master separately

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: The less variables/fields we have the better.

If you want, you can introduce a method, like:
```
func (m *apiAvailabilitymetrics) previousProbeFailed() bool {
}
```

that will return what you want and use it where you want.
Let's not introduce unnecessary vars that you have to keep in sync.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: This is the problem:
https://github.com/kubernetes/perf-tests/pull/1162/files#diff-769e85d1bdfa456cbbe72288f2ba2ab5R140-R142

If you have gazillions failure in a row at the end of the test, you will not include it in your summary.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Understood. Do we want the health details for each master at every probe? If yes, then I will have to change the code - which I reused from the probes example that you had sent.
So, when SSH is not supported, then what we will have is only cluster level metrics.

It's only when ssh is supported that I can get individual level metrics. Here, currently from the re-used logic, if **any** of the master fails, the method returns a false for api server availability. Instead, I will have to ssh into each master, get the success/failure and report it back and update in a map where I will store metrics against each master. 

The only thing is, if we have too many masters then is there a chance of the probes overlapping with each other? Can that have a bearing on the metrics reported?

And lastly, is it a fair thing to say that if ssh isn't supported then all we get is cluster level metrics, and in the other case, both cluster level and individual master level metrics?

wojtek-t: I think we should start with something and iterate, rather than trying to get a perfect solution from the very beginning:
- I'm fine initially that only return all metrics for ssh-supported providers [and that should be hidden behind flag anyway]
- re overlapping probes - that also sounds fine at least initially 

A clear TODOs should be added though to emhpasize the problems we're aware of.

vamossagar12: hey @wojtek-t , I have made the changes. I couldn't push them because formatting changes are failing for a recently pushed commit to probes.go. I have rectified it here:
https://github.com/kubernetes/perf-tests/pull/1190

Once merged, I will push my changes.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: I made that change.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: I added the comparision during the gather flow as well.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Thanks for merging. I pushed the latest changes.

vamossagar12: hi @wojtek-t , did you get a chance to look at the latest changes?

wojtek-t:
@vamossagar12 - sorry for delay, I'm pretty overloaded. I will try to look into it today or tomorrow

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+	"sync"
+	"time"
+)
+
+const (
+	apiAvailabilityName = "ApiAvailability"
```
wojtek-t, line 32: s/ApiAvailability/APIAvailability/

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+func init() {
+	if err := measurement.Register(apiAvailabilityName, createApiAvailabilityMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", apiAvailabilityName, err)
+	}
+}
+
+func createApiAvailabilityMeasurement() measurement.Measurement {
```
wojtek-t, line 41: s/Api/API/  (same everywhere else)

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+
+type apiAvailabilityOutput struct {
+	ClusterMetrics             bool          `json:"clusterMetrics"`
+	HostIp                     string        `json:"hostIp"`
+	AvailabilityPercentage     float32       `json:"availabilityPercentage"`
+	LongestUnavailableDuration time.Duration `json:"longestUnavailableDuration"`
+}
+
+type apiAvailabilityMetrics struct {
+	numProbes                  int
```
wojtek-t, line 53: numProbes seems unneded - it's numSuccess+numFailures

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	numSuccess                 int
```
wojtek-t, line 54: numSuccesses (plural, for consistency with failures)

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+type apiAvailabilityMeasurement struct {
+	isRunning           bool
+	stopCh              chan struct{}
+	hosts               []string
+	clusterLevelMetrics *apiAvailabilityMetrics
+	hostLevelMetrics    map[string]*apiAvailabilityMetrics
+	wg                  sync.WaitGroup
+}
+
+func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.MeasurementConfig) {
```
wojtek-t, line 69: This PR requires rebase - e.g. MeasurementConfig was renamed to Config.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+				}
+				return
+			}
+
+		}
+	}()
+	return nil, nil
+}
+
+func (a *apiAvailabilityMeasurement) updateAvailabilityMetrics(apiServerAvailable bool, metrics *apiAvailabilityMetrics) {
```
wojtek-t, line 125: nit: move above the start method to keep this stuff together

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.MeasurementConfig) {
+	for _, host := range a.hosts {
+		// update the number of probes
+		a.hostLevelMetrics[host].numProbes++
+		// SSH and check the health of the host
+		command := fmt.Sprintf("curl -s -k %slocalhost:%v/healthz", "https://", 443)
+		_, err := measurementutil.SSH(command, host+":22", "")
+		if err != nil {
```
wojtek-t, line 76: Does it really do what you want? If the command fails with non-zero error code, is the err always returned?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+		a.updateAvailabilityMetrics(false, a.clusterLevelMetrics)
+	} else {
+		a.updateAvailabilityMetrics(true, a.clusterLevelMetrics)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) start(config *measurement.MeasurementConfig, SSHToMasterSupported bool, probeDuration int) ([]measurement.Summary, error) {
+	a.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(a.hosts) < 1 {
+		return nil, goerrors.Errorf("Api Server measurement can't execute due to no MasterIps")
```
wojtek-t, line 99: APIAvailability measurement ...

s/MasterIps/MasterIPs/

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	k8sClient := config.ClusterFramework.GetClientSets().GetClient()
+	a.isRunning = true
+	a.stopCh = make(chan struct{})
+	a.wg.Add(1)
+	go func() {
```
wojtek-t, line 106: nit: add an empty line before this go func - it's separating intialization from actual execution

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+		for {
+			select {
+			case <-a.stopCh:
+				return
+			case <-time.After(time.Duration(probeDuration)):
+				a.updateClusterAvailabilityMetrics(k8sClient)
+				if SSHToMasterSupported {
+					a.updateMasterAvailabilityMetrics(k8sClient, config)
+				}
+				return
```
wojtek-t, line 117: that seems incorrect - we don't want to return after the first probe, right?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (a *apiAvailabilityMeasurement) updateAvailabilityMetrics(apiServerAvailable bool, metrics *apiAvailabilityMetrics) {
+	if apiServerAvailable {
+		metrics.updateSuccessMetrics()
+	} else {
+		metrics.updateFailureMetrics()
+	}
+}
+
+// If the consecutive failed probes counter value > 0, then it means we are in the middle of an unavailability period
```
wojtek-t, line 133: nit: please make this line shorter

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+// for the api server
+func (a *apiAvailabilityMetrics) previousProbeFailure() bool {
+	return a.consecutiveFailedProbes > 0
+}
+
+// Check if the recently concluded consecutive failed number of probes is
+// higher than the max consecutive failed number of probes so far
+// if yes, then Update max consecutive failed probes
+func (a *apiAvailabilityMetrics) mayBeUpdateMaxConsecutiveFailures() {
```
wojtek-t, line 142: updateMaxConsecutiveFailuresIfNeeded

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	if a.consecutiveFailedProbes > a.maxConsecutiveFailedProbes {
+		a.maxConsecutiveFailedProbes = a.consecutiveFailedProbes
+	}
+}
+
+func (a *apiAvailabilityMetrics) updateFailureMetrics() {
```
wojtek-t, line 148: Can you maybe please move apiAvailabilityMetrics type and its methods to a separate file?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	// increment total number of failed probes count, current number of failed probes by 1
```
wojtek-t, line 149: remove this comment - don't describe what the code is doing - only describe why if that isn't obvious (in this case it is, so no comment is needed)

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	a.numFailures++
+	a.consecutiveFailedProbes++
+}
+
+func (a *apiAvailabilityMetrics) updateSuccessMetrics() {
+	// if the probe succeeded, then increase the number of successful probes count by 1
```
wojtek-t, line 155: same here and below

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+		a.mayBeUpdateMaxConsecutiveFailures()
+		// reset current consecutive failed number of probes
+		a.consecutiveFailedProbes = 0
+	}
+}
+
+// Execute starts the api-server healthz probe end point from start action and
+// collects availability metrics in gather.
+func (a *apiAvailabilityMeasurement) Execute(config *measurement.MeasurementConfig) ([]measurement.Summary, error) {
+
```
wojtek-t, line 168: nit: remove this empty line

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+			summaries = a.createMasterAvailabilitySummary(probeFrequency, summaries)
+		}
+		return summaries, nil
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) createClusterAvailabilitySummary(probeFrequency int, summaries []measurement.Summary) []measurement.Summary {
+	a.clusterLevelMetrics.mayBeUpdateMaxConsecutiveFailures()
```
wojtek-t, line 201: let's not do this here - instead let's make it a defer in stop() function (or not a defer - just done after wg.Wait()

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	clusterAvailabilityMetricsOutput := a.buildApiAvailabilityMetricsOutput(probeFrequency, a.clusterLevelMetrics, true, "")
+	content, err := util.PrettyPrintJSON(clusterAvailabilityMetricsOutput)
+	if err == nil {
```
wojtek-t, line 204: don't ignore the error - if error != nil this function should return an error

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+		summaries = append(summaries, measurement.CreateSummary(apiAvailabilityName, "json", content))
+	}
+	return summaries
+}
+
+func (a *apiAvailabilityMeasurement) createMasterAvailabilitySummary(probeFrequency int, summaries []measurement.Summary) []measurement.Summary {
+	for _, host := range a.hosts {
+		a.hostLevelMetrics[host].mayBeUpdateMaxConsecutiveFailures()
+		hostAvailabilityMetricsOutput := a.buildApiAvailabilityMetricsOutput(probeFrequency, a.hostLevelMetrics[host], false, host)
+		content, err := util.PrettyPrintJSON(hostAvailabilityMetricsOutput)
```
wojtek-t, line 214: don't ignore the error - if error != nil this function should return an error

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	content, err := util.PrettyPrintJSON(clusterAvailabilityMetricsOutput)
+	if err == nil {
+		summaries = append(summaries, measurement.CreateSummary(apiAvailabilityName, "json", content))
+	}
+	return summaries
+}
+
+func (a *apiAvailabilityMeasurement) createMasterAvailabilitySummary(probeFrequency int, summaries []measurement.Summary) []measurement.Summary {
+	for _, host := range a.hosts {
+		a.hostLevelMetrics[host].mayBeUpdateMaxConsecutiveFailures()
```
wojtek-t, line 212: Same here - this should be done in stop() function

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	// Check the availability of the cluster by issuing a REST call to /healthz end point
+	_, err := c.CoreV1().RESTClient().Get().AbsPath("/healthz").DoRaw()
+	if err != nil {
+		a.updateAvailabilityMetrics(false, a.clusterLevelMetrics)
+	} else {
+		a.updateAvailabilityMetrics(true, a.clusterLevelMetrics)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) start(config *measurement.MeasurementConfig, SSHToMasterSupported bool, probeDuration int) ([]measurement.Summary, error) {
```
wojtek-t, line 96: start shouldn't return measurement.Summary - simply return just error

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+		content, err := util.PrettyPrintJSON(hostAvailabilityMetricsOutput)
+		if err == nil {
+			summary := measurement.CreateSummary(apiAvailabilityName, "json", content)
+			summaries = append(summaries, summary)
+		}
+	}
+	return summaries
+}
+
+func (a *apiAvailabilityMeasurement) buildApiAvailabilityMetricsOutput(probeFrequency int, metrics *apiAvailabilityMetrics, clusterMetrics bool, hostIp string) *apiAvailabilityOutput {
```
wojtek-t, line 223: this shouldn't be apiAvailabilityMeasurement method (a isn't used anywhere).

Instead, make this a method of apiAvailabilityMetrics

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+	// Gather availability metrics
+	availabilityPercentage := (float32(metrics.numSuccess) / float32(metrics.numProbes)) * 100
+	longestUnavailableDuration := metrics.maxConsecutiveFailedProbes * probeFrequency
+	availabilityMetrics := &apiAvailabilityOutput{AvailabilityPercentage: availabilityPercentage, LongestUnavailableDuration: time.Duration(longestUnavailableDuration), ClusterMetrics: clusterMetrics}
```
wojtek-t, line 227: nit: please don't try to put it into single line

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,248 @@
...
+import (
+	"fmt"
+	goerrors "github.com/go-errors/errors"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+	"sync"
+	"time"
```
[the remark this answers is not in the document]
vamossagar12: done

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Do you mean that err won't capture failure of command with non-zero error code?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Sure, would it make sense to move them to the utils package?

vamossagar12:
Thanks @wojtek-t I have more or less implemented the changes you suggested. Just had a couple of questions post which I can push the changes.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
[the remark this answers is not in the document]
wojtek-t: not done

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: Yes - I think so (which is kind-of by-design):
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/util/ssh.go#L65
https://github.com/kubernetes/perf-tests/blob/2eeb004780c3f0ec99d0e03e6e44261e03017d6f/clusterloader2/vendor/k8s.io/kubernetes/pkg/ssh/ssh.go#L217

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: They are pretty specific to this measurement, so I don't think utils package is the best.
At some point, we should probably refactor each measurement to its own subpackage, but that's independent from this PR for sure.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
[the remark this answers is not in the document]
vamossagar12: yeah sorry about that. i haven't pushed the changes yet

vamossagar12:
@wojtek-t i made the changes but not able to push them. Keep getting this error after rebasing with the latest from master:

.git/hooks/pre-push: line 11: ./verify/verify-boilerplate.sh: No such file or directory

I see some changes have been made recently in the /verify folder but not able to figure out what the problem is.

wojtek-t:
@vamossagar12 - the verify-boilerplace.sh script still exists and wasn't touched:
https://github.com/kubernetes/perf-tests/blob/master/verify/verify-boilerplate.sh

BTW - the hooks must be something in your local environment - I don't have them set up.

vamossagar12:
> @vamossagar12 - the verify-boilerplace.sh script still exists and wasn't touched:
> https://github.com/kubernetes/perf-tests/blob/master/verify/verify-boilerplate.sh
> 
> BTW - the hooks must be something in your local environment - I don't have them set up.

Well, actually I set the hooks up since they were mentioned in the README of perf-tests. If it isn't needed, I can remove it. Would it have any implications though?

wojtek-t:
> Well, actually I set the hooks up since they were mentioned in the README of perf-tests. If it isn't needed, I can remove it. Would it have any implications though?

We have presubmits anyway - so these should be caught by them anyway.

vamossagar12:
> > Well, actually I set the hooks up since they were mentioned in the README of perf-tests. If it isn't needed, I can remove it. Would it have any implications though?
> 
> We have presubmits anyway - so these should be caught by them anyway.

Alright. Removed the local hooks folder and was able to push now.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+	hostLevelMetrics    map[string]*apiAvailabilityMetrics
+	wg                  sync.WaitGroup
+}
+
+func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.Config) {
+	for _, host := range a.hosts {
+		// SSH and check the health of the host
+		command := fmt.Sprintf("curl -s -k %slocalhost:%v/healthz", "https://", 443)
+		sshResult, err := measurementutil.SSH(command, host+":22", "")
+		if err != nil || sshResult.Code != 0 {
```
wojtek-t, line 62: What does curl return, if the server return say 500?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+			a.updateAvailabilityMetrics(false, a.hostLevelMetrics[host])
+		} else {
+			a.updateAvailabilityMetrics(true, a.hostLevelMetrics[host])
+		}
+	}
+}
+
+func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
+	// Check the availability of the cluster by issuing a REST call to /healthz end point
+	_, err := c.CoreV1().RESTClient().Get().AbsPath("/healthz").DoRaw()
```
wojtek-t, line 72: Don't we need to parse the response to get the returned code?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) createClusterAvailabilitySummary(probeFrequency int) error {
+	a.clusterLevelMetrics.updateMaxConsecutiveFailuresIfNeeded()
+	clusterAvailabilityMetricsOutput := a.clusterLevelMetrics.buildAPIAvailabilityMetricsOutput(probeFrequency, true, "")
+	content, err := util.PrettyPrintJSON(clusterAvailabilityMetricsOutput)
+	if err == nil {
```
wojtek-t, line 144: if err != nil {
  return err
}
// the rest of logic

[less nesting and easier to follow]

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+func (a *apiAvailabilityMeasurement) createMasterAvailabilitySummary(probeFrequency int) error {
+	createMasterAvailabilitySummaryErrors := errors.NewErrorList()
+	for _, host := range a.hosts {
+		a.hostLevelMetrics[host].updateMaxConsecutiveFailuresIfNeeded()
+		hostAvailabilityMetricsOutput := a.hostLevelMetrics[host].buildAPIAvailabilityMetricsOutput(probeFrequency,false, host)
+		content, err := util.PrettyPrintJSON(hostAvailabilityMetricsOutput)
+		if err == nil {
```
wojtek-t, line 158: if err != nil {
   ...
   continue
}
// rest of logic

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+			summary := measurement.CreateSummary(apiAvailabilityName, "json", content)
+			a.summaries = append(a.summaries, summary)
+		} else{
+			createMasterAvailabilitySummaryErrors.Append(createMasterAvailabilitySummaryErrors, err)
```
wojtek-t, line 162: I think you only want to append err

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type apiAvailabilityOutput struct {
+	ClusterMetrics             bool          `json:"clusterMetrics"`
+	HostIP                     string        `json:"hostIP"`
+	AvailabilityPercentage     float32       `json:"availabilityPercentage"`
+	LongestUnavailableDuration time.Duration `json:"longestUnavailableDuration"`
+}
+
+// number of consecutive failed probes is > 0 signifies previous probe was a failure
```
wojtek-t, line 37: // previousProbeFailed returns whether the last probe failed healthcheck.

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+func (a *apiAvailabilityMetrics) previousProbeFailure() bool {
```
wojtek-t, line 38: previousProbeFailed

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+	return a.consecutiveFailedProbes > 0
+}
+
+// Check if the recently concluded consecutive failed number of probes is
```
wojtek-t, line 42: follow the gostyle: comments should start with method name.

clusterloader2/pkg/measurement/common/api_availability_metrics.go
wojtek-t: Actually - let's remove this method completely - see my comment below.

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (a *apiAvailabilityMetrics) updateFailureMetrics() {
+	a.numFailures++
+	a.consecutiveFailedProbes++
+}
+
+func (a *apiAvailabilityMetrics) updateSuccessMetrics() {
+	a.numSuccesses++
+	if a.previousProbeFailure() {
```
wojtek-t, line 58: let's remove this check and do that unconditionally - it's not needed.

if the previous attempt was success, consecutiveFailedProbes=0, but maxing with 0 and setting it to 0 don't hurt

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: I think it would be part of the stdout. So, might want to check for the status. I am not totally sure of the behaviour of the ssh session which executes the command. Is there a way to test?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Yeah that's needed. Not totally sure what the /healthz end point out put be though. Any way to find it?

vamossagar12:
> Also please squash commits

Done. Had a couple of questions.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: You can start locally a server that will always respond 500 and call this code against it.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,207 @@
...
+			a.updateAvailabilityMetrics(false, a.hostLevelMetrics[host])
+		} else {
+			a.updateAvailabilityMetrics(true, a.hostLevelMetrics[host])
+		}
+	}
+}
+
+func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
+	// Check the availability of the cluster by issuing a REST call to /healthz end point
+	_, err := c.CoreV1().RESTClient().Get().AbsPath("/healthz").DoRaw(context.TODO())
```
wojtek-t, line 73: I think what you want is this one:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/cmd/server/testing/testserver.go#L174-L179

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: I was trying to write a test for this with the same ssh code. I have docker for desktop running on my local and wanted to ssh into a locally running kube-api server. The master IP that I am passing is kubernetes.docker.internal. I see that in the configs, there is an option to pass a MasterIP. But, on a locally running kube cluster, masterIP is local, right?

Note that, I am able to run the /healthz end points by setting up a clientset with the same kube config file and the below example you provided works.

vamossagar12: @wojtek-t I was little jammed with work this week but tried some more things today. I set up a docker container running a Flask application inside and deployed it through kubernetes. i also set up a mechanism to ssh into the container. 

Then why I ran it through the test case which invokes this logic, i found that any error in web app is returned in sshResult.stderr. I tried with 404 and got a big html like string (perhaps flask returns that). Note that sshResult.stdout is empty.

In the happy scenario, the sshResult.stdout has the response and sshResult.stderr is empty. 

So, to answer the original question, can I also make this assumption that the presence of stderr indicates the /healthz end point failed and vice versa in the other case? Should I also look to parse the stdout result?

PS Unfortunately, I still couldn't find a way to ssh into the kube-apiserver container. Otherwise I could have validated directly.

vamossagar12: hey @wojtek-t i am presuming you're busy... Just reaching out to ask did you get a chance to look at the comments?

wojtek-t: Sorry for delay.
It would be better to check it directly, but I don't want to block the whole PR on this thing.
So if you could open an issue to check the semantics of ssh and document it better and here in the code add a TODO (linked to this issue) to validate if the logic of relying on stderr non-empty is ok, I'm fine with that (resolving it should be a blocker for enabling the measurement though, but not necessary for merging this PR).

vamossagar12: Thanks @wojtek-t . As suggested, created a new issue to understand ssh client semantics:
https://github.com/kubernetes/perf-tests/issues/1343 and added a TODO here.
Just one thing, you asked me to document it you meant in the newly created issue?

vamossagar12: hey @wojtek-t I have the made the changes to this PR based upon my findings mentioned here:
https://github.com/kubernetes/perf-tests/issues/1343

Plz review whenever you have the time :)

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+			return nil, err
+		}
+		return a.summaries, nil
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) createClusterAvailabilitySummary(probeFrequency int) error {
+	a.clusterLevelMetrics.updateMaxConsecutiveFailuresIfNeeded()
```
wojtek-t, line 149: Instead of doing this here, let's:
- remove it from here (and createMasterAvailabilitySummary)
- let's explicitly add it `updateFailureMetrics`, i.e.;

```
func (a *apiAvailabilityMetrics) updateFailureMetrics() {
	a.numFailures++
	a.consecutiveFailedProbes++
        updateMaxConsecutiveFailuresIfNeeded()
}
```

With that - you don't need to call it anywhere elese.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+		a.summaries = append(a.summaries, measurement.CreateSummary(apiAvailabilityName, "json", content))
+		return nil
+	}
+}
+
+func (a *apiAvailabilityMeasurement) createMasterAvailabilitySummary(probeFrequency int) error {
+	createMasterAvailabilitySummaryErrors := errors.NewErrorList()
+	for _, host := range a.hosts {
+		a.hostLevelMetrics[host].updateMaxConsecutiveFailuresIfNeeded()
+		hostAvailabilityMetricsOutput := a.hostLevelMetrics[host].buildAPIAvailabilityMetricsOutput(probeFrequency,false, host)
```
wojtek-t, line 164: nit: space before 'false"

vamossagar12: Done.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: Done. That makes sense.

vamossagar12:
> Sorry for delay - I was on vacation for last 2.5 weeks.
> 
> I added some comments, but i'm mostly fine with what is now. Please change the PR out of Draft and let's proceed with it.

No worries :) I have made the changes and also removed the PR out of draft.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (a *apiAvailabilityMeasurement) createMasterAvailabilitySummary(probeFrequency int) error {
+	createMasterAvailabilitySummaryErrors := errors.NewErrorList()
+	for _, host := range a.hosts {
+		hostAvailabilityMetricsOutput := a.hostLevelMetrics[host].buildAPIAvailabilityMetricsOutput(probeFrequency, false, host)
+		content, err := util.PrettyPrintJSON(hostAvailabilityMetricsOutput)
+		if err != nil {
+			createMasterAvailabilitySummaryErrors.Append(err)
+		} else{
```
wojtek-t, line 166: Please fix gofmt:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1162/pull-perf-tests-verify-lint/1285548209413820419/build-log.txt

[both here and in the second file]

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,207 @@
...
+		a.updateAvailabilityMetrics(true, a.clusterLevelMetrics)
+	} else {
+		a.updateAvailabilityMetrics(false, a.clusterLevelMetrics)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMasterSupported bool, probeDuration int) error {
+	a.hosts = config.ClusterFramework.GetClusterConfig().MasterIPs
+	if len(a.hosts) < 1 {
+		return goerrors.Errorf("API Server measurement can't execute due to no MasterIPs")
```
wojtek-t, line 86: nit: APIAvailability measurement can't start due to lack of master IPs

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (a *apiAvailabilityMeasurement) stop(SSHToMasterSupported bool, probeFrequency int) error {
+	if !a.isRunning {
+		return nil
+	}
+	close(a.stopCh)
+	a.wg.Wait()
+
+	errList := errors.NewErrorList()
+	err := a.createClusterAvailabilitySummary(probeFrequency)
```
wojtek-t, line 185: Sorry - I should have caught that earlier.

Actually I would really prefer to have a single summary rather than N summaries for every master and cluster separately (see my comment below).

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type apiAvailabilityMetrics struct {
+	numSuccesses               int
+	numFailures                int
+	maxConsecutiveFailedProbes int
+	consecutiveFailedProbes    int
+}
+
+type apiAvailabilityOutput struct {
```
wojtek-t, line 30: Actually, I would prefer the output to be:

```
type apiAvailabilitySummary struct {
  IP                                           string
  AvailabilityPercentage           float32
  LongestUnavailableDuration time.Duration
}

type apiAvailabilityOutput struct {
  ClusterMetrics apiAvailabilitySummary
  MasterMetrics []apiAvailabilitySummary
}
```

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -0,0 +1,81 @@
...
+func (a *apiAvailabilityMetrics) updateFailureMetrics() {
+	a.numFailures++
+	a.consecutiveFailedProbes++
+	a.updateMaxConsecutiveFailuresIfNeeded()
+}
+
+func (a *apiAvailabilityMetrics) updateSuccessMetrics() {
+	a.numSuccesses++
+	if a.consecutiveFailedProbes > 0 {
+		a.updateMaxConsecutiveFailuresIfNeeded()
```
wojtek-t, line 55: this is no longer needed - you already checked that in updateFailureMetrics and here you don't bump numConsecutiveFailures

wojtek-t: So basically, simply unconditionally set consecutiveFailedProbes = 0

clusterloader2/pkg/measurement/common/api_availability_metrics.go
vamossagar12: hi @wojtek-t , slight confusion here. The final output for MasterMetrics is an array here. If you see the output of the measurement that would be returned, all outputs would be consolidated into 1. 

https://github.com/kubernetes/perf-tests/pull/1162/files#diff-769e85d1bdfa456cbbe72288f2ba2ab5R54-R55

Do you think this isn't similar to what you prefer?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
vamossagar12: I ran gofmt -s for these 2 files but this error still got thrown. Is that the right command?

wojtek-t: gofmt -w -s

vamossagar12: thank you.. ran the command and pushed again.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,211 @@
...
+func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.Config, provider provider.Provider) {
+	for _, host := range a.hosts {
+		// SSH and check the health of the host
+		command := fmt.Sprintf("curl -f -s -k %slocalhost:%v/healthz", "https://", 443)
+		sshResult, err := measurementutil.SSH(command, host+":22", provider)
+		if err != nil || sshResult.Code != 0 {
+			a.updateAvailabilityMetrics(false, a.hostLevelMetrics[host])
+		} else {
+			a.updateAvailabilityMetrics(true, a.hostLevelMetrics[host])
+		}
```
wojtek-t, line 68: Let's switch to:
```
availability := err != nil && sshResult.Code != 0
a.udpateAvailabilityMetrics(availability, a.hostLevelMetrics[host])
```

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,211 @@
...
+func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
+	// Check the availability of the cluster by issuing a REST call to /healthz end point
+	result := c.CoreV1().RESTClient().Get().AbsPath("/healthz").Do(context.TODO())
+	status := 0
+	result.StatusCode(&status)
+	if status == 200 {
+		a.updateAvailabilityMetrics(true, a.clusterLevelMetrics)
+	} else {
+		a.updateAvailabilityMetrics(false, a.clusterLevelMetrics)
+	}
```
wojtek-t, line 81: a.updateAvailabilitymetrics(status == 200, a.clusterLevelMetrics)

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+		}
+	}()
+	return nil
+}
+
+// Execute starts the api-server healthz probe end point from start action and
+// collects availability metrics in gather.
+func (a *apiAvailabilityMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+
```
wojtek-t, line 117: nit: remove this empty line

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,211 @@
...
+	if !SSHToMasterSupported {
+		klog.Infof("ssh to master nodes not supported. Measurement would have only Cluster Level Metrics")
+	}
+
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+
```
wojtek-t, line 128: nit: remove this empty line too

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) createClusterAvailabilitySummary(probeFrequency int) error {
+	clusterAvailabilityMetricsOutput := a.clusterLevelMetrics.buildAPIAvailabilityMetricsOutput(probeFrequency, true, "")
+	content, err := util.PrettyPrintJSON(clusterAvailabilityMetricsOutput)
+	if err != nil {
+		return err
+	} else {
```
wojtek-t, line 157:
```
if err != nil {
  return err
}
a.summaries = append...
return nil
```

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: This isn't yet addressed.

clusterloader2/pkg/measurement/common/api_availability_metrics.go
wojtek-t: I care about 'availability' summary as a whole. Yes - I want a single summary, instead of N separate ones.
Alternatively, we may consider two measurements: one for the cluster, second for all masters.

vamossagar12: I see. I might want to make some tweaks to the apiAvailabilityMeasurement struct in that case. Currently it looks like this:

```
type apiAvailabilityMeasurement struct {
	isRunning           bool
	stopCh              chan struct{}
	hosts               []string
	summaries           []measurement.Summary
	clusterLevelMetrics *apiAvailabilityMetrics
	hostLevelMetrics    map[string]*apiAvailabilityMetrics
	wg                  sync.WaitGroup
}
```
So, if I understand correctly, instead of an array of summaries, you want 1 cluster level summary and array of MasterLevel summaries in a single object. I would also need to remove the clusterLevelMetrics and hostLevelMetrics from the above struct.

Regarding 2 separate measurements, do we expose them as separate measurements- which seems slightly confusing to me. With a unified measurement, a user would be able to get all availability metrics in one summary.

wojtek-t: Sorry - I meant "two summaries" - one for the cluster, second for all masters.
But, I think having a single summary is cleaner.

vamossagar12: ok. so in that case, i will make the code changes to have one summary in the format you explained above.

vamossagar12: @wojtek-t I have made the code changes to expose only 1 summary for both ClusterLevel and MasterMetrics. Plz review when you get the chance..

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+type apiAvailabilityMeasurement struct {
+	isRunning             bool
+	stopCh                chan struct{}
+	hosts                 []string
+	summaries             []measurement.Summary
+	clusterLevelMetrics   *apiAvailabilityMetrics
+	hostLevelMetrics      map[string]*apiAvailabilityMetrics
+	apiAvailabilityOutput *apiAvailabilityOutput
```
wojtek-t, line 56: let's remove this field - this concenptually isn't part of the measurement. See also my comments below.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (a *apiAvailabilityMeasurement) stopAndSummarize(SSHToMasterSupported bool, probeFrequency int) error {
+	if !a.isRunning {
+		return nil
+	}
+	close(a.stopCh)
+	a.wg.Wait()
+
+	a.createClusterAvailabilitySummary(probeFrequency)
```
wojtek-t, line 163: Given apiAvailabilityOutput shouldn't be a measurement-level field, let's simply do:
```
output := apiAvailabilityOutput{}
output.createClusterSummary(a.clusterLevelMetrics, probeFrequency)
if SSHToMasterSupported {
  output.createMastersSummary(a.hostLevelMetrics, probeFrequency)
}
```

[which also means removing the createSummary methods above and making the output methods]

vamossagar12:
Thanks @wojtek-t , I made those changes as well.

PR #1168, opened 2020-04-09
ClusterLoader - add kubelet port override option

coltonstapper:
This is intended for clusters whose Kubelets
are not configured with the default 10250 port.


oxddr:
/ok-to-test
/lgtm
/approve

Thanks for the change!

PR #1171, opened 2020-04-10
ClusterLoader - update to Go modules 

coltonstapper:
Update to Kubernetes 1.16.4
Re-generate vendor


clusterloader2/pkg/framework/client/objects.go
```diff
@@ -277,7 +277,7 @@ func PatchObject(dynamicClient dynamic.Interface, namespace string, name string,
 		if err != nil {
 			return fmt.Errorf("creating patch diff error: %v", err)
 		}
-		_, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.MergePatchType, patch, metav1.UpdateOptions{})
+		_, err = dynamicClient.Resource(gvr).Namespace(namespace).Patch(obj.GetName(), types.MergePatchType, patch, metav1.PatchOptions{})
```
coltonstapper, line 280: https://github.com/kubernetes/client-go/blob/kubernetes-1.16.4/dynamic/interface.go#L40
Uses PatchOptions now

clusterloader2/pkg/measurement/common/metrics_for_e2e.go
```diff
@@ -99,7 +99,7 @@ func (*metricsForE2EMeasurement) String() string {
 	return metricsForE2EName
 }
 
-func filterMetrics(m *metrics.MetricsCollection) {
+func filterMetrics(m *metrics.Collection) {
```
coltonstapper, line 102: https://github.pie.apple.com/kubernetes/kubernetes/blob/master/test/e2e/framework/metrics/metrics_grabber.go#L217
Defined as Collection object now

coltonstapper:
I've done real cluster testing and verified the package compiles and I can run random profiles, I've asked in the sig-scalability channel about opinions on the versions and the `legacyIsMasterNode` logic. If anyone has any thoughts feel free to post here.

wojtek-t:
Let's see how the tests will go.

/ok-to-test

coltonstapper:
@oxddr tests pass and got verbal ok from @wojtek-t about copying the now-private legacyIsMasterNode for now. There will be a follow-up PR to move to k8s 1.18 and go 1.13, let me know if there's anything else you'd like to see for this PR/if you want me to squash my commits.

@mm4tt checked in the vendor directory as you suggested so that it's compatible with Travis build.

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -42,7 +41,7 @@ const (
 	e2eSchedulingDurationMetricName           = model.LabelValue(schedulermetric.SchedulerSubsystem + "_e2e_scheduling_duration_seconds_bucket")
 	schedulingAlgorithmDurationMetricName     = model.LabelValue(schedulermetric.SchedulerSubsystem + "_scheduling_algorithm_duration_seconds_bucket")
 	frameworkExtensionPointDurationMetricName = model.LabelValue(schedulermetric.SchedulerSubsystem + "_framework_extension_point_duration_seconds_bucket")
-	schedulingLatencyMetricName               = model.LabelValue(schedulermetric.SchedulerSubsystem + "_" + schedulermetric.DeprecatedSchedulingDurationName)
+	schedulingLatencyMetricName               = model.LabelValue(schedulermetric.SchedulerSubsystem + "_" + schedulermetric.DeprecatedSchedulingLatencyName)
```
coltonstapper, line 44: Confirmed these are the same values: https://github.com/kubernetes/kubernetes/commit/70b80ae28a2770778e811a1121582aa378c577b3#diff-bb6f47a66f0de6549afa80b167861d8eR32

coltonstapper:
@wojtek-t tests pass, any other things you'd like me to do here?

clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
```diff
@@ -27,8 +27,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
-	"k8s.io/kubernetes/pkg/util/system"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	clutil "k8s.io/perf-tests/clusterloader2/pkg/util"
```
oxddr, line 31: nit: cl2util - we refer `clusterloader2` as `cl2`.

coltonstapper:
Thanks @oxddr, addressed your comments

clusterloader2/go.mod
```diff
@@ -0,0 +1,72 @@
+module k8s.io/perf-tests/clusterloader2
+
+go 1.12
+
+replace k8s.io/api => k8s.io/api v0.16.4
```
wojtek-t, line 5: Forgive me my godeps ignorance, but why those are v0...?
Shouldn't those directly reflect the k8s releases, and thus be 1.16.4?

clusterloader2/go.mod
```diff
@@ -0,0 +1,72 @@
...
+	github.com/prometheus/client_model v0.2.0
+	github.com/prometheus/common v0.4.1
+	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.4.0
+	golang.org/x/crypto v0.0.0-20190611184440-5c40567a22f8
+	golang.org/x/sync v0.0.0-20190423024810-112230192c58
+	golang.org/x/time v0.0.0-20181108054448-85acf8d2951c
+	gopkg.in/yaml.v2 v2.2.4
+	k8s.io/api v0.16.4
+	k8s.io/apimachinery v1.16.4
```
wojtek-t, line 67: hmm - why here it's 1.16.4 and in replace section 0.16.4?

Why do we even need replace ones? Can't we put everything here?

clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
```diff
@@ -27,8 +27,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
-	"k8s.io/kubernetes/pkg/util/system"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	cl2 "k8s.io/perf-tests/clusterloader2/pkg/util"
```
wojtek-t, line 31: cl2 is misleading - the previous is also cl2.
Can you please name alias it by "pkgutil"? (we generally use convention to use suffix of path as an alias).

BTW - it might be useful to move this part to a separate PR, since it's unrelated and the risk of breaking is very low.

clusterloader2/go.mod
coltonstapper: k8s.io/api follow the same tagging scheme as client-go: https://github.com/kubernetes/client-go#compatibility-your-code---client-go

clusterloader2/go.mod
coltonstapper: @wojtek-t some of the versioning does the 1.x.y scheme and some do the 0.x.y scheme. Also, starting the initial go mod project is difficult because if you try to "go get" or "go mod download" the bits of the Kubernetes project they try to pull in dependencies from the wrong/mismatched places. There's a lot of discussion in https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597 about this, and there's a script there that many people use to start a project on the desired k8s version.

If I straight try to download the apimachinery 0.16.4 version, for example, we get:
```
% go get k8s.io/apimachinery@v0.16.4
go: finding k8s.io/apimachinery v0.0.0
go: k8s.io/apimachinery@v0.0.0: unknown revision v0.0.0
go: error loading module requirements
%
```

It's possible this is all because we're on 1.12 and these kinds of errors go away once we upgrade to >= Go 1.13

clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
coltonstapper: @wojtek-t it has to be coupled with this PR otherwise the build fails because legacyIsMasterNode is no longer available publicly

clusterloader2/pkg/util/cluster.go
```diff
@@ -158,3 +158,23 @@ func GetMasterIPs(c clientset.Interface, addressType corev1.NodeAddressType) ([]
 	}
 	return ips, nil
 }
+
+// LegacyIsMasterNode returns true if given node is a registered master according
```
wojtek-t, line 162: Can you move introducing this function and replacing the old usage to a separate PR (that will be merged first)?

It's obviously correct, and with go-modules, we've seen different problems in the past in different places, so the risk of rollback is higher.

coltonstapper: Oh merge that PR first, I understand now! Thanks

clusterloader2/go.mod
mm4tt: Thanks for explaining, @coltonstapper.
FTR, there is nothing prohibiting us to upgrade to go1.13 even now. There is WIP to upgrade k/k repo to go1.14, so we could even target 1.14. That being said, I'm totally fine with starting with 1.12 and upgrading later. Thanks for doing this! This is great!

/lgtm
/hold 
Leaving for Wojtek for a final approval and lift of the hold

coltonstapper:
Thanks for the reviews @wojtek-t @mm4tt, rebased on the legacyIsMasterNode and this should be good to go now.

clusterloader2/go.mod
coltonstapper: Thanks @mm4tt I'll be fast-following with the Golang upgrade so we can be getting to a newer version of Kubernetes as well.

wojtek-t: I hate the fact how many dependencies we have (e.g. why we're relying on sample-apiserver - that sounds crazy. Event the dependency on apiserver itself is crazy - we should just depend on client-go and apimachinery and the API itself).
Anyway - that's for another discussion...

wojtek-t:
/hold cancel

Thanks!

coltonstapper:
@wojtek-t we good to merge?

coltonstapper:
Oh cool I'll rebase and remove the old metric

wojtek-t:
Yeah - I don't know why it didn't merge after I cancelled hold on Friday.

Issue #1169, opened 2020-04-09
ClusterLoader enhancement: use existing namespaces in a test


yuanchen8911:
clusterloader automatically creates and manages namespaces, but we need to use existing namespaces in some testing scenarios. For example, an internal controller has to create a bunch of namespaces with quota limits first for any tests in our environment.  There might be other use cases for using pre-created namespaces too. 

We have implemented an enhancement to clusterloader in support of using existing namespaces by adding a new keyword "_namespacesPrefix_" in the testconfig file.  The naming of namespaces must still follow the convention [prefix-]-1, -2, 3.... and managed by clusterloader. If a prefix of namespaces is defined and the corresponding namespaces have been created, clusterloader will skip the namespace creation. Would it be a useful feature?  If it is, we would like to submit a PR. 

Thanks for your feedback!

wojtek-t:
At the high level I agree it would be useful. The case with "prepared namespaces" is definitely a valid one. Another usecase that I see is actually running multiple tests against the same set of namespaces.

So yes - I'm will to accept the feature at the high level.

If we get to details, I would like to avoid introducing a dedicated field for that.
Instead, I think that we should:
- deprecate the existing AutomanagedNamespaces field:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/api/types.go#L45
 [we've never figured out versioning policy, but we shouldn't delete immediately for sure, at least to let us update all the tests - @mm4tt ]
- introduce a struct to define namespaces; I need to think more about it, but the first attempt would be something like:

```
type NamespacesConfig struct {
  // optional, if set CL will use it instead its own 
  Prefix string
  // whether to adopt existing namespaces if they exist
  Adopt bool
[the reason I would like to make it explicit is to ensure that someone won't mess up something inconcsciously - if Adopt will be set to false (default), existing namespaces will result in error and should stop the test]
  // Number of namespaces that should exist after creation phase.
  Number int32
}
```

As mentioned above, we may want to tune/change that, but that's my first attempt.

Apart from that, the feature makes sense to me.

yuanchen8911:
@wojtek-t Yes, the introduction of a namespace structure is certainly a better way.  So the logic would look like.

1.  Check whether the namespaces with names $Prefix-1, -2, -3, ...  exist.

2.  If the namespaces exist,  report an error and stop the test if _Adopt_ is not set  or use the existing namespaces if _Adopt_ is set.  Otherwise,  create the namespaces.

[...]

PR #1173, opened 2020-04-14
clusterloader: enhance namespace configurations

yuanchen8911:
This PR is to separate namespace specific configurations from cluster level configurations (e.g., kubeconfig, master-ip). The cluster configurations belong to config.ClusterConfig and are specified as the command line options while namespace configuration is a part of api.Config and deigned in the testconfig.yaml file.

1.  A new structure _NamespaceConfig_ is defined for namespace parameters as a part of api.Config and the values are specified in the testconfig yaml file.  If a parameter is not specified, a default value is used.
2. A new parameter "_Prefix_" is added to specify the prefix of auto-created namespaces. Clusterloader uses the specified prefix if set otherwise uses the default "_test-\<random string\>_".
3. Add two files _api.default.go_ and _api.validate.go_  for setting default parameters and checking the validness of namespace parameters. 
4. _"automanagedNamespaces"_  in testconfig file and _"delete-stale-namespaces"_ and _"delete-automananged-namespaces"_ in the command line can still be used with deprecate warnings. 

New structure for namespace configuration
```
// NamespaceConfig defines namespace parameters.
type NamespaceConfig struct {
        Number int32  `json: number` 
        Prefix string  `json: prefix`
        DeleteStaleNamespaces *bool `json.deleteStaleNamespaces`
        DeleteAutomanagedNamespaces *bool `json.deleteAutomanagedNamespaces`
}
```
An example namespace configuration in the testconfig file.
```
namespace:
   number: 1
   namePrefix: "test"
   EnableExistingNamespaces: false
   DeleteStaleNamespaces: false
```
An issue https://github.com/kubernetes/perf-tests/issues/1169 was opened and discussed with @wojtek-t. The support of pre-created existing namespaces will be another PR after this PR is approved and merged.

### DIFF
--- a/clusterloader2/api/default.go
+++ b/clusterloader2/api/default.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+// SetDefaults set the default configuration parameters.
+func (conf *Config) SetDefaults() {
+	conf.Namespace.SetDefaults()
+
+	// TODO: remove after deprecated automanagedNamespaces is disabled.
+	if conf.Namespace.Number == 1 && conf.AutomanagedNamespaces > 1 {
+		conf.Namespace.Number = conf.AutomanagedNamespaces
+	}
+}
+
+// SetDefaults specifies the default values for namespace parameters.
+func (ns *NamespaceConfig) SetDefaults() {
+	if ns.Number == 0 {
+		ns.Number = 1
+	}
+
+	if ns.Prefix == "" {
+		ns.Prefix = fmt.Sprintf("test-%s", util.RandomDNS1123String(6))
+	}
+
+	defaultDeleteStaleNS := false
+	if ns.DeleteStaleNamespaces == nil {
+		ns.DeleteStaleNamespaces = &defaultDeleteStaleNS
+	}
+
+	defaultDeleteAutoNS := true
+	if ns.DeleteAutomanagedNamespaces == nil {
+		ns.DeleteAutomanagedNamespaces = &defaultDeleteAutoNS
+	}
+}
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -41,8 +41,11 @@ type TestScenario struct {
 type Config struct {
 	// Name of the test case.
 	Name string `json: name`
-	// AutomanagedNamespaces is a number of automanaged namespaces.
-	AutomanagedNamespaces int32 `json: automanagedNamespaces`
+	// Deprecated: a number of automanaged namespaces.
+	// Use Namespace.Number instead.
+	AutomanagedNamespaces int32 `json: automanagedNamespaces,omitempty`
+	// Namespace is a structure for namespace configuration.
+	Namespace NamespaceConfig `json: namespace`
 	// Steps is a sequence of test steps executed in serial.
 	Steps []Step `json: steps`
 	// TuningSets is a collection of tuning sets that can be used by steps.
@@ -104,6 +107,19 @@ type Object struct {
 // ListUnknownObjectOptions struct specifies options for listing unknown objects.
 type ListUnknownObjectOptions struct {
 	LabelSelector *metav1.LabelSelector `json: labelSelector`
+}
+
+// NamespaceConfig defines namespace parameters.
+type NamespaceConfig struct {
+	// Number is a number of automanaged namespaces.
+	Number int32 `json: number,omitempty`
+	// NamePrefix is the name prefix of automanaged namespaces.
+	// It's optional, if set CL will use it, otherwise generate one with random string.
+	Prefix string `json: prefix,omitempty`
+	// DeleteStaleNamespaces specifies whether or not delete stale namepaces
+	DeleteStaleNamespaces *bool `json: deleteStaleNamespaces,omitempty`
+	// DeleteAutomanangedNamespaces specifies whether or not delete namepaces after a test
+	DeleteAutomanagedNamespaces *bool `json: deleteAutomanagedNamespaces,omitempty`
 }
 
 // NamespaceRange specifies the range of namespaces [Min, Max].
--- a/clusterloader2/api/validation.go
+++ b/clusterloader2/api/validation.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+)
+
+// Validate checks and verifies the configuration parameters.
+func (conf *Config) Validate() *errors.ErrorList {
+	return conf.Namespace.Validate()
+}
+
+// Validate checks and verifies the namespace parameters.
+func (ns *NamespaceConfig) Validate() *errors.ErrorList {
+	errList := errors.NewErrorList()
+	if ns.Number <= 0 {
+		errList.Append(fmt.Errorf("number of namespaces: %d was less than 1", ns.Number))
+	}
+
+	if errList.IsEmpty() {
+		return nil
+	}
+
+	return errList
+}
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -66,8 +66,10 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdCertificatePath, "etcd-certificate", "ETCD_CERTIFICATE", "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt", "Path to the etcd certificate on the master machine")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdKeyPath, "etcd-key", "ETCD_KEY", "/etc/srv/kubernetes/pki/etcd-apiserver-server.key", "Path to the etcd key on the master machine")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
-	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "Whether to delete all stale namespaces before the test execution.")
-	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteAutomanagedNamespaces, "delete-automanaged-namespaces", "DELETE_AUTOMANAGED_NAMESPACES", true, "Whether to delete all automanaged namespaces after the test execution.")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "DEPRECATED: Whether to delete all stale namespaces before the test execution.")
+	flags.MarkDeprecated("delete-stale-namespaces", "specify deleteStaleNamespaces in testconfig file instead.")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteAutomanagedNamespaces, "delete-automanaged-namespaces", "DELETE_AUTOMANAGED_NAMESPACES", true, "DEPRECATED: Whether to delete all automanaged namespaces after the test execution.")
+	flags.MarkDeprecated("delete-automanaged-namespaces", "specify deleteAutomanagedNamespaces in testconfig file instead.")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.MasterName, "mastername", "MASTER_NAME", "", "Name of the masternode")
 	// TODO(#595): Change the name of the MASTER_IP and MASTER_INTERNAL_IP flags and vars to plural
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
--- a/clusterloader2/pkg/test/simple_test_executor.go
+++ b/clusterloader2/pkg/test/simple_test_executor.go
@@ -52,9 +52,10 @@ func createSimpleTestExecutor() TestExecutor {
 
 // ExecuteTest executes test based on provided configuration.
 func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *errors.ErrorList {
-	ctx.GetClusterFramework().SetAutomanagedNamespacePrefix(fmt.Sprintf("test-%s", util.RandomDNS1123String(6)))
+	ctx.GetClusterFramework().SetAutomanagedNamespacePrefix(conf.Namespace.Prefix)
 	klog.Infof("AutomanagedNamespacePrefix: %s", ctx.GetClusterFramework().GetAutomanagedNamespacePrefix())
-	defer cleanupResources(ctx)
+
+	defer cleanupResources(ctx, conf)
 	ctx.GetTuningSetFactory().Init(conf.TuningSets)
 
 	stopCh := make(chan struct{})
@@ -103,7 +104,7 @@ func (ste *simpleTestExecutor) ExecuteTestSteps(ctx Context, conf *api.Config) *
 	if len(automanagedNamespacesList) > 0 {
 		return errors.NewErrorList(fmt.Errorf("pre-existing automanaged namespaces found"))
 	}
-	var deleteStaleNS = ctx.GetClusterFramework().GetClusterConfig().DeleteStaleNamespaces
+	var deleteStaleNS = *conf.Namespace.DeleteStaleNamespaces
 	if len(staleNamespaces) > 0 && deleteStaleNS {
 		klog.Warning("stale automanaged namespaces found")
 		if errList := ctx.GetClusterFramework().DeleteNamespaces(staleNamespaces); !errList.IsEmpty() {
@@ -111,7 +112,7 @@ func (ste *simpleTestExecutor) ExecuteTestSteps(ctx Context, conf *api.Config) *
 		}
 	}
 
-	err = ctx.GetClusterFramework().CreateAutomanagedNamespaces(int(conf.AutomanagedNamespaces))
+	err = ctx.GetClusterFramework().CreateAutomanagedNamespaces(int(conf.Namespace.Number))
 	if err != nil {
 		return errors.NewErrorList(fmt.Errorf("automanaged namespaces creation failed: %v", err))
 	}
@@ -371,10 +372,10 @@ func isErrsCritical(*errors.ErrorList) bool {
 	return false
 }
 
-func cleanupResources(ctx Context) {
+func cleanupResources(ctx Context, conf *api.Config) {
 	cleanupStartTime := time.Now()
 	ctx.GetMeasurementManager().Dispose()
-	if ctx.GetClusterFramework().GetClusterConfig().DeleteAutomanagedNamespaces {
+	if *conf.Namespace.DeleteAutomanagedNamespaces {
 		if errList := ctx.GetClusterFramework().DeleteAutomanagedNamespaces(); !errList.IsEmpty() {
 			klog.Errorf("Resource cleanup error: %v", errList.String())
 			return
--- a/clusterloader2/pkg/test/test.go
+++ b/clusterloader2/pkg/test/test.go
@@ -61,5 +61,19 @@ func RunTest(clusterFramework, prometheusFramework *framework.Framework, cluster
 	if err != nil {
 		return errors.NewErrorList(fmt.Errorf("config reading error: %v", err))
 	}
+
+	// TODO: remove them after the deprecated command options are removed.
+	if testConfig.Namespace.DeleteStaleNamespaces == nil {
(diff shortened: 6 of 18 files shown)


clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -42,7 +42,11 @@
 {{$RESTART_COUNT_THRESHOLD_OVERRIDES:= DefaultParam .RESTART_COUNT_THRESHOLD_OVERRIDES ""}}
 
 name: access-tokens
-automanagedNamespaces: {{$namespaces}}
+namespace:
+  count: {{$namespaces}}
+  prefix:
+  deleteStaleNamespaces: false
+  deleteAutomanagedNamespaces: true
```
wojtek-t, line 49: I really don't like the fact we need to set those values in every test - we need to smartly default them instead and rely on the defaulting whenever it is possible.

clusterloader2/api/types.go
```diff
@@ -41,8 +41,8 @@ type TestScenario struct {
 type Config struct {
 	// Name of the test case.
 	Name string `json: name`
-	// AutomanagedNamespaces is a number of automanaged namespaces.
-	AutomanagedNamespaces int32 `json: automanagedNamespaces`
+	// Namespace is a structure for namespace configuration.
+	Namespace *NamespaceConfig `json: namespace`
```
wojtek-t, line 45: We can't do that in a single step. We should:
- add this field and mark the AutomanagedNamespaces field as deprecated (though still respect it somehow)
- allow people for doing the transition
- only in some time get rid of this field

clusterloader2/api/types.go
```diff
@@ -106,6 +106,22 @@ type ListUnknownObjectOptions struct {
...
+	// NamePrefix is the name prefix of automanaged namespaces.
+	// It's optional, if set CL will use it, otherwise generate one with random string.
+	Prefix string `json: prefix`
+	// DeleteStaleNamespaces specifies whether or not delete stale namepaces
+	DeleteStaleNamespaces bool `json: deleteStaleNamespaces`
+	// DeleteAutomanangedNamespaces specifies whether or not delete namepaces after a test
+	DeleteAutomanagedNamespaces bool `json: deleteAutomanagedNamespaces`
+	// EnableExistingNamespace specifies whether to use existing namespaces
+	// if they exist  (TO DO)
+	EnableExistingNamespaces bool `json: enableExistingNamespaces`
```
wojtek-t, line 122: Let's not add fields that are not yet supported.

clusterloader2/cmd/clusterloader.go
```diff
@@ -66,8 +66,6 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdCertificatePath, "etcd-certificate", "ETCD_CERTIFICATE", "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt", "Path to the etcd certificate on the master machine")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdKeyPath, "etcd-key", "ETCD_KEY", "/etc/srv/kubernetes/pki/etcd-apiserver-server.key", "Path to the etcd key on the master machine")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
-	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "Whether to delete all stale namespaces before the test execution.")
```
wojtek-t, line 69: I'm probably fine with deleting those. But it would be great if we could provide a defaulting in a way that would accept the CL2_DELETE_STALE_NAMESPACES env var as overwrite for it.
Didin't think about it much - @mm4tt - do you think it would be possible?

clusterloader2/testing/access-tokens/config.yaml
yuanchen8911: OK, will set the following default values. 
```
count=1
prefix= ""
deleteStaleNamespaces=false
deleteAutomanagedNamespace=true
```

clusterloader2/api/types.go
yuanchen8911: sure,  we'll keep it as as deprecated.

clusterloader2/api/types.go
yuanchen8911: will delete it.

clusterloader2/cmd/clusterloader.go
yuanchen8911: To maintain back compatibility,  it might be good to keep it for the time being. I'll undo the deletion.

clusterloader2/api/types.go
yuanchen8911: The following default values are used if a parameter is not specified in either new or old way. 
```
     count: 1
     prefix: ""
     deleteStaleNamespaces: false
     deleteAutomananagedNamespaces: true
```

clusterloader2/api/types.go
yuanchen8911: deleted it.

clusterloader2/cmd/clusterloader.go
yuanchen8911: To maintain backward compatibility, the current command line options and environment variables "delete-stale-namespaces" and "delete-automanaged-namespaces" are kept with deprecated warnings.  automananagedNamespaces can still be used to specify the number of namespaces in testconfig file. If a parameter is not specified, a default value is used. If a parameter is specified in both new and old ways, the value using the new way is used.

clusterloader2/api/types.go
yuanchen8911: AutomanangedNamespaces is kept.  If the number of namespaces is specified by both AutoMananagedNamespaces and the new structure, the value of new structure will be used. If neither defines the parameter, a default value of 1 will be used.  The new code also checks the value and reports an error if it's less than 1.

yuanchen8911:
@wojtek-t Thanks for the very helpful feedback! Here is a summary of the updated version. I hope it addressed most of your comments and concerns. 

1. **Backward compatibility**: the use of command line options (and environment variables) to specify "_delete-stale-namespace_" and "_delete-automanaged-namespaces_" and the use of automanagedNamespaces to define the number of namespaces in testconfig file can still be used with deprecated warnings or flag. So scripts and configuration files for the current version should continue to work without any changes.

2. **Default values**: defined the following default value if a parameter is not defined either way.
```
    count: 1
     prefix: ""
     deleteStaleNamespaces: false
     deleteAutomananagedNamespaces: true
```

3. If a parameter is specified in both old and new ways, the value defined using the new way is used.

4. Check the value of automanagedNamespaces and report an error if the specified value for it is less than 1.

clusterloader2/api/types.go
```diff
@@ -106,6 +109,19 @@ type ListUnknownObjectOptions struct {
 	LabelSelector *metav1.LabelSelector `json: labelSelector`
 }
 
+// NamespaceConfig defines namespace parameters.
+type NamespaceConfig struct {
+	// Count is a number of automanaged namespaces.
+	Count *int32 `json: count,omitempty`
```
wojtek-t, line 115: Let's remove the pointer here.
Basically, "0" is the bad value (the test doesn't make sense for count=0). So we don't really need to distinguish it from nil - we can simply default if set to 0.

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -318,6 +320,52 @@ func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, na
...
+		conf.Namespace.Count = conf.AutomanagedNamespaces
+	}
+
+	var defaultNamespaces int32 = 1
+	if conf.Namespace.Count == nil {
+		conf.Namespace.Count = &defaultNamespaces
+	}
+	if *conf.Namespace.Count <= 0 {
+		return errors.NewErrorList(fmt.Errorf("number of namespaecs: %d was less than 1", *conf.Namespace.Count))
+	}
```
wojtek-t, line 338: This technically has nothing to do with executor.
So far we never have a need for it, but I would prefer to:
- in api/ directory introduce the default.go file
- in that file define functions like:
```
func (c *Config) SetDefaults() {
  c.Namespace.SetDefaults()
}

func (n *NamespaceConfig) SetDefaults() {
  // defaulting logic that you have here
}
```

On top of that, introduce a `validation.go` where you would introduce define:
```
func (c *Config) Validate() errors.ErrorList {
  return c.Namespace.Validate()
}

func (n *NamespaceConfig) Validate() errors.ErrorList {
  // return what you have here
}
```

And call defaulting and validation here:
https://github.com/kubernetes/perf-tests/blob/d9bdf46ed8414e87b3c50d49f14a64e9e8970051/clusterloader2/pkg/test/test.go#L63

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -318,6 +320,52 @@ func (ste *simpleTestExecutor) ExecuteObject(ctx Context, object *api.Object, na
...
+
+	// For backwarad compatibility with deprecated options "delete-stale-namespaces" and "delete-automananged-namespaces".
+	// Remove in the future versions
```
wojtek-t, line 341: We don't have any formal deprecation policy, so let's remove it (it's not perfect, but it will be much cleaner).

clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -42,7 +42,11 @@
 {{$RESTART_COUNT_THRESHOLD_OVERRIDES:= DefaultParam .RESTART_COUNT_THRESHOLD_OVERRIDES ""}}
 
 name: access-tokens
-automanagedNamespaces: {{$namespaces}}
+namespace:
+  count: {{$namespaces}}
+  prefix:
```
wojtek-t, line 47: With the defaulting you added, neither of those below should be needed (you only need to specify count).

clusterloader2/pkg/test/simple_test_executor.go
yuanchen8911: > This technically has nothing to do with executor.
> So far we never have a need for it, but I would prefer to:
> 
> * in api/ directory introduce the default.go file
> * in that file define functions like:
> 
> ```
> func (c *Config) SetDefaults() {
>   c.Namespace.SetDefaults()
> }
> 
> func (n *NamespaceConfig) SetDefaults() {
>   // defaulting logic that you have here
> }
> ```
> 
> On top of that, introduce a `validation.go` where you would introduce define:
> 
> ```
> func (c *Config) Validate() errors.ErrorList {
>   return c.Namespace.Validate()
> }
> 
> func (n *NamespaceConfig) Validate() errors.ErrorList {
>   // return what you have here
> }
> ```
> 
This would certainly lead to a neater implementations, but I have two comments/questions. 

1. Supporting the deprecated command line options _delete-stale-namespaces_ would be difficult because SetDefaults does not have access to config.ClusterConfig.DeleteStaleNamespaces. I'd suggest we just delete this command flag.

2.  To continue support automanagedNamespaces in the testconfig file would be tricky too. Without using *int32, when Count's value is 1 (the default), it's impossible to tell if it's explicitly set or the default value is used. In practice, a user should only specify the number in one or the other way, but not both.  In case both are specified (not equal to 1), we'll use the Count's value.
```
@wojtek-t

File: clusterloader2/testing/access-tokens/config.yaml (hunk 8 above)
Comment (yuanchen8911) [reply clusterloader2/testing/access-tokens/config.yaml]: done

File: clusterloader2/pkg/test/simple_test_executor.go (hunk 7 above)
Comment (yuanchen8911) [reply clusterloader2/pkg/test/simple_test_executor.go]: done,  new implementation in api.default.go and api.validate.go as suggested.

File: clusterloader2/pkg/test/simple_test_executor.go (hunk 6 above)
Comment (yuanchen8911) [reply clusterloader2/pkg/test/simple_test_executor.go]: done

**Comment by yuanchen8911** [comment]
@wojtek-t Based on your feedback,  I've made the following changes. 

1. Added two files default.go and validate.go for setting the default values and checking the validness of the parameters for namespaces respectively.
2. Use _int32_ and _bool_  instead of _*int32_ and _*bool_ in NamespaceConfig.
3. To maintain the compatibility,  _"automanagedNamespaces"_ can still be used to specify the number of namespaces in testconfig file, but the use of command line options to specify " _delete-stale-namespaces_" and "_delete-automanaged-namespaces_" are removed  because it's hard to implement it in a neat way.  If you have a good idea, please let me know.
4. Removed additional namespace parameters in config*.yaml as you suggested.

File: clusterloader2/testing/access-tokens/config.yaml (hunk 1 above)
Comment (yuanchen8911) [reply clusterloader2/testing/access-tokens/config.yaml]: removed

File: clusterloader2/api/default.go [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+package api
+
+import (
+	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+// SetDefaults set the default configuration parameters.
+func (conf *Config) SetDefaults() {
+	//To support deprecated "automanangedNamespaces" in testconfig
```
wojtek-t, line 26: Let's remove this comment.

clusterloader2/api/default.go
```diff
@@ -0,0 +1,38 @@
...
+	conf.AutomanagedNamespaces = 1
```
wojtek-t, line 27: I don't think it's needed. Below we default `namespace.number`, so that one will be used anyway..
I think we should remove it.

clusterloader2/api/default.go
```diff
@@ -0,0 +1,38 @@
...
+
+	conf.Namespace.SetDefaults()
+}
+
+// SetDefaults specifies the default values for namespace parameters.
+func (ns *NamespaceConfig) SetDefaults() {
+	ns.Number = 1
```
wojtek-t, line 34: That is incorrect. You should default to 1, only if it was previously set to 0.

clusterloader2/api/default.go
```diff
@@ -0,0 +1,38 @@
...
+	ns.Prefix = fmt.Sprintf("test-%s", util.RandomDNS1123String(6))
```
wojtek-t, line 35: This also is incorrect - this should be conditional

clusterloader2/api/default.go
```diff
@@ -0,0 +1,38 @@
...
+	ns.DeleteStaleNamespaces = false
```
wojtek-t, line 36: Same here. And this has to be a pointer, because otherwise you have no way to distinguish between unset value and explicitly set to false.
[It's also a reason  why I previously commented about removing a pointer only for Number field.]

clusterloader2/api/types.go
```diff
@@ -106,6 +109,19 @@ type ListUnknownObjectOptions struct {
...
 
+// NamespaceConfig defines namespace parameters.
+type NamespaceConfig struct {
+	// Number is a number of automanaged namespaces.
+	Number int32 `json: number,omitempty`
+	// NamePrefix is the name prefix of automanaged namespaces.
+	// It's optional, if set CL will use it, otherwise generate one with random string.
+	Prefix string `json: prefix,omitempty`
+	// DeleteStaleNamespaces specifies whether or not delete stale namepaces
+	DeleteStaleNamespaces bool `json: deleteStaleNamespaces,omitempty`
```
wojtek-t, line 120: As mentioned above - these bool fields should be pointers.

clusterloader2/api/validation.go
```diff
@@ -0,0 +1,54 @@
...
+		return err
+	}
+
+	if conf.AutomanagedNamespaces > 1 && conf.Namespace.Number > 1 {
+		return errors.NewErrorList(fmt.Errorf("number of namespaces was set twice."))
+	}
+
+	// To support deprecated automanagedNamespaces.
+	if conf.AutomanagedNamespaces > 1 {
+		conf.Namespace.Number = conf.AutomanagedNamespaces
```
wojtek-t, line 41: Validation should not change anything.

clusterloader2/pkg/config/codec.go
```diff
@@ -35,12 +35,11 @@ var (
 )
 
 // convertToConfig converts array of bytes into test config.
-func convertToConfig(raw []byte) (*api.Config, error) {
-	var config api.Config
-	if err := yaml.NewYAMLOrJSONDecoder(bytes.NewBuffer(raw), 4096).Decode(&config); err != nil {
-		return nil, fmt.Errorf("decoding failed: %v", err)
+func convertToConfig(config *api.Config, raw []byte) error {
```
wojtek-t, line 38: Why did you change this?

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -52,9 +52,11 @@ func createSimpleTestExecutor() TestExecutor {
 
 // ExecuteTest executes test based on provided configuration.
 func (ste *simpleTestExecutor) ExecuteTest(ctx Context, conf *api.Config) *errors.ErrorList {
-	ctx.GetClusterFramework().SetAutomanagedNamespacePrefix(fmt.Sprintf("test-%s", util.RandomDNS1123String(6)))
+
```
wojtek-t, line 55: remove this empty line

clusterloader2/pkg/test/test.go
```diff
@@ -57,9 +58,23 @@ func RunTest(clusterFramework, prometheusFramework *framework.Framework, cluster
 	}
 	ctx := CreateContext(clusterLoaderConfig, clusterFramework, prometheusFramework, state.NewState(), mapping)
 	testConfigFilename := filepath.Base(clusterLoaderConfig.TestScenario.ConfigPath)
-	testConfig, err := ctx.GetTemplateProvider().TemplateToConfig(testConfigFilename, mapping)
+
+	// Set the default values for configuration.
+	var testConfig api.Config
+	testConfig.SetDefaults()
```
wojtek-t, line 64: I'm not a fan of this pattern. You should leave what was here as it was and simply do:
```
testConfig.SetDetauls()
if err := testConfig.Validate(); err != nil {
  // ....
}
```

clusterloader2/pkg/test/test.go
```diff
@@ -57,9 +58,23 @@ func RunTest(clusterFramework, prometheusFramework *framework.Framework, cluster
...
+	// Support the deprecated command line options
```
wojtek-t, line 65: please add a TODO to remove them.

clusterloader2/api/default.go
yuanchen8911: @wojtek-t  I'm not sure about this one. My implementation is to set the defaults before parsing the testconfig yaml file. The default value will be overridden if it's set in testconfig.
I think it would be a neater implementation. If we set the defaults after unmarshalling the testconfig, we have to check if "0" or false is set or unset using pointers.  The same to the delete namespace flags. I already tested it with different cases and worked as expected. 

https://github.com/yuanchen8911/perf-tests/blob/master/clusterloader2/pkg/test/test.go
```
# Directory Structure

- profile: clusterloader template files for different test cases
- profile/override: test case-specific parameter files
- script: scirpts to run tests and manage kubernetes resources
- config: kubelets config files
```

clusterloader2/api/default.go
yuanchen8911: Same as the above comment, the defaults are set first. If a filed is specified in the yaml file and the default value will be overridden.

clusterloader2/api/default.go
yuanchen8911: Same here. The defaults are set using the command line flags or default values before parsing the yaml file and will be overridden by the values in testconfig file. This is an simple and clean implementation in my opinion.

clusterloader2/pkg/config/codec.go
yuanchen8911: @wojtek-t  Because the default values are set before parsing the testcong yaml file. So  I'll need to define testConfig in test/test.go and pass the pointer to TemplateToConfig in config/template_provider.go. The current version is to define testConfig in config/codec.go and return the pointer to it.

clusterloader2/api/default.go
wojtek-t: I comment on that too - I'm not a fan of this approach. I much more prefer the more standard approach (and what k8s itself is doing internally), which is what I suggested:
- read the config
- for fields that are unset, propagate with defaults

clusterloader2/pkg/test/test.go
yuanchen8911: The implementation logic is 

1. set defaults (1 for namespaces,   delete ns flags from command line defaults or settings
2. decode yaml test config
3.  verify the validness of the values

So testConfig needs to be defined  first.

wojtek-t: As mentioned in: https://github.com/kubernetes/perf-tests/pull/1173#discussion_r410413673
I don't think what you did is what we really want. We don't want to come up with different patterns for things that were already solved in Kubernetes itself.

clusterloader2/api/default.go
yuanchen8911: My current implementation is to set the defaults first and override them if set later.

yuanchen8911: OK, I'll make the changes.

yuanchen8911:
@wojtek-t  OK, I'll make the changes by decoding the yaml file first and setting the defaults if not set.

yuanchen8911:
@wojtek-t The implementation logic was modified as you suggested. Please review. Thanks!
1. Parse testconfig
2. Set defaults for unset parameters

To support current command line options delete-stale-namepsace and delete-automanaged-namespaces, their values are as the default values before call SetDefault. Add "TODO:  remove them in the future versions."

clusterloader2/api/default.go
yuanchen8911: done

clusterloader2/pkg/test/test.go
yuanchen8911: changed as suggested. Please review. Thanks.

clusterloader2/pkg/test/test.go
yuanchen8911: Done

clusterloader2/api/default.go
yuanchen8911: made the change

clusterloader2/api/types.go
yuanchen8911: Changed

clusterloader2/api/validation.go
yuanchen8911: Changed it.

clusterloader2/pkg/config/codec.go
yuanchen8911: Changed it to parse confit first and then set defaults for unset Parma.

clusterloader2/pkg/test/simple_test_executor.go
yuanchen8911: Deleted it

clusterloader2/api/validation.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Validate checks and verifies the configuration parameters.
+func (conf *Config) Validate() *errors.ErrorList {
+	return conf.Namespace.Validate()
+}
+
+// Validate checks and verifies the namespace parameters.
+func (ns *NamespaceConfig) Validate() *errors.ErrorList {
+	if ns.Number <= 0 {
+		return errors.NewErrorList(fmt.Errorf("number of namespaces: %d was less than 1", ns.Number))
```
wojtek-t, line 32: We would like to return all errors. I know in the current form it doesn't matter. But for future extensibility, let's change:
```
result := errors.NewErrorList()
if ns.Number <= 0 {
   // Append
}

if result.IsEmpty() {
  return nil
}
return result
}
```

wojtek-t:
/lgtm
/approve

Thanks!

clusterloader2/api/validation.go
yuanchen8911: @wojtek-t  Thank, the following change was made as suggested. 

```
        if ns.Number <= 0 {
                errList.Append(fmt.Errorf("number of namespaces: %d was less than 1", ns.Number))
        }

        if errList.IsEmpty() {
                return nil
        }

        return errList

```

Issue #1110, opened 2020-03-05
Investigate making containerd an official scalability CRI recommendation


mm4tt:
We've got a lot of premises that containerd performance is better comparing to docker. E.g. 

1. Pod throughput

    **Docker**
    http://perf-dash.k8s.io/#/?jobname=docker-node-throughput&metriccategoryname=E2E&metricname=PodStartup&Metric=pod_startup
    ![2aT6DcSAOYG](https://user-images.githubusercontent.com/2604887/75969747-ace2b000-5ecf-11ea-9108-bcf9961c88cb.png)

    **Containerd**
    http://perf-dash.k8s.io/#/?jobname=containerd-node-throughput&metriccategoryname=E2E&metricname=PodStartup&Metric=pod_startup
    ![W4MqW0DqV0C](https://user-images.githubusercontent.com/2604887/75969755-aeac7380-5ecf-11ea-9165-95956b8b3861.png)

2. Recent findings when working on merging density test into load test - https://github.com/kubernetes/perf-tests/issues/1007

We should investigate it more thoroughly, e.g. see what would be the impact of switching our scale tests to containerd. If those experiments turn out successful, we should consider making containerd an official SIG Scalability recommendation.

mm4tt:
/remove-lifecycle rotten
We're still working on that

wojtek-t:
/remove-lifecycle stale

That kind-of already happened, because we updated the default in startup scripts.
We should just clean up stuff a bit (our test configs - containerd is already a default) and then update the documentation.

jprzychodzen:
Pending items

* Remove containerd specific present `preset-e2e-scalability-containerd`. It can be removed when branch 1.20 will be removed from supported branches. It will be about ~8 months from now.
* Update sig-scalability documentation. Expalining ContainerD vs Docker around https://github.com/kubernetes/community/tree/master/sig-scalability should cover this.

wojtek-t:
FIxed.

PR #1175, opened 2020-04-16
[ClusterLoader2] Enable custom timeouts for probes

jprzychodzen:
Allow probes measurement to use custom timeout for pod readiness. Required to adapt containerd for tests

Ref. #1110 

/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/measurement/common/probes/probes.go
+++ b/clusterloader2/pkg/measurement/common/probes/probes.go
@@ -38,7 +38,8 @@ const (
 	manifestsPathPrefix = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/"
 
 	checkProbesReadyInterval = 15 * time.Second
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -37,8 +37,8 @@ const (
 
 	manifestsPathPrefix = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/common/probes/manifests/"
 
-	checkProbesReadyInterval = 15 * time.Second
-	checkProbesReadyTimeout  = 5 * time.Minute
+	defaultCheckProbesReadyInterval = 15 * time.Second
```
wojtek-t, line 40: Let's not expose the interval as a var - timeout seems to be enough for now.

jprzychodzen: Make sense. We don't need it and we can easily expose it if needed.

PR #1178, opened 2020-04-17
Pass proper GCP project with snapshots storage to PrometheusController

tosi3k:
When using pod-utils for SIG scalability test jobs, clusterloader was attempting to snapshot Prometheus' PDs to Prow cluster project. The proposed flag `experimental-snapshot-project` (which defaults to `PROJECT` environment variable value which is set [here](https://github.com/kubernetes/test-infra/blob/5f62df0ca89fd93fdcc3f1a7a76019bb115ffa47/kubetest/main.go#L775)) is added to be utilized by gcloud command that performs snapshotting.

/sig scalability

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -130,8 +130,14 @@ func (pc *PrometheusController) snapshotPrometheusDiskIfEnabled() error {
 
 func (pc *PrometheusController) trySnapshotPrometheusDisk(pdName, snapshotName, zone string) error {
 	klog.Info("Trying to snapshot Prometheus' persistent disk...")
-	klog.Infof("Snapshotting PD %q into snapshot %q in zone %q", pdName, snapshotName, zone)
-	cmd := exec.Command("gcloud", "compute", "disks", "snapshot", pdName, "--zone", zone, "--snapshot-names", snapshotName)
+	project := pc.clusterLoaderConfig.PrometheusConfig.SnapshotProject
+	if project == "" {
+		// This should never happen when run from kubetest with a GCE/GKE Kubernetes
+		// provider - kubetest always propagates PROJECT env var in such situations.
+		return fmt.Errorf("unknown project - please set --experimental-snapshot-project flag")
+	}
+	klog.Infof("Snapshotting PD %q into snapshot %q in project %q in zone %q", pdName, snapshotName, project, zone)
+	cmd := exec.Command("gcloud", "compute", "disks", "snapshot", pdName, "--project", project, "--zone", zone, "--snapshot-names", snapshotName)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
 		klog.Errorf("Creating disk snapshot failed: %v\nCommand output: %q", err, string(output))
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -57,6 +57,7 @@ func InitFlags(p *config.PrometheusConfig) {
 	flags.BoolEnvVar(&p.ScrapeNodeExporter, "prometheus-scrape-node-exporter", "PROMETHEUS_SCRAPE_NODE_EXPORTER", false, "Whether to scrape node exporter metrics.")
 	flags.BoolEnvVar(&p.ScrapeKubelets, "prometheus-scrape-kubelets", "PROMETHEUS_SCRAPE_KUBELETS", false, "Whether to scrape kubelets. Experimental, may not work in larger clusters. Requires heapster node to be at least n1-standard-4, which needs to be provided manually.")
 	flags.BoolEnvVar(&p.ScrapeKubeProxy, "prometheus-scrape-kube-proxy", "PROMETHEUS_SCRAPE_KUBE_PROXY", true, "Whether to scrape kube proxy.")
+	flags.StringEnvVar(&p.SnapshotProject, "experimental-snapshot-project", "PROJECT", "", "GCP project used where disks and snapshots are located.")
```
wojtek-t, line 60: I think you need to default it somewhere to not break current behavior.

tosi3k: @wojtek-t we rely on the `$PROJECT` environment variable being set - this is done by kubetest [here](https://github.com/kubernetes/test-infra/blob/5f62df0ca89fd93fdcc3f1a7a76019bb115ffa47/kubetest/main.go#L775).

tosi3k: It's either the value of `--gcp-project` flag being passed to kubetest or the project being assigned by boskos (if `--gcp-project` was not provided).

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -130,8 +130,9 @@ func (pc *PrometheusController) snapshotPrometheusDiskIfEnabled() error {
 
 func (pc *PrometheusController) trySnapshotPrometheusDisk(pdName, snapshotName, zone string) error {
 	klog.Info("Trying to snapshot Prometheus' persistent disk...")
-	klog.Infof("Snapshotting PD %q into snapshot %q in zone %q", pdName, snapshotName, zone)
-	cmd := exec.Command("gcloud", "compute", "disks", "snapshot", pdName, "--zone", zone, "--snapshot-names", snapshotName)
+	project := pc.clusterLoaderConfig.PrometheusConfig.SnapshotProject
+	klog.Infof("Snapshotting PD %q into snapshot %q in project %q in zone %q", pdName, snapshotName, project, zone)
```
wojtek-t, line 134: Please add a check that project != "". Sth like.
```
if project == "" {
  // This should never happen when run from prow - kubetest should always propagate PROJECT env var.
  return fmt.Errorf("unknown project - please set --experimental-snapshot-project flag")
}
```

clusterloader2/pkg/prometheus/prometheus.go
tosi3k: Resolving after offline discussion.

clusterloader2/pkg/prometheus/experimental.go
tosi3k: Done.

PR #1183, opened 2020-04-21
Cleanup clusterloader2 load testing experimental config file

maniSbindra:
As a part of #1125, default value of the variables
ENABLE_CONFIGMAPS, ENABLE_DAEMONSETS, ENABLE_JOBS, ENABLE_SECRETS, and ENABLE_STATEFULSETS are set to true, and these values are not overriden in any of the tests. The clusterloader2 load testing  [experimental-config.yaml](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/experimental-config.yaml) file can now be be simplified as discussed in https://github.com/kubernetes/perf-tests/issues/1036#issuecomment-617133148


wojtek-t:
Hmm - apparently I was wrong and this config is used (even though it wasn't modified for the last 7 months or so).
@mm4tt - what is it used for now comparing the the regular load test?

wojtek-t:
Ohh I see now - it's merging load and density...

@maniSbindra - sorry for that; apparently this is still used, so if you could actually revert this change and migrate actually clean up the unused vars the same way you did for other tests, it would be great!

maniSbindra:
Thanks will make the required changes.

maniSbindra:
Hi @wojtek-t  I have pushed the requested changes. I have updated the PR and Title to reflect the changes

PR #1184, opened 2020-04-21
Changed perfdash service type to NodePort

bartsmykla:
/cc @mm4tt 

As perfdash is currently deployed at the `aaa` cluster
at community owned infrastructure and we don't have to use
services with LoadBalancer types it's good to change it
in the source repository and then removing unnecessary
manual instructions from the k/k8s.io [repository](https://github.com/kubernetes/k8s.io/tree/master/perf-dash.k8s.io)

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

### DIFF
--- a/perfdash/perfdash-service.yaml
+++ b/perfdash/perfdash-service.yaml
@@ -11,4 +11,4 @@ spec:
   - name: status
     port: 80
     targetPort: status
-  type: LoadBalancer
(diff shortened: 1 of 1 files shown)


bartsmykla:
@mm4tt thanks for looking! :-) Would you look at kubernetes/k8s.io#778 too please?

PR #1185, opened 2020-04-22
Change query function to rate for etcd_network_client_grpc_sent_bytes_total metric

tosi3k:
`rate` query function better reflects the change in time in case of `etcd_network_client_grpc_sent_bytes_total` metric than `irate`.

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
+++ b/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
@@ -103,7 +103,7 @@ ETCD_PANELS = [
     d.simple_graph("etcd leader", "etcd_server_is_leader", legend="{{instance}}"),
     d.simple_graph(
         "etcd bytes sent",
-        "irate(etcd_network_client_grpc_sent_bytes_total[1m])",
+        "rate(etcd_network_client_grpc_sent_bytes_total[1m])",
         yAxes=g.single_y_axis(format=g.BYTES_PER_SEC_FORMAT),
         legend="{{instance}}",
     ),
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -103,7 +103,7 @@ def api_call_latency(title, verb, scope, threshold):
     d.simple_graph("etcd leader", "etcd_server_is_leader", legend="{{instance}}"),
     d.simple_graph(
         "etcd bytes sent",
-        "irate(etcd_network_client_grpc_sent_bytes_total[1m])",
+        "rate(etcd_network_client_grpc_sent_bytes_total[1m])",
```
oxddr, line 106: Could you provide an explanation why `rate` is prefered over `irate` in this case? I suspect the reason is not obvious. Let's save next person time on figuring it out :)

tosi3k: Experience shows that `irate` doesn't work good with this metric. When I conducted A/B testing between two different 5k runs [here](https://github.com/kubernetes/kubernetes/issues/89961#issuecomment-617625083), graphs of the aforementioned metric looked completely different with `irate` applied. With `rate` function the graphs looked pretty much the same.

From Prometheus' [doc](https://prometheus.io/docs/prometheus/latest/querying/functions/#irate):
`irate` should only be used when graphing volatile, fast-moving counters. Use `rate` for alerts and slow-moving counters, as brief changes in the rate can reset the `FOR` clause and graphs consisting entirely of rare spikes are hard to read.

PR #1188, opened 2020-04-23
Add aggregated graphs to master dashboard

jprzychodzen:
/assign @mm4tt 

Add a set of new graphs to Graphana dashboard, using `quantile_over_time(0.99,` query.

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
+++ b/clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
@@ -14,60 +14,70 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import namedtuple
 from grafanalib import core as g
 import defaults as d
 
 
-def api_call_latency(title, verb, scope, threshold):
-    return d.Graph(
-        title=title,
-        targets=[
-            g.Target(expr=str(threshold), legendFormat="threshold"),
-            g.Target(
-                expr=d.one_line(
-                    """
+def api_call_latency_panel(expression):
+    def api_call_latency(title, verb, scope, threshold):
+        return d.Graph(
+            title=title,
+            targets=[
+                g.Target(expr=str(threshold), legendFormat="threshold"),
+                g.Target(
+                    expr=d.one_line(expression % {"verb": verb, "scope": scope}
+                                    ),
+                    # TODO(github.com/grafana/grafana/issues/19410): uncomment once fixed
+                    # legendFormat="{{verb}} {{scope}}/{{resource}}",
+                ),
+            ],
+            yAxes=g.single_y_axis(format=g.SECONDS_FORMAT),
+        )
+
+    return [
+        api_call_latency(
+            title="Read-only API call latency (percentaile=99, scope=resource, threshold=1s)",
+            verb="GET",
+            scope="namespace",
+            threshold=1,
+        ),
+        api_call_latency(
+            title="Read-only API call latency (percentaile=99, scope=namespace, threshold=5s)",
+            verb="LIST",
+            scope="namespace",
+            threshold=5,
+        ),
+        api_call_latency(
+            title="Read-only API call latency (percentaile=99, scope=cluster, threshold=30s)",
+            verb="LIST",
+            scope="cluster",
+            threshold=30,
+        ),
+        api_call_latency(
+            title="Mutating API call latency (threshold=1s)",
+            verb=d.any_of("CREATE", "DELETE", "PATCH", "POST", "PUT"),
+            scope=d.any_of("namespace", "cluster"),
+            threshold=1,
+        ),
+    ]
+
+API_CALL_LATENCY_PANELS = api_call_latency_panel("""
 apiserver:apiserver_request_latency_1m:histogram_quantile{
   quantile="0.99",
   verb=~"%(verb)s",
   scope=~"%(scope)s",
   resource=~"${resource:regex}s*",
-}"""
-                    % {"verb": verb, "scope": scope}
-                ),
-                # TODO(github.com/grafana/grafana/issues/19410): uncomment once fixed
-                # legendFormat="{{verb}} {{scope}}/{{resource}}",
-            ),
-        ],
-        yAxes=g.single_y_axis(format=g.SECONDS_FORMAT),
-    )
+}""")
 
-
-CLUSTERLOADER_PANELS = [
-    api_call_latency(
-        title="Read-only API call latency (percentaile=99, scope=resource, threshold=1s)",
-        verb="GET",
-        scope="namespace",
-        threshold=1,
-    ),
-    api_call_latency(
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -486,6 +497,7 @@ def api_call_latency(title, verb, scope, threshold):
     refresh="",
     rows=[
         d.Row(title="Clusterloader", panels=CLUSTERLOADER_PANELS),
+        d.Row(title="Clusterloader aggregated with qunatile", panels=CLUSTERLOADER_PANELS_AGGREGATED, collapse=True),
```
mm4tt, line 500: s/qunatile/quantile/

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -486,6 +497,7 @@ def api_call_latency(title, verb, scope, threshold):
     refresh="",
     rows=[
         d.Row(title="Clusterloader", panels=CLUSTERLOADER_PANELS),
```
oxddr, line 499: At the moment those graphs have little to do with clusterloader. Can we just rename it to `API call latency` or sth similar?

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
jprzychodzen: Done.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
jprzychodzen: Sure, make sense. Renamed.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -41,34 +34,52 @@ def api_call_latency(title, verb, scope, threshold):
...
         threshold=1,
     ),
 ]
 
+API_CALL_LATENCY_PANELS = [api_call_latency(title=n.title, verb=n.verb, scope=n.scope, threshold=n.threshold, expression="""
+apiserver:apiserver_request_latency_1m:histogram_quantile{
+  quantile="0.99",
+  verb=~"%(verb)s",
+  scope=~"%(scope)s",
+  resource=~"${resource:regex}s*",
```
oxddr, line 71: I should have add this earlier, but could you extract panel generation to a function e.g. `def api_call_latency_panel(expression)`? I think this would be cleaner that setting so many const in this file.

jprzychodzen: Added function.

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -41,34 +34,55 @@ def api_call_latency(title, verb, scope, threshold):
...
         threshold=30,
     ),
-    api_call_latency(
+    ApiCallLatencyPanelArgs(
         title="Mutating API call latency (threshold=1s)",
         verb=d.any_of("CREATE", "DELETE", "PATCH", "POST", "PUT"),
         scope=d.any_of("namespace", "cluster"),
         threshold=1,
     ),
 ]
```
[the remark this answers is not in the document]
jprzychodzen: Moved everything used only by `api_call_latency_panel` to function

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -14,60 +14,75 @@
...
-    return d.Graph(
-        title=title,
-        targets=[
-            g.Target(expr=str(threshold), legendFormat="threshold"),
-            g.Target(
-                expr=d.one_line(
-                    """
+
+
+def api_call_latency_panel(expression):
```
oxddr, line 24: please leave only 2 blank lines

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
```diff
@@ -14,60 +14,75 @@
...
+            threshold=30,
+        ),
+        ApiCallLatencyPanelArgs(
+            title="Mutating API call latency (threshold=1s)",
+            verb=d.any_of("CREATE", "DELETE", "PATCH", "POST", "PUT"),
+            scope=d.any_of("namespace", "cluster"),
+            threshold=1,
+        ),
+    ]
+    return [api_call_latency(title=n.title, verb=n.verb, scope=n.scope, threshold=n.threshold, expression=expression) for n in API_CALL_LATENCY_PANELS_TEMPLATE]
```
oxddr, line 68: This should be written as `api_call_latency_panels_template`. All caps named variables are considered to be consts in Python. 

Since you will be changing this, I'd go for a shorter name (e.g. `args`, `panel_args`)

jprzychodzen: Actually, it is const, so I do not plan to change it.

Why do you think it's not const?

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
jprzychodzen: Done

clusterloader2/pkg/prometheus/manifests/dashboards/master-dashboard.dashboard.py
oxddr: Const are module-level variables, that should never changed. The fact that it's not modified, doesn't make it a const - it's in-function variable, that simple doesn't change. 

Function variables should be typed as functions: using underscores. Please see this PEP: https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions.

jprzychodzen: Refering to Pep: https://www.python.org/dev/peps/pep-0008/#constants
```
Constants are usually defined on a module level and written in all capital letters with underscores separating words.
```
They do no need to be defined on a module level, although I agree that this is the most common use case.

Could you point out what is the difference between all caps written variable on module level and on function level, and why one you are considering as a const and other as a regular mutable variable?

jprzychodzen: Inlined variable so it does not need to have any name.

oxddr: O, that's even nicer! 

If you're still interested in discussion about naming convention, feel free to reach out to me offline.

PR #1194, opened 2020-04-28
Add CL2_ prefix to LOAD_TEST_THROUGHPUT

mm4tt:
This is to make it overridable via env


clusterloader2/testing/load/kubemark/500_nodes/override.yaml
```diff
@@ -1 +1,2 @@
-LOAD_TEST_THROUGHPUT: 50
+CL2_LOAD_TEST_THROUGHPUT: 50
+
```
wojtek-t, line 2: nit: remove this empty line

mm4tt: done, thanks

PR #1197, opened 2020-04-29
Add support for kube-dns dns

jprzychodzen:
/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
+++ b/clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
@@ -17,7 +17,7 @@
 from grafanalib import core as g
 import defaults as d
 
-DNS_LATENCY_PANEL = [
+PROBER_PANEL = [
     d.Graph(
         title="In-cluster DNS latency SLI",
         targets=d.show_quantiles(
@@ -74,18 +74,18 @@ DNS_LATENCY_PANEL = [
     ),
 ]
 
-COREDNS_PANELS = [
+SERVICE_PANELS = [
     d.Graph(
-        title="coredns: # running",
+        title="Service: # running",
         targets=[
             d.Target(
-                expr='count(container_memory_usage_bytes{namespace="kube-system", container="coredns"}) by (container, namespace)'
+                expr='count(process_resident_memory_bytes{namespace="kube-system", job="kube-dns"}) by (job, namespace)'
             )
         ],
         nullPointMode="null",
     ),
     d.Graph(
-        title="coredns: memory usage",
+        title="Service: memory usage",
         targets=d.min_max_avg(
             base='process_resident_memory_bytes{namespace="kube-system", job="kube-dns"}',
             by=["job", "namespace"],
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
```diff
@@ -74,18 +74,18 @@
     ),
 ]
 
-COREDNS_PANELS = [
+KUBEDNS_PANELS = [
```
mborsz, line 77: I'm not sure if I understand why you need to remove coredns? Can't we have dashboards for both (potentially one of them will be empty depending on the environment we are using)?

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorLegacyKubeDNS.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: monitoring.coreos.com/v1
```
mborsz, line 1: What happens if we have no kube-dns in the system?

Will clusterloader block on waiting for this servicemontior to be healthy?

clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
jprzychodzen: CoreDNS is not removed, data from CoreDNS will be shown in the same panel as data for kube-dns,.

Actually, we will have single dashboard with kube-dns and CoreDNS data - they both use the same ServiceMonitor job name so we can use the same queries.

If we want to see the difference between them, we can query over 'pods' , but we should not have both CoreDNS and kube-dns running in the same cluster.

I prefer to rename from specific implementation to service name - CoreDNS also uses this service name in it's template.

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorLegacyKubeDNS.yaml
jprzychodzen: No, it will just go on.

Logs from presubmit confirm that - we are using CoreDNS, and ClusterLoader is not blocked.

clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
oxddr: If this panels is intended to view in-cluster DNS data, then I'd rename it to "In-cluster DNS" instead of using "kube-dns". This can be misleading.

jprzychodzen: Actually, we are already on DNS dashboard so it's obvious that we are discussing DNS related values. 'In-cluster' is not verbose enough, at there is no information if which part we are referring to.

Renamed both panels as client/service, because this is the data that is shown.

Left DNS related info in dashboard titles, to increase readability in case of screenshoting graphs.

clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
```diff
@@ -17,7 +17,7 @@
 from grafanalib import core as g
 import defaults as d
 
-DNS_LATENCY_PANEL = [
+CLIENT_PANEL = [
```
oxddr, line 20: s/client/prober

oxddr:
/hold 
one more small change

clusterloader2/pkg/prometheus/manifests/dashboards/dns.dashboard.py
jprzychodzen: Done

Issue #1020, opened 2020-02-05
[ClusterLoader2] Prometheus metrics corrupted


jprzychodzen:
While running tests with ClusterLoader2, sometimes created snapshot lacks data. Probably, this is related to creating a snapshot of a running disk.

For small tests (~15 minutes long) this can result in missing metrics for specific tests run.

jkaniuk:
I've just bumped into that as well.
It happens relatively frequently for 100 nodes tests - 25% of runs out of random 8 runs?

Prometheus logs when data is missing entirely:
```
tsdb msg="last page of the wal is torn, filling it with zeros" segment=/prometheus-db/wal/00000000
tsdb msg="deleting all segments newer than corrupted segment" segment=0
```
Missing entirely:
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1225352227909013505
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-networkpolicies/1223218919838322690

Looks good:
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1225338388027019266 (24m of data)
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1225309197101436929 (23m of data)
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1225294850157449222 (25m of data)
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-networkpolicies/1225307936805359618 (43m of data)
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-networkpolicies/1225126496062935042 (43m of data)
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability-networkpolicies/1225035640324755464 (30m of data - duration of Prometheus stack running)

~20-40m of data expected:
```
W0206 08:56:09.385] I0206 08:56:09.380792   13158 prometheus.go:133] Setting up prometheus stack
...
[...]

PR #1199, opened 2020-04-29
Snapshot Prometheus disk after tearing Prometheus down

jkaniuk:
Gracefully tear down Prometheus retaining PD, then snapshot and delete PD.

Fix: #1020

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/spf13/pflag"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -146,3 +149,42 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk(pdName, snapshotName,
...
+		return err
+	}
+	if pc.diskMetadata.name == "" || pc.diskMetadata.zone == "" {
+		klog.Errorf("Missing zone or PD name, aborting deletion")
+		klog.Infof("PD name=%s, zone=%s", pc.diskMetadata.name, pc.diskMetadata.zone)
+		return fmt.Errorf("missing zone or PD name, aborting deletion")
+	}
+	// Delete Prometheus disk
+	return wait.Poll(
+		20*time.Second,
```
oxddr, line 164: Please extract this interval as const too.

jkaniuk: Done

PR #1200, opened 2020-04-29
update Readme.md for clusterloader2 to replace APIResponsiveness with APIResponsivenessPrometheusSimple

sonyafenge:
ref https://github.com/kubernetes/perf-tests/pull/915

### DIFF
--- a/clusterloader2/README.md
+++ b/clusterloader2/README.md
@@ -86,10 +86,10 @@ export CL2_ACCESS_TOKENS_QPS=5
 
 Currently available measurements are:
 - **APIResponsivenessPrometheusSimple** \
-This measurement creates summary for latency and number for server api calls based on the data collected by the prometheus-operator/prometheus server. 
+This measurement creates percentiles of latency and number for server api calls based on the data collected by the prometheus server. 
 Api calls are divided by resource, subresource, verb and scope. \
 This measurement verifies if [API call latencies SLO] is satisfied.
-If prometheus-operator/prometheus server is not available, the measurement will be skipped.
+If prometheus server is not available, the measurement will be skipped.
 - **APIResponsivenessPrometheus** \
 This measurement creates summary for latency and number for server api calls
 based on the data collected by the prometheus server.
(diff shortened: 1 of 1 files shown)


clusterloader2/README.md
```diff
@@ -85,10 +85,11 @@ export CL2_ACCESS_TOKENS_QPS=5
 ## Measurement
 
 Currently available measurements are:
-- **APIResponsiveness** \
-This measurement creates summary for latency and number for server api calls.
+- **APIResponsivenessPrometheusSimple** \
+This measurement creates summary for latency and number for server api calls based on the data collected by the prometheus-operator.
```
wojtek-t, line 89: s/prometheus-operator/prometheus server/

sonyafenge: could you please clarify your comments? do you mean replace prometheus-operator to "prometheus-operator/prometheus server"?

wojtek-t: yes - sorry for not being clear

sonyafenge: thanks and just updated.

clusterloader2/README.md
```diff
@@ -85,10 +85,11 @@ export CL2_ACCESS_TOKENS_QPS=5
 ## Measurement
 
 Currently available measurements are:
-- **APIResponsiveness** \
-This measurement creates summary for latency and number for server api calls.
+- **APIResponsivenessPrometheusSimple** \
+This measurement creates summary for latency and number for server api calls based on the data collected by the prometheus-operator/prometheus server.
```
wojtek-t, line 89: Sorry - it seems I wasn't clear again.

I was simply suggesting to remove "prometheus-operator" and just leave prometheus server.
See lines 94-98 - those two should be consistent.

sonyafenge: so what is the difference between APIResponsivenessPrometheusSimple and  APIResponsivenessPrometheus? if the description is same, why we need both of them?

wojtek-t: The readme is misleading. 

The simple simple gets percentiles over the whole test run.
The non-simple - implements the window-based official definition: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md

sonyafenge: thanks for your information. just updated and added percentiles to APIResponsivenessPrometheusSimple

Issue #986, opened 2020-01-16
Take into account number of API calls above SLO threshold when failing test


oxddr:
Over last year, we had quiet a few failed tests, because of single pair (verb, resource) exceeding SLO threshold. Most of them, were for cases where number of API call was also small (<100). We should consider taking number of API calls into account when failing test. 

I'd add a requirement that at least 2 calls must be above threshold to fail the test. This way we won't fail test because of a single API call exceeding the threshold. API call latency is collected as a histogram, so we could hack around it to get number of from high latency buckets in a specif time range. 

@mm4tt thoughts, before I start to dig on how to implement it?

wojtek-t:
> I'd add a requirement that at least 2 calls must be above threshold to fail the test.

That sounds reasonable to me.

mm4tt:
/remove-lifecycle stale

We most likely would like to do something like this. Or something else but addressing the same problem that Janek describe. The problem still exists and makes our tests flaky

oxddr:
/assign 

I am planning to have a look at this issue this or next week.

wojtek-t:
This has been enabled and rolled out.

/close

PR #1209, opened 2020-05-05
Improve api_responsiveness_prometheus

oxddr:
Improvements:
1.  Added unit tests to cover all functionality
2. Refactor code, so it's IMO structured better.

/assign @mm4tt 

One remaining thing is to move each measurement into it's own package. Currently, due to similarity of measurements, I've hit names collision few times while working on this PR.

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -14,10 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/*
-TODO(krzysied): This measurement should replace api_responsiveness.go.
-*/
-
 package slos
 
 import (
@@ -47,8 +43,7 @@ const (
 
 	currentAPICallMetricsVersion = "v1"
 
-	// TODO(krzysied): figure out why we're getting non-capitalized proxy and fix this
-	filters = `resource!="events", verb!~"WATCH|WATCHLIST|PROXY|proxy|CONNECT"`
+	filters = `resource!="events", verb!~"WATCH|WATCHLIST|PROXY|CONNECT"`
 
 	// latencyQuery matches description of the API call latency SLI and measure 99th percentaile over 5m windows
 	//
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -47,8 +43,7 @@ const (
 
 	currentAPICallMetricsVersion = "v1"
 
-	// TODO(krzysied): figure out why we're getting non-capitalized proxy and fix this
```
mm4tt, line 50: Is this figured out? :)

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -77,56 +72,45 @@ func init() {
 	}
 }
 
-type apiCall struct {
+type apiCallMetric struct {
```
mm4tt, line 75: Can we move it down? Usually you define a type and immediately below the methods for that type. Can we stick to such convention?

mm4tt: Or ideally, given this file has 300 lines of code, can we extract the apiCallMetric type to a separate file?

oxddr: I plan to send a separate PR that moves each SLO measurement to its own package (for the reason stated in PR's description). Once that's done I'd split this into two files. WDYT?

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
oxddr: I recall this was an issue fixed in k/k long time ago. If not, this indicates an issue in k/k which should be fixed. So both ways, we should remove this.

oxddr:
ref https://github.com/kubernetes/perf-tests/issues/986

mm4tt:
/approve
/lgtm

/hold
/assign @jkaniuk 
Adding current oncall as FYI, please void the hold whenever you want

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -309,3 +271,40 @@ func apiCallToPerfData(apicalls *apiResponsiveness) *measurementutil.PerfData {
...
 	return perfData
 }
+
+func (m *apiCallMetrics) sorted() []*apiCallMetric {
+	all := make([]*apiCallMetric, 0)
+	for _, v := range m.metrics {
+		all = append(all, v)
+	}
+	sort.Slice(all, func(i, j int) bool {
+		return all[i].Latency.Perc99 < all[j].Latency.Perc99
```
wojtek-t, line 281: This is sorting it in increasing order. Which isn't what we want, because in line 188 you now don't print the top metrics..

Am I missing something?

oxddr: Good catch. Fix with tests in https://github.com/kubernetes/perf-tests/pull/1229.

PR #1210, opened 2020-05-05
Add experiment: ALLOWED_SLOW_API_CALLS

oxddr:
/assign @mm4tt


clusterloader2/testing/experiments/allowed_slow_api_calls.yaml
```diff
@@ -0,0 +1 @@
+ALLOWED_SLOW_API_CALLS: 1
```
mm4tt, line 1: Cl2 supports ingesting env variables starting with `CL2_` prefix directly without a need for any files. E.g. see the `CL2_SCHEDULER_THROUGHPUT_THRESHOLD` example - https://github.com/kubernetes/test-infra/blob/3437f66887c1d15380f0cbdb33af1d6f80d32f81/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L82

Can we go the same way? Going with file means adding an `--override` line in all tests (20+ tests?)

oxddr: Sure, I forgot about this functionality.

PR #1211, opened 2020-05-05
WIP - Use golang 1.14.2 in Travis presubmit

oxddr:


.travis.yml
```diff
@@ -9,7 +9,7 @@ language: go
 
 matrix:
     include:
-        - go: 1.11
+        - go: 1.14.2
```
wojtek-t, line 12: Do we even run it?
I don't see this presubmit on any PRs...

oxddr: Me neither. It seems we don't anymore. Which kinda surprising, especially we don't have equivalent presubmit checks on Prow.

PR #1212, opened 2020-05-05
[WIP] Prometheus snapshots - print error when project is not set (1.14)

jkaniuk:
Print error explicitly when snapshotting failed due to lack of `PROJECT` variable set

/cc jkaniuk

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -128,6 +128,7 @@ func (pc *PrometheusController) snapshotPrometheusDiskIfEnabled() error {
 		snapshotRetryTimeout,
 		func() (bool, error) {
 			err := pc.trySnapshotPrometheusDisk(pc.diskMetadata.name, snapshotName, pc.diskMetadata.zone)
+			klog.Error(err)
 			// Poll() stops on error so returning nil
 			return err == nil, nil
 		})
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -139,7 +139,9 @@ func (pc *PrometheusController) trySnapshotPrometheusDisk(pdName, snapshotName,
 	if project == "" {
 		// This should never happen when run from kubetest with a GCE/GKE Kubernetes
 		// provider - kubetest always propagates PROJECT env var in such situations.
-		return fmt.Errorf("unknown project - please set --experimental-snapshot-project flag")
+		err := fmt.Errorf("unknown project - please set --experimental-snapshot-project flag")
+		klog.Error(err)
```
wojtek-t, line 143: we shouldn't klog error and return it - it often leads to multiple logs about the same thing...

jkaniuk: I have not decided yet if I want to merge it, I am using it to debug jobs.
But right now we do not have any logs.

jkaniuk: But yes - there is a better place to log it, I will push a fix

wojtek-t:
Closing - it was merged in master. If needed, we should cherrypick that commit.

PR #1213, opened 2020-05-05
Do not fail API call latency SLO if there is less than X slow calls

oxddr:
This requires changes done in: https://github.com/kubernetes/perf-tests/pull/1209
/hold

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -59,6 +59,15 @@ const (
 	// countQuery %v should be replaced with (1) filters and (2) query window size.
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
+	countSlowQuery = "sum(rate(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
+
+	// exclude all buckets of 1s and shorter
+	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT", le!~"0.\\d+|1"`
+	// exclude all buckets below or equal 5s
+	filterNamespaceList = `scope!="cluster", verb="LIST", le!~"[01234](.\\d+)?|5"`
+	// exclude all buckets below or equal 30s
+	filterClusterList = `scope="cluster", verb="LIST", le!~"[12]?[0-9](.\\d+)?|30"`
+
 	latencyWindowSize = 5 * time.Minute
 
 	// Number of metrics with highest latency to print. If the latency exceeeds SLO threshold, a metric is printed regardless.
@@ -79,7 +88,7 @@ type apiCallMetric struct {
 	Scope       string                        `json:"scope"`
 	Latency     measurementutil.LatencyMetric `json:"latency"`
 	Count       int                           `json:"count"`
-	FailedCount int                           `json:"failedCount"`
+	SlowCount   int                           `json:"slowCount"`
 }
 
 type apiCallMetrics struct {
@@ -106,7 +115,12 @@ func (a *apiResponsivenessGatherer) Gather(executor QueryExecutor, startTime tim
 		measurement.CreateSummary(summaryName, "json", content),
 	}
 
-	badMetrics := a.validateAPICalls(config.Identifier, apiCalls)
+	allowedSlowCalls, err := util.GetIntOrDefault(config.Params, "allowedSlowCalls", 0)
+	if err != nil {
+		return nil, err
+	}
+
+	badMetrics := a.validateAPICalls(config.Identifier, allowedSlowCalls, apiCalls)
 	if len(badMetrics) > 0 {
 		err = errors.NewMetricViolationError("top latency metric", fmt.Sprintf("there should be no high-latency requests, but: %v", badMetrics))
 	}
@@ -124,6 +138,7 @@ func (a *apiResponsivenessGatherer) IsEnabled(config *measurement.Config) bool {
 func (a *apiResponsivenessGatherer) gatherAPICalls(executor QueryExecutor, startTime time.Time, config *measurement.Config) (*apiCallMetrics, error) {
 	measurementEnd := time.Now()
 	measurementDuration := measurementEnd.Sub(startTime)
+	promDuration := measurementutil.ToPrometheusTime(measurementDuration)
 
 	useSimple, err := util.GetBoolOrDefault(config.Params, "useSimpleLatencyQuery", false)
 	if err != nil {
@@ -132,7 +147,6 @@ func (a *apiResponsivenessGatherer) gatherAPICalls(executor QueryExecutor, start
 
 	var latencySamples []*model.Sample
 	if useSimple {
-		promDuration := measurementutil.ToPrometheusTime(measurementDuration)
 		quantiles := []float64{0.5, 0.9, 0.99}
 		for _, q := range quantiles {
 			query := fmt.Sprintf(simpleLatencyQuery, q, filters, promDuration)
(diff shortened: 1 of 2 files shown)


oxddr:
ref https://github.com/kubernetes/perf-tests/issues/986

wojtek-t:
@oxddr Can you please rebase to make it easier to review?

oxddr:
> janluk Can you please rebase to make it easier to review?

It's not ready yet for the review yet :) But sure, will rebase it before asking for one.

oxddr:
@wojtek-t this PR is ready for the review

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -59,6 +59,16 @@ const (
 	// countQuery %v should be replaced with (1) filters and (2) query window size.
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
+	countSlowQuery = "sum(rate(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
+
+	// Bucket definitions: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L84
+	// exclude all buckets of 1s and shorter
+	filterGetAndMutating = filters + `, verb!="LIST", le!~"0.\\d+|1"`
+	// include all buckets above 5s
+	filterNamespaceList = `resource!="events", scope!="cluster", verb="LIST", le=~"[6789]|\\d0|\\+Inf"`
```
wojtek-t, line 68: It's not strictly related to this PR, but I suggest removing "resource!=events" finally.
We added it 5 years ago, because we have 1s for every opertation back then and there were too many events.
But currently, this is no longer the case, so let's remove events.
@mm4tt - FYI

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -59,6 +59,16 @@ const (
 	// countQuery %v should be replaced with (1) filters and (2) query window size.
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
+	countSlowQuery = "sum(rate(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
+
+	// Bucket definitions: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L84
+	// exclude all buckets of 1s and shorter
+	filterGetAndMutating = filters + `, verb!="LIST", le!~"0.\\d+|1"`
+	// include all buckets above 5s
```
wojtek-t, line 67: This is a bit error prone, if buckets will change in k/k.
I would prefer removing all buckets <=5s (even though the regexp will be slightly more difficult)
I think it will be something
"0.\\d+|[1234](.\\d+)*|5"

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -59,6 +59,16 @@ const (
...
+	filterNamespaceList = `resource!="events", scope!="cluster", verb="LIST", le=~"[6789]|\\d0|\\+Inf"`
+	// include only 40s, 50s, 60s and +Inf buckets
```
wojtek-t, line 69: Same here - can we instead exclude <=30?

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
wojtek-t: Might be a separate PR to remove it from the existing place (and then got rid of it here in this PR).

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -265,6 +310,8 @@ func (m *apiCallMetrics) ToPerfData() *measurementutil.PerfData {
 				"Subresource": apicall.Subresource,
 				"Scope":       apicall.Scope,
 				"Count":       fmt.Sprintf("%v", apicall.Count),
+				// FailedCount should be considered implementation details
+				// and there no need to include it in PerfData
```
wojtek-t, line 314: I disagree - this may affect whether the measurement fails or not, so I would like to see that in summary.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -287,9 +334,12 @@ func (m *apiCallMetrics) buildKey(resource, subresource, verb, scope string) str
 	return fmt.Sprintf("%s|%s|%s|%s", resource, subresource, verb, scope)
 }
 
-func (ap *apiCallMetric) Validate() error {
+func (ap *apiCallMetric) Validate(allowedSlowCalls int) error {
 	threshold := ap.getSLOThreshold()
 	if err := ap.Latency.VerifyThreshold(threshold); err != nil {
+		if allowedSlowCalls > 0 && ap.SlowCount <= allowedSlowCalls {
```
wojtek-t, line 340: nit: You don't need to check "allowedSlowCalls > 0" - if it's <=0, then the second part will fail anyway.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
oxddr: I am going to remove it now then.

oxddr: https://github.com/kubernetes/perf-tests/pull/1237

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
oxddr: Zero indicates count of slow API calls is turned off and SlowCount won't be populated (and the second check will pass, not fail).

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
oxddr: Added.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -59,6 +59,16 @@ const (
 	// countQuery %v should be replaced with (1) filters and (2) query window size.
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
+	countSlowQuery = "sum(rate(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
+
+	// Bucket definitions: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L84
```
wojtek-t, line 64: nit: this is no longer needed, since we don't rely on actual buckets.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -59,6 +59,16 @@ const (
...
+	// exclude all buckets of 1s and shorter
+	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT", le!~"0.\\d+|1"`
+	// exclude all buckets below or equal 5s
+	filterNamespaceList = `scope!="cluster", verb="LIST", le!~"[01234](.\\d+)?|5"`
+	// exclude all buckets below or equal 30s
+	filterClusterList = `scope="cluster", verb="LIST", le!~"[12]?[0-9](.\\d+)?|30"`
```
wojtek-t, line 70: nice - I like the filters much more now

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -153,30 +168,49 @@ func (a *apiResponsivenessGatherer) gatherAPICalls(executor QueryExecutor, start
...
 	}
-	return newFromSamples(latencySamples, countSamples)
+
+	allowedSlowCalls, err := util.GetIntOrDefault(config.Params, "allowedSlowCalls", 0)
+	if err != nil {
+		return nil, err
+	}
+
+	countSlowSamples := make([]*model.Sample, 0)
+	if allowedSlowCalls != 0 {
```
wojtek-t, line 192: Actually, I think it would be useful to propagate that no matter if this is going to be later used or not - this metric is pretty interesting by itself.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
wojtek-t: Thanks for pointing. I added a comment at that place now, that it would be useful to gather it no matter if the param is set or not - that is interesting metric by itself. When done, this will no longer be needed.

oxddr: Long term this check will be removed and failed count will be measured no matter whether we will use it to make decision on failures or not. 

I've added this guard for now, so we can gradually enable this feature in our tests and minimize breaking things (as per https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/experiments.md)

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
oxddr: Indeed, they look cleaner :)

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
wojtek-t: OK - Then please add a TODO in line 191 to get rid of the if and do that unconditionally once we prove it's stable.

PR #1217, opened 2020-05-05
Print errors handling Prometheus disk and snapshot

jkaniuk:
Currently some error messages are silently ignored.

/cc wojtek-t


clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -128,6 +128,7 @@ func (pc *Controller) snapshotPrometheusDiskIfEnabled() error {
 		snapshotRetryTimeout,
 		func() (bool, error) {
 			err := pc.trySnapshotPrometheusDisk(pc.diskMetadata.name, snapshotName, pc.diskMetadata.zone)
+			klog.Error(err)
```
wojtek-t, line 131: I would prefer sth like:
```
if err != nil {
  klog.Errorf("Trying to snapshot prometheus disk failed: %v", err)
}
```

it's more informative

jkaniuk: Done

PR #1218, opened 2020-05-05
Set PROMETHEUS_SCRAPE_KUBE_PROXY default to true in 1.15

jkaniuk:
Continuing effort of unifying Prometheus handling after backporting pkg/prometheus.

Fix problem with disabling kube-proxy scraping. 

/cc wojtek-t


jkaniuk:
Tested manually on GKE 1.15 - works fine

Issue #829, opened 2019-10-04
Update Netperf to use modern client-go


phenixblue:
the Netperf code references an outdated version of client-go and has references to older versions of `k8s.io/api/core/v1`.

PR #1219, opened 2020-05-05
Updating launch.go file to use updated client-go api

Sebuckle:
Resolves issue #829

### DIFF
--- a/network/benchmarks/netperf/launch.go
+++ b/network/benchmarks/netperf/launch.go
@@ -32,6 +32,7 @@ import (
 	"os"
 	"strings"
 	"time"
+	"context"
 
 	api "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -93,7 +94,7 @@ func setupClient() *kubernetes.Clientset {
 
 // getMinions : Only return schedulable/worker nodes
 func getMinionNodes(c *kubernetes.Clientset) *api.NodeList {
-	nodes, err := c.Core().Nodes().List(
+	nodes, err := c.CoreV1().Nodes().List(context.TODO(),
 		metav1.ListOptions{
 			FieldSelector: "spec.unschedulable=false",
 		})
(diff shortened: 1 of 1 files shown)


oxddr:
Hey @Sebuckle, please sign CLA first :)

oxddr:
Oh, actually it seems the bot changes its mind.

/ok-to-test

Sebuckle:
It looks like the tests are currently failing when running with GO111MODULE=off because it's using an older version of the client-go api, which takes different arguments. With GO111MODULE=on, some test files are missing, but this is looks consistent with other test runs as well. 

@oxddr Is there a "golden path" for users to setup their machine? I.e. should they always have one setting for GO111MODULE?

oxddr:
In the perf-tests repo we have a mixture of packages that are using Go modules and some other still reling on the old approach with vendor or Godeps. `network` seems to belong to the second group. There are two options here on how to proceed:

1. `network` is migrated to Go modules (as-is, in a sepearate PR). Then updating to a newer version of `client-go` would be a matter of updating `go.mod` file.
1. vendor/Godeps are updated to a specific version of the `client-go` you want to use.

In the light of https://github.com/kubernetes/perf-tests/pull/1232/files, the first option is preferred. 

Regarding your local setup, I don't think it's possible to do anything other that 1) remember which packages require which `GO111MODULE` setting (some may have `Makefiles` that will take care of this for you) 2) rely on presubmits to test your code.

I hope that helps.

oxddr:
@Buck I've created PR to migrate this package to Go module: https://github.com/kubernetes/perf-tests/pull/1240.

network/benchmarks/netperf/launch.go
```diff
@@ -32,6 +32,7 @@ import (
 	"os"
 	"strings"
 	"time"
+	"context"
```
wojtek-t, line 35: please sort alphabetically

network/benchmarks/netperf/launch.go
```diff
@@ -93,7 +94,7 @@ func setupClient() *kubernetes.Clientset {
 
 // getMinions : Only return schedulable/worker nodes
 func getMinionNodes(c *kubernetes.Clientset) *api.NodeList {
-	nodes, err := c.Core().Nodes().List(
+	nodes, err := c.CoreV1().Nodes().List(context.TODO(),
```
wojtek-t, line 97: network Godeps/ are pretty old now - so to make it work, you need to update godeps (or ideally upgrade to go modules and then update)

PR #1229, opened 2020-05-06
Fix incorrect order of apiCalls in API latency SLO measurement

oxddr:
Add test to verify log statements (they are on of the interfaces for this
measurement). Also take advantage of t.Run method in tests.

/assign @wojtek-t


wojtek-t:
/lgtm

Thanks!

PR #1235, opened 2020-05-06
Clusterloader2: make the prometheus manifest path configurable

yuanchen8911:
In the current implementation, the prometheus manifest file path is relative to` $GOPATH`. This PR enables to configure the path prefix through a new command line flag `--prometheus-manifest-path`. If not specified, the default value `$GOPATH` is used.  The prometheus manifest file path parameters are changed to PrometheusConfig's fields.

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -41,13 +41,9 @@ import (
 const (
 	namespace                    = "monitoring"
 	storageClass                 = "ssd"
-	coreManifests                = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/*.yaml"
-	defaultServiceMonitors       = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/default/*.yaml"
-	masterIPServiceMonitors      = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/master-ip/*.yaml"
 	checkPrometheusReadyInterval = 30 * time.Second
 	checkPrometheusReadyTimeout  = 15 * time.Minute
 	numK8sClients                = 1
-	nodeExporterPod              = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/exporters/node-exporter.yaml"
 )
 
 // InitFlags initializes prometheus flags.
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -78,6 +82,21 @@ type Controller struct {
...
 
+// CompleteConfig completes Prometheus manifest file path config
+func CompleteConfig(p *config.PrometheusConfig) {
+	if p.ManifestPath != "" {
+		coreManifests = p.ManifestPath + "/*.yaml"
+		defaultServiceMonitors = p.ManifestPath + "/default/*.yaml"
+		masterIPServiceMonitors = p.ManifestPath + "/master-ip/*.yaml"
+		nodeExporterPod = p.ManifestPath + "/exporter/node-exporter.yaml"
+	} else {
+		coreManifests = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests/*.yaml"
```
oxddr, line 93: Please keep the prefix `$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests` as const.  With this the whole code can be simplified to:

```
prefix := p.ManifestPath
if p.ManifestPath == "" {
  prefix = defaultPrometheusManifestPath
}
coreManifests = prefix + "/*.yaml"
defaultServiceMonitors = prefix + "/default/*.yaml"
masterIPServiceMonitors = prefix + "/master-ip/*.yaml"
nodeExporterPod = prefix + "/exporter/node-exporter.yaml"
```

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -78,6 +82,21 @@ type Controller struct {
 	ssh util.SSHExecutor
 }
 
+// CompleteConfig completes Prometheus manifest file path config
+func CompleteConfig(p *config.PrometheusConfig) {
```
oxddr, line 86: I'd name this function more precisely, e.g. `SetupManifests`. We can always rename it later on if more configuration will be added.

wojtek-t: I actually like CompleteConfig - it's consistent with many other places across the whole k8s.

clusterloader2/pkg/prometheus/prometheus.go
wojtek-t: +1

clusterloader2/pkg/prometheus/prometheus.go
oxddr: This seems to be matter of preference. But if `CompleteConfig` common in k8s, then it's fine to leave it as is.

yuanchen8911: Changed as suggested, thanks.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -78,6 +83,18 @@ type Controller struct {
...
+	prefix := p.ManifestPath
+	if prefix == "" {
+		prefix = defaultPrometheusManifestPath
+	}
+	coreManifests = prefix + "/*.yaml"
+	defaultServiceMonitors = prefix + "/default/*.yaml"
+	masterIPServiceMonitors = prefix + "/master-ip/*.yaml"
+	nodeExporterPod = prefix + "/exporter/node-exporter.yaml"
```
wojtek-t, line 95: s/exporter/exporters/

[This is the test failure cause]

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -39,15 +39,19 @@ import (
...
+	namespace                     = "monitoring"
+	storageClass                  = "ssd"
+	checkPrometheusReadyInterval  = 30 * time.Second
+	checkPrometheusReadyTimeout   = 15 * time.Minute
+	numK8sClients                 = 1
+	defaultPrometheusManifestPath = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests"
+)
+
+var (
+	coreManifests           string
```
wojtek-t, line 51: TBH - I'm not a fan of global variables.

Instead, we should add fields to PrometheusConfig and fill them in CompleteConfig.

yuanchen8911: @wojtek-t  Thanks! I fixed the bug and moved the global config parameters to PrometheusConfig. The const "DefaultPrometheusPath" was removed too and use the pflag default value instead. Please review it.

clusterloader2/pkg/prometheus/prometheus.go
yuanchen8911: Thanks!

PR #1238, opened 2020-05-07
Clusterloader2: Change the default IncludeFile path

yuanchen8911:
This PR changes the default` IncludeFile `path prefix from the clusterloader binary location to the clusterloader source path, which logically makes more sense.  It also checks if the `IncludeFile` path in the manifest file is an absolute path or not. If it's an absolute path, the default path prefix will not be joined.

### DIFF
--- a/clusterloader2/pkg/config/template_functions.go
+++ b/clusterloader2/pkg/config/template_functions.go
@@ -32,6 +32,8 @@ import (
 	"k8s.io/klog"
 )
 
+const defaultIncludeFilepath = "$GOPATH/src/k8s.io/perf-tests/clusterloader2"
+
 func init() {
 	rand.Seed(time.Now().UnixNano())
 }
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/config/template_functions.go
```diff
@@ -203,20 +203,19 @@ func defaultParam(param, defaultValue interface{}) interface{} {
...
 	fileStr, ok := file.(string)
 	if !ok {
 		return "", fmt.Errorf("incorrect argument type: got: %T want: string", file)
 	}
-
-	ex, err := os.Executable()
-	if err != nil {
-		return "", fmt.Errorf("unable to determine executable path: %v", err)
+	if !filepath.IsAbs(fileStr) {
+		defaultPathPrefix := "$GOPATH/src/k8s.io/perf-tests/clusterloader2"
```
wojtek-t, line 214: make that a constant

PR #1246, opened 2020-05-11
Add validation for Prometheus config

jprzychodzen:
Without this, we will only know at the end of the test that the project is not configured

/assign @jkaniuk

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	clerrors "k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/flags"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -33,10 +33,12 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	clerrors "k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/flags"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
+
```
wojtek-t, line 41: nit: remove

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -59,6 +61,14 @@ func InitFlags(p *config.PrometheusConfig) {
 	flags.StringEnvVar(&p.ManifestPath, "prometheus-manifest-path", "PROMETHEUS_MANIFEST_PATH", "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/prometheus/manifests", "Path to the prometheus manifest files.")
 }
 
+func ValidatePrometheusFlags(p *config.PrometheusConfig) *clerrors.ErrorList {
+	errList := clerrors.NewErrorList()
+	if *shouldSnapshotPrometheusDisk && p.SnapshotProject == "" {
+		errList.Append(fmt.Errorf("Requesting snapshot, but snapshot project not configured. Use --experimental-snapshot-project flag"))
```
wojtek-t, line 67: nit: Errorf() should start with small letter (as those are usually joined) - this is the convension;

so s/Requesting/requesting/

clusterloader2/pkg/prometheus/prometheus.go
jprzychodzen: Done

clusterloader2/pkg/prometheus/prometheus.go
jprzychodzen: Done

jprzychodzen:
Fixed nits, added comment to public method.

PR #1247, opened 2020-05-12
Add flag --skip-steps

jprzychodzen:
This change enables runetime test config change - skipping any number of
steps without a need to modify config file.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/modifier"
 	"os"
 	"path"
 	"time"
@@ -101,6 +102,7 @@ func initFlags() {
 	flags.StringArrayVar(&testOverridePaths, "testoverrides", []string{}, "Paths to the config overrides file. The latter overrides take precedence over changes in former files.")
 	flags.StringVar(&testSuiteConfigPath, "testsuite", "", "Path to the test suite config file")
 	initClusterFlags()
+	modifier.InitFlags(&clusterLoaderConfig.ModifierConfig)
 	prometheus.InitFlags(&clusterLoaderConfig.PrometheusConfig)
 }
 
--- a/clusterloader2/pkg/modifier/modifier.go
+++ b/clusterloader2/pkg/modifier/modifier.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modifier
+
+import (
+	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/pkg/config"
+	"k8s.io/perf-tests/clusterloader2/pkg/flags"
+)
+
+// Modifier mutates provided test
+type Modifier interface {
+	ChangeTest(*api.Config)
+}
+
+// InitFlags allows setting configuration with flags
+func InitFlags(m *config.ModifierConfig) {
+	flags.StringArrayVar(&m.SkipSteps, "skip-steps", []string{}, "Name of steps to skip in test")
+}
+
+// NewModifier creates new Modifier according to provided configuration
+func NewModifier(m *config.ModifierConfig) Modifier {
+	return &simpleModifier{skipSteps: m.SkipSteps}
+}
+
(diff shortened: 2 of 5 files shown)


clusterloader2/pkg/test/modifier_test.go
```diff
@@ -0,0 +1,45 @@
+package test
+
+import (
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/perf-tests/clusterloader2/api"
+)
+
+var _ = ginkgo.Describe("Modifier", func() {
```
wojtek-t, line 9: Let's not introduce ginkgo to our tests - it's simple unit test and ;et's stick to basic unit tests please

clusterloader2/cmd/clusterloader.go
```diff
@@ -101,6 +101,7 @@ func initFlags() {
 	flags.StringArrayVar(&testOverridePaths, "testoverrides", []string{}, "Paths to the config overrides file. The latter overrides take precedence over changes in former files.")
 	flags.StringVar(&testSuiteConfigPath, "testsuite", "", "Path to the test suite config file")
 	initClusterFlags()
+	test.InitModifierFlags()
```
wojtek-t, line 104: modifier.InitFlags()

[put the logic in modifier package]

clusterloader2/go.mod
```diff
@@ -53,6 +53,7 @@ replace k8s.io/sample-controller => k8s.io/sample-controller v0.18.0
 require (
 	github.com/go-errors/errors v1.0.1
 	github.com/onsi/ginkgo v1.11.0
+	github.com/onsi/gomega v1.7.0
```
wojtek-t, line 56: please don't....

clusterloader2/pkg/test/modifier_test.go
jprzychodzen: Removed ginkgo

clusterloader2/cmd/clusterloader.go
jprzychodzen: Moved to package modifier

clusterloader2/go.mod
jprzychodzen: Removed.

clusterloader2/pkg/modifier/modifier.go
```diff
@@ -0,0 +1,45 @@
...
+
+type Modifier interface {
+	ChangeTest(*api.Config)
+}
+
+var (
+	skipSteps []string
+)
+
+func InitModifierFlags() {
```
wojtek-t, line 32: InitFlags (it's modifier package, we know these are Modifier ones)

jprzychodzen:
Re-commited after `gofmt` linter fix

clusterloader2/pkg/modifier/modifier.go
```diff
@@ -0,0 +1,61 @@
...
+	"k8s.io/perf-tests/clusterloader2/api"
+	"k8s.io/perf-tests/clusterloader2/pkg/flags"
+)
+
+type Modifier interface {
+	ChangeTest(*api.Config)
+}
+
+var (
+	skipSteps []string
```
wojtek-t, line 29: I'm not a fan of global variables.

I think that creating `Config` type and adding it as a field to cclusterLoaderConfig would be cleaner.

clusterloader2/pkg/modifier/modifier.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func InitModifierFlags() {
+	flags.StringArrayVar(&skipSteps, "skip-steps", []string{}, "Name of steps to skip in test")
+}
+
+func NewTestModifier() Modifier {
```
wojtek-t, line 36: NewModifier() ?

clusterloader2/pkg/modifier/modifier.go
jprzychodzen: Done.

clusterloader2/pkg/modifier/modifier.go
jprzychodzen: Renamed

clusterloader2/pkg/modifier/modifier.go
jprzychodzen: Make sense, tried to follow convection - some flags are using global variables.

There is mix in the code with this approach. Added ModifierConfig with the same approach as PrometheusConfig - PTAL if this makes sense.

clusterloader2/pkg/modifier/modifier.go
```diff
@@ -0,0 +1,62 @@
...
+	flags.StringArrayVar(&m.SkipSteps, "skip-steps", []string{}, "Name of steps to skip in test")
+}
+
+// NewModifier creates new Modifier according to provided configuration
+func NewModifier(m *config.ModifierConfig) Modifier {
+	return &simpleModifier{skipSteps: m.SkipSteps}
+}
+
+type simpleModifier struct {
+	skipSteps []string
```
mm4tt, line 41: NIT: I'd make this field of type map[string]bool to have a O(1) lookup, but given the # steps is usually < 20-30, I don't think it would change a lot :)

jprzychodzen: Thanks for the tip!

Ideally, flags would support maps as a first class citizens, then it would be easier to use maps when they are better choice.

Issue #1248, opened 2020-05-13
Potential segfault in runtimeobjects.GetNumObjectsMatchingSelector method


mm4tt:
It happened in one of our 1.16 tests, but the vulnerability is present in all release branches.

Stacktrace:
```
W0513 00:57:50.893] panic: runtime error: invalid memory address or nil pointer dereference
W0513 00:57:50.893] [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x11bd734]
W0513 00:57:50.893] 
W0513 00:57:50.893] goroutine 672187 [running]:
W0513 00:57:50.893] k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects.GetNumObjectsMatchingSelector.func1(0xc021068bc0, 0x15)
W0513 00:57:50.894] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go:409 +0x1d4
W0513 00:57:50.894] k8s.io/perf-tests/clusterloader2/pkg/framework/client.RetryFunction.func1(0x14f6fe0, 0x1674701, 0x0)
W0513 00:57:50.894] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/framework/client/objects.go:113 +0x32
W0513 00:57:50.894] k8s.io/perf-tests/clusterloader2/vendor/k8s.io/apimachinery/pkg/util/wait.ExponentialBackoff(0x5f5e100, 0x4008000000000000, 0x0, 0x6, 0x0, 0xc081ec9180, 0x0, 0x0)
W0513 00:57:50.894] 	/go/src/k8s.io/perf-tests/clusterloader2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:284 +0x51
W0513 00:57:50.894] k8s.io/perf-tests/clusterloader2/pkg/framework/client.RetryWithExponentialBackOff(...)
W0513 00:57:50.895] 	/go/src/k8s.io/perf-tests/clusterloader2/pkg/framework/client/objects.go:59
W0513 00:57:50.895] k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects.GetNumObjectsMatchingSelector(0x17af880, 0xc00ab37a80, 0xc05db3d950, 0xe, 0x0, 0x0, 0xc0527bf880, 0x2, 0xc021068c00, 0x16, ...)
[...]

PR #1249, opened 2020-05-13
Fix segfault in GetNumObjectsMatchingSelector

mm4tt:
If the List method returns error the list object is nil and accessing list.Items results in segfault.

Ref. https://github.com/kubernetes/perf-tests/issues/1248


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -396,7 +396,9 @@ func GetNumObjectsMatchingSelector(c dynamic.Interface, namespace string, resour
 	var numObjects int
 	listFunc := func() error {
 		list, err := c.Resource(resource).Namespace(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector.String()})
-		numObjects = len(list.Items)
+		if err == nil {
```
wojtek-t, line 399: The below would be more canonical...:

```
if err != nil {
  return err
}
numObject = len(list.Items)
return nil
```

mm4tt: Done

PR #1250, opened 2020-05-13
Fix segfault in GetNumObjectsMatchingSelector

mm4tt:
If the List method returns error the list object is nil and accessing list.Items results in segfault.

Ref. https://github.com/kubernetes/perf-tests/issues/1248

(cherry picked from commit 5b365e5bfb6624bece619551fb95d0520d050dfd)


mm4tt:
Failing presubmit should get fixed with https://github.com/kubernetes/test-infra/pull/17589

PR #1257, opened 2020-05-13
Enable Kubemark nodes objects size increase in golang tests

tosi3k:
`build-and-push-k8s-at-golang-tip` builds Kubernetes v1.18 that does not contain k/k [PR](https://github.com/kubernetes/kubernetes/pull/90806) that increases Kubemark nodes objects. The proposed cherry-pick is a quick workaround for that issue.

/sig scalability

### DIFF
--- a/golang/build-kubernetes.sh
+++ b/golang/build-kubernetes.sh
@@ -64,6 +64,11 @@ function build_kubernetes {
   cd /go/src/k8s.io/kubernetes/build/build-image
   git checkout $K8S_COMMIT
 
+  # The git cherry-pick command below takes a commit from k8s v1.19 that
+  # artifically increases Kubemark cluster nodes objects.
+  # TODO: Get rid of this cherry-pick once we start testing aginst k8s v1.19+.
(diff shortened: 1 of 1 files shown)


golang/build-kubernetes.sh
```diff
@@ -64,6 +64,11 @@ function build_kubernetes {
   cd /go/src/k8s.io/kubernetes/build/build-image
   git checkout $K8S_COMMIT
 
+  # TODO: The git cherry-pick command below takes a commit from k8s v1.19 that
+  # artifically increases Kubemark cluster nodes objects.
+  # Hence, cherry-picking becomes redundant after we start testing k8s v1.19.
```
mm4tt, line 69: I'd rephrase it to:
```
# The git cherry-pick command below takes a commit from k8s v1.19 that
# artificially increases Kubemark cluster nodes objects.
# TODO: Get rid of this cherry-pick once we start testing against k8s 1.19+
```

tosi3k: Done.

mm4tt:
Thanks!

/lgtm
/approve

PR #1258, opened 2020-05-14
Enable allowedSlowCalls feature in density and load

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/986


oxddr:
/lgtm
/hold

Let's wait for the presubmit to pass, so we can double check logs, whether it worked as expected.

mm4tt:
/hold cancel

Test passed, we didn't have any SLO violation and all slowApiCalls were 0. But it definitely doesn't break anything so we're good to merge it. Thanks Janek!

PR #1260, opened 2020-05-14
Solve the problem with faulty WaitForControllerPods

### LINKED ISSUE #1259: WaitForControlledPods is racy for deletions
In WaitForControllerPods::gather method, we first determine what RV we should be synced to to ensure that all checkers are already registered:
https://github.com/kubernetes/perf-tests/blob/628c8a54ba8041bfa3231a89576e93f1a5be85b6/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go#L181

However, if we delete an object, it's no longer in the list, we don't wait for it.
In extreme case, when all objects are deleted, that function returns maxResourceVersion=0, so all the conditions after that are trivially met.

We should fix it to ensure that all checkers are really registered before getting to handlingGroup.Wait() even for deletions.

@mm4tt @kubernetes/sig-scalability-bugs

wojtek-t:
Fix #1259

/assign @mm4tt

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -375,8 +375,8 @@ func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(runtim
 // - wait until set of objects in local cache is equal to the one returned from LIST call
 //   Note, that we can't compare whole objects, as in the meantime local cache may have
 //   gone ahead of the original LIST (e.g. due to status updates of existing objects).
-// - equal set of objects guarantees that all deletions are already reflected in our local
-//   cache, and at that point we know that its resource version is fresh enough - this
+//   Equal set of objects guarantees that all deletions are already reflected in out local
+//   cache, and at this point we know that is resource version is fresh enough - this
 //   solves the deletions case
 //   Note that the fact that the store is synced doesn't mean that all handlers were already
 //   fired, so we still need to wait for them.
@@ -419,6 +419,11 @@ func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion(sy
 	cond := func() (bool, error) {
 		currentKeys := w.informer.GetStore().ListKeys()
 		sort.Strings(currentKeys)
+		// The following check means that we are not resistant to situations
+		// if something will create/delete objects. However, this can't be
+		// clusterloader, as measurements don't interfere with phases, so
+		// it can only be independent users/controllers. We're fine with
+		// accepting this risk.
 		return reflect.DeepEqual(objectKeys, currentKeys), nil
 	}
 	if err := wait.Poll(checkControlledPodsInterval, syncTimeout, cond); err != nil {
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -18,6 +18,9 @@ package common
 
 import (
 	"fmt"
+	"reflect"
+	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


wojtek-t:
/hold

Until we figure out if there aren't any bugs in the implementation.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -358,12 +364,24 @@ func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(runtim
 }
 
 // getObjectCountAndMaxVersion returns number of objects that satisfy measurements parameters
-// and maximal resource version of these objects.
-// These two values allow to properly handle all object operations:
-// - When create/delete operation are called we expect the exact number of objects.
-// - When objects is updated we expect to receive event referencing this specific version.
-//   Using maximum from objects resource versions assures that all updates will be processed.
```
wojtek-t, line 365: @krzysied - if you're interested why your approach didn't work :)

wojtek-t:
/hold cancel

@mm4tt - PTAL

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -358,12 +364,24 @@ func (w *waitForControlledPodsRunningMeasurement) updateOpResourceVersion(runtim
...
+// We can't guarantee exact value fo maximal resource version, as deletions are problematic.
+// When the object is deleted, by definition it will not be returned by the LIST request,
+// and thus we cannot infer resource version of the deletion itself.
+// To solve this problem, we:
+// - take maximum from resource versions of returned objects - this solves the creations
+//   updates case
+// - wait until set of objects in local cache is equal to the one returned from LIST call
+//   Note, that we can't compare whole objects, as in the meantime local cache may have
+//   gone ahead of the original LIST (e.g. due to status updates of existing objects).
+// - equal set of objects guarantees that all deletions are already reflected in our local
```
mm4tt, line 378: This doesn't look like the other steps, i.e.

1. take maximum rv..
2. wait until...
4. return ...

Shouldn't this be part of step 2?

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -384,11 +403,35 @@ func (w *waitForControlledPodsRunningMeasurement) getObjectCountAndMaxVersion()
...
 		if version > maxResourceVersion {
 			maxResourceVersion = version
 		}
 	}
+	sort.Strings(objectKeys)
+
+	cond := func() (bool, error) {
+		currentKeys := w.informer.GetStore().ListKeys()
+		sort.Strings(currentKeys)
+		return reflect.DeepEqual(objectKeys, currentKeys), nil
```
mm4tt, line 422: Hmm, there is a risk that if someone creates/deletes a resource that we're watching, between the List and the Poll in this method the condition will never be satisfied, e.g. 

1. We list: a,b,c
2. Someone adds d
3. Informer is at a,b,c,d
4. Condition will be always false

But I guess, it's fine. Given that a step can be either measurement or phase, CL2 itself will never modifying anything when this code.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
wojtek-t: done

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
wojtek-t: Yeah - that was my thinking. Added explicit comment.

mm4tt:
FTR, this presubmit failed on timeouting WaitForControlledPodsRunning after thsi PR has merged- https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1263/pull-perf-tests-clusterloader2/1261274584590585856/

mm4tt:
This is the line that returned error -https://github.com/kubernetes/perf-tests/blob/a80a8d97ca8be06e375a14d38cd325354921ffc3/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go#L196

Now, I'm thinking about it. Isn't it that bookmarks will break this? As the informer RV from bookmark can be newer than any RV that we'll record in "handle" method.

Anyway, given that there is a strong singal that this PR can break something and that it's Friday, I'm going to revert it

PR #1267, opened 2020-05-16
attempt2: Solve the problem with faulty WaitForControlledPods

### LINKED ISSUE #1259: WaitForControlledPods is racy for deletions

wojtek-t:
Fix #1259

/assign @mm4tt

/hold

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -26,7 +26,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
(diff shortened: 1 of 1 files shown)


wojtek-t:
Density test failed on unrelated issue (pod startup time latency: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1267/pull-perf-tests-clusterloader2/1261620111752564736/)

/retest

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -178,13 +182,31 @@ func (w *waitForControlledPodsRunningMeasurement) gather(syncTimeout time.Durati
...
+	//   NOTE: we could try waiting for the informer state to be the same and use the
+	//   resourceVersion from there, but then existence of bookmarks and the fact that our
+	//   informer doesn't necessary follow all objects of a given type can break that.
+	//   See #1259 for more details.
+
 	cond := func() (bool, error) {
-		return w.opResourceVersion >= maxResourceVersion, nil
+		w.lock.Lock()
+		defer w.lock.Unlock()
+		return w.opResourceVersion >= maxResourceVersion && objectKeys.Equal(w.objectKeys), nil
```
mm4tt, line 209: To confirm, we cannot wait just for `objectKeys.Equal(...)` because it can easily break for the "update" case, right?

wojtek-t: Right - the maxRV is needed for update case.

wojtek-t:
/hold cancel

In case of issues, we should revert it, but it seems to work fine.
@jprzychodzen as oncall - FYI

PR #1270, opened 2020-05-20
Clusterloader: customize logging levels

yuanchen8911:
Clusterloader generates a lot of output on the console. The information is helpful but the large amount of very detailed runtime measurement data can be overwhelming and may hide some desired and  important information. This PR changes the detailed runtime measurement logs to V(2).

### DIFF
--- a/clusterloader2/pkg/measurement/util/phase_latency.go
+++ b/clusterloader2/pkg/measurement/util/phase_latency.go
@@ -119,7 +119,7 @@ func (o *ObjectTransitionTimes) printLatencies(latencies []LatencyData, header s
 	if index < 0 {
 		index = 0
 	}
-	klog.Infof("%s: %d %s: %v", o.name, len(latencies)-index, header, latencies[index:])
+	klog.V(2).Infof("%s: %d %s: %v", o.name, len(latencies)-index, header, latencies[index:])
 	var thresholdString string
 	if threshold != time.Duration(0) {
 		thresholdString = fmt.Sprintf("; threshold %v", threshold)
--- a/clusterloader2/pkg/measurement/util/wait_for_nodes.go
+++ b/clusterloader2/pkg/measurement/util/wait_for_nodes.go
@@ -54,7 +54,7 @@ func WaitForNodes(clientSet clientset.Interface, stopCh <-chan struct{}, options
 		case <-time.After(options.WaitForNodesInterval):
 			nodeCount = getNumReadyNodes(ps.List())
 			if options.EnableLogging {
-				klog.Infof("%s: node count (selector = %v): %d", options.CallerName, options.Selector.String(), nodeCount)
+				klog.V(2).Infof("%s: node count (selector = %v): %d", options.CallerName, options.Selector.String(), nodeCount)
 			}
 			if options.MinDesiredNodeCount <= nodeCount && nodeCount <= options.MaxDesiredNodeCount {
 				return nil
--- a/clusterloader2/pkg/measurement/util/wait_for_pods.go
+++ b/clusterloader2/pkg/measurement/util/wait_for_pods.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 )
@@ -89,7 +89,7 @@ func WaitForPods(clientSet clientset.Interface, stopCh <-chan struct{}, options 
 				klog.Errorf("%s: %s: %d pods appeared: %v", options.CallerName, options.Selector.String(), len(addedPods), strings.Join(addedPods, ", "))
 			}
 			if options.EnableLogging {
-				klog.Infof("%s: %s: %s", options.CallerName, options.Selector.String(), podsStatus.String())
+				klog.V(2).Infof("%s: %s: %s", options.CallerName, options.Selector.String(), podsStatus.String())
 			}
 			// We allow inactive pods (e.g. eviction happened).
 			// We wait until there is a desired number of pods running and all other pods are inactive.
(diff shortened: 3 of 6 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -79,6 +80,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubemarkRootKubeConfigPath, "kubemark-root-kubeconfig", "KUBEMARK_ROOT_KUBECONFIG", "",
 		"Path the to kubemark root kubeconfig file, i.e. kubeconfig of the cluster where kubemark cluster is run. Ignored if provider != kubemark")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.APIServerPprofByClientEnabled, "apiserver-pprof-by-client-enabled", "APISERVER_PPROF_BY_CLIENT_ENABLED", true, "Whether apiserver pprof endpoint can be accessed by Kubernetes client.")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.LoggingLevel, "logging-level", "LOGGING_LEVEL", "2", "0: errors and warnings, 1: info, 2: detailed info.")
```
wojtek-t, line 83: I'm against introducing this flag. klog already has the --v flag that you can use for this purpose

yuanchen8911: Thanks,  I'll remove it and just add V() logging control.  @wojtek-t

yuanchen8911: Any idea why the test failed? Thanks.

clusterloader2/cmd/clusterloader.go
```diff
@@ -95,6 +96,8 @@ func validateClusterFlags() *errors.ErrorList {
 }
 
 func initFlags() {
+	const defaultLoggingLevel = "2"
```
wojtek-t, line 99: This is default anyway - no need to do that. Please remove

clusterloader2/pkg/flags/flags.go
```diff
@@ -35,6 +35,11 @@ func init() {
 
 var flags []flagFunc
 
+// Set sets the default value of the named command-line flag.
+func Set(name, value string) error {
+	return pflag.Set(name, value)
+}
```
wojtek-t, line 41: Shouldn't be needed - please remove

clusterloader2/pkg/chaos/nodes.go
```diff
@@ -121,7 +121,7 @@ func (k *NodeKiller) pickNodes() ([]v1.Node, error) {
 	for i := range prometheusPods {
 		if prometheusPods[i].Spec.NodeName != "" {
 			nodesHasPrometheusPod.Insert(prometheusPods[i].Spec.NodeName)
-			klog.Infof("%s: Node %s removed from killing. Runs pod %s", k, prometheusPods[i].Spec.NodeName, prometheusPods[i].Name)
+			klog.V(1).Infof("%s: Node %s removed from killing. Runs pod %s", k, prometheusPods[i].Spec.NodeName, prometheusPods[i].Name)
```
wojtek-t, line 124: While I understand the motivation of changing verbotisty of all logs (I didn't think about all changes in this PR), I have a concern that it won't persist for long time. i.e. my fear is that we will soon add a new `klog.Infof(...)` logs that will break your assumptions again.

So while I agree with switching some logs to V(2), I'm not sure that changing other to V(1) actually makes sense.
@mm4tt - thoughts?

clusterloader2/cmd/clusterloader.go
yuanchen8911: OK, but note that the default is 0.

clusterloader2/pkg/chaos/nodes.go
yuanchen8911: It would be helpful to have a guideline. The current levels were mainly set based on the importance and the amount of output.  My original motivation was to be able to disable the large amount of runtime measurement information generated in wait_for_pods.go, wait_for_pvs.go, wait_for_pvcs.go and wait_for_nodes, the detailed latency metrics (phase_latency.go:127) and the cluster nodes information (./util/cluster.go:70) and make them V(2).   I'll undo those V(1) changes and keep V(2).

yuanchen8911: Removed the default level set and reverted all the V(1) changes.

mm4tt: Yeah, switching some (the ones spamming the logs most) to V(2) make sense. 
We'll still run our CI/CD tests with full logs, as it's really helpful for debugging.  But I understand that other users may want to set verbosity to something higher.

mm4tt:
I've had a different idea about reducing spam in cl2 logs some time ago, see https://github.com/kubernetes/kubernetes/issues/88546#issuecomment-599976255. It would benefit also our CI/CD runs, but it would require changes in klog which are not likely to happen anytime soon.

Your solution doesn't address ci/cd runs, we cannot reduce verbosity there as it would make debugging harder. But it solves the problem for the other user. Nice!

/lgtm
/approve

wojtek-t:
@mm4tt - given the default is --v=0, this actually broke our tests (i.e. we no longer have those logs).

mm4tt:
Oops, dunno why I thought the default is 2...
Reverting in https://github.com/kubernetes/perf-tests/pull/1271

@yuanchen8911, sorry about that, could you sent a roll-forward with the fixed default?

yuanchen8911:
> Oops, dunno why I thought the default is 2...
> Reverting in #1271
> 
> @yuanchen8911, sorry about that, could you sent a roll-forward with the fixed default?

@mm4tt  I'll change the default to V(2). What should I do next?  Submit a new PR?

mm4tt:
Yeah, please send a new PR, if you haven't already.

PR #1271, opened 2020-05-21
Revert "Clusterloader: change the detailed measurement logs to V(2)"

mm4tt:
This reverts commit bf383bc8887d16206bef698f04eb443c35859c33.


yuanchen8911:
@mm4tt @wojtek-t I can change the default to V(2). How should I submit the change?  Thanks.

wojtek-t:
> @mm4tt @wojtek-t I can change the default to V(2). How should I submit the change? Thanks.

We should not change it here - we should rather change it somewhere in test-infra (either in our presets and/or somewhere in test-infra). Requires a bit of investigation...

Issue #980, opened 2020-01-14
Do not rely on flags, etc. to control logging in CL2


oxddr:
Example cleanup: https://github.com/kubernetes/perf-tests/pull/979, but there are other places with similar issue. 

/area clusterloader

oxddr:
Example area to cleanup: [`EnableLogging`](https://github.com/kubernetes/perf-tests/search?q=EnableLogging&unscoped_q=EnableLogging) from WairForX measurements.

/help-wanted
/good-first-issue

All this occurrences should be replace with klog calls, with lower priority (2? 3?). We should rely on klog to control what's being logged.

PR #1274, opened 2020-05-26
Clusterloader:  enable/disable measurement logging

yuanchen8911:
This PR supports custom measurement logging. A new command line flag "--enable-measurement-logging" is added for enable and disable the detailed measurement logging. The default is true. 

It's a re-implementation of the previous PR #1270 .  which was merged and then reverted.  Please refer to #1270  and #1271.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -18,10 +18,11 @@ package main
 
 import (
 	"fmt"
-	"k8s.io/perf-tests/clusterloader2/pkg/modifier"
 	"os"
 	"path"
 	"time"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/modifier"
 
 	"k8s.io/kubernetes/pkg/master/ports"
 
@@ -79,6 +80,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubemarkRootKubeConfigPath, "kubemark-root-kubeconfig", "KUBEMARK_ROOT_KUBECONFIG", "",
 		"Path the to kubemark root kubeconfig file, i.e. kubeconfig of the cluster where kubemark cluster is run. Ignored if provider != kubemark")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.APIServerPprofByClientEnabled, "apiserver-pprof-by-client-enabled", "APISERVER_PPROF_BY_CLIENT_ENABLED", true, "Whether apiserver pprof endpoint can be accessed by Kubernetes client.")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.EnableMeasurementLogging, "enable-measurement-logging", "ENABLE_MEASUREMENT_LOGGING", true, "Whether to enable detailed measurement logging.")
 }
 
 func validateClusterFlags() *errors.ErrorList {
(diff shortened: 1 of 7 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -95,6 +95,8 @@ func validateClusterFlags() *errors.ErrorList {
 }
 
 func initFlags() {
+	// Set the default logging level to 2"
+	flags.Set("v", "2")
```
wojtek-t, line 99: I don't think that's the right way to go - see:
https://github.com/kubernetes/perf-tests/pull/1271#issuecomment-633546818

yuanchen8911: @wojtek-t  Sorry, I didn't see your comment. I'm not familiar with test-infra. Does it make sense to set v(2) in  [clusterloader2/api/default.go](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/api/default.go), which is called by [clusterloader2/pkg/test.go](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/test/test.go)?

```
func (conf *Config) SetDefaults() {
       flags.Set("V", defaultLogLevel)
   ...
 }
```

yuanchen8911: @wojtek-t  Sorry, I didn't see your comments.  I agree we need more thoughts on how to manage the verbosity level. 

Since we just wanted to remove the detailed measurement logging information, instead of setting the v levels, we can use the the existing field _EnableLogging_ , which is already defined in struct _WaitForPodOptions_ (_WaitForNodeOptions_, _WaitForPVSOptions_, and _WaitForPVCSOptions_). The current version always sets it  true. The PR adds a new flag _--enable-measurement-logging_ and sets the value of _EnableLogging_ according to the flag. 
```
type WaitForPodOptions struct {
    Selector            *ObjectSelector
    DesiredPodCount     int
    EnableLogging       bool
    CallerName          string
    WaitForPodsInterval time.Duration
}
```

clusterloader2/pkg/config/cluster.go
```diff
@@ -53,6 +53,7 @@ type ClusterConfig struct {
 	// using kubernetes client. If false, clusterloader will avoid collecting kube-apiserver profiles.
 	APIServerPprofByClientEnabled bool
 	KubeletPort                   int
+	EnableMeasurementLogging      bool
```
wojtek-t, line 56: Actually - this really goes against where we would like to go: https://github.com/kubernetes/perf-tests/issues/980

yuanchen8911: How is the field “EnableLogging” in the measurement option structure supposed to be used? It’s hardcoded and always true. There should be a way to set it false. 

If v(2) is preferred, how about simply setting the default  to 2 in api/default.go?

wojtek-t: It's not supposed to be logged - I/we would like to get rid of it and rely on verbosity flags for klog (exactly what you started doing).

I don't think we should be setting the --v in the code, that seems pretty misleading to users.
But I just looked into it, and it seems all our usage is going via this script:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/run-e2e.sh#L37

So if you set it there explicitly, that should work fine for us.

yuanchen8911: Thanks, I’ll take a look.

yuanchen8911:
@wojtek-t  I submitted a new PR that uses V levels and sets the default v(2) in the e2e test script. Please refer to #1298.  Closed this one.

PR #1279, opened 2020-05-27
Migrate to master and non-daemons mode

wojtek-t:


wojtek-t:
/hold

Requires rebase once https://github.com/kubernetes/perf-tests/pull/1278 merges

wojtek-t:
/hold cancel

Actually - let'e merge without waiting for that to have it in todays run.

PR #1280, opened 2020-05-27
Overrides cleanup - part2

wojtek-t:
/assign @mm4tt


clusterloader2/testing/experimental/storage/pod-startup/suite.yaml
```diff
@@ -45,3 +51,9 @@
   overridePaths:
     - testing/experimental/storage/pod-startup/volume-types/secret/override.yaml
     - testing/experimental/storage/pod-startup/max_volumes_per_node/override.yaml
+
+#- identifier: persistent-vol-per-node
```
mm4tt, line 55: Not sure if I follow why we're adding this comments. Could you explain?

wojtek-t: I want to clean those up at once - see: https://github.com/kubernetes/test-infra/pull/17727/files#diff-1db66d4d90bd960aa1cd23062e094030L2

PR #1181, opened 2020-04-20
clusterloader: enable the use of pre-created namespaces

yuanchen8911:
This PR is to enhance clusterloader to support the use of existing namespaces in a test.

A new flag "_EnableExistingNamespaces_" is added. If set, pre-created namespaces with the specified prefix will be used if existing. If the specified namespace doesn't exist, a new one is created. 

Please refer to PR #1173 # and Issue #1169 for the related discussions.

PR #1282, opened 2020-05-28
Add namespace configuration to test configs

jprzychodzen:
This PR enables changes from #1181 and #1173 as a environment variable.

When setting custom namespace prefix, additionally this namespace will not be deleted and we are accepting situation if it's already created. With this variable, one can easily run  tests divided over different CL2 runs over the same namespace.

/assign @mborsz 
/cc @wojtek-t

### DIFF
--- a/clusterloader2/testing/access-tokens/config.yaml
+++ b/clusterloader2/testing/access-tokens/config.yaml
@@ -41,8 +41,17 @@
 {{$ENABLE_RESTART_COUNT_CHECK := DefaultParam .ENABLE_RESTART_COUNT_CHECK false}}
 {{$RESTART_COUNT_THRESHOLD_OVERRIDES:= DefaultParam .RESTART_COUNT_THRESHOLD_OVERRIDES ""}}
 
+# Check if custom namespace prefix is set
+{{$NAMESPACE_PREFIX := DefaultParam .CL2_NAMESPACE_PREFIX ""}}
+
 name: access-tokens
 namespace:
+{{if $NAMESPACE_PREFIX}}
+  deleteStaleNamespaces: false
+  deleteAutomanagedNamespaces: false
+  enableExistingNamespaces: true
+  prefix: {{$NAMESPACE_PREFIX}}
+{{end}}
   number: {{$namespaces}}
 tuningSets:
   - name: Sequence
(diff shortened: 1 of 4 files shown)


clusterloader2/testing/access-tokens/config.yaml
```diff
@@ -41,8 +41,17 @@
 {{$ENABLE_RESTART_COUNT_CHECK := DefaultParam .ENABLE_RESTART_COUNT_CHECK false}}
 {{$RESTART_COUNT_THRESHOLD_OVERRIDES:= DefaultParam .RESTART_COUNT_THRESHOLD_OVERRIDES ""}}
 
+# Check if custom namespace prefix is set
+{{$NAMESPACE_PREFIX := DefaultParam .CL2_NAMESPACE_PREFIX ""}}
+
 name: access-tokens
 namespace:
+{{if $NAMESPACE_PREFIX}}
```
wojtek-t, line 49: I'm not 100% convinced that this should be guarded by NAMESPACE_PREFIX - intuitively if I set namespace-prefix, I dn't expect  those other parameters to be set for me...

jprzychodzen: 1. I agree that there could be a better name about it.
2. There are two configuration that we are interested in - default testing, that we are already using, with deleteStaleNamespace/deleteAutomanagedNamespaces/enableExistingNamespaces being true/true/false, and partial testing, that uses false/false/true.

Solution: rename to CL2_RUN_PARTIAL_TEST_IN_NAMESPACE?

wojtek-t: Yeah - that name sounds better.
@mm4tt - or you maybe have better ideas how to make the config simpler?

mm4tt: Hmm, ideally one could simply re-use the test config as is and specify namespace setting via flags, but it looks like we abandoned this approach. 

Now, the only thing that comes to my mind is to get rid of the `if` block and instead introduce variables for `deleteStaleNamespaces`, `deleteAutomanagedNamespaces`, etc. with proper defaults.  That way, Kuba can run his test with all these envs overridden to achieve what he needs.

If you're going to go with what you currently have, then +1 to naming it in a more self-documenting way.

jprzychodzen:
#1308 was merged, closing this PR as not needed anymore.

PR #1291, opened 2020-05-29
Rename storageUrl to storageURL

mborsz:
It got renamed in https://github.com/kubernetes/perf-tests/commit/5d15f0f796e371284b47a1831a5d06b73afb6bc4#diff-824455981b31a867d75ca77e17833a13L43

I think it's the only place, so we should be good to simply rename here instead of revert of that change.


mm4tt:
/lgtm
/approve

Thanks!

PR #1295, opened 2020-05-29
clusterloader2: add "--from-from-cluster" flag to support in-cluster client config if "--kubeconfig" is empty

gyuho:


clusterloader2/cmd/clusterloader.go
```diff
@@ -61,7 +60,7 @@ var (
 )
 
 func initClusterFlags() {
-	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file (optional, if empty, creates using in-cluster client config)")
```
wojtek-t, line 63: I'm fine with supporting incluster config, but I think this is a bit too implicit.

How about adding an explicit "run-from-cluster" (or sth along those lines) flag and:
- if true, create in-cluster config and validate kubeconfig is unset
- if false, use kubeconfig (and validate it is set)

?

gyuho: Good point. Will add that flag.

clusterloader2/cmd/clusterloader.go
```diff
@@ -61,7 +60,8 @@ var (
 )
 
 func initClusterFlags() {
-	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file")
+	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.KubeConfigPath, "kubeconfig", "KUBECONFIG", "", "Path to the kubeconfig file, --run-from-cluster must be false")
+	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.RunFromCluster, "run-from-cluster", "RUN_FROM_CLUSTER", false, "creates using in-cluster client config, --kubeconfig must be unset")
```
wojtek-t, line 64: nit: the description text seems strange (some noun is missing; I guess you meant

`Whether to use in-cluster client-config to create a client, --kubeconfig must be unset`

clusterloader2/cmd/clusterloader.go
```diff
@@ -83,9 +83,20 @@ func initClusterFlags() {
...
 	errList := errors.NewErrorList()
-	if clusterLoaderConfig.ClusterConfig.KubeConfigPath == "" {
-		errList.Append(fmt.Errorf("no kubeconfig path specified"))
+
+	// if '--run-from-cluster=true', create in-cluster config and validate kubeconfig is unset
+	// if '--run-from-cluster=false', use kubeconfig (and validate it is set)
+	switch clusterLoaderConfig.ClusterConfig.RunFromCluster {
+	case true:
+		if clusterLoaderConfig.ClusterConfig.KubeConfigPath != "" {
+			errList.Append(fmt.Errorf("unexpected kubeconfig path specified %q", clusterLoaderConfig.ClusterConfig.KubeConfigPath))
```
wojtek-t, line 92: ... specified %q when --run-from-cluster is set

PR #1297, opened 2020-05-30
the project is deleted!

wyyxd2017:
the project is deleted!


clusterloader2/vendor/github.com/imdario/mergo/README.md
```diff
@@ -58,7 +58,6 @@ If Mergo is useful to you, consider buying me a coffee, a beer or making a month
 - [go-task/task](https://github.com/go-task/task)
 - [sensu/uchiwa](https://github.com/sensu/uchiwa)
 - [ory/hydra](https://github.com/ory/hydra)
-- [sisatech/vcli](https://github.com/sisatech/vcli)
```
wojtek-t, line 61: This is vendor directory managed by go modules. This particular dependency is coming from github.com/imdario/mergo library. It should be fixed there and then that dependency should be updated.

PR #1299, opened 2020-06-01
[ClusterLoader2][Load test][AutomountServiceAccountToken] Add enabling/disabling flags

jprzychodzen:
Allow to enable/disable AutomountServiceAccountToken per resource type.

/assign @mborsz

### DIFF
--- a/clusterloader2/testing/load/deployment.yaml
+++ b/clusterloader2/testing/load/deployment.yaml
@@ -34,6 +34,9 @@ spec:
             mountPath: /var/secret
       dnsPolicy: Default
       terminationGracePeriodSeconds: 1
+{{if Chance (DefaultParam .CL2_DEPLOYMENTS_DISABLE_SERVICE_ACCOUNT_CHANCE 0)}}
+      automountServiceAccountToken: false
+{{- end}}
       # Add not-ready/unreachable tolerations for 15 minutes so that node
       # failure doesn't trigger pod deletion.
       tolerations:
@@ -52,4 +55,3 @@ spec:
         - name: secret
           secret:
             secretName: {{.BaseName}}-{{.Index}}
-
--- a/clusterloader2/pkg/config/template_functions.go
+++ b/clusterloader2/pkg/config/template_functions.go
@@ -43,6 +43,7 @@ func GetFuncs() template.FuncMap {
 	return template.FuncMap{
 		"AddFloat":      addFloat,
 		"AddInt":        addInt,
+		"Chance":        chance,
 		"DefaultParam":  defaultParam,
 		"DivideFloat":   divideFloat,
 		"DivideInt":     divideInt,
@@ -98,6 +99,11 @@ func toFloat64(val interface{}) float64 {
 		}
 	}
 	panic(fmt.Sprintf("cannot cast %v to float64", val))
+}
+
+// chance returns true/false with some probability for true
+func chance(i interface{}) bool {
+	return rand.Float64() < toFloat64(i)
 }
 
 // randInt returns pseudo-random int in [0, i].
--- a/clusterloader2/pkg/config/template_functions_test.go
+++ b/clusterloader2/pkg/config/template_functions_test.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"bytes"
+	"testing"
+	"text/template"
+)
+
+func TestChance(t *testing.T) {
+	for i := 0; i < 1000; i++ {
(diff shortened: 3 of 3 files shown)


clusterloader2/testing/load/daemonset.yaml
```diff
@@ -1,4 +1,5 @@
 {{$Image := DefaultParam .Image "k8s.gcr.io/pause:3.1"}}
+{{$AutomountServiceAccountToken := DefaultParam .AutomountServiceAccountToken true}}
```
mm4tt, line 2: Can we do instead:
```
{{$AutomountServiceAccountToken := DefaultParam .CL2_LOAD_DAEMON_SETS_AUTOMOUNT_SERVICE_ACCOUNT_TOKEN true}}
```

And revert all the changes in config.yaml?

The rule of thumb I'd like to apply here to make the load config easier to read (mostly shorter): use `templateFillMap` if test uses same template with different configs, e.g. `CpuRequest` is a good example in density test. Otherwise, create `CL2_*` env var inside the template to make the config.yaml shorter.

wojtek-t: +1

I also don't see particular value in allowing to configure mounting SA`s differently for different objects. Rather - we could set some percentage-targets controlled by a single thing (i.e. ~X% of all pods that we create will have default SA mounted).

jprzychodzen: > And revert all the changes in config.yaml?

Sure, make sense. I didn't know that envs are passed to all object templates.

> I also don't see particular value in allowing to configure mounting SA`s differently for different objects.

It seems that DaemonSet should be considered in a different way - we are passing template only for a single namespace, to this percentage will have a huge impact on a performance test.

Also, other Pods are no uniform - for STS we are mounting PVC, for deployment we are mounting configmap/secret. I do not know if this impacts performance of watches.

> Rather - we could set some percentage-targets controlled by a single thing (i.e. ~X% of all pods that we create will have default SA mounted).

Sure, I can change bool flag to % flag, however I still prefer to keep them divided per kind.

clusterloader2/testing/load/deployment.yaml
```diff
@@ -34,6 +34,9 @@ spec:
             mountPath: /var/secret
       dnsPolicy: Default
       terminationGracePeriodSeconds: 1
+{{if not ( Chance (DefaultParam .CL2_LOAD_DEPLOYMENTS_AUTOMOUNT_SERVICE_ACCOUNT_TOKEN_CHANCE 1 ))}}
```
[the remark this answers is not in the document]
jprzychodzen: Yes, it would have the same meaning, but this would mean that parameter would be renamed to something like CL2_LOAD_DEPLOYMENTS_AUTOMOUNT_SERVICE_ACCOUNT_TOKEN_CHANCE_TO_NOT_BE_MOUNTED, and I prefer to have a simpler API from client perspective.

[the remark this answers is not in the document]
wojtek-t: CL2_DEPLOYMENT_DISABLE_AUTOMOUNT_DEFAULT_SERVICE_ACCOUNT_CHANCE

this is actually more user-friendly, because it is enabled by default, to you want to specify percentage chance of changing default behavior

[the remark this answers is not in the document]
jprzychodzen: I was aiming in preventing adding a negation to a flag, but I can get an argument about changin default state.

Updated to `CL2_*_DISABLE_AUTOMOUNT_DEFAULT_SERVICE_ACCOUNT_CHANCE`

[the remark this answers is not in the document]
wojtek-t: What we should target it not "prevent negation', but think what is default.
By default automounting is enabled. So you don't need any action to enabled it. The action is to disable it, and thus the var should be around disabling, not enabling it.

clusterloader2/testing/load/daemonset.yaml
```diff
@@ -28,3 +28,6 @@ spec:
             memory: "10M"
       priorityClassName: daemonset-priorityclass-0 # Name is autogenerated, hence the -0 prefix.
       terminationGracePeriodSeconds: 1
+{{if Chance (DefaultParam .CL2_LOAD_DAEMON_SETS_DISABLE_AUTOMOUNT_DEFAULT_SERVICE_ACCOUNT_CHANCE 0)}}
```
wojtek-t, line 31: It's hard to imagine longer name - can we make it shorter?
- remove load (we will be merging tests sooner or later)
- remove automount (it seems quite obvious from the context)
- remove default probably too

so
CL2_DAEMON_SETS_DISABLE_SERVICE_ACCOUNT_CHANCE

(still long, but not that long)

wojtek-t: BTW - do we really need this param for everything? Maybe having it from just deployments is enough, since these are 90%+ of all pods?

jprzychodzen: Done, fixed too long names.

jprzychodzen: There are two use cases that we are enable for testing:

1. Limited number of default SA to non-zero value (here chance is required and probably we could only depend only on deployments)
1. Disable all default SA - here we need a knob to tune other kinds

For use case 2. we could either provide additional boolean value that will disable all SA, provide per kind boolean value, or provide per kind chance value.

The middle ground between 1. and 2. is having per kind chance value, as done in this PR and this way we can keep consistency. It's also the most flexible way.

jprzychodzen:
Thanks for reviews, it seems that this change is not needed anymore.

PR #1300, opened 2020-06-02
Copy missed kubemark 100 nodes override

jkaniuk:
Copy missed ./testing/load/kubemark/throughput_override.yaml to ./testing/overrides/kubemark_100_nodes.yaml

/cc wojtek-t mm4tt


clusterloader2/testing/overrides/kubemark_100_nodes.yaml
```diff
@@ -0,0 +1 @@
+CL2_LOAD_TEST_THROUGHPUT: 20
```
wojtek-t, line 1: There is already one for it - it's this one: https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/overrides/kubemark_load_throughput.yaml

PR #1302, opened 2020-06-02
Add http method to retrieve scheduler metrics

weiling61:
Add http method to retrieve scheduler metrics. In a secure cluster, remote ssh to master is not always allowed. With the additional http connection, user can run Clusterloader tests on a node within the cluster or on master node.


weiling61:
account link error will pull again

PR #1303, opened 2020-06-02
Add HTTP method to connect to scheduler

weiling61:
Add HTTP method to connect o scheduler to retrieve metrics. In some secure clusters, ssh to its master node is not allowed. With http connection method, user can run ClusterLoader test on a node within the cluster or on master node.


weiling61:
account setting issue

PR #1309, opened 2020-06-03
Add cluster OOMs tracking in clusterloader2

tosi3k:
This PR adds a measurement tracking OOM events in clusterloader2 emitted by `node-problem-detector` and enables it in density and load tests.

Confirmation that it works: [cl2 presubmit job results](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1309/pull-perf-tests-clusterloader2/1268900097278611456/) (there was a deployment with an OOMing container in the PR version of this presubmit).

### DIFF
--- a/clusterloader2/docs/oom_detection.md
+++ b/clusterloader2/docs/oom_detection.md
@@ -1,0 +1,51 @@
+# Clusterloader2 - tracking OOMs
+
+In order to track pods OOMs (out of memory) during clusterloader2 tests
+execution, TestMetrics has been augmented with `ClusterOOMsTracker` measurement.
+This outputs an appropriate summary containing basic information regarding OOMs
+inside the cluster like the PID of OOMing process or the name of a node where
+the OOM happened.
+
+Given that `ClusterOOMsTracker` is based on Kubernetes events and events are
+best effort by their nature, the reported summary is not guaranteed to
+accurately describe what really happened - some of the OOMs may be missed.
+
+## Enabling OOMs tracking
+
+Firstly, ensure that the `TestMetrics` measurement is added to `Starting
+measurements` and `Collecting measurements` steps of your test config. Next,
+add `clusterOOMsTrackerEnabled` parameter and set it to `true` in both steps
+configuration.
+
+Sample configuration in `Starting measurements` step:
+
+```yaml
+- name: Starting measurements
+  measurements:
+    ...
+    - Identifier: TestMetrics
+      Method: TestMetrics
+      Params:
+        action: start
+        clusterOOMsTrackerEnabled: true
+```
+
+Sample configuration in `Collecting measurements` step:
+
+```yaml
+- name: Collecting measurements
+  measurements:
+    ...
+    - Identifier: TestMetrics
+      Method: TestMetrics
+      Params:
+        action: gather
+        clusterOOMsTrackerEnabled: true
+```
+
+## Further debugging steps
+
+`ClusterOOMsTracker` watches for events emitted by `node-problem-detector` when
+an OOM occurs. Such events contain only a fraction of information that may be
+useful for debugging - for more, check `systemd.log` files of an appropriate
+node for the name of OOMing pod/container or the container's memory limit.
--- a/clusterloader2/pkg/measurement/common/bundle/test_metrics.go
+++ b/clusterloader2/pkg/measurement/common/bundle/test_metrics.go
@@ -81,6 +81,9 @@ func createTestMetricsMeasurement() measurement.Measurement {
 	if metrics.systemPodMetrics, err = measurement.CreateMeasurement("SystemPodMetrics"); err != nil {
 		klog.Errorf("%v: systemPodMetrics creation error: %v", metrics, err)
 	}
+	if metrics.clusterOOMsTracker, err = measurement.CreateMeasurement("ClusterOOMsTracker"); err != nil {
+		klog.Errorf("%v: clusterOOMsTracker creation error: %v", metrics, err)
+	}
 	return &metrics
 }
 
@@ -99,6 +102,7 @@ type testMetrics struct {
 	controllerManagerCPUProfile    measurement.Measurement
 	controllerManagerMemoryProfile measurement.Measurement
 	systemPodMetrics               measurement.Measurement
+	clusterOOMsTracker             measurement.Measurement
 }
 
 // Execute supports two actions. start - which sets up all metrics.
@@ -181,6 +185,8 @@ func (t *testMetrics) Execute(config *measurement.Config) ([]measurement.Summary
 		appendResults(&summaries, errList, summary, executeError(t.controllerManagerMemoryProfile.String(), action, err))
 		summary, err = execute(t.systemPodMetrics, config)
 		appendResults(&summaries, errList, summary, executeError(t.systemPodMetrics.String(), action, err))
+		summary, err = execute(t.clusterOOMsTracker, config)
+		appendResults(&summaries, errList, summary, executeError(t.clusterOOMsTracker.String(), action, err))
 	case "gather":
 		summary, err := execute(t.etcdMetrics, actionGatherConfig)
 		appendResults(&summaries, errList, summary, executeError(t.etcdMetrics.String(), action, err))
@@ -210,6 +216,8 @@ func (t *testMetrics) Execute(config *measurement.Config) ([]measurement.Summary
 		appendResults(&summaries, errList, summary, executeError(t.controllerManagerMemoryProfile.String(), action, err))
 		summary, err = execute(t.systemPodMetrics, config)
 		appendResults(&summaries, errList, summary, executeError(t.systemPodMetrics.String(), action, err))
+		summary, err = execute(t.clusterOOMsTracker, config)
+		appendResults(&summaries, errList, summary, executeError(t.clusterOOMsTracker.String(), action, err))
 	default:
 		return summaries, fmt.Errorf("unknown action %v", action)
 	}
--- a/clusterloader2/pkg/measurement/common/ooms_tracker.go
+++ b/clusterloader2/pkg/measurement/common/ooms_tracker.go
@@ -1,0 +1,214 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"regexp"
+	"sync"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	clusterOOMsTrackerEnabledParamName = "clusterOOMsTrackerEnabled"
+	clusterOOMsTrackerName             = "ClusterOOMsTracker"
+	informerTimeout                    = time.Minute
+	oomEventReason                     = "OOMKilling"
+)
+
+var (
+	oomEventMsgRegex = regexp.MustCompile(`Kill process (\d+) \((.+)\) score \d+ or sacrifice child\nKilled process \d+ .+ total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`)
+)
+
+func init() {
+	if err := measurement.Register(clusterOOMsTrackerName, createClusterOOMsTrackerMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", clusterOOMsTrackerName, err)
+	}
+}
+
+func createClusterOOMsTrackerMeasurement() measurement.Measurement {
+	return &clusterOOMsTrackerMeasurement{}
+}
+
+type clusterOOMsTrackerMeasurement struct {
+	selector  *measurementutil.ObjectSelector
+	msgRegex  *regexp.Regexp
+	isRunning bool
+	stopCh    chan struct{}
+	oomsLock  sync.Mutex
+	ooms      []oomEvent
+}
+
+// TODO: Reevaluate if we can add new fields here when node-problem-detector
+// starts using new events.
+type oomEvent struct {
+	Node          string    `json:"node"`
+	Process       string    `json:"process"`
+	ProcessMemory string    `json:"memory"`
+	ProcessID     string    `json:"pid"`
+	Time          time.Time `json:"time"`
+}
+
+func (m *clusterOOMsTrackerMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	clusterOOMsTrackerEnabled, err := util.GetBoolOrDefault(config.Params, clusterOOMsTrackerEnabledParamName, false)
+	if err != nil {
+		return nil, fmt.Errorf("problem with getting %s param: %w", clusterOOMsTrackerEnabledParamName, err)
+	}
+	if !clusterOOMsTrackerEnabled {
+		klog.Info("skipping tracking of OOMs in the cluster")
+		return nil, nil
+	}
+
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, fmt.Errorf("problem with getting %s param: %w", "action", err)
+	}
+
+	switch action {
+	case "start":
+		if err = m.start(config.ClusterFramework.GetClientSets().GetClient()); err != nil {
+			return nil, fmt.Errorf("starting cluster OOMs measurement problem: %w", err)
+		}
+		return nil, nil
+	case "gather":
+		m.oomsLock.Lock()
+		defer m.oomsLock.Unlock()
+		summary, err := m.gather()
+		if err != nil {
+			return nil, fmt.Errorf("gathering cluster OOMs measurement problem: %w", err)
+		}
+		return summary, m.validate()
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (m *clusterOOMsTrackerMeasurement) Dispose() {
+	m.stop()
+}
+
+func (m *clusterOOMsTrackerMeasurement) String() string {
+	return clusterOOMsTrackerName
+}
+
+func (m *clusterOOMsTrackerMeasurement) start(c clientset.Interface) error {
+	if m.isRunning {
+		klog.Infof("%s: cluster OOMs tracking measurement already running", m)
+		return nil
+	}
+	klog.Infof("%s: starting cluster OOMs tracking measurement...", m)
+	m.initFields()
+	// Watching for OOM events from node-problem-detector below.
+	i := informer.NewInformer(
+		c,
+		"events",
+		m.selector,
+		m.handleOOMEvent,
+	)
+	if err := informer.StartAndSync(i, m.stopCh, informerTimeout); err != nil {
+		return fmt.Errorf("problem with OOM events informer starting: %w", err)
+	}
+	return nil
+}
+
+func (m *clusterOOMsTrackerMeasurement) initFields() {
+	m.isRunning = true
+	m.stopCh = make(chan struct{})
+	m.ooms = make([]oomEvent, 0)
+	m.selector = &measurementutil.ObjectSelector{
+		FieldSelector: fields.Set{"reason": oomEventReason}.AsSelector().String(),
+		Namespace:     metav1.NamespaceAll,
+	}
+	m.msgRegex = oomEventMsgRegex
+}
+
+func (m *clusterOOMsTrackerMeasurement) stop() {
+	if m.isRunning {
+		m.isRunning = false
+		close(m.stopCh)
+	}
+}
+
+func (m *clusterOOMsTrackerMeasurement) gather() ([]measurement.Summary, error) {
+	klog.Infof("%s: gathering cluster OOMs tracking measurement", clusterOOMsTrackerName)
+	if !m.isRunning {
+		return nil, fmt.Errorf("measurement %s has not been started", clusterOOMsTrackerName)
+	}
+
+	m.stop()
+
+	content, err := util.PrettyPrintJSON(struct {
+		Ooms []oomEvent `json:"ooms"`
+	}{
+		Ooms: m.ooms,
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("OOMs PrettyPrintJSON problem: %w", err)
+	}
+
+	summary := measurement.CreateSummary(clusterOOMsTrackerName, "json", content)
+	return []measurement.Summary{summary}, nil
+}
+
+func (m *clusterOOMsTrackerMeasurement) validate() error {
+	if len(m.ooms) == 0 {
+		return nil
+	}
+	return fmt.Errorf("OOMs recorded: %+v", m.ooms)
+}
+
+func (m *clusterOOMsTrackerMeasurement) handleOOMEvent(_, obj interface{}) {
+	if obj == nil {
+		return
+	}
+	event, ok := obj.(*corev1.Event)
+	if !ok {
+		return
+	}
+
+	oom := oomEvent{
+		Node: event.InvolvedObject.Name,
+	}
+	if !event.EventTime.IsZero() {
+		oom.Time = event.EventTime.Time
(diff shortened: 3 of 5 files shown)


clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, err
+	}
+
+	switch action {
+	case "start":
+		return []measurement.Summary{}, m.start(config.ClusterFramework.GetClientSets().GetClient())
+	case "gather":
+		summary, err := m.gather()
+		if err != nil {
+			return summary, err
```
mm4tt, line 103: NIT: `return nil, err` is more idiomatic

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	selector  *measurementutil.ObjectSelector
+	msgRegex  *regexp.Regexp
+	isRunning bool
+	stopCh    chan struct{}
+	ooms      []oomEvent
+}
+
+type oomEvent struct {
+	Node          string    `json:"node"`
+	Process       string    `json:"process"`
```
mm4tt, line 76: I'm not sure whether this will be the "process" in the descriptive way we hope to have. E.g. here is an example event.
```
- apiVersion: v1
  count: 16
  eventTime: null
  firstTimestamp: "2020-06-08T15:16:32Z"
  involvedObject:
    kind: Node
    name: gke-cluster-1-default-pool-4b0893d6-mr7x
    uid: gke-cluster-1-default-pool-4b0893d6-mr7x
  kind: Event
  lastTimestamp: "2020-06-08T15:16:47Z"
  message: |-
    (combined from similar events): Memory cgroup out of memory: Kill process 918454 (exe) score 10000 or sacrifice child
    Killed process 918454 (exe) total-vm:7972kB, anon-rss:4kB, file-rss:0kB, shmem-rss:0kB
  metadata:
    creationTimestamp: "2020-06-08T15:16:32Z"
    name: gke-cluster-1-default-pool-4b0893d6-mr7x.16169a7fe54f4e6b
    namespace: default
    resourceVersion: "8174"
    selfLink: /api/v1/namespaces/default/events/gke-cluster-1-default-pool-4b0893d6-mr7x.16169a7fe54f4e6b
    uid: da4e9029-c406-4d86-92d4-d9cb2246d837
  reason: OOMKilling
  reportingComponent: ""
  reportingInstance: ""
  source:
    component: kernel-monitor
    host: gke-cluster-1-default-pool-4b0893d6-mr7x
  type: Warning
```

The Process here will be parsed as "exe" which doesn't give us much. To be clear, I didn't name the pod/container `exe` or anything like that. 
Ideally, we'd have a pod name somewhere here, but it looks like it's not exported anywhere by node-problem-detector. I'm fine with that as long as there is an instruction documented somewhere on how to connect these OOM events with terminated pods, e.g. can we get it from systemd / kubelet logs?

jkaniuk: When checking for OOMs directly, I've also worked with processes, not containers/pods.
It is usually descriptive enough (pod_nanny being an exception).
Documenting how to debug it further would be nice (probably by checking kubelet logs (via log viewer) for restarts).

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	if obj == nil {
+		return
+	}
+	event, ok := obj.(*corev1.Event)
+	if !ok {
+		return
+	}
+	oom := oomEvent{
+		Node: event.InvolvedObject.Name,
+		Time: event.FirstTimestamp.Time,
```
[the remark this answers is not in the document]
tosi3k: Apparently, OOM events do not have EventTime field set, unfortunately. To emit events, `node-problem-detector` launches this part of k8sexporter [here](https://github.com/kubernetes/node-problem-detector/blob/452818cef8be2858f3c37f884b3583ba2463846a/pkg/exporters/k8sexporter/k8s_exporter.go#L72) which (indirectly) uses [this](https://github.com/kubernetes/client-go/blob/be97aaa976ad58026e66cd9af5aaf1b006081f09/tools/record/event.go#L347) function to create an event without `EventTime` field being set. Past manual tests of this change also notice only events with empty `EventTime` field as well.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+		selector: &measurementutil.ObjectSelector{
+			FieldSelector: fieldSelector,
+			Namespace:     metav1.NamespaceAll,
+		},
+		msgRegex: regexp.MustCompile(oomEventMsgRegexPattern),
+	}
+}
+
+type clusterOomsTrackerMeasurement struct {
+	// lock for synchronized access to ooms field
```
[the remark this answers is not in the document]
tosi3k: Done - renamed the field in regard to Jacek's suggestion.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+type oomEvent struct {
+	Node          string    `json:"node"`
+	Process       string    `json:"process"`
+	ProcessMemory string    `json:"memory"`
+	ProcessID     int       `json:"pid"`
+	Time          time.Time `json:"time"`
+}
+
+func (m *clusterOomsTrackerMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	clusterOomsTrackerEnabled, err := util.GetBoolOrDefault(config.Params, clusterOomsTrackerEnabledFlagName, false)
```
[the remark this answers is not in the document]
tosi3k: You're right, it isn't - renamed it accordingly.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+
+	summary := measurement.CreateSummary(clusterOomsTrackerName, "json", content)
+	return []measurement.Summary{summary}, nil
+}
+
+func (m *clusterOomsTrackerMeasurement) validate() error {
+	if len(m.ooms) == 0 {
+		return nil
+	}
+	return fmt.Errorf("OOMs recorded: %+v", m.ooms)
```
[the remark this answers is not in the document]
tosi3k: Done - applied the suggestion file-wide.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (m *clusterOomsTrackerMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	clusterOomsTrackerEnabled, err := util.GetBoolOrDefault(config.Params, clusterOomsTrackerEnabledFlagName, false)
+	if err != nil {
+		return nil, err
+	}
+	if !clusterOomsTrackerEnabled {
+		klog.Info("skipping tracking of OOMs in the cluster")
+		return []measurement.Summary{}, nil
```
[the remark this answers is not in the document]
tosi3k: I checked _bundle/test_metrics.go_ source file and apparently we can!
Changed the return statements accordingly.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	oom := oomEvent{
+		Node: event.InvolvedObject.Name,
+		Time: event.FirstTimestamp.Time,
+	}
+	if match := m.msgRegex.FindStringSubmatch(event.Message); len(match) == 4 {
+		pid, err := strconv.Atoi(match[1])
+		if err != nil {
+			// This would actually never happen because of the regex pattern
+			// in which the first capture group corresponds to digits only.
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/testing/density/config.yaml
```diff
@@ -20,6 +20,7 @@
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_SYSTEM_POD_METRICS:= DefaultParam .ENABLE_SYSTEM_POD_METRICS true}}
+{{$ENABLE_CLUSTER_OOMS_TRACKER := DefaultParam .CL2_ENABLE_CLUSTER_OOMS_TRACKER true}}
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	clusterOomsTrackerEnabledFlagName = "clusterOomsTrackerEnabled"
+	clusterOomsTrackerName            = "ClusterOomsTracker"
+	informerTimeout                   = time.Minute
+	oomEventReason                    = "OOMKilling"
+	oomEventMsgRegexPattern           = `Kill process (\d+) \((.+)\) score \d+ or sacrifice child\nKilled process \d+ .+ total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`
```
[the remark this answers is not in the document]
tosi3k: Good point. Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	Process       string    `json:"process"`
+	ProcessMemory string    `json:"memory"`
+	ProcessID     int       `json:"pid"`
+	Time          time.Time `json:"time"`
+}
+
+func (m *clusterOomsTrackerMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	clusterOomsTrackerEnabled, err := util.GetBoolOrDefault(config.Params, clusterOomsTrackerEnabledFlagName, false)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	m.stopCh = make(chan struct{})
+	m.ooms = make([]oomEvent, 0)
+	// Watching for OOM events from node-problem-detector below.
+	i := informer.NewInformer(
+		c,
+		"events",
+		m.selector,
+		m.handleOomEvent,
+	)
+	return informer.StartAndSync(i, m.stopCh, informerTimeout)
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	m.stop()
+
+	content, err := util.PrettyPrintJSON(struct {
+		Ooms []oomEvent `json:"ooms"`
+	}{
+		Ooms: m.ooms,
+	})
+
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
tosi3k: Using `goerrors.Errorf` in this case, but basically it's done ;).

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	event, ok := obj.(*corev1.Event)
+	if !ok {
+		return
+	}
+	oom := oomEvent{
+		Node: event.InvolvedObject.Name,
+		Time: event.FirstTimestamp.Time,
+	}
+	if match := m.msgRegex.FindStringSubmatch(event.Message); len(match) == 4 {
```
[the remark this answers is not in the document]
tosi3k: Good point. Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+		pid, err := strconv.Atoi(match[1])
+		if err != nil {
+			// This would actually never happen because of the regex pattern
+			// in which the first capture group corresponds to digits only.
+			return
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	if err := measurement.Register(clusterOomsTrackerName, createClusterOomsTrackerMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", clusterOomsTrackerName, err)
+	}
+}
+
+func createClusterOomsTrackerMeasurement() measurement.Measurement {
+	fieldSelector := fields.Set{
+		"reason": oomEventReason,
+	}.AsSelector().String()
+	return &clusterOomsTrackerMeasurement{
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+		selector: &measurementutil.ObjectSelector{
+			FieldSelector: fieldSelector,
+			Namespace:     metav1.NamespaceAll,
+		},
+		msgRegex: regexp.MustCompile(oomEventMsgRegexPattern),
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+}
+
+type clusterOomsTrackerMeasurement struct {
+	// lock for synchronized access to ooms field
+	lock      sync.Mutex
```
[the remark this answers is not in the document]
tosi3k: Done, good point.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (m *clusterOomsTrackerMeasurement) gather() ([]measurement.Summary, error) {
+	klog.Infof("%s: gathering cluster OOMs tracking measurement", clusterOomsTrackerName)
+	if !m.isRunning {
+		return nil, fmt.Errorf("measurement %s has not been started", clusterOomsTrackerName)
+	}
+
+	m.stop()
+
```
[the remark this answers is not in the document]
tosi3k: I put it inside "gather" case of a switch statement inside `Execute` function instead to prevent a situation in which `m.gather()` reads a different `m.ooms` field than `m.validate()`. Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: I put a small Markdown entry about the measurement in the newest version of this PR - PTAL.

@mm4tt @jkaniuk shall I get rid of the "Process" field altogether then? If so, I think we can leave out "ProcessMemory" field as well as it's not the actual memory limit enforced on the pod but total virtual memory used up by the process. I would keep the rest of the `oomEvent` struct fields, though - "Node" is essential for narrowing the search for essential information in logs, "Pid" may be convenient for grepping and "Time" is also convenient for having a time reference when the OOM was recorded.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,216 @@
...
+package common
+
+import (
+	"regexp"
+	"strconv"
+	"sync"
+	"time"
+
+	goerrors "github.com/go-errors/errors"
```
jkaniuk, line 25: Why don't we use idiomatic way instead of a separate library?
https://blog.golang.org/go1.13-errors#TOC_3.3.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,216 @@
...
+	if err := informer.StartAndSync(i, m.stopCh, informerTimeout); err != nil {
+		return goerrors.Errorf("problem with OOM events informer starting: %v", err)
+	}
+	return nil
+}
+
+func (m *clusterOomsTrackerMeasurement) initFields() {
+	m.isRunning = true
+	m.stopCh = make(chan struct{})
+	m.ooms = make([]oomEvent, 0)
```
jkaniuk, line 140: NIT: is there actually a benefit to create slice at all?

clusterloader2/pkg/measurement/common/ooms_tracker.go
jkaniuk: "check `kern.log` as well as `systemd.log` files of an appropriate node for the name of OOMing pod/container or the pod's memory limit, for instance."

Did you verify that this info is there? Do we need to check both files?

Process is quite important, please leave it as it is.
If ProcessMemory is the same value as reported in kern.log, it was handy for me in the past and did match container limit (exceeding it a little bit).
I would not remove any of those fields.

clusterloader2/pkg/measurement/common/ooms_tracker.go
mm4tt: Nice, I wasn't aware that the is an idiomatic way :)
+1 to using it instead of 3rd part library

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Actually, `systemd.log` duplicates information from `kern.log`. Moreover, it contains an addition of logs from kubelet and container runtime regarding the OOMing container/pod, hence I'll just leave `systemd.log` as the suggested file to look at in the doc.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: It's not necessary as one can append to a `nil` slice in Golang as well it seems. The only benefit I see is that when there are no OOMs, `PrettyPrintJSON` will print out a JSON summary like that:
```
{
    ooms: []
}
```
instead of
```
{
    ooms: null
}
```
but it's super minor stuff.

tosi3k: After offline discussion we agreed to leave it as it is. Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
jkaniuk: Yep, pretty printing JSON is a valid reason.

clusterloader2/docs/oom_detection.md
```diff
@@ -0,0 +1,48 @@
...
+        clusterOomsTrackerEnabled: true
+```
+
+## Further debugging steps
+
+`ClusterOomsTracker` watches for events emitted by `node-problem-detector` when
+an OOM occurs. Such events contain only a fraction of information that may be
+useful for debugging - for more, check `systemd.log` files of an appropriate
+node for the name of OOMing pod/container or the pod's memory limit, for
+instance.
```
jkaniuk, line 48: I would remove ", for instance."
Nit: memory limit is set on a container, not a pod, but in a pod definition.

tosi3k: Done.

clusterloader2/docs/oom_detection.md
```diff
@@ -0,0 +1,47 @@
+# Clusterloader2 - tracking OOMs
+
+In order to track pods OOMs (out of memory) during clusterloader2 tests
+execution, TestMetrics has been augmented with `ClusterOomsTracker` measurement.
+This outputs an appropriate summary containing basic information regarding OOMs
+inside the cluster like the PID of OOMing process or the name of a node where
+the OOM happened.
```
wojtek-t, line 7: I think I wrote it somewhere already, but please mention it explicitly here that:
given that oom-detector is based on k8s events and events are best-effort by their nature, the report is not guaranteed to accurately describe what really happened (some ooms may be missed)

clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -81,6 +81,9 @@ func createTestMetricsMeasurement() measurement.Measurement {
 	if metrics.systemPodMetrics, err = measurement.CreateMeasurement("SystemPodMetrics"); err != nil {
 		klog.Errorf("%v: systemPodMetrics creation error: %v", metrics, err)
 	}
+	if metrics.clusterOomsTracker, err = measurement.CreateMeasurement("ClusterOomsTracker"); err != nil {
```
wojtek-t, line 84: Please call it clusterOOMsTracker 

i.e. OOM should all be capital letters.

here and everywhere else (including doc)

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,216 @@
...
+
+const (
+	clusterOomsTrackerEnabledParamName = "clusterOomsTrackerEnabled"
+	clusterOomsTrackerName             = "ClusterOomsTracker"
+	informerTimeout                    = time.Minute
+	oomEventReason                     = "OOMKilling"
+)
+
+var (
+	oomEventMsgRegex = regexp.MustCompile(`Kill process (\d+) \((.+)\) score \d+ or sacrifice child\nKilled process \d+ .+ total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`)
```
wojtek-t, line 45: this is quite fragile.

how do we protect ourselves from this becoming outdated?

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,216 @@
...
+	return &clusterOomsTrackerMeasurement{}
+}
+
+type clusterOomsTrackerMeasurement struct {
+	selector  *measurementutil.ObjectSelector
+	msgRegex  *regexp.Regexp
+	isRunning bool
+	stopCh    chan struct{}
+	ooms      []oomEvent
+	oomsLock  sync.Mutex
```
wojtek-t, line 64: nit: move the mutex above fields that it is protecting

clusterloader2/pkg/measurement/common/ooms_tracker.go
wojtek-t: I really hope that in not that far future once we migrate to new events API, the relatedObject will contain all the information about the Pod itself.
Might be useful to add a comment to evaluate if we can get more once NPD is migrated to new events (hopefully in 1.20?)

jkaniuk:
/lgtm

Could you run additional test on the newest version?

clusterloader2/pkg/measurement/common/ooms_tracker.go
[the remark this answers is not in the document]
wojtek-t: EventTime is used be the new API, FirstTimestamp is the old API one.
This will change in the future once we migrate.

You may want to do the same what we did in pod-startup SLO to address both at the same time:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go#L192-L196

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	oom := oomEvent{
+		Node: event.InvolvedObject.Name,
+		Time: event.FirstTimestamp.Time,
+	}
+
+	match := m.msgRegex.FindStringSubmatch(event.Message)
+	if len(match) != 4 {
+		klog.Warningf(`unexpected OOM event message regex mismatch; event message contents: "%v"`, event.Message)
```
wojtek-t, line 200: should we do something else?
Tht warning may suggest that OOM events have a new format.

Maybe the better thing to do is to basically report an OOM with not all fields being set to present the problem?

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,216 @@
...
+		return
+	}
+
+	pid, err := strconv.Atoi(match[1])
```
wojtek-t, line 204: What's the point of it? Can't we simply write PID as a string and avoid this conversion?
From read POV it doesn't matter much.

wojtek-t:
/hold

I had some comments that I would like to see at least answered before it will merge.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -0,0 +1,214 @@
...
+type clusterOOMsTrackerMeasurement struct {
+	selector  *measurementutil.ObjectSelector
+	msgRegex  *regexp.Regexp
+	isRunning bool
+	stopCh    chan struct{}
+	oomsLock  sync.Mutex
+	ooms      []oomEvent
+}
+
+// Reevaluate if we can add new fields here when node-problem-detector starts
```
wojtek-t, line 66: nit: TODO: Reevaluate...
[to make it explicit that it's an AI]

tosi3k: Done.

Issue #1275, opened 2020-05-27
Support MasterAndNonDaemonSystemPods option in resource gathere


wojtek-t:
Currently resource gathere supports three options: master, master and nodes, master and dns:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go#L39

We use master-and-dns option in large scale to avoid overloading CL2.

However, what we really want is to avoid scraping pods running on every node, but we would really benefit from having metrics from other system pods, like metrics-server.

We should introduce a new option like "master and non daemon-set system pods", that would be scraping all pods from kube-system whose ownerReference kind is not equal to DasemonSet or Nod (static pods, kube-proxy being an example).

wojtek-t:
Done.
And enabled in https://github.com/kubernetes/perf-tests/pull/1279

/close

PR #1324, opened 2020-06-16
Migrate to master and non-daemons mode (1.18)

jkaniuk:
Cherry-pick: #1279

Ref #1275

/cc @wojtek-t, @mborsz

### DIFF
--- a/clusterloader2/testing/density/2000_nodes/override.yaml
+++ b/clusterloader2/testing/density/2000_nodes/override.yaml
@@ -1,1 +1,1 @@
-NODE_MODE: masteranddns
+NODE_MODE: masterandnondaemons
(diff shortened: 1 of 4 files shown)


wojtek-t:
/lgtm
/approve

@jkaniuk @jprzychodzen - why presubmits aren't firing on release branches?

jprzychodzen:
> why presubmits aren't firing on release branches?

It seems reverted in kubernetes/test-infra#16360. Probably we should try to enable them again. To do this we need to book a support of someone to interfere in a case of a broken configuration.

jkaniuk:
> @jkaniuk @jprzychodzen - why presubmits aren't firing on release branches?

I've filed a bug about this but it was deprioritized ;)

PR #1340, opened 2020-06-22
Add a note about Kubemark in OOMs detection doc

tosi3k:
Documents how OOMs tracking works in Kubemark tests.

/sig scalability
/cc jkaniuk
/cc wojtek-t

### DIFF
--- a/clusterloader2/docs/oom_detection.md
+++ b/clusterloader2/docs/oom_detection.md
@@ -10,6 +10,9 @@ Given that `ClusterOOMsTracker` is based on Kubernetes events and events are
 best effort by their nature, the reported summary is not guaranteed to
 accurately describe what really happened - some of the OOMs may be missed.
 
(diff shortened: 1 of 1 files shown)


clusterloader2/docs/oom_detection.md
```diff
@@ -10,6 +10,9 @@ Given that `ClusterOOMsTracker` is based on Kubernetes events and events are
 best effort by their nature, the reported summary is not guaranteed to
 accurately describe what really happened - some of the OOMs may be missed.
 
+For Kubemark clusters where nodes are hollow and `node-problem-detector` is
```
jkaniuk, line 13: In Kubemark tests clusters nodes are hollow ... , so this will only track OOMs occuring in master components.

tosi3k: Done.

PR #1341, opened 2020-06-23
Cherry-pick --skip-steps from master

jprzychodzen:
Original message

Add flag --skip-steps

This change enables runetime test config change - skipping any number of
steps without a need to modify config file.

/assign @mborsz

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/modifier"
 	"os"
 	"path"
 	"time"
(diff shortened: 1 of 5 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"k8s.io/perf-tests/clusterloader2/pkg/modifier"
```
mborsz, line 21: This should go below, but I see it was like that in the origin PR: https://github.com/kubernetes/perf-tests/pull/1247

wojtek-t:
@jprzychodzen - was this cherrypicked to 1.18? If not - please cherrypick, the branches should be more-or-less continuous.

jprzychodzen:
@wojtek-t It was cherrypicked in https://github.com/kubernetes/perf-tests/pull/1256

PR #1342, opened 2020-06-23
Exec command in the middle of performance test.

mborsz:
This can be used to e.g. test cluster-wide operations like rolling update or dumping some debug info.

/assign @wojtek-t

### DIFF
--- a/clusterloader2/pkg/util/util.go
+++ b/clusterloader2/pkg/util/util.go
@@ -71,6 +71,11 @@ func GetMap(dict map[string]interface{}, key string) (map[string]interface{}, er
 	return getMap(dict, key)
 }
 
+// GetStringArray tries to return value from map of type map. If value doesn't exist, error is returned.
+func GetStringArray(dict map[string]interface{}, key string) ([]string, error) {
+	return getStringArray(dict, key)
+}
+
 // GetStringOrDefault tries to return value from map cast to string type. If value doesn't exist default value is used.
 func GetStringOrDefault(dict map[string]interface{}, key string, defaultValue string) (string, error) {
 	value, err := getString(dict, key)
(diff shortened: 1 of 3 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -50,6 +50,9 @@
 # Probe measurements shared parameter
 {{$PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT := DefaultParam .CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT "5m"}}
 
+# Command to be executed
+{{$EXEC_COMMAND := DefaultParam .EXEC_COMMAND nil}}
```
mm4tt, line 54: Let's use the `.CL2_EXEC_COMMAND` to make it configurable via env var

clusterloader2/pkg/util/util.go
```diff
@@ -129,6 +134,28 @@ func getMap(dict map[string]interface{}, key string) (map[string]interface{}, er
 	return mapValue, nil
 }
 
+func getStringArray(dict map[string]interface{}, key string) ([]string, error) {
+	value, exists := dict[key]
+	if !exists || value == nil {
+		return nil, &ErrKeyNotFound{key}
+	}
+
+	mapValueInt, ok := value.([]interface{})
```
wojtek-t, line 143: Why "Int"? These aren't ints..

Did you mean int to be shortcut of internal? If so, let's make it explicit.

Also maybe just "sliceValue" or sth like that?

clusterloader2/pkg/util/util.go
```diff
@@ -129,6 +134,28 @@ func getMap(dict map[string]interface{}, key string) (map[string]interface{}, er
...
+	if !ok {
+		return nil, fmt.Errorf("type assertion error: %v (%T) is not a []string", value, value)
+	}
+
+	var mapValue []string
+	for _, valueInt := range mapValueInt {
```
wojtek-t, line 149: Same here about Int - maybe just "val" or sth like that?

clusterloader2/pkg/util/util.go
mborsz: Int like "Interface", renamed anyway

mborsz:
PTAL

clusterloader2/testing/load/config.yaml
mborsz: It's array, it cannot be modified by ENV anyway

mborsz: But will rename

PR #1346, opened 2020-06-24
Do not track OOMs happening before test start in clusterloader2

tosi3k:
Current OOMs tracker's informer picks up OOMs that are not relevant during performance testing.

/sig scalability
/cc jkaniuk
/cc wojtek-t

### DIFF
--- a/clusterloader2/pkg/measurement/common/ooms_tracker.go
+++ b/clusterloader2/pkg/measurement/common/ooms_tracker.go
@@ -58,6 +58,7 @@ type clusterOOMsTrackerMeasurement struct {
 	selector  *measurementutil.ObjectSelector
 	msgRegex  *regexp.Regexp
 	isRunning bool
+	startTime time.Time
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -132,6 +133,7 @@ func (m *clusterOOMsTrackerMeasurement) start(c clientset.Interface) error {
 	if err := informer.StartAndSync(i, m.stopCh, informerTimeout); err != nil {
 		return fmt.Errorf("problem with OOM events informer starting: %w", err)
 	}
+	m.startTime = time.Now()
```
jkaniuk, line 136: could you move it to initFields under m.isRunning = true ?
I think both should be initialized in the same place.

tosi3k: Makes sense; done.

PR #1348, opened 2020-06-24
Add ignore mechanism for specified processes to OOM tracking

tosi3k:
/sig scalability

PR #1350, opened 2020-06-25
OOM tracking: backport #1348, #1346 and #1340 (release-1.18)

tosi3k:
Backports #1348, #1346 and #1340.

/sig scalability
/cc jkaniuk


jkaniuk:
Could you import it as separate commits (all in this PR) ?

PR #1356, opened 2020-06-26
[1.18] Backport Exec measurement from master

jprzychodzen:
Not merged cleanly, some changes required (different type names)

Original message:

Exec command in the middle of performance test.

This can be used to e.g. test cluster-wide operations like rolling
update or dumping some debug info.

/cc @mborsz 
/assign @mm4tt

### DIFF
--- a/clusterloader2/pkg/util/util.go
+++ b/clusterloader2/pkg/util/util.go
@@ -66,6 +66,11 @@ func GetBool(dict map[string]interface{}, key string) (bool, error) {
 	return getBool(dict, key)
 }
 
+// GetStringArray tries to return value from map of type map. If value doesn't exist, error is returned.
+func GetStringArray(dict map[string]interface{}, key string) ([]string, error) {
+	return getStringArray(dict, key)
+}
+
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/util/util.go
```diff
@@ -111,6 +121,41 @@ func GetBoolOrDefault(dict map[string]interface{}, key string, defaultValue bool
 	return value, err
 }
 
+func getMap(dict map[string]interface{}, key string) (map[string]interface{}, error) {
```
mborsz, line 124: Do we need this? I don't think it's a part of the origin PR

jprzychodzen: Good questions - this is what git suggested to me as a diff when I was cherry-picking d50a7c4983fe4f95e88ace79997ac39cf7958417. You are right, it's not needed. I will remove it.

However I'm not sure what are the best practices if we are doing a number of backports between master and release-1.18 - what is the best way to avoid conflicts.

mm4tt:
/approve
/hold

Leaving for @mborsz  for final LGTM

PR #1357, opened 2020-06-26
Start DnsLookupLatency measurement despite timing out

tosi3k:
Scale tests flake from time to time because of timing out during starting prober pods in `DnsLookupLatency` measurement, as mentioned in https://github.com/kubernetes/kubernetes/issues/87143#issuecomment-573746949. This PR will start the measurement in such scenarios and make `DnsLookupLatency` measurement `gather` step not report a second error (`measurement call DnsLookupLatency - DnsLookupLatency error: measurement DnsLookupLatency has not been started`).

/sig scalability
/cc jkaniuk
/cc wojtek-t
/cc mm4tt

### DIFF
--- a/clusterloader2/pkg/measurement/common/probes/probes.go
+++ b/clusterloader2/pkg/measurement/common/probes/probes.go
@@ -173,6 +173,13 @@ func (p *probesMeasurement) start(config *measurement.Config) error {
 		return err
 	}
 	if err := p.waitForProbesReady(config); err != nil {
+		if err.Error() == wait.ErrWaitTimeout.Error() {
+			// Start the measurement nonetheless as timing out during waiting for
+			// probe pods to be ready is almost always because of a small fraction
+			// of pods not visible via Prometheus.
+			p.startTime = time.Now()
+			return fmt.Errorf("%w (most probable reason: not all probe pods are visible via Prometheus)", err)
+		}
 		return err
 	}
 	p.startTime = time.Now()


clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -173,6 +173,13 @@ func (p *probesMeasurement) start(config *measurement.Config) error {
 		return err
 	}
 	if err := p.waitForProbesReady(config); err != nil {
+		if err.Error() == wait.ErrWaitTimeout.Error() {
+			// Start the measurement nonetheless as timing out during waiting for
+			// probe pods to be ready is almost always because of a small fraction
```
wojtek-t, line 178: Let's not talk abotu "small fraction" - in small clusters there is one probe and then if it's not ready it's 100% of them.

Let's only say that that this usually means that not all probes are visible via promethes.

jkaniuk: it is at least 3 pods for currently supported 100 nodes. 2 nodes for small clusters (supported in the future)

wojtek-t: Still - at least 1 has to be down, and 1 out of 3 is not a "small fraction".

jkaniuk: lets skip "small", just a fraction

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -173,6 +173,13 @@ func (p *probesMeasurement) start(config *measurement.Config) error {
 		return err
 	}
 	if err := p.waitForProbesReady(config); err != nil {
+		if err.Error() == wait.ErrWaitTimeout.Error() {
```
jkaniuk, line 176: What about timing out for all pods? I would expect a full fail then.

current code:

	expectedTargets := p.replicasPerProbe * len(p.config.ProbeLabelValues)
	return prometheus.CheckAllTargetsReady()

my proposition:

	allTargets := p.replicasPerProbe * len(p.config.ProbeLabelValues)
	if fraction > 0 {
		allowedUnhealthyTargets := max(1, fraction * allTargets)
	}

not sure if that fraction should be same for all probe types
what about ping servers and clients - would % work for that as well?

maybe instead of CheckAllTargetsReady, that could be used (not sure if probes are unready or not visible at all in prometheus):

	CheckTargetsReady(k8sClient kubernetes.Interface, selector func(Target) bool, minActiveTargets, minReadyTargets int)

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -173,6 +173,13 @@ func (p *probesMeasurement) start(config *measurement.Config) error {
...
+			// Start the measurement nonetheless as timing out during waiting for
+			// probe pods to be ready is almost always because of a small fraction
+			// of pods not visible via Prometheus.
+			p.startTime = time.Now()
```
jkaniuk, line 180: Setting it in two places separately calls for inconsistencies in the future.

lets restructure:

	err := waitForProbes
	if err != nil && err is fundamental {
		return fmt.Errorf(err)
	}
	// comment that most of the probes started
	startTime = now
	if err != nil {
		return fmt.Errorf(err)
	}
	return nil

mm4tt:
I'm against this change until we understand why the probe targets are not becoming ready. 
Let's cherrypick the probe logging changes to older releases. It should help us understand the issue.

tosi3k:
Closing as the problem doesn't appear to be happening anymore.

/close

Issue #1244, opened 2020-05-08
Run more linters with golangci-lint


oxddr:
Currently we run only `golint` and `gofmt`. We should at least enable linters which are enabled by default in `golangci-lint`: 
```
$ golangci-lint help linters
Enabled by default linters:
deadcode: Finds unused code [fast: true, auto-fix: false]
errcheck: Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases [fast: true, auto-fix: false]
gosimple (megacheck): Linter for Go source code that specializes in simplifying a code [fast: true, auto-fix: false]
govet (vet, vetshadow): Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string [fast: true, auto-fix: false]
ineffassign: Detects when assignments to existing variables are not used [fast: true, auto-fix: false]
staticcheck (megacheck): Staticcheck is a go vet on steroids, applying a ton of static analysis checks [fast: true, auto-fix: false]
structcheck: Finds unused struct fields [fast: true, auto-fix: false]
typecheck: Like the front-end of a Go compiler, parses and type-checks Go code [fast: true, auto-fix: false]
unused (megacheck): Checks Go code for unused constants, variables, functions and types [fast: false, auto-fix: false]
varcheck: Finds unused global variables and constants [fast: true, auto-fix: false]
```

Enabling some of them (I've checked `errcheck` and `deadcode`) would require fixing outstanding issues. 

Work here should be done incrementally, one-by-one. 

/good-first-issue
/help-wanted

bitcodr:
@oxddr I'd like to work on it

oxddr:
@bitcodr that's great to hear. I think the best way to proceed is to pick up one of the linters from above (e.g. `deadcode` of `errcheck`), enable it and fix all reported issue and send me a PR for a review.

wfernandes:
Hey @bitcodr Are you perhaps still working on this? If so feel free to `/assign` this issue. Thanks.

wojtek-t:
Adding gofmt would be especially useful.

satishbellapu:
@oxddr  Cleared errors for few of the modules, need some help on "clusterloader2" on few errors,

[...]

PR #1363, opened 2020-07-01
Run more linters with golangci-lint #1244

satishbellapu:
In a series of iterative commits, this is the first commit enabling gofmt and also cleared few of other lint errors with respective to deadcode.


.golangci.yml
```diff
@@ -5,7 +5,7 @@ linters:
   disable-all: true
   enable:
     - gofmt
-    - golint
     # TODO(oxddr): enable more checkers
+    # - golint
     # - deadcode
-    # - errcheck
+    # - errcheck
```
mm4tt, line 11: NIT: Add a newline

clusterloader2/cmd/clusterloader.go
```diff
@@ -149,7 +149,7 @@ func completeConfig(m *framework.MultiClientSet) error {
 		}
 	}
 	if len(clusterLoaderConfig.ClusterConfig.MasterIPs) == 0 {
-		masterIPs, err := util.GetMasterIPs(m.GetClient(), corev1.NodeExternalIP)
+		masterIPs, err := util.GetMasterIPs(m.GetClient(), NodeExternalIP)
```
mm4tt, line 152: Why introducing another, redundant constant instead of using corev1.NodeExternalIp?

clusterloader2/pkg/config/template_provider.go
```diff
@@ -225,7 +225,10 @@ func GetMapping(clusterLoaderConfig *ClusterLoaderConfig) (map[string]interface{
 	if err != nil {
 		return nil, errors.NewErrorList(goerrors.Errorf("mapping creation error: %v", err))
 	}
-	MergeMappings(mapping, envMapping)
+	err = MergeMappings(mapping, envMapping)
+	if err != nil {
+		return nil, errors.NewErrorList(fmt.Errorf("mapping creation error: %v", err))
```
mm4tt, line 230: Use %w to wrap errors

network/benchmarks/netperf/nptest/nptest.go
```diff
@@ -509,13 +509,18 @@ func (t *NetPerfRPC) ReceiveOutput(data *WorkerOutput, reply *int) error {
 
 func serveRPCRequests(port string) {
 	baseObject := new(NetPerfRPC)
-	rpc.Register(baseObject)
-	rpc.HandleHTTP()
-	listener, e := net.Listen("tcp", ":"+port)
-	if e != nil {
-		log.Fatal("listen error:", e)
+	err := rpc.Register(baseObject)
+	if err == nil {
```
mm4tt, line 513: if err != nil {
  log.Fatal(...)
}
...
?

mm4tt:
Thanks!

.golangci.yml
```diff
@@ -5,7 +5,7 @@ linters:
   disable-all: true
   enable:
     - gofmt
-    - golint
```
wojtek-t, line 8: Let's not remove golint

.golangci.yml
wojtek-t, line 11: nit - leave the end-line as it was

clusterloader2/cmd/clusterloader.go
```diff
@@ -50,6 +48,8 @@ import (
 const (
 	dashLine        = "--------------------------------------------------------------------------------"
 	nodesPerClients = 100
+	NodeExternalIP  = "ExternalIP"
+	NodeInternalIP  = "InternalIP"
```
wojtek-t, line 52: I don't like this change - let's use the constants from corev1 instead of introducing our own ones.

clusterloader2/pkg/tuningset/global_qps_load.go
```diff
@@ -59,7 +60,10 @@ func newGlobalQPSLoad(params *api.GlobalQPSLoad) *globalQPSLoad {
 func (ql *globalQPSLoad) Execute(actions []func()) {
 	var wg wait.Group
 	for i := range actions {
-		ql.limiter.Wait(context.TODO())
+		err := ql.limiter.Wait(context.TODO())
+		if err != nil {
+			log.Fatal("QPS load exec error ", err)
```
wojtek-t, line 65: we don't use log in clusterloader - let's use klog if you need

that said - I don't think it deserves fatal - let's simply log an error.

network/benchmarks/netperf/nptest/nptest.go
wojtek-t, line 513: if err != nil {
  return
}
// the rest here

[Let's avoid long nesting where possible.]

perfdash/downloader.go
```diff
@@ -53,12 +48,12 @@ func serveHTTPObject(res http.ResponseWriter, req *http.Request, obj interface{}
 	if err != nil {
 		res.Header().Set("Content-type", "text/html")
 		res.WriteHeader(http.StatusInternalServerError)
-		res.Write([]byte(fmt.Sprintf("<h3>Internal Error</h3><p>%v", err)))
+		fmt.Fprintln(res, "<h3>Internal Error</h3><p>%v", err)
```
wojtek-t, line 51: Why do we need to change this?

clusterloader2/cmd/clusterloader.go
satishbellapu: This is the error we are getting when used the corev1 constants...
 
perf-tests/clusterloader2/cmd/clusterloader.go:153:54: cannot use corev1.NodeExternalIP (constant "ExternalIP" of type v1.NodeAddressType) as v1.NodeAddressType value in argument to util.GetMasterIPs

.golangci.yml
satishbellapu: All errors are not cleared with golint at, hence commented for now, let me know if you want me to enable ?

clusterloader2/cmd/clusterloader.go
satishbellapu: This is the error we are getting when used the corev1 constants...
 
perf-tests/clusterloader2/cmd/clusterloader.go:153:54: cannot use corev1.NodeExternalIP (constant "ExternalIP" of type v1.NodeAddressType) as v1.NodeAddressType value in argument to util.GetMasterIPs

clusterloader2/pkg/config/template_provider.go
satishbellapu: Sorry throughout the clusterloader2 the errors are wrapped with %v, so i want to blend in.

clusterloader2/pkg/tuningset/global_qps_load.go
satishbellapu: Added klog.Errorf

perfdash/downloader.go
satishbellapu: This is fail on - errcheck lint,

perfdash/downloader.go:51:12: Error return value of `res.Write` is not checked (errcheck)
		res.Write([]byte(fmt.Sprintf("<h3>Internal Error</h3><p>%v", err)))

.golangci.yml
wojtek-t: Let's enable it - we'are still not blocking on it for nw.

clusterloader2/cmd/clusterloader.go
wojtek-t: This is strange - but probably some casting would be enough to fix. Let's not introduce new constants for existing ones.

PR #1365, opened 2020-07-01
Fix wrong type in Errorf format (1.17)

jkaniuk:
Fix wrong type in Errorf format

/cc @mm4tt


jkaniuk:
Checking presubmit
/hold

PR #1368, opened 2020-07-02
Remove duplicate OOM event reports in OOM tracker

tosi3k:
Perform deduplication of OOM events with the same `ResourceVersion`.

/sig scalability
/cc mm4tt
/cc jkaniuk
/cc wojtek-t

### DIFF
--- a/clusterloader2/pkg/measurement/common/ooms_tracker.go
+++ b/clusterloader2/pkg/measurement/common/ooms_tracker.go
@@ -56,14 +56,15 @@ func createClusterOOMsTrackerMeasurement() measurement.Measurement {
 }
 
 type clusterOOMsTrackerMeasurement struct {
-	selector       *measurementutil.ObjectSelector
-	msgRegex       *regexp.Regexp
-	isRunning      bool
-	startTime      time.Time
-	stopCh         chan struct{}
-	processIgnored map[string]bool
-	oomsLock       sync.Mutex
-	ooms           []oomEvent
+	selector                *measurementutil.ObjectSelector
+	msgRegex                *regexp.Regexp
+	isRunning               bool
+	startTime               time.Time
+	stopCh                  chan struct{}
+	lock                    sync.Mutex
+	processIgnored          map[string]bool
+	resourceVersionRecorded map[string]bool
+	ooms                    []oomEvent
 }
 
 // TODO: Reevaluate if we can add new fields here when node-problem-detector
@@ -98,8 +99,8 @@ func (m *clusterOOMsTrackerMeasurement) Execute(config *measurement.Config) ([]m
 		}
 		return nil, nil
 	case "gather":
-		m.oomsLock.Lock()
-		defer m.oomsLock.Unlock()
+		m.lock.Lock()
+		defer m.lock.Unlock()
 		return m.gather()
 	default:
 		return nil, fmt.Errorf("unknown action %v", action)
@@ -145,6 +146,7 @@ func (m *clusterOOMsTrackerMeasurement) initFields(config *measurement.Config) e
 		Namespace:     metav1.NamespaceAll,
 	}
 	m.msgRegex = oomEventMsgRegex
+	m.resourceVersionRecorded = make(map[string]bool)
 
 	ignoredProcessesString, err := util.GetStringOrDefault(config.Params, clusterOOMsIgnoredProcessesParamName, "")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -213,6 +215,15 @@ func (m *clusterOOMsTrackerMeasurement) handleOOMEvent(_, obj interface{}) {
 		return
 	}
 
+	m.lock.Lock()
```
mm4tt, line 218: Let's do it more granularly, i.e.
```
  m.lock.Lock()
  m.resourceVersionRecorded[event.ObjectMeta.ResourceVersion] = true
  m.lock.Unlock()
```

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -213,6 +215,15 @@ func (m *clusterOOMsTrackerMeasurement) handleOOMEvent(_, obj interface{}) {
...
+	defer m.lock.Unlock()
+
+	if m.resourceVersionRecorded[event.ObjectMeta.ResourceVersion] {
+		klog.Warningf("caught an OOM event with already recorded resource version: %+v", event)
```
mm4tt, line 222: Given that it may happen on the re-list, we shouldn't log this as warning. Warning means that something unexpected happened, it's not the case. Let's log as info or not log at all

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -230,7 +241,5 @@ func (m *clusterOOMsTrackerMeasurement) handleOOMEvent(_, obj interface{}) {
 		klog.Warningf(`unrecognized OOM event message pattern; event message contents: "%v"`, event.Message)
 	}
 
-	m.oomsLock.Lock()
```
mm4tt, line 233: Same here, let's lock as narrowly as possible

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -145,6 +146,7 @@ func (m *clusterOOMsTrackerMeasurement) initFields(config *measurement.Config) e
 		Namespace:     metav1.NamespaceAll,
 	}
 	m.msgRegex = oomEventMsgRegex
+	m.resourceVersionRecorded = make(map[string]bool)
```
jkaniuk, line 149: Do we need a map?
Is there a guarantee that events will be delivered with incrementing RV, so that keeping last seen RV would be enough?

jkaniuk: If so - why not compare with the last Event in ooms[] slice, that would be enough with O(1) access.

clusterloader2/pkg/measurement/common/ooms_tracker.go
jkaniuk: I disagree - there are few lines along each other, all needing a lock. Why would  you lock them separately?
It is not a heavy parallel scenario.

jkaniuk: Keep in mind that acquiring a lock has its cost too. There should be a balance between number of locks and amount of code guarded by it.

tosi3k: I share Jacek's view on that - critical section ought to be as small as possible but in this particular case we need to have mutual exclusion applied to the whole piece of code (and, as stated, it shouldn't hurt us performance-wise).

If we were to do it more granularly (meaning we lock the read to `m.resourceVersionRecorded` and the write to `m.ooms` separately), we might end up with an undesired scenario in which two goroutines executing `handleOOMEvent` method that receive the same OOM event (with the same `ResourceVersion` that was previously unregistered) save that very OOM to `m.ooms` twice.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: > Is there a guarantee that events will be delivered with incrementing RV, so that keeping last seen RV would be enough?

Not so sure about that, from [K8s docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#the-resourceversion-parameter):

>For example, clients must not assume resource versions are numeric, and may only compare two resource version for equality (i.e. must not compare resource versions for greater-than or less-than relationships).

That said, we might just save the last registered resource version and check it for equality whenever we meet a new OOM event in `handleOOMEvent` function.

wojtek-t: In general, we're sending events in the order. The exceptions are events from resync (which we explicitly disable setting period to zero) and the events coming from relist (which we actually observe).

So while the map is not the most efficient implementation, I actually think the map is the simplest one and we can proceed with it.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Done - I replaced logging with a comment containing a rephrasing of your remark ;).

clusterloader2/pkg/measurement/common/ooms_tracker.go
mm4tt: I took a look at the Informer's code and there is only one go-routing for handling the events. It makes sense, otherwise a scenario like Antek described could happen. Given that, I agree with you that a single Lock/Unlock is better here

PR #1374, opened 2020-07-06
* Added support for GitHub bearer token

micahhausler:
GitHub's API will rate limit unauthenticated requests from IPs with high request
rates, but will allow requests that include authentication. This adds
support for reading in a `GITHUB_TOKEN` environment variable and adding
it as a bearer token to requests against the GitHub API

I got the message when trying to test this tool on my corporate network.
```json
{
  "message": "API rate limit exceeded for xxx.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```

### DIFF
--- a/perfdash/github-configs-fetcher.go
+++ b/perfdash/github-configs-fetcher.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"k8s.io/klog"
 	"net/http"
+	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


shyamjvs:
/lgtm
/approve

Thanks Micah!

PR #1370, opened 2020-07-03
Slow API calls - backport #1262, #1258, #1213, #1229, #1209, #915, #923 and #914 (release-1.17)

tosi3k:
Backport #1262, #1258, #1213, #1229, #1209, #915, #923 and #914 to get rid of API call measurement flakes due to single slow API calls breaking the latency SLOs.

I needed to cherry-pick #914 which gets rid of a deprecated `APIResponsiveness` altogether for a much easier backporting process - for 1.17 release already we do not collect anything of interest therein (take a look at artifacts in [testgrid/#gce-cos-1.17-scalability-100](https://k8s-testgrid.appspot.com/sig-scalability-gce#gce-cos-1.17-scalability-100) and [testgrid/#kubemark-1.17-500](https://k8s-testgrid.appspot.com/sig-scalability-kubemark#kubemark-1.17-500)).

1.17 test jobs will still require an override file changing the default value of `ALLOWED_SLOW_API_CALLS`.

/sig scalability
/cc jkaniuk
/cc mm4tt
/cc wojtek-t

PR #1379, opened 2020-07-08
Add override file for slow API calls allowance

tosi3k:
This is to make it possible for 1.17 jobs to override the default value of `ALLOWED_SLOW_API_CALLS` which is set to 0 in density and load configs.

For the record: PR with slow API calls improvement cherry-pick in `release-1.17`: https://github.com/kubernetes/perf-tests/pull/1370.

/sig scalability
/cc mm4tt


mm4tt:
Can we instead name the variable CL2_ALLOWED_SLOW_API_CALLS so it can be set via env variable?

tosi3k:
> Can we instead name the variable CL2_ALLOWED_SLOW_API_CALLS so it can be set via env variable?

We could, but it'd need backporting https://github.com/kubernetes/perf-tests/pull/921. I could give it a try soon, as that would make lots of future backports a bit easier.

tosi3k:
Closing this PR, as we've backported `CL2_*` env var handling to release-1.17 in #1391 to enable slow API calls in a bit more convenient fashion.

PR #1385, opened 2020-07-15
Use gofmt to fix lint issues and update perfdash README

mirahmed753:
In this PR: 
- I fix the lint issues from PR #1382. My gofmt/golint was not working properly. The PR was merged, but it failed the verify-lint test
- I updated the readme to reflect that perfdash is no longer using godeps since we moved to go modules

PR #1396, opened 2020-07-17
Add AWS S3 functionality for perfdash

mirahmed753:
In this PR, I added functionality to get data from AWS S3 by:
- adding s3 specific functions to pull build numbers and files from the specified s3 bucket
- adding s3 specific flags (flag to specify region)
- adding mode flag to specify whether to get data from s3 or gcs (the default is gcs)

I also moved s3 and gcs functions to their own files (s3_metrics_bucket.go and gcs_metrics_bucket.go). I tested the changes on local and on an EKS cluster; metrics were successfully pulled from S3 and displayed on perfdash. 😄

### DIFF
--- a/perfdash/gcs_metrics_bucket.go
+++ b/perfdash/gcs_metrics_bucket.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"strconv"
+	"strings"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	"k8s.io/klog"
+)
+
+// GCSMetricsBucket that creates a Google Cloud Storage client to fetch data.
+type GCSMetricsBucket struct {
+	client  *storage.Client
+	bucket  *storage.BucketHandle
+	logPath string
+}
+
+// NewGCSMetricsBucket creates a new GCSMetricsBucket.
+func NewGCSMetricsBucket(bucket, path, credentialPath string) (MetricsBucket, error) {
+	ctx := context.Background()
+	authOpt := option.WithoutAuthentication()
+	if credentialPath != "" {
+		authOpt = option.WithCredentialsFile(credentialPath)
+	}
+	c, err := storage.NewClient(ctx, authOpt)
+	if err != nil {
+		return nil, err
+	}
+	b := c.Bucket(bucket)
+	return &GCSMetricsBucket{
+		client:  c,
+		bucket:  b,
+		logPath: path,
+	}, nil
+}
+
+// GetBuildNumbers fetches the build numbers from a GCS Bucket.
+func (b *GCSMetricsBucket) GetBuildNumbers(job string) ([]int, error) {
+	var builds []int
+	ctx := context.Background()
+	jobPrefix := joinStringsAndInts(b.logPath, job) + "/"
+	klog.Infof("%s", jobPrefix)
+	it := b.bucket.Objects(ctx, &storage.Query{
+		Prefix:    jobPrefix,
+		Delimiter: "/",
+	})
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		if attrs.Prefix == "" {
+			continue
+		}
+		build := strings.TrimPrefix(attrs.Prefix, jobPrefix)
+		build = strings.TrimSuffix(build, "/")
+		buildNo, err := strconv.Atoi(build)
+		if err != nil {
+			return nil, fmt.Errorf("unknown build name convention: %s", build)
+		}
+		builds = append(builds, buildNo)
+	}
+	return builds, nil
+}
+
+// ListFilesInBuild fetches the files in the build from GCS.
+func (b *GCSMetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	var files []string
+	ctx := context.Background()
+	jobPrefix := joinStringsAndInts(b.logPath, job, buildNumber, prefix)
+	it := b.bucket.Objects(ctx, &storage.Query{
+		Prefix: jobPrefix,
+	})
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		files = append(files, attrs.Name)
+	}
+	return files, nil
+}
+
+// ReadFile reads the file contents from the GCS bucket.
+func (b *GCSMetricsBucket) ReadFile(job string, buildNumber int, path string) ([]byte, error) {
+	ctx := context.Background()
+	filePath := joinStringsAndInts(b.logPath, job, buildNumber, path)
+	rc, err := b.bucket.Object(filePath).NewReader(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer rc.Close()
+
+	data, err := ioutil.ReadAll(rc)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
--- a/perfdash/metrics-downloader.go
+++ b/perfdash/metrics-downloader.go
@@ -17,9 +17,7 @@ limitations under the License.
 package main
 
 import (
-	"context"
 	"fmt"
-	"io/ioutil"
 	"path"
 	"path/filepath"
 	"sort"
@@ -28,15 +26,12 @@ import (
 	"sync"
 
 	"k8s.io/klog"
-
-	"cloud.google.com/go/storage"
-	"google.golang.org/api/iterator"
-	"google.golang.org/api/option"
 	"k8s.io/kubernetes/test/e2e/perftype"
 )
 
 // DownloaderOptions is an options for Downloader.
 type DownloaderOptions struct {
+	Mode               string
 	ConfigPaths        []string
 	GithubConfigDirs   []string
 	DefaultBuildsCount int
@@ -80,7 +75,7 @@ func (g *Downloader) getData() (JobToCategoryData, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to refresh config: %v", err)
 	}
-	klog.Infof("Getting Data from GCS...")
+	klog.Infof("Getting Data from %v...", options.Mode)
 	result := make(JobToCategoryData)
 	var resultLock sync.Mutex
 	var wg sync.WaitGroup
@@ -135,7 +130,7 @@ func (g *Downloader) getJobData(wg *sync.WaitGroup, result JobToCategoryData, re
 					for _, artifact := range artifacts {
 						metricsFileName := filepath.Base(artifact)
 						resultCategory := getResultCategory(metricsFileName, filePrefix, categoryLabel, artifacts)
-						testDataResponse, err := g.MetricsBkt.GetFile(job, buildNumber,
+						testDataResponse, err := g.MetricsBkt.ReadFile(job, buildNumber,
 							fmt.Sprintf("artifacts/%v", metricsFileName))
 						if err != nil {
 							klog.Errorf("Error when reading response Body: %v", err)
@@ -180,104 +175,7 @@ func getBuildData(result JobToCategoryData, prefix string, category string, labe
 type MetricsBucket interface {
 	GetBuildNumbers(job string) ([]int, error)
 	ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error)
-	GetFile(job string, buildNumber int, path string) ([]byte, error)
-}
-
-// GCSMetricsBucket that creates a Google Cloud Storage client to fetch data.
-type GCSMetricsBucket struct {
-	client  *storage.Client
-	bucket  *storage.BucketHandle
-	logPath string
-}
-
-// NewGCSMetricsBucket creates a new GCSMetricsBucket.
-func NewGCSMetricsBucket(bucket, path, credentialPath string) (MetricsBucket, error) {
-	ctx := context.Background()
-	authOpt := option.WithoutAuthentication()
-	if credentialPath != "" {
-		authOpt = option.WithCredentialsFile(credentialPath)
-	}
-	c, err := storage.NewClient(ctx, authOpt)
-	if err != nil {
-		return nil, err
-	}
-	b := c.Bucket(bucket)
-	return &GCSMetricsBucket{
-		client:  c,
-		bucket:  b,
-		logPath: path,
-	}, nil
-}
-
-// GetBuildNumbers fetches the build numbers from a GCS Bucket.
-func (b *GCSMetricsBucket) GetBuildNumbers(job string) ([]int, error) {
-	var builds []int
-	ctx := context.Background()
-	jobPrefix := joinStringsAndInts(b.logPath, job) + "/"
-	klog.Infof("%s", jobPrefix)
-	it := b.bucket.Objects(ctx, &storage.Query{
-		Prefix:    jobPrefix,
-		Delimiter: "/",
-	})
-	for {
-		attrs, err := it.Next()
-		if err == iterator.Done {
-			break
-		}
-		if err != nil {
-			return nil, err
-		}
-		if attrs.Prefix == "" {
-			continue
-		}
-		build := strings.TrimPrefix(attrs.Prefix, jobPrefix)
-		build = strings.TrimSuffix(build, "/")
-		buildNo, err := strconv.Atoi(build)
-		if err != nil {
-			return nil, fmt.Errorf("unknown build name convention: %s", build)
-		}
-		builds = append(builds, buildNo)
-	}
-	return builds, nil
-}
-
-// ListFilesInBuild fetches the files in the build from GCS.
-func (b *GCSMetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
-	var files []string
-	ctx := context.Background()
-	jobPrefix := joinStringsAndInts(b.logPath, job, buildNumber, prefix)
-	it := b.bucket.Objects(ctx, &storage.Query{
-		Prefix: jobPrefix,
-	})
-	for {
-		attrs, err := it.Next()
-		if err == iterator.Done {
-			break
-		}
-		if err != nil {
-			return nil, err
-		}
-
-		files = append(files, attrs.Name)
-	}
-	return files, nil
-}
-
-// GetFile fetches the file from the GCS bucket.
-func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]byte, error) {
-	ctx := context.Background()
-	filePath := joinStringsAndInts(b.logPath, job, buildNumber, path)
-	rc, err := b.bucket.Object(filePath).NewReader(ctx)
-	if err != nil {
-		return nil, err
-	}
-	defer rc.Close()
-
-	data, err := ioutil.ReadAll(rc)
-	if err != nil {
-		return nil, err
-	}
-	return data, nil
+	ReadFile(job string, buildNumber int, path string) ([]byte, error)
 }
 
 func joinStringsAndInts(pathElements ...interface{}) string {
--- a/perfdash/perfdash.go
+++ b/perfdash/perfdash.go
@@ -23,15 +23,17 @@ import (
 	"os"
 	"time"
 
-	"k8s.io/klog"
-
 	"github.com/spf13/pflag"
+	"k8s.io/klog"
 )
 
 const (
 	pollDuration = 10 * time.Minute
 	errorDelay   = 10 * time.Second
 	maxBuilds    = 100
+
+	s3Mode  = "s3"
+	gcsMode = "gcs"
 )
 
 var options = &DownloaderOptions{}
@@ -51,9 +53,13 @@ var (
 
 	// Google GCS Specific flags
 	credentialPath = pflag.String("credentialPath", "", "Path to the gcs credential json")
+
+	// AWS S3 Specific flags
+	awsRegion = pflag.String("aws-region", "us-west-2", "AWS region of the S3 bucket")
 )
 
 func initDownloaderOptions() {
+	pflag.StringVar(&options.Mode, "mode", gcsMode, "Storage provider from which to download metrics from. Options are 's3' or 'gcs'. The default is 'gcs'.")
 	pflag.BoolVar(&options.OverrideBuildCount, "force-builds", false, "Whether to enforce number of builds to process as passed via --builds flag. "+
 		"This would override values defined by \"perfDashBuildsCount\" label on prow job")
 	pflag.IntVar(&options.DefaultBuildsCount, "builds", maxBuilds, "Total builds number")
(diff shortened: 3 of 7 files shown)


perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+// NewS3MetricsBucket creates a new S3MetricsBucket.
+func NewS3MetricsBucket(bucket, path, region string) (MetricsBucket, error) {
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(region)},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	s3Client := s3.New(sess)
```
ellistarn, line 304: consider inlining this.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
 
+// S3MetricsBucket that creates a AWS S3 client to fetch data.
+type S3MetricsBucket struct {
+	client  *s3.S3
+	bucket  *string
+	logPath string
+}
+
+// NewS3MetricsBucket creates a new S3MetricsBucket.
+func NewS3MetricsBucket(bucket, path, region string) (MetricsBucket, error) {
```
ellistarn, line 296: just a nit, if it has a path it isn't a bucket anymore. Since you're specifying a path, this object is more of an S3MetricsDirectory.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+
+	return &S3MetricsBucket{
+		client:  s3Client,
+		bucket:  aws.String(bucket),
+		logPath: path,
+	}, nil
+}
+
+// GetBuildNumbers fetches the build numbers from an S3 Bucket.
+func (fetcher *S3MetricsBucket) GetBuildNumbers(job string) ([]int, error) {
```
ellistarn, line 314: consider naming fetcher "b" or "s". It's typical go style for pointer receivers to be a single letter for brevity.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+
+	return builds, nil
+}
+
+// ListFilesInBuild fetches the files in the build from S3.
+func (fetcher *S3MetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	var files []string
+	jobPrefix := joinStringsAndInts(fetcher.logPath, job, buildNumber, prefix)
+
+	params := &s3.ListObjectsV2Input{
```
ellistarn, line 347: consider inlining

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+		Bucket: fetcher.bucket,
+		Prefix: aws.String(jobPrefix),
+	}
+
+	result := func(objects *s3.ListObjectsV2Output, lastPage bool) bool {
```
ellistarn, line 352: consider inlining.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+		for _, key := range objects.Contents {
+			files = append(files, *key.Key)
+		}
+
+		return true
```
ellistarn, line 357: I'm not familiar with how this paging functionality works. My guess is that returning true will exit early and potentially not list all of the pages. It's a pretty huge assumption about how this works, but should this be `return lastPage`?

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+	}
+
+	err := fetcher.client.ListObjectsV2Pages(params, result)
+	if err != nil {
```
ellistarn, line 361: When you have a single return that is an error, consider the more concise pattern
```
if err := b.client.ListObjectsV2Pages(params, result); err != nil {
  return nil, fmt.Errorf("while listing objects %v", err.Error()) // this error annotation pattern yields really clean error messages.
}
```

perfdash/metrics-downloader.go
```diff
@@ -19,19 +19,25 @@ package main
...
-	"k8s.io/klog"
 	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
 
+	"k8s.io/klog"
+
```
micahhausler, line 32: nit: most Go code will just have 2 groups of import blocks; stdlib and other.

perfdash/metrics-downloader.go
micahhausler: I'd support renaming `path` to `pathPrefix` or just `prefix`

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+	err := fetcher.client.ListObjectsV2Pages(params, result)
+	if err != nil {
+		return nil, err
+	}
+
+	return files, nil
+}
+
+// GetFile fetches the file from the S3 bucket.
+func (fetcher *S3MetricsBucket) GetFile(job string, buildNumber int, path string) ([]byte, error) {
```
ellistarn, line 369: So there are two paths here? the logPath is the prefix and the job itself has its own path? Seems like an awkward interface to me. Perhaps not worth the effort to change for both providers.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
 	return data, nil
 }
 
+// S3MetricsBucket that creates a AWS S3 client to fetch data.
+type S3MetricsBucket struct {
```
ellistarn, line 289: All of this should go in a separate s3_metrics_bucket.go file.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+	}
+
+	return files, nil
+}
+
+// GetFile fetches the file from the S3 bucket.
+func (fetcher *S3MetricsBucket) GetFile(job string, buildNumber int, path string) ([]byte, error) {
+	filePath := joinStringsAndInts(fetcher.logPath, job, buildNumber, path)
+
+	params := &s3.GetObjectInput{
```
ellistarn, line 372: likewise, re: inlining.

perfdash/perfdash.go
```diff
@@ -46,10 +47,14 @@ var (
...
 	// Storage Service Bucket and Path flags
 	logsBucket = pflag.String("logsBucket", "kubernetes-jenkins", "Name of the data bucket")
-	logsPath = pflag.String("logsPath", "logs", "Path to the logs inside the logs bucket")
+	logsPath   = pflag.String("logsPath", "logs", "Path to the logs inside the logs bucket")
 
 	// Google GCS Specific flags
 	credentialPath = pflag.String("credentialPath", "", "Path to the gcs credential json")
+
+	// AWS S3 Specific flags
+	s3Mode    = pflag.Bool("s3Mode", false, "If true, download metrics from AWS S3")
```
ellistarn, line 56: consider using an enum under the flag "provider". --provider=s3

perfdash/perfdash.go
```diff
@@ -80,12 +85,15 @@ func run() error {
 		options.DefaultBuildsCount = maxBuilds
 	}
 
-	gcs, err := NewGCSMetricsBucket(*logsBucket, *logsPath, *credentialPath)
+	metricsBkt, err := NewGCSMetricsBucket(*logsBucket, *logsPath, *credentialPath)
```
ellistarn, line 88: enums will make this type of object instantiation nice and clean w/ a switch statement or chain of ifs.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+type S3MetricsBucket struct {
+	client  *s3.S3
+	bucket  *string
+	logPath string
+}
+
+// NewS3MetricsBucket creates a new S3MetricsBucket.
+func NewS3MetricsBucket(bucket, path, region string) (MetricsBucket, error) {
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(region)},
```
micahhausler, line 298: nit: Can you turn on verbose credential chain debugging?
```suggestion
		Region:                        aws.String(region),
	        CredentialsChainVerboseErrors: aws.Bool(true)},
```

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+	jobPrefix := joinStringsAndInts(fetcher.logPath, job) + "/"
+	klog.Infof("%s", jobPrefix)
+
+	params := &s3.ListObjectsV2Input{
+		Bucket:    fetcher.bucket,
+		Prefix:    aws.String(jobPrefix),
+		Delimiter: aws.String("/"),
+	}
+
+	objects, _ := fetcher.client.ListObjectsV2(params)
```
micahhausler, line 325: Why ignore the error here?

perfdash/metrics-downloader.go
micahhausler, line 372: Channeling Ellis, inlining.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+// GetFile fetches the file from the S3 bucket.
+func (fetcher *S3MetricsBucket) GetFile(job string, buildNumber int, path string) ([]byte, error) {
+	filePath := joinStringsAndInts(fetcher.logPath, job, buildNumber, path)
+
+	params := &s3.GetObjectInput{
+		Bucket: fetcher.bucket,
+		Key:    aws.String(filePath),
+	}
+
+	req, err := fetcher.client.GetObject(params)
```
micahhausler, line 377: I'd call this `resp` instead of `req` (response instead of request)

perfdash/perfdash.go
```diff
@@ -19,11 +19,12 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"k8s.io/klog"
 	"net/http"
 	"os"
 	"time"
 
+	"k8s.io/klog"
```
micahhausler, line 26: nit: import grouping (stdlib, everything else)

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+	}
+
+	var reader io.Reader = req.Body
+	data, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
```
micahhausler, line 389: This can be simplified, but you'll still want to close after reading

```suggestion
	defer resp.Body.Close()
	return ioutil.ReadAll(resp.Body)
}
```

perfdash/perfdash.go
```diff
@@ -46,10 +47,14 @@ var (
 
 	// Storage Service Bucket and Path flags
 	logsBucket = pflag.String("logsBucket", "kubernetes-jenkins", "Name of the data bucket")
-	logsPath = pflag.String("logsPath", "logs", "Path to the logs inside the logs bucket")
+	logsPath   = pflag.String("logsPath", "logs", "Path to the logs inside the logs bucket")
```
micahhausler, line 50: Did `gofmt` add this?

perfdash/perfdash.go
micahhausler, line 56: We'll probably want a `mode` flag with options of either `gcs` or `s3` with gcs being the default, and some validation

perfdash/perfdash.go
```diff
@@ -46,10 +47,14 @@ var (
...
 
 	// Google GCS Specific flags
 	credentialPath = pflag.String("credentialPath", "", "Path to the gcs credential json")
+
+	// AWS S3 Specific flags
+	s3Mode    = pflag.Bool("s3Mode", false, "If true, download metrics from AWS S3")
+	awsRegion = pflag.String("awsRegion", "us-west-2", "AWS region where s3 bucket lives")
```
micahhausler, line 57: Can we call this `aws-region`?
```suggestion
	awsRegion = pflag.String("aws-region", "us-west-2", "AWS region of the S3 bucket")
```

perfdash/metrics-downloader.go
mirahmed753: Inlined using session.Must

perfdash/metrics-downloader.go
mirahmed753: Renamed `path` to `pathPrefix`
I totally agree with the naming though in regards to "if it has a path, it isn't a bucket anymore." I would be open to refactoring in the future if that's something we're looking for.

perfdash/metrics-downloader.go
ellistarn: Let your editor manage all this and we can use the same import management rules.

perfdash/metrics-downloader.go
JacobGabrielson, line 369: Very minor.. I'd consider renaming this to something like `ReadFile` or `ReadJob` - since it returns the `[]bytes` from the file rather than (say) a handle to the file, etc.

perfdash/metrics-downloader.go
```diff
@@ -279,6 +285,109 @@ func (b *GCSMetricsBucket) GetFile(job string, buildNumber int, path string) ([]
...
+		}
+		builds = append(builds, buildNo)
+	}
+
+	return builds, nil
+}
+
+// ListFilesInBuild fetches the files in the build from S3.
+func (fetcher *S3MetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	var files []string
```
JacobGabrielson, line 344: very minor... I'd consider moving the declaration of `files` up into the return declaration..

perfdash/metrics-downloader.go
mirahmed753: The prefix is changing for what we are attempting to get. For instance:
1. `logPath` refers to part of the prefix where the jobs live (`logs`)
2. `logPath` + `job` refers to part of the prefix where the builds for the `job` lives (`logs/k8s-job-name`)
3. `logPath` + `job` + `buildNumber` refers to where the info for the builds live (`logs/k8s-job-name/1234567910`)
4. Finally, we have the file path ~> The metric files that are under `logs/k8s-job-name/1234567910/artifacts/`

Everything between logs/.../artifacts are variable; I can totally see why this can be awkward though

perfdash/metrics-downloader.go
mirahmed753: Good point, I added the error back

perfdash/metrics-downloader.go
mirahmed753: Gotcha, super nice + clean

perfdash/perfdash.go
mirahmed753: Yeah. My gofmt wasn't on/working when I submitted a previous PR so the formatting was off. I submitted another PR to fix that and it was just approved (#1385)

perfdash/metrics-downloader.go
mirahmed753: This is a very good point. I'm going to make this change since `GetFile` is rather misleading. FWIW, there's some more refactoring that needs to happen, and I'll be happy to take that on if we deem it necessary :)

perfdash/metrics-downloader.go
mirahmed753: It's the opposite actually. Returning false will exit early. IIRC, the function exits when there are no more pages to read, so the `lastPage` variable is kind of useless in this case...

In fact, if I do a `return lastPage`, the func will exit early if it isn't the lastPage. However, I can instead do a `return !lastPage` to achieve the same thing the func is currently doing.

perfdash/metrics-downloader.go
```diff
@@ -62,6 +57,7 @@ func NewDownloader(opt *DownloaderOptions, bkt MetricsBucket) *Downloader {
 // TODO(random-liu): Only download and update new data each time.
 func (g *Downloader) getData() (JobToCategoryData, error) {
 	configPaths := make([]string, len(g.Options.ConfigPaths))
+	fmt.Println(len(configPaths))
```
wojtek-t, line 60: Let's not use fmt - use klog for consistency.

perfdash/perfdash.go
```diff
@@ -81,12 +86,19 @@ func run() error {
 		options.DefaultBuildsCount = maxBuilds
 	}
 
-	gcs, err := NewGCSMetricsBucket(*logsBucket, *logsPath, *credentialPath)
+	var metricsBkt MetricsBucket
```
wojtek-t, line 89: let's not use those artificial shortcuts.

metricsBucket

perfdash/perfdash.go
```diff
@@ -81,12 +86,19 @@ func run() error {
...
+	var err error
+
+	if options.Mode == s3Mode {
```
wojtek-t, line 92:
```
switch options.Mode {
case gceMode:
  metricsBucket, err = ...
case s3Mode:
  metricsBucket, err = ...
defalt:
  return fmt.Errorf("unexpected mode: %s", options.Mode)
}
```

perfdash/perfdash.go
```diff
@@ -81,12 +86,19 @@ func run() error {
...
+		metricsBkt, err = NewS3MetricsBucket(*logsBucket, *logsPath, *awsRegion)
+	} else {
+		metricsBkt, err = NewGCSMetricsBucket(*logsBucket, *logsPath, *credentialPath)
+	}
```
wojtek-t, line 96: Since you touch this code - let's not panic and simply return an error below.

perfdash/metrics-downloader.go
mirahmed753: I was using this line for debugging and forgot to remove it - will do so now

perfdash/gcs_metrics_bucket.go
```diff
@@ -0,0 +1,127 @@
...
+	jobPrefix := joinStringsAndInts(b.logPath, job) + "/"
+	klog.Infof("%s", jobPrefix)
+	it := b.bucket.Objects(ctx, &storage.Query{
+		Prefix:    jobPrefix,
+		Delimiter: "/",
+	})
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
```
ellistarn, line 71: Break statements seems like a failure of flow control.

Is it better to do something like 
for attrs, err:= it.Next();  err != iterator.Done {
}

perfdash/gcs_metrics_bucket.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		builds = append(builds, buildNo)
+	}
+	return builds, nil
+}
+
+// ListFilesInBuild fetches the files in the build from GCS.
+func (b *GCSMetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	var files []string
+	ctx := context.Background()
```
ellistarn, line 93: likewise

perfdash/gcs_metrics_bucket.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil, fmt.Errorf("unknown build name convention: %s", build)
+		}
+		builds = append(builds, buildNo)
+	}
+	return builds, nil
+}
+
+// ListFilesInBuild fetches the files in the build from GCS.
+func (b *GCSMetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	var files []string
```
ellistarn, line 92: If you name your return argument, you don't need to instantiate this.

i.e. 

func (b *GCSMetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) (**files** []string, error) {

perfdash/gcs_metrics_bucket.go
```diff
@@ -0,0 +1,127 @@
...
+	ctx := context.Background()
+	jobPrefix := joinStringsAndInts(b.logPath, job, buildNumber, prefix)
+	it := b.bucket.Objects(ctx, &storage.Query{
+		Prefix: jobPrefix,
+	})
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
```
ellistarn, line 101: likewise.

perfdash/gcs_metrics_bucket.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return builds, nil
+}
+
+// ListFilesInBuild fetches the files in the build from GCS.
+func (b *GCSMetricsBucket) ListFilesInBuild(job string, buildNumber int, prefix string) ([]string, error) {
+	var files []string
+	ctx := context.Background()
+	jobPrefix := joinStringsAndInts(b.logPath, job, buildNumber, prefix)
+	it := b.bucket.Objects(ctx, &storage.Query{
```
ellistarn, line 95: Consider putting this on one line 

it := b.bucket.Objects(ctx, &storage.Query{ Prefix: jobPrefix })

perfdash/gcs_metrics_bucket.go
```diff
@@ -0,0 +1,127 @@
...
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		files = append(files, attrs.Name)
```
ellistarn, line 107: totally optional, but you might be able to write an IterateAppend function for both of these flows

```
IterateAppend(b.bucket.Objects(ctx, &storage.Query{ ... }), func (attrs Attr[], attr Attr) error {
 // specific appending logic
}, files)

func IterateAppend(it Iterator, reducer func([]Attr, Attr) error, accumulator []Attr) {
 // generic for loop and error check
}
```

This is much more functional style, which is how I tend to code. Happy to chat if you're interested, or feel free to skip. https://runestone.academy/runestone/books/published/fopp/Iteration/TheAccumulatorPattern.html

perfdash/perfdash.go
mirahmed753: That's a good point. Done.

ellistarn:
sorry I got a little tunneled on the green code. Feel free to ignore those if you're just moving code.

perfdash/metrics-downloader.go
```diff
@@ -80,7 +75,11 @@ func (g *Downloader) getData() (JobToCategoryData, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to refresh config: %v", err)
 	}
-	klog.Infof("Getting Data from GCS...")
+	if options.Mode == s3Mode {
+		klog.Infof("Getting Data from S3...")
+	} else {
+		klog.Infof("Getting Data from GCS...")
+	}
```
wojtek-t, line 82: just replace with:
```
klog.Infof("Getting Data from %v", options.Mode)
```

perfdash/perfdash.go
```diff
@@ -81,12 +87,23 @@ func run() error {
...
+	case gcsMode:
+		metricsBucket, err = NewGCSMetricsBucket(*logsBucket, *logsPath, *credentialPath)
+	case s3Mode:
+		metricsBucket, err = NewS3MetricsBucket(*logsBucket, *logsPath, *awsRegion)
+	default:
+		return fmt.Errorf("unexpected mode: %s", options.Mode)
+	}
+
 	if err != nil {
 		panic(err)
```
wojtek-t, line 103: since you touch this code - can you please return the error here, instead of panic-ing?

mirahmed753: Misunderstood your earlier comment about this. I think I got it right this time hopefully, please take a look :)

PR #1401, opened 2020-07-24
Make schedulerThroughput measurement SLO violation failing a test customizable

tosi3k:
Addressing issue https://github.com/kubernetes/kubernetes/issues/93217 where problems with this measurement arose due to master VMs in 100 nodes tests lacking CPU time for kube-scheduler in the density test after changes in https://github.com/kubernetes/kubernetes/pull/92840.

We maintain the default behavior of failing the test when SLO violation occurs but will make 100 nodes tests not fail when the violation happens.

/sig scalability
/cc jkaniuk
/cc wojtek-t

### DIFF
--- a/clusterloader2/pkg/measurement/common/scheduling_throughput.go
+++ b/clusterloader2/pkg/measurement/common/scheduling_throughput.go
@@ -82,7 +82,21 @@ func (s *schedulingThroughputMeasurement) Execute(config *measurement.Config) ([
 		if err != nil {
 			klog.Warningf("error while getting threshold param: %v", err)
 		}
-		return s.gather(threshold)
+		enableViolations, err := util.GetBoolOrDefault(config.Params, "enableViolations", true)
+		if err != nil {
+			klog.Warningf("error while getting enableViolations param: %v", err)
+		}
+		summary, err := s.gather(threshold)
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -82,7 +82,15 @@ func (s *schedulingThroughputMeasurement) Execute(config *measurement.Config) ([
 		if err != nil {
 			klog.Warningf("error while getting threshold param: %v", err)
 		}
-		return s.gather(threshold)
+		enableViolations, err := util.GetBoolOrDefault(config.Params, "enableViolations", true)
+		if err != nil {
+			klog.Warningf("error while getting enableViolations param: %v", err)
+		}
+		summary, err := s.gather(threshold)
+		if err != nil && errors.IsMetricViolationError(err) && !enableViolations {
```
wojtek-t, line 90: Can you please make it consistent with what is here:
https://github.com/kubernetes/perf-tests/blob/4d71498446f84fd2b371e1506748fb2a986c0d7f/clusterloader2/pkg/measurement/common/slos/prometheus_measurement.go#L89

PR #1415, opened 2020-08-15
set nullPointMode to null as default

islamtaha:


mm4tt:
/ok-to-test
/lgtm
/approve

Thanks! This is great

Issue #1406, opened 2020-07-28
APIPrometheusReponsiveness returns incorrect values for SlowCount


wojtek-t:
This is the example:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1287819459200815104

```
[measurement call APIResponsivenessPrometheus - APIResponsivenessPrometheusSimple error: top latency metric: there should be no high-latency requests, but: [got: &{Resource:replicasets Subresource: Verb:POST Scope:namespace Latency:perc50: 26.916666ms, perc90: 48.45ms, perc99: 1.442499999s Count:325 SlowCount:11}; expected perc99 <= 1s]]
```

However, clearly there aren't 11 slow call. Looking into kube-apiserver logs:
```
// out.txt contains list of POST replicasets calls from load test
wojtekt@wojtekt:~/Downloads$ cat out.txt | grep "[0-9]s"
I0727 19:05:13.351431      11 httplog.go:89] "HTTP" verb="POST" URI="/apis/apps/v1/namespaces/test-r5b2iw-1/replicasets" latency="1.797940226s" userAgent="kube-controller-manager/v1.20.0 (linux/amd64) kubernetes/00da04b/system:serviceaccount:kube-system:deployment-controller" srcIP="[::1]:38634" resp=201
I0727 19:05:13.353515      11 httplog.go:89] "HTTP" verb="POST" URI="/apis/apps/v1/namespaces/test-r5b2iw-1/replicasets" latency="1.717497009s" userAgent="kube-controller-manager/v1.20.0 (linux/amd64) kubernetes/00da04b/system:serviceaccount:kube-system:deployment-controller" srcIP="[::1]:38634" resp=201
I0727 19:05:13.354785      11 httplog.go:89] "HTTP" verb="POST" URI="/apis/apps/v1/namespaces/test-r5b2iw-1/replicasets" latency="1.722224098s" userAgent="kube-controller-manager/v1.20.0 (linux/amd64) kubernetes/00da04b/system:serviceaccount:kube-system:deployment-controller" srcIP="[::1]:38634" resp=201
[...]

PR #1417, opened 2020-08-19
New method of counting slow queries

jkaniuk:
Fix: #1406

/cc @mm4tt 

Need to test it first
/hold

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -60,14 +60,14 @@ const (
 	// countQuery %v should be replaced with (1) filters and (2) query window size.
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
-	countSlowQuery = "sum(rate(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
+	countFastQuery = "sum(increase(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
 
 	// exclude all buckets of 1s and shorter
-	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT", le!~"0.\\d+|1"`
+	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT|LIST", le="1"`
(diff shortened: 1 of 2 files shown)


mm4tt:
Please ping me when it's ready for review

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -60,14 +60,14 @@ const (
 	// countQuery %v should be replaced with (1) filters and (2) query window size.
 	countQuery = "sum(increase(apiserver_request_duration_seconds_count{%v}[%v])) by (resource, subresource, scope, verb)"
 
-	countSlowQuery = "sum(rate(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
+	countFastQuery = "sum(increase(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
 
 	// exclude all buckets of 1s and shorter
-	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT", le!~"0.\\d+|1"`
+	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT|LIST", le="1"`
```
wojtek-t, line 66: does le="1" work ok? Will it count all buckets?

jkaniuk: le="1" counts only this single bucket, which contain all request that did fit in 1s

So idea is to treat this bucket as fast queries, and subtract it from overall count to get slow queries.

wojtek-t: Makes sense - thanks.

So I'm actually wondering what the previous query was doing - it was choosing one of the buckets?

Anyway - thanks a lot for attacking this problem - this approach looks reasonable, let me know when this will be ready for final review.

jkaniuk: To be honest - I do not understand the previous approach.
But results were somewhat close (less than 2x away) to the new approach.

jkaniuk: Actually, old logic frequently show 0 SlowQueries even when the measurement fails. It was not noticed before as SlowQueries was always set to 0 if slow queries were not used.

wojtek-t:
/lgtm
/approve

Thanks!

PR #1419, opened 2020-08-20
cl2/density: add pod-pod anti-affinity benchmarks

adtac:
/cc @ahg-g @mm4tt 
/sig scheduling

(Will update with results in a bit.)

### DIFF
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -38,6 +38,9 @@
 # failure won't fail the test. See https://github.com/kubernetes/kubernetes/issues/73461#issuecomment-467338711
 {{$saturationDeploymentHardTimeout := MaxInt $saturationDeploymentTimeout 1200}}
 
+{{$saturationDeploymentSpec := DefaultParam .SATURATION_DEPLOYMENT_SPEC "deployment.yaml"}}
+{{$latencyDeploymentSpec := DefaultParam .LATENCY_DEPLOYMENT_SPEC "deployment.yaml"}}
+
 # Probe measurements shared parameter
 {{$PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT := DefaultParam .CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT "5m"}}
 
@@ -123,7 +126,7 @@ steps:
     tuningSet: Uniform5qps
     objectBundle:
     - basename: saturation-deployment
-      objectTemplatePath: deployment.yaml
+      objectTemplatePath: {{$saturationDeploymentSpec}}
       templateFillMap:
         Replicas: {{$podsPerNamespace}}
         Group: saturation
@@ -174,7 +177,7 @@ steps:
     tuningSet: Uniform5qps
     objectBundle:
     - basename: latency-deployment
-      objectTemplatePath: deployment.yaml
+      objectTemplatePath: {{$latencyDeploymentSpec}}
       templateFillMap:
         Replicas: 1
         Group: latency
@@ -197,7 +200,7 @@ steps:
     tuningSet: Uniform5qps
     objectBundle:
     - basename: latency-deployment
-      objectTemplatePath: deployment.yaml
+      objectTemplatePath: {{$latencyDeploymentSpec}}
 
 - name: Waiting for latency pods to be deleted
   measurements:
@@ -222,7 +225,7 @@ steps:
     tuningSet: Uniform5qps
     objectBundle:
     - basename: saturation-deployment
-      objectTemplatePath: deployment.yaml
+      objectTemplatePath: {{$saturationDeploymentSpec}}
 
 - name: Waiting for saturation pods to be deleted
   measurements:
--- a/clusterloader2/testing/density/high-density-config.yaml
+++ b/clusterloader2/testing/density/high-density-config.yaml
@@ -35,6 +35,9 @@
 # failure won't fail the test. See https://github.com/kubernetes/kubernetes/issues/73461#issuecomment-467338711
 {{$saturationDeploymentHardTimeout := MaxInt $saturationDeploymentTimeout 1200}}
 
+{{$saturationDeploymentSpec := DefaultParam .SATURATION_DEPLOYMENT_SPEC "deployment.yaml"}}
+{{$latencyDeploymentSpec := DefaultParam .LATENCY_DEPLOYMENT_SPEC "deployment.yaml"}}
+
 name: density
 namespace:
   number: {{$namespaces}}
@@ -113,7 +116,7 @@ steps:
     tuningSet: Uniform5qps
     objectBundle:
     - basename: saturation-deployment
-      objectTemplatePath: deployment.yaml
+      objectTemplatePath: {{$saturationDeploymentSpec}}
       templateFillMap:
         Replicas: {{$podsPerNamespace}}
         Group: saturation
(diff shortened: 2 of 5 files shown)


clusterloader2/testing/density/suite.yaml
```diff
@@ -0,0 +1,8 @@
+- identifier: vanilla
+  configPath: testing/density/config.yaml
+  overridePaths: []
+
+- identifier: pod-pod-anti-affinity
```
ahg-g, line 5: lets rename to pod-anti-affinity (drop the first pod), here and in the directory names as well.

clusterloader2/testing/density/pod-pod-anti-affinity/overrides.yaml
```diff
@@ -0,0 +1,7 @@
+SATURATION_DEPLOYMENT_SPEC: pod-pod-anti-affinity/deployment.yaml
+LATENCY_DEPLOYMENT_SPEC: pod-pod-anti-affinity/deployment.yaml
+
+# The default CL2_SCHEDULER_THROUGHPUT_THRESHOLD value of 90 is too high for
+# pod-pod anti-affinity. Use a low threshold until a good baseline can be
+# identified and then increase the threshold incrementally.
+CL2_SCHEDULER_THROUGHPUT_THRESHOLD: 40
```
ahg-g, line 7: is this for an alert? if so, please clarify that in the comment and I think we should keep it at 90 unless this will be executed for older versions.

clusterloader2/testing/density/suite.yaml
```diff
@@ -0,0 +1,8 @@
+- identifier: vanilla
```
ahg-g, line 1: out of curiosity, where is this suite.yaml file being consumed? and how do we ensure that this only gets executed on kubemark?

clusterloader2/testing/density/suite.yaml
adtac: my thinking behind this was that we'll be adding pod-to-node affinity too (similar naming as the integration tests)

but since pod will be a common prefix, I think we can drop it :+1:

clusterloader2/testing/density/suite.yaml
adtac: It's not used here, but we'll need to update kubernetes/test-infra's scalability tests to use the suite instead of directly running the config.yaml

clusterloader2/testing/density/suite.yaml
ahg-g: I think we can use pod-anti-affinity/pod-affinity and node-anti-affinity/node-affinity.

clusterloader2/testing/density/pod-pod-anti-affinity/overrides.yaml
adtac: oops, this is an artifact from my testing when the throughput was below 50 for unrelated reasons. removing.

clusterloader2/testing/density/config.yaml
```diff
@@ -15,7 +15,7 @@
 # This results in decreased probability of choosing the same node again.
 {{$LATENCY_POD_CPU := DefaultParam .LATENCY_POD_CPU 100}}
 {{$LATENCY_POD_MEMORY := DefaultParam .LATENCY_POD_MEMORY 350}}
-{{$MIN_LATENCY_PODS := 500}}
+{{$MIN_LATENCY_PODS := MinInt (MultiplyInt $PODS_PER_NODE .Nodes) 500}}
```
adtac, line 18: `PODS_PER_NODE * Nodes` will be greater than 500 in all cases that we currently run; existing tests are therefore unaffected. However, in the requiredDuringScheduling anti-affinity tests, we cannot schedule more than one pod per node, so we need this.

mm4tt:
> /hold
> 
> investigating an issue in my local tests, will report back

SG, please ping when the PR is ready to review

adtac:
Somewhere along the process, I changed `podAffinityTerm` to `podAntiAffinityTerm`, which does not exist even in pod anti-affinity. Anyway, I've fixed that and I've re-run the experiments locally and everything looks good. Here are the results:

```
~/tmp/cl2/032 $ cat SchedulingThroughput_density_2020-08-24T19:47:33Z.json                                                                                                                                                                                                                                                                          
{                                                                                                                                                                                                                                                                                                                                                                            
  "perc50": 66.6,                                                                                                                                                                                                                                                                                                                                                            
  "perc90": 103.8,                                                                                                                                                                                                                                                                                                                                                           
  "perc99": 103.8,                                                                                                                                                                                                                                                                                                                                                           
  "max": 103.8                                                                             
}                                                                                          
```

Interestingly, there's a considerable drop in throughput when scheduling 3000 pods with preferred anti-affinity on 100 nodes. The ~66.6 pods/sec is reproducible across multiple experiments (in the 65-72 range across 5 runs), so it's definitely not measurement error or noise.

/hold cancel

clusterloader2/testing/density/suite.yaml
```diff
@@ -0,0 +1,8 @@
...
+  configPath: testing/density/config.yaml
+  overridePaths: []
```
mm4tt, line 3: Let's create a separate suite just for scheduler benchmarks. We'll run the vanilla as usual and start running the scheduler suite once a day or so.

ahg-g:
> Somewhere along the process, I changed `podAffinityTerm` to `podAntiAffinityTerm`, which does not exist even in pod anti-affinity. Anyway, I've fixed that and I've re-run the experiments locally and everything looks good. Here are the results:
> 
> ```
> ~/tmp/cl2/032 $ cat SchedulingThroughput_density_2020-08-24T19:47:33Z.json                                                                                                                                                                                                                                                                          
> {                                                                                                                                                                                                                                                                                                                                                                            
>   "perc50": 66.6,                                                                                                                                                                                                                                                                                                                                                            
>   "perc90": 103.8,                                                                                                                                                                                                                                                                                                                                                           
>   "perc99": 103.8,                                                                                                                                                                                                                                                                                                                                                           
>   "max": 103.8                                                                             
> }                                                                                          
> ~/tmp/cl2/033 $ cat SchedulingThroughput_density_2020-08-24T19:55:56Z.json 
> {
>   "perc50": 71.2,
>   "perc90": 100.2,
>   "perc99": 100.2,
>   "max": 100.2
> }
> ```
> 
> Interestingly, there's a considerable drop in throughput when scheduling 3000 pods with preferred anti-affinity on 100 nodes. The ~66.6 pods/sec is reproducible across multiple experiments (in the 65-72 range across 5 runs), so it's definitely not measurement error or noise.
> 
> /hold cancel

Thanks Adi, I wonder what master size was used?

clusterloader2/testing/density/suite.yaml
adtac: Done.

adtac:
>Thanks Adi, I wonder what master size was used?

The kubemark master was a n1-standard-4 (4 vCPUs, 16 GB memory). Note that this was the same master used in vanilla tests, which produced a ~100 pods/sec throughput.

clusterloader2/testing/density/suite.yaml
```diff
@@ -0,0 +1,4 @@
+- identifier: pod-anti-affinity
```
mm4tt, line 1: Can we name the file `scheduler-suite.yaml` or something like this?

Other than that it looks good to me

/approve

adtac: Done.

PR #1420, opened 2020-08-23
Add countErrorMargin to WaitForControlledPods

jkaniuk:
Add countErrorMargin option to WaitForControlledPods

When using preemptibles on large scale, number of nodes do not become stable, so reaching DesiredPodCount could take a very long time.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -74,10 +74,15 @@ func createWaitForControlledPodsRunningMeasurement() measurement.Measurement {
 }
 
 type waitForControlledPodsRunningMeasurement struct {
-	apiVersion            string
-	kind                  string
-	selector              *measurementutil.ObjectSelector
-	operationTimeout      time.Duration
+	apiVersion       string
+	kind             string
+	selector         *measurementutil.ObjectSelector
+	operationTimeout time.Duration
+	// countErrorMargin orders measurement to wait for number of pods to be in
+	// <desired count - countErrorMargin, desired count> range
+	// When using preemptibles on large scale, number of ready nodes is not stable
+	// and reaching DesiredPodCount could take a very long time.
+	countErrorMargin      int
 	stopCh                chan struct{}
 	isRunning             bool
 	queue                 workerqueue.Interface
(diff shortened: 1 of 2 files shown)


jkaniuk:
Tested manually, work as expected.

/cc @mm4tt 
/cc @wojtek-t

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -78,6 +78,7 @@ type waitForControlledPodsRunningMeasurement struct {
 	kind                  string
 	selector              *measurementutil.ObjectSelector
 	operationTimeout      time.Duration
+	countErrorMargin      int
```
mm4tt, line 81: I know we haven't been doing it so far, but can we start documenting the fields here? They are, in a way, forming an external API (measurement api exposed in the yaml files) and deserve a sentence of documentation if meaning is not 100% clear.

mm4tt (commented):
/lgtm
/approve

/hold

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
jkaniuk: Done

mm4tt: Documentation in golang should be "a complete sentence that begins with the name of the element it describes" - https://blog.golang.org/godoc

mm4tt (commented):
/lgtm

Thanks!

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -482,13 +491,18 @@ func (w *waitForControlledPodsRunningMeasurement) waitForRuntimeObject(obj runti
 	o.lock.Lock()
 	defer o.lock.Unlock()
 	w.handlingGroup.Start(func() {
+		var minDesiredPodCount int
+		if w.countErrorMargin > 0 {
+			minDesiredPodCount = int(runtimeObjectReplicas) - w.countErrorMargin
```
wojtek-t, line 496: I would do that slightly different, i.e. always set
minDesiredPodCount = runtimeObjectReplicas - w.countErrorMargin

[i.e always treat it as min] - I think the code would be a bit simpler below]

PR #1428, opened 2020-08-27
fix namespace num method

zackzhangkai:
Obviously, the namespaces = nodes * namespaces/per-node

Signed-off-by: zackzhangkai <zackzhang@yunify.com>

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -29,7 +29,7 @@
 {{$CUSTOM_API_CALL_THRESHOLDS := DefaultParam .CUSTOM_API_CALL_THRESHOLDS ""}}
 {{$ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS := DefaultParam .CL2_ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS false}}
 #Variables
-{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -29,7 +29,7 @@
 {{$CUSTOM_API_CALL_THRESHOLDS := DefaultParam .CUSTOM_API_CALL_THRESHOLDS ""}}
 {{$ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS := DefaultParam .CL2_ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS false}}
 #Variables
-{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+{{$namespaces := MultiplyInt .Nodes $NODES_PER_NAMESPACE}}
```
wojtek-t, line 32: This isn't correct - number of namespaces is #node divided by nodes per namespace.

zackzhangkai: for example: you have 4 nodes（Nodes）, and specify that each node has 3 namespaces（NODES_PER_NAMESPACE）, then, the namespaces numbers should be:  4x3 = 12 namespaces， am I right?

wojtek-t: That's not what "NODES_PER_NAMESPACE" is for.

What you're talking about would be "NAMESPACES_PER_NODE", which we don't really have.

zackzhangkai: Got it , thanks a lot.

PR #1429, opened 2020-08-27
fix curl with cacert

zackzhangkai:
when curl with cert, cacert is necessary, such as:

```
curl --cert /etc/ssl/etcd/ssl/node-node1.pem --key /etc/ssl/etcd/ssl/node-node1-key.pem --cacert /etc/ssl/etcd/ssl/ca.pem  https://192.168.0.2:2379/metrics
```

Signed-off-by: zackzhangkai <zackzhang@yunify.com>

### DIFF
--- a/clusterloader2/pkg/measurement/common/etcd_metrics.go
+++ b/clusterloader2/pkg/measurement/common/etcd_metrics.go
@@ -200,15 +200,15 @@ func (e *etcdMetricsMeasurement) getEtcdMetrics(host string, provider provider.P
 
 	// Use old endpoint if new one fails, "2379" is hard-coded here as well, it is kept as is since
 	// we don't want to bloat the cluster config only for a fall-back attempt.
-	etcdCert, etcdKey, etcdHost := os.Getenv("ETCD_CERTIFICATE"), os.Getenv("ETCD_KEY"), os.Getenv("ETCD_HOST")
+	etcdCert, etcdKey, etcdCaCert, etcdHost := os.Getenv("ETCD_CERTIFICATE"), os.Getenv("ETCD_KEY"), os.Getenv("ETCD_CACERT"), os.Getenv("ETCD_HOST")
 	if etcdHost == "" {
 		etcdHost = "localhost"
 	}
-	if etcdCert == "" || etcdKey == "" {
-		klog.Warning("empty etcd cert or key, using http")
+	if etcdCert == "" || etcdKey == "" || etcdHost == "" {
+		klog.Warning("empty etcd cert or key or CAcert, using http")
 		cmd = fmt.Sprintf("curl http://%s:2379/metrics", etcdHost)
 	} else {
-		cmd = fmt.Sprintf("curl -k --cert %s --key %s https://%s:2379/metrics", etcdCert, etcdKey, etcdHost)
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -200,15 +200,15 @@ func (e *etcdMetricsMeasurement) getEtcdMetrics(host string, provider provider.P
...
+	etcdCert, etcdKey, etcdCaCert, etcdHost := os.Getenv("ETCD_CERTIFICATE"), os.Getenv("ETCD_KEY"), os.Getenv("ETCD_CACERT"), os.Getenv("ETCD_HOST")
 	if etcdHost == "" {
 		etcdHost = "localhost"
 	}
 	if etcdCert == "" || etcdKey == "" {
 		klog.Warning("empty etcd cert or key, using http")
 		cmd = fmt.Sprintf("curl http://%s:2379/metrics", etcdHost)
 	} else {
-		cmd = fmt.Sprintf("curl -k --cert %s --key %s https://%s:2379/metrics", etcdCert, etcdKey, etcdHost)
+		cmd = fmt.Sprintf("curl -k --cert %s --key %s --cacert %s https://%s:2379/metrics", etcdCert, etcdKey, etcdCaCert, etcdHost)
```
mm4tt, line 211: What if ETCD_CERTIFICATE and ETCD_KEY are set but ETCD_CACART is not? I think it'll make the command fail

zackzhangkai: e.g.
```
# curl --cert /etc/ssl/etcd/ssl/node-node1.pem --key /etc/ssl/etcd/ssl/node-node1-key.pem   https://192.168.0.2:2379/metrics
curl: (60) Peer's Certificate issuer is not recognized.
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
```

mm4tt: In the situation I described the command will be 
```
curl --cert /etc/ssl/etcd/ssl/node-node1.pem --key /etc/ssl/etcd/ssl/node-node1-key.pem --cacert   https://192.168.0.2:2379/metrics
```

zackzhangkai:
```suggestion
		cmd = fmt.Sprintf("curl -k --cert %s --key %s --cacert %s https://%s:2379/metrics", etcdCert, etcdKey, etcdCaCert, etcdHost)
```
you shoud set this env
```
export ETCD_CERTIFICATE=/etc/ssl/etcd/ssl/node-node1.pem
export ETCD_KEY=/etc/ssl/etcd/ssl/node-node1-key.pem
export ETCD_HOST=192.168.0.2
export ETCD_CACERT=/etc/ssl/etcd/ssl/ca.pem
```

zackzhangkai: fix it ,  if ETCD_CERTIFICATE and ETCD_KEY are set but ETCD_CACART is not, then use http

PR #1436, opened 2020-08-27
Add per step reporting in CL2

tosi3k:
Introduce a level of granularity in the performance tests reporting that is shown in the Prow job test results. Every step in each of the test configs will be reported separately.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -22,9 +22,6 @@ import (
 	"path"
 	"time"
 
-	ginkgoconfig "github.com/onsi/ginkgo/config"
-	ginkgoreporters "github.com/onsi/ginkgo/reporters"
-	ginkgotypes "github.com/onsi/ginkgo/types"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
@@ -297,13 +294,8 @@ func main() {
 		klog.Exitf("Error while preloading images: %v", err)
 	}
 
-	suiteSummary := &ginkgotypes.SuiteSummary{
-		SuiteDescription:           "ClusterLoaderV2",
-		NumberOfSpecsThatWillBeRun: len(testConfigPaths),
-	}
-	junitReporter := ginkgoreporters.NewJUnitReporter(path.Join(clusterLoaderConfig.ReportDir, "junit.xml"))
-	junitReporter.SpecSuiteWillBegin(ginkgoconfig.GinkgoConfig, suiteSummary)
-	testsStart := time.Now()
+	testReporter := test.CreateSimpleReporter(path.Join(clusterLoaderConfig.ReportDir, "junit.xml"), "ClusterLoaderV2")
+	testReporter.BeginTestSuite()
 	if testSuiteConfigPath != "" {
 		testSuite, err := config.LoadTestSuite(testSuiteConfigPath)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


clusterloader2/pkg/test/interface.go
```diff
@@ -64,3 +66,14 @@ type Executor interface {
 	ExecutePhase(ctx Context, phase *api.Phase) *errors.ErrorList
 	ExecuteObject(ctx Context, object *api.Object, namespace string, replicaIndex int32, operation OperationType) *errors.ErrorList
 }
+
+// TestReporter is an interface for reporting tests results.
+type TestReporter interface {
+	SetTestID(name string)
+	GetTestID() string
+	GetNumberOfFailedSpecs() int
```
tosi3k, line 74: TODO: Rename the method to sth like `GetNumberOfFailedTestItems()`.

clusterloader2/cmd/clusterloader.go
```diff
@@ -297,31 +295,25 @@ func main() {
...
 
-	suiteSummary := &ginkgotypes.SuiteSummary{
-		SuiteDescription:           "ClusterLoaderV2",
-		NumberOfSpecsThatWillBeRun: len(testConfigPaths),
-	}
-	junitReporter := ginkgoreporters.NewJUnitReporter(path.Join(clusterLoaderConfig.ReportDir, "junit.xml"))
-	junitReporter.SpecSuiteWillBegin(ginkgoconfig.GinkgoConfig, suiteSummary)
-	testsStart := time.Now()
+	testReporter := test.CreateSimpleTestReporter(path.Join(clusterLoaderConfig.ReportDir, "junit.xml"), "ClusterLoaderV2")
+	testReporter.BeginTestSuite()
```
wojtek-t, line 299: Before submitt, can you please double check that this will also work fine if run for the suite?
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml#L2
is a good example

tosi3k: Good catch - I didn't notice that we actually use `--testsuite` flag in our test jobs.

I will come back here with the results confirming it also works for this type of tests setup.

tosi3k: I ran a manual test and it looks okay apart from the fact that some steps do not have a name in the [config](https://github.com/kubernetes/perf-tests/blob/2afae16bdd662ef636404cf165df864f90dc1b81/clusterloader2/testing/experimental/storage/pod-startup/config.yaml) that test suite leverages. Because of that, some of the reported test cases have the same identifier from my PR's point of view and Prow treats them as duplicates and filters them out.

I propose that we add names for every step in that config and implement a mechanism in CL2 that imposes providing a name for each of the steps. What do you think?

mm4tt: For now, can we make a change in CL2 that if a step doesn't have name then a name is generated, e.g. something like: "Step N [autogenerated, please name your step]"? Can we do it in a separate CL? It should be fairly small change

We can start naming the steps later and maybe eventually start enforcing steps to have name.

tosi3k: Sounds good to me; will come back with an appropriate PR here.

tosi3k: PR with step names autogeneration: https://github.com/kubernetes/perf-tests/pull/1472.

tosi3k: Done.

clusterloader2/pkg/test/interface.go
mm4tt: What about this TODO?

clusterloader2/pkg/test/interface.go
```diff
@@ -64,3 +66,15 @@ type Executor interface {
...
+
+// TestReporter is an interface for reporting tests results.
+type TestReporter interface {
+	SetTestName(name string)
+	GetTestName() string
+	GetNumberOfFailedSpecs() int
+	GetTestSuiteDescription() string
+	BeginTestSuite()
+	EndTestSuite()
+	ReportTestStepFinish(stepSummary *ginkgotypes.SpecSummary)
```
mm4tt, line 78: Can we make this interface more abstract and not depend on a specific implementation (`gingotypes.SpecSummary`)?
I believe the `gingotypes.SpecSummary` can be constructed from the following primitives: stepName, duration, error, can't it? It could be done in the implementation of this interface

clusterloader2/pkg/test/interface.go
```diff
@@ -64,3 +66,15 @@ type Executor interface {
...
+	ReportTestFinish(testSummary *ginkgotypes.SpecSummary)
```
mm4tt, line 79: Same here

mm4tt:
Also golint seems to be failing with
```
pkg/test/interface.go:71:6: type name will be used as test.TestReporter by other packages, and that stutters; consider calling this Reporter (golint)
```

clusterloader2/pkg/test/interface.go
tosi3k: Done.

clusterloader2/pkg/test/interface.go
tosi3k: Done.

clusterloader2/pkg/test/interface.go
tosi3k: Done.

tosi3k:
> Also golint seems to be failing with
> 
> ```
> pkg/test/interface.go:71:6: type name will be used as test.TestReporter by other packages, and that stutters; consider calling this Reporter (golint)
> ```

Not anymore ;) done.

mm4tt:
/lgtm
/approve

This is amazing! Thanks!

PR #1437, opened 2020-08-27
Clean up steps in density test config

tosi3k:
Merge two saturation pods steps into one and fix the names of saturation-related steps.

/assign mm4tt


wojtek-t:
/lgtm
/approve

Thanks!

PR #1440, opened 2020-08-28
fix deployment template variable & fix ds to tolerate

zackzhangkai:
1. deloyment.yaml should get the variable,  or it will failed with error of "nil"
2. daemonset shoud set the tolerations , or if some nodes has a taint , such as master usally has a taint that forbidden to schedule general pod on it, the daemonset will fail.


clusterloader2/testing/load/config.yaml
```diff
@@ -490,6 +490,10 @@ steps:
     objectBundle:
     - basename: big-deployment
       objectTemplatePath: deployment.yaml
+      templateFillMap:
+        ReplicasMin: {{MultiplyInt $BIG_GROUP_SIZE 0.5}}
```
mm4tt, line 494: This is not needed for the delete step. Note that this test runs continuously without any issues - https://k8s-testgrid.appspot.com/sig-scalability-gce#gce-cos-master-scalability-100

clusterloader2/testing/load/daemonset.yaml
```diff
@@ -28,5 +28,7 @@ spec:
           requests:
             cpu: 10m
             memory: "10M"
+      tolerations:
+      - operator: Exists
```
mm4tt, line 32: Not sure if I follow this. FTR, DS has some tolerations added automatically, see https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#taints-and-tolerations

zackzhangkai: it can add add tolerations automatically with only the specified ones. If I have a master tained other than these, then it will fail.

clusterloader2/testing/load/config.yaml
zackzhangkai: it really does need it , if do the test in local cluster, or it will fail with "cannot convert nil to int64"

wojtek-t: We run this test periodically (tens of times a day on different cluster sizes). It is NOT needed.

PR #1441, opened 2020-08-28
Increase page size to 5000 in slo-monitor

mborsz:
Change page size in slo-monitor from 500 to 5000.

kube-apiserver internally uses the page size as a chunk size for a single etcd requests. In cases like this one (a fieldSelector that matches only a tiny fraction of all obects) it needs to scan all objects which takes a lot of time when 500 page size is used.

In my benchmark with 200k events, this change seems to decrease request latency 3x.

/assign @wojtek-t

### DIFF
--- a/slo-monitor/Makefile
+++ b/slo-monitor/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 PACKAGE = k8s.io/perf-tests/slo-monitor
-TAG = 0.12.0
+TAG = 0.13.0
 # Image should be pulled from k8s.gcr.io, which will auto-detect
 # region (us, eu, asia, ...) and pull from the closest.
 REPOSITORY?=k8s.gcr.io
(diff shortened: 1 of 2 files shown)


wojtek-t:
/lgtm
/approve

Thanks!

mborsz:
This doesn't work. The pg.PageSize here is never used as NewWatcherWithHandler is implementing pagination internally, so it sets page size to 500 there and our value is ignored.

We need to somehow pass our 5000 limit there and remove the pagination inside of &cache.ListWatch handler as it's noop anyway.

wojtek-t:
The pages is now built-into refelctor - so we may remove this one.

And once this is removed, you can use WatchListPageSize option of the reflector:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/reflector.go#L269

[you may need to update dependencies though...]

mborsz:
Yes, this is exactly what I'm going to do: removing the pagination from here and setting WatchListPageSize in reflector. I will need to send a contribution to client-go's "controller" to pass WatchListPageSize value from cache.Config to r.Reflector (currently, we have no access to reflector when controller is used).

mborsz:
I'm also going to check why we do use 'controller' from cache instead of using Reflector directly or other library available there

PR #1442, opened 2020-08-28
perfdash: add scheduler throughput/latency labels

adtac:
This label will be used to view the graphs for data generated by
different tests defined in cl2/testing/density/scheduler.

Related to #1414

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -179,16 +179,30 @@ var (
 			}},
 		},
 		"Scheduler": {
-			"SchedulingLatency": []TestDescription{{
-				Name:             "density",
-				OutputFilePrefix: "SchedulingMetrics",
-				Parser:           parseSchedulingLatency,
-			}},
-			"SchedulingThroughput": []TestDescription{{
-				Name:             "density",
-				OutputFilePrefix: "SchedulingThroughput",
-				Parser:           parseSchedulingThroughputCL,
-			}},
+			"SchedulingLatency": []TestDescription{
+				// `density_*` items need to be before the `density` item because of
+				// how data file prefixes work. Same applies to SchedulingThroughput.
+				{
+					Name:             "density_pod-anti-affinity",
+					OutputFilePrefix: "SchedulingMetrics",
+					Parser:           parseSchedulingLatency("pod-anti-affinity"),
+				}, {
+					Name:             "density",
+					OutputFilePrefix: "SchedulingMetrics",
+					Parser:           parseSchedulingLatency("density"),
+				},
+			},
+			"SchedulingThroughput": []TestDescription{
+				{
+					Name:             "density_pod-anti-affinity",
+					OutputFilePrefix: "SchedulingThroughput",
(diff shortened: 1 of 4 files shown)


perfdash/parser.go
```diff
@@ -258,32 +258,34 @@ type schedulingMetrics struct {
 	FrameworkExtensionPointDuration map[string]latencyMetric `json:"frameworkExtensionPointDuration"`
 }
 
-func parseOperationLatency(latency latencyMetric, operationName string) perftype.DataItem {
-	perfData := perftype.DataItem{Unit: "ms", Labels: map[string]string{"Operation": operationName}, Data: make(map[string]float64)}
+func parseOperationLatency(latency latencyMetric, testName string, operationName string) perftype.DataItem {
+	perfData := perftype.DataItem{Unit: "ms", Labels: map[string]string{"TestName": testName, "Operation": operationName}, Data: make(map[string]float64)}
```
adtac, line 262: Will adding the TestName label in future parses of "density" workloads affect the graphs? Or should I default to not adding the TestName label if `testName == "density"` to keep DataItems compatible with old data?

mm4tt: Parsed data is not stored everywhere. If we deploy a new perf-dash version it will parse all runs and build the internal data from scratch. Thus, I think it should work, but it'd be good to double check by running the perf-dash locally for this change - https://github.com/kubernetes/perf-tests/tree/master/perfdash#local-development

adtac: Screenshot_2020-09-03 Screenshot

Local instance runs fine, but since only the density test exists right now under the kubemark jobs, there isn't any data. Once https://github.com/kubernetes/test-infra/pull/19044 is merged and once a few executions of that job complete, I'll check again and report back.

mm4tt:
Sounds good, please ping when it's ready

perfdash/config.go
```diff
@@ -179,16 +179,28 @@ var (
...
-				Parser:           parseSchedulingLatency,
-			}},
-			"SchedulingThroughput": []TestDescription{{
-				Name:             "density",
-				OutputFilePrefix: "SchedulingThroughput",
-				Parser:           parseSchedulingThroughputCL,
-			}},
+			"SchedulingLatency": []TestDescription{
+				{
+					Name:             "density_pod-anti-affinity",
```
adtac, line 184: `density_*` items need to be before the `density` item because of how data file prefixes work.

adtac:
Screenshot_2020-09-08 Screenshot(1)

@mm4tt See above for a screenshot of a local instance of perfdash showing the throughput data for the new 100-node job (pod-anti-affinity).

perfdash/config.go
mm4tt: Nice finding! Would you mind documenting it in a comment?

mm4tt (commented):
/approve

perfdash/config.go
adtac: Done.

mm4tt:
I'll deploy it shortly

mm4tt:
Done

PR #1443, opened 2020-08-28
Allow per-job path override for clusterloader2 results dir

chemikadze:
Some jobs may have layout that is placing results below root artifacts dir.

### DIFF
--- a/perfdash/metrics-downloader.go
+++ b/perfdash/metrics-downloader.go
@@ -121,7 +121,7 @@ func (g *Downloader) getJobData(wg *sync.WaitGroup, result JobToCategoryData, re
 			for testLabel, testDescriptions := range categoryMap {
 				for _, testDescription := range testDescriptions {
 					filePrefix := fmt.Sprintf("%v_%v", testDescription.OutputFilePrefix, testDescription.Name)
-					searchPrefix := fmt.Sprintf("artifacts/%v", filePrefix)
+					searchPrefix := g.artifactName(tests, filePrefix)
 					artifacts, err := g.MetricsBkt.ListFilesInBuild(job, buildNumber, searchPrefix)
 					if err != nil || len(artifacts) == 0 {
 						klog.Infof("Error while looking for %s* in %s build %v: %v", searchPrefix, job, buildNumber, err)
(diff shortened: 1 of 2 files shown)


mm4tt:
Please rebase
/ok-to-test

Could you provide a sample prow config of cl2 job that outputs results below root artifacts dir? 
What is the reason for that? Cannot you change the config to have CL2 results in /artifacts dir?

chemikadze:
> Could you provide a sample prow config of cl2 job that outputs results below root artifacts dir?
> What is the reason for that? Cannot you change the config to have CL2 results in /artifacts dir?

Don't have OSS example for that, but essentially that's a non-kubetest job that runs clusterloader2, providing --report-dir that's under ARTIFACTS to logically isolate cl2 output from anything else. It is possible to modify it to put it to root dir, though it would be neat to keep that grouping and make perfdash to be able to see past results.

One hypotetical extension of this use case I could see is multiple clusterloader2 runs per job (let say with different parameters), with multiple perfDashArtifactsDir per job specified.

chemikadze:
Rebased, thanks!

perfdash/metrics-downloader.go
```diff
@@ -130,10 +130,10 @@ func (g *Downloader) getJobData(wg *sync.WaitGroup, result JobToCategoryData, re
 					for _, artifact := range artifacts {
 						metricsFileName := filepath.Base(artifact)
 						resultCategory := getResultCategory(metricsFileName, filePrefix, categoryLabel, artifacts)
-						testDataResponse, err := g.MetricsBkt.ReadFile(job, buildNumber,
-							fmt.Sprintf("artifacts/%v", metricsFileName))
+						fileName := g.artifactName(tests, metricsFileName)
+						testDataResponse, err := g.MetricsBkt.ReadFile(job, buildNumber, fileName)
 						if err != nil {
-							klog.Infof("Error when reading response Body: %v", err)
+							klog.Errorf("Error when reading response Body for %q: %v", fileName, err)
```
wojtek-t, line 136: IIRC, those aren't necessary errors - please revert to Infof.

wojtek-t (commented):
One minor nit - other than that lgtm.

/approve

Issue #1099, opened 2020-03-03
Migrate perf-tests to Go modules


oxddr:
Yesterday I tried to release new version of perfdash, but failed to build a binary locally due to a missing dependency. I had a look into dependencies and it's a mess. We still use Godeps, which are now deprecated in favor of Go modules. We should migrate to go modules fully, drop support for Godeps and possibly ditch 'vendor/' dirs too (@mm4tt is there any reason we keep them?). 

There are couple of reasons for migrating to Go modules fully and dropping support for Godeps. 
1. `perfdash` can't be built without local changes to Makefile. 
1. Little visibility into dependencies version (`Godeps.json` is way less readable). 
1. Vast majority of dependencies is old. `perfdash` code still relies on v1.11.0-alpha.0-2155-g80bd7510df of k/k
1. `godep` is required to run perfdash tests. With Go modules we can use bare `go`. 
1. Tests run via Makefile under `perfdash/` are run differently than ones in presubmit. 

The only issue with migrating to go modules might be with relying on k/k as a module (https://github.com/kubernetes/kubernetes/issues/79384). There are existing workaround though. 

/priority critical-urgent

mm4tt:
In general, I'm in favor of this change. 
I don't think it's critical-urgent, though. Everything seems to be working (I'm quite sure that building perfdash works if you use proper go version and disable GOMODULES). And it definitely works without any issues for clusterloader2.

Downgrading the priority to important-soon
/priority important-soon

IIRC, we still have vendor dirs in some places because of the travis presubmits. But we know what to do to solve that problem (we should migrate our presubmits to prow, @jprzychodzen created one for building stuff that allowed us to get rid of vendor dir for probes/ and access-token util).

Janek, do you want to do it? Alternatively, I think it could be a good first issue (assuming we limit the scope somehow, e.g. to starting with perf-dash). In that case we should lay out a better plan.

oxddr:
/remove-priority critical-urgent
With https://github.com/kubernetes/perf-tests/pull/1100 merged, this is indeed not that urgent, as we will be able to again build `perfdash` at HEAD. 

I tried to check if this can be done with a simple PR, but I think there is more work needed to make sure we don't break anything. I could prepare a plan on how to do it, what is exactly in the scope (e.g. migration off Travis to Prow) and then we can discuss timeline. I suspect most of the work from the plan, could be a "good first issue" then.

mm4tt:
Sounds good, thanks!

[...]

PR #1447, opened 2020-09-01
feat: migrate benchmark package to golang modules

kushthedude:
Signed-off-by: Kush Trivedi <kushthedude@gmail.com>

Refers #1099


benchmark/go.mod
```diff
@@ -0,0 +1,35 @@
+module k8s.io/perf-tests/benchmark
+
+go 1.15
+
+require (
+	github.com/aclements/go-moremath v0.0.0-20190830160640-d16893ddf098 // indirect
```
wojtek-t, line 6: Why do we remove all of these from vendor/ ?

In other packages already using go modules (like clusterloader2/ vendor seem to be fully propagated...)

kushthedude: With this all the packages will be stored in GOPATH rather than in vendor. Like how perfdash, util-images does, I don't know why we still have a vendor folder in clusterloader2 when we have already migrated to go modules.

wojtek-t: That's actually desired to have vendor - we do exactly the same in the main repo:
https://github.com/kubernetes/kubernetes/tree/master/vendor

kushthedude: Got it I added the `go mod vendor`

kushthedude:
Tests Passed 😄

kushthedude:
Thanks @wojtek-t!

PR #1449, opened 2020-09-02
Increase page size in slo-monitor to 5000 (attempt #2).

mborsz:
The previous attempt to increase page size was incorrect (was changing a pagination that is noop).

It seems to work now:

```
I0902 09:54:33.023988       1 httplog.go:90] verb="GET" URI="/api/v1/pods?limit=5000&resourceVersion=0" latency=3.396776ms resp=200 UserAgent="slo-monitor/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="[::1]:49440":
I0902 09:54:34.031390       1 httplog.go:90] verb="GET" URI="/api/v1/events?fieldSelector=involvedObject.kind%3DPod%2Csource%3Dkubelet&limit=5000&resourceVersion=0" latency=34.948446ms resp=200 UserAgent="slo-monitor/v0.0.0 (linux/amd64) kubernetes/$Format" srcIP="[::1]:49440":
```

/assign @wojtek-t

PR #1454, opened 2020-09-03
[WIP] Use 5000 page size in OOMTracker

mborsz:
Experimental change to use 5000 page size in OOMTracker

The goal is to do similar change as https://github.com/kubernetes/perf-tests/pull/1449 but for OOMTrackerMeasurement.


wojtek-t:
@mborsz - are we going to proceed with this approach? Or should we close it?

PR #1457, opened 2020-09-07
Pin python and pip to v2 version

mm4tt:
Fixes https://github.com/kubernetes/kubernetes/issues/94473


jkaniuk:
Python2 is deprecated. Could you update to Python 3?

PR #1463, opened 2020-09-11
chore: remove vendor modules from benchmark 

kushthedude:
Signed-off-by: Kush Trivedi <kushthedude@gmail.com>

**What type of PR is this?**

/area dependency
- This PR removes vendor modules for benchmark sub-module.
- We don't need vendor modules as even go modules dependency management.

Refers #1099

```release-note
NONE
```


mm4tt:
Sorry, but we do we need the vendor dir? It's obsolete and redundant with the new `go mod` dep management system. In case of clusterloader2 we needed it for backward compatibility (travis presubmits), but it's not the case here. It's also not the case anymore for cl2, as we got rid of the travis presubmits. 

We should also remove the vendor dir from clusterloader2 and other places (e.g. benchmarks)

wojtek-t:
> Sorry, but we do we need the vendor dir? It's obsolete and redundant with the new go mod dep management system. In case of clusterloader2 we needed it for backward compatibility (travis presubmits), but it's not the case here. It's also not the case anymore for cl2, as we got rid of the travis presubmits.

> We should also remove the vendor dir from clusterloader2 and other places (e.g. benchmarks)

@mm4tt - That's not obvious to me - we have vendor/ in k/k repo and we don't have travis there at all.
There were other reasons for having it, and unless you know all of them, let's not remove them.

mm4tt:
> > Sorry, but we do we need the vendor dir? It's obsolete and redundant with the new go mod dep management system. In case of clusterloader2 we needed it for backward compatibility (travis presubmits), but it's not the case here. It's also not the case anymore for cl2, as we got rid of the travis presubmits.
> 
> > We should also remove the vendor dir from clusterloader2 and other places (e.g. benchmarks)
> 
> Matt Matejczyk - That's not obvious to me - we have vendor/ in k/k repo and we don't have travis there at all.
> There were other reasons for having it, and unless you know all of them, let's not remove them.

I don't know what are the reasons for k/k repo. I'd guess it has something to do with the complicated build system we have for kubernetes. e.g. doing the magic for the /staging dir requires all files to be on disk.

But I know that the only reason for k/perf-tests to keep the vendor dir was the travis presubmit. Given that we don't have the travis presubmit anymore, we're free to remove it. And we should do it, as it'll get rid of the redundancy and make some operations easier, e.g. updating the dependencies will be a few line, easy-to-review change instead of a change with 1k+ modified files.

wojtek-t:
> But I know that the only reason for k/perf-tests to keep the vendor dir was the travis presubmit. Given that we don't have the travis presubmit anymore, we're free to remove it. And we should do it, as it'll get rid of the redundancy and make some operations easier, e.g. updating the dependencies will be a few line, easy-to-review change instead of a change with 1k+ modified files.

I guess I disagree with it. Yes - the PRs are smaller, but it also makes introducing some backdoors (e.g. security vulnerabilities) super easy, because noone will be reviewing this code any more...

mm4tt:
> > But I know that the only reason for k/perf-tests to keep the vendor dir was the travis presubmit. Given that we don't have the travis presubmit anymore, we're free to remove it. And we should do it, as it'll get rid of the redundancy and make some operations easier, e.g. updating the dependencies will be a few line, easy-to-review change instead of a change with 1k+ modified files.
> 
> I guess I disagree with it. Yes - the PRs are smaller, but it also makes introducing some backdoors (e.g. security vulnerabilities) super easy, because noone will be reviewing this code any more...

I think it's quite the opposite. Consider bumping some k8s dep to the next minor, e.g.
```
k8s.io/apimachinery v0.18.0
```

to
```
k8s.io/apimachinery v0.19.0
```

Then the go mod change will be as above, a one liner. It's impossible to introduce any backdoors there as to do that someone would have to hack the k8s repo and place them there.

On the other hand, with vendor dir, it'll be hundreds of files, and it's super easy to smuggle something there - no one really reviews the thousand of lines "update vendor" changes. 

The other argument is that having both vendor and go.mod file creates a huge redundancy in the system. It's easy to imagine a situation where someone updates one, but forget about the other. It can lead to some very hard to debug bugs.

Unless we don't have a strong reason to have both go.mod and vendor dir (and we don't) we shouldn't keep both.

wojtek-t:
> Then the go mod change will be as above, a one liner. It's impossible to introduce any backdoors there as to do that someone would have to hack the k8s repo and place them there.

We have tens of random dependencies, which I have no reason to believe into by default.

> Unless we don't have a strong reason to have both go.mod and vendor dir (and we don't) we shouldn't keep both.

I'm actually on the opposite - until we fully understand why there are both in k/k, let's not remove.

mm4tt:
> > Then the go mod change will be as above, a one liner. It's impossible to introduce any backdoors there as to do that someone would have to hack the k8s repo and place them there.
> 
> We have tens of random dependencies, which I have no reason to believe into by default.
> 

Where do we have "tens of random dependencies"? We have 16 deps in the clusterloader2's go.mod, I don't see a single random one.

> > Unless we don't have a strong reason to have both go.mod and vendor dir (and we don't) we shouldn't keep both.
> 
> I'm actually on the opposite - until we fully understand why there are both in k/k, let's not remove.

I already explained one of the reasons. K/k is not a good example of typical go project when it comes to dependency management, we do a lot of things in our own quirky ways there. It doesn't make sense to look at it as an example.
Looking at the more typical repo, k/test-infra, the vendor dir has been removed there over a year ago - https://github.com/kubernetes/test-infra/pull/14036

wojtek-t:
> I already explained one of the reasons. K/k is not a good example of typical go project when it comes to dependency management, we do a lot of things in our own quirky ways there. It doesn't make sense to look at it as an example.

One example of reason is not a good explanation - the fact that one reason doesn't work for us, doesn't mean that other reasons will not work.

That said:

> Looking at the more typical repo, k/test-infra, the vendor dir has been removed there over a year ago - kubernetes/test-infra#14036

That seems like a much stronger reason to me. Conceptually perf-tests repo is quite similar to test-infra so that looks like a good argument to me.

kushthedude:
Should I amend the PR to move away from vendor modules?

mm4tt:
> Should I amend the PR to move away from vendor modules?

Yes :)

benchmark/go.sum
```diff
@@ -94,6 +94,7 @@ github.com/go-openapi/validate v0.18.0/go.mod h1:Uh4HdOzKt19xGIGm1qHf/ofbX1YQ4Y+
 github.com/go-ozzo/ozzo-validation v3.5.0+incompatible/go.mod h1:gsEKFIVnabGBt6mXmxK0MoFy+cZoTJY6mu5Ll3LVLBU=
 github.com/godbus/dbus v0.0.0-20151105175453-c7fdd8b5cd55/go.mod h1:/YcGZj5zSblfDWMMoOzV4fas9FZnQYTkDnsGvmh2Grw=
 github.com/gogo/protobuf v0.0.0-20171007142547-342cbe0a0415/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7atdtwQ=
+github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekfRL6a582T4P37/31XEstQ5p58=
```
mm4tt, line 97: Why do we touch go.sum? Also, we do we have two entries for glog?

mm4tt:
It's a huge change. Let's do them separately - cl2 & benchmark. It'll be a bit easier.

benchmark/go.sum
kushthedude: https://github.com/golang/go/issues/28456 The go.sum is still confusing as `go mod tidy` doesn't clear up the duplicate lines for modules rather it added another line.

kushthedude:
> It's a huge change. Let's do them separately - cl2 & benchmark. It'll be a bit easier.

I agree, removing the commits related to benchmark

kushthedude:
@mm4tt I don't guess the test failure is due to the changes are done in the PR, it is unable to fetch a dependency from bitbucket 😕

mm4tt:
Yeah, this is being addressed in https://github.com/kubernetes/perf-tests/pull/1482 as we speak.
Please wait for that PR to be merged then retest

mm4tt:
The PR seems to be touching both cl2 and benchmark files. Please fix

mm4tt:
Please also update the PR description

kushthedude:
> The PR seems to be touching both cl2 and benchmark files. Please fix

BUILD files are removed from cl2 which were already present in .gitignore. If the following files are to be committed then should the gitignore be amended?

wojtek-t:
> BUILD files are removed from cl2 because they were listed in .gitignore. If the following files are to be committed then should the gitignore be amended?

Please don't mix them. It's fine to remove those but in a separate PR.

kushthedude:
Got it, amending the PR

kushthedude:
Fixed!

mm4tt:
/lgtm
/approve

Thanks!

PR #1464, opened 2020-09-11
cl2/density: add pod affinty, pod topology spreading workloads to scheduler suite

adtac:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: This PR adds the pod-affinity test to the suite of scheduler benchmark(s) we have.

Related to https://github.com/kubernetes/perf-tests/issues/1414

**Special notes for your reviewer**: No.

```release-note
NONE
```

/sig scheduling
/sig testing
/cc @ahg-g @mm4tt

### DIFF
--- a/clusterloader2/testing/density/scheduler-suite.yaml
+++ b/clusterloader2/testing/density/scheduler-suite.yaml
@@ -11,3 +11,8 @@
   configPath: testing/density/config.yaml
   overridePaths:
   - testing/density/scheduler/pod-anti-affinity/overrides.yaml
+
+- identifier: pod-topology-spread
+  configPath: testing/density/config.yaml
+  overridePaths:
+  - testing/density/scheduler/pod-topology-spread/overrides.yaml
--- a/clusterloader2/testing/density/scheduler-suite.yaml
+++ b/clusterloader2/testing/density/scheduler-suite.yaml
@@ -1,3 +1,8 @@
+- identifier: pod-affinity
+  configPath: testing/density/config.yaml
+  overridePaths:
+  - testing/density/scheduler/pod-affinity/overrides.yaml
+
 - identifier: pod-anti-affinity
   configPath: testing/density/config.yaml
   overridePaths:
--- a/clusterloader2/testing/density/scheduler-suite.yaml
+++ b/clusterloader2/testing/density/scheduler-suite.yaml
@@ -1,3 +1,7 @@
+- identifier: vanilla
+  configPath: testing/density/config.yaml
+  overridePaths: []
+
 - identifier: pod-affinity
   configPath: testing/density/config.yaml
   overridePaths:
--- a/clusterloader2/testing/density/scheduler/pod-affinity/deployment.yaml
+++ b/clusterloader2/testing/density/scheduler/pod-affinity/deployment.yaml
@@ -1,0 +1,46 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  labels:
+    group: {{.Group}}
+spec:
+  replicas: {{.Replicas}}
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        name: {{.Name}}
+        group: {{.Group}}
+    spec:
+      affinity:
+        podAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  group: {{.Group}}
+              topologyKey: "kubernetes.io/hostname"
+            weight: 1
+      containers:
+      - image: k8s.gcr.io/pause:3.1
+        imagePullPolicy: IfNotPresent
+        name: {{.Name}}
+        ports:
+        resources:
+          requests:
+            cpu: {{.CpuRequest}}
+            memory: {{.MemoryRequest}}
+      # Add not-ready/unreachable tolerations for 15 minutes so that node
+      # failure doesn't trigger pod deletion.
+      tolerations:
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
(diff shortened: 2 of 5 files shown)


adtac:
I've tested this locally with my personal cluster to make sure everything looks correct, but I don't think my local throughput/latency metrics will be useful to anyone as I'm temporarily using a real cluster to test with instead of kubemark (because my kubemark set up broke for some reason).

clusterloader2/testing/density/config.yaml
```diff
@@ -15,7 +15,7 @@
 # This results in decreased probability of choosing the same node again.
 {{$LATENCY_POD_CPU := DefaultParam .LATENCY_POD_CPU 100}}
 {{$LATENCY_POD_MEMORY := DefaultParam .LATENCY_POD_MEMORY 350}}
-{{$MIN_LATENCY_PODS := 500}}
+{{$MIN_LATENCY_PODS := DefaultParam .MIN_LATENCY_PODS 500}}
```
ahg-g, line 18: why is this needed?

ahg-g (commented):
/ok-to-test

clusterloader2/testing/density/config.yaml
adtac: Even though there exists no such override in-tree or in this PR, making this overridable makes it easier to test multiple cl2 executions quickly without having to wait for 3000 saturation pods / 500 latency pods.

adtac:
Added the vanilla workload to the suite for two reasons:

1. Everything else is kept constant if we use the scheduler suite instead of a separate job for the vanilla workload (even though the two jobs are very similar right now).
1. Convenience: it's easier to switch between different workloads whilst maintaining the same metrics in perf-dash if the job name doesn't have to change (changing job name resets other dropdowns).

ahg-g:
> Added the vanilla workload to the suite for two reasons:
> 
> 1. Everything else is kept constant if we use the scheduler suite instead of a separate job for the vanilla workload (even though the two jobs are very similar right now).
> 2. Convenience: it's easier to switch between different workloads whilst maintaining the same metrics in perf-dash if the job name doesn't have to change (changing job name resets other dropdowns).

for 5k runs this may not be cheap though, but differing to Matt on this.

/lgtm from the scheduler

adtac:
Added the pod topology spreading workload as well. Re-pasting the commit message:

    cl2/density: add pod topology spread workload to scheduler suite
    
    This pod topology spread workload uses the hostname as the topology key
    create a spreading deployment with a max skew of 1. This is effectively
    a round-robin.
    
    The hostname was used as the topology key due to the lack of a better
    label. Ideally, I would've wanted to benchmark zone spreading with
    topology.kubernetes.io/zone as the topology key, but since all Kubemark
    kubemark nodes have identical labels except for kubernetes.io/hostname,
    this is the only viable option. In the future, we can look into the
    following options to make this more even representative of a real
    workload:
    
    1. Make Kubemark hollow nodes be actually present in different GCP
       zones, thereby guaranteeing that each node has a different
       topology.kubernetes.io/zone.
    2. Set a dummy label on hollow nodes to be used as the spreading
       topology key at Kubemark cluster creation time. This would probably
       be a kubetest change.
    3. Allow setting and clearing node labels as a part of clusterloader.
    
    I think (3) is the cleanest option if we end up doing this, but whether
    or not setting node labels is within the scope of clusterloader is
    debatable.

adtac:
Scheduling 1000 saturation pods onto a personal cluster, I've verified that spreading works as expected -- after scheduling each pod, the difference between the maximum and minimum number of pods running on each node is never greater than 1 (except for once or twice, but that's expected due to `ScheduleAnyway`).

clusterloader2/testing/density/config.yaml
mm4tt: Given that it's not used anywhere here it's a bit "yagni" to me. 
I'm fine with submitting this if you need it for manual testing, but in that case let's move it to a separate PR

clusterloader2/testing/density/scheduler-suite.yaml
```diff
@@ -1,4 +1,18 @@
+- identifier: vanilla
```
mm4tt, line 1: Do we know how long the test will take at 5k node scale? What about the whole suite? If we don't please DM me, I'll help you with running the suite manually

I'm tentatively supportive for having the vanilla being part of the suite, but before I make the final call I'd like to have more data.

adtac: According to past data: the throughput test (saturation deployment) is expected to take around ~25 minutes while the latency test is expected to take around ~11 hours. The test in total takes around ~13 hours, unfortunately :(

I think that's a bit much for a convenience feature, yes?

clusterloader2/testing/density/config.yaml
adtac: Done: https://github.com/kubernetes/perf-tests/pull/1467

clusterloader2/testing/density/scheduler-suite.yaml
mm4tt: Wow, 11 hours? Why does it take that long? The whole vanilla density test takes only 110 min  - https://k8s-testgrid.appspot.com/sig-scalability-kubemark#kubemark-5000&grid=beta&graph-metrics=test-duration-minutes
FPAQY9mqKpFqgb4

adtac: Ah nevermind, `testing/density/config.yaml` finishes in ~ 1h50m. The other 10+ hours is `testing/load/config.yaml`. Finally, `testing/access-tokens/config.yaml` also runs in the same job and it finishes in ~ 6m. I'm not fully aware of what `testing/load/config.yaml` does differently.

Since the scheduler suite uses `test/density/config.yaml`, I don't foresee any running time length issues with the 5k job :+1:

PR #1476, opened 2020-09-17
Collect resource usage from non-joined master nodes

chemikadze:
Some providers do not have master nodes joined to a cluster, so
fallback to SSH is added for such providers.

Resource gatherer will merge nodes registered in a cluster with
master ips, deduplicating by address, and will use old method of
collecting container ids and gathering metrics, when possible.

For master ips that are not joined to a cluster as a node, gatherer
will use provider's SSH method, and will collect metrics of all
containers registered on that node, including ones running
as part of static pods.

Kubelet package is extended with ssh stats gatherer, and generalized
way to provide container filtering (either list based, or match-all).

### DIFF
--- a/clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
+++ b/clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
@@ -30,6 +30,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/kubelet"
 	"k8s.io/perf-tests/clusterloader2/pkg/provider"
 	pkgutil "k8s.io/perf-tests/clusterloader2/pkg/util"
 )
@@ -71,6 +72,7 @@ type ContainerResourceGatherer struct {
 type ResourceGathererOptions struct {
 	InKubemark                        bool
 	Nodes                             NodesSet
+	MasterIPs                         []string
 	ResourceDataGatheringPeriod       time.Duration
 	MasterResourceDataGatheringPeriod time.Duration
 }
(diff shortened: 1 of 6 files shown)


clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
```diff
@@ -146,18 +148,37 @@ func NewResourceUsageGatherer(c clientset.Interface, host string, port int, prov
...
 		}
 
+		for _, address := range options.MasterIPs {
+			nodesToConsider[address] = true
+		}
+		masterIPs := sets.NewString(options.MasterIPs...)
+		nodeNames := sets.NewString(masterIPs.List()...)
 		for _, node := range nodeList.Items {
-			if options.Nodes == AllNodes || masterNodes.Has(node.Name) || nodesToConsider[node.Name] {
+			nodeNames.Insert(node.Name)
```
mm4tt, line 157: Looks like we're mixing node IPs and node names. I don't like it :)

mm4tt (commented):
/ok-to-test

Could you rephrase the second paragraph of the PR description:
```
Because pods are not visible via API, all pods are
matched for non-joined hosts.
```

I don't fully understand it. I'd also like to be able to tell from the description what this PR does.
From a cursory glance, I assume it allows to collect resource usage from non-joined master nodes by having them listed in the masterIPs env/flag, is it right?

clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
chemikadze: Good callout - also potentially may cause node to appear twice (if both appearing as node, and specified as ip). Will look into it.

chemikadze:
> From a cursory glance, I assume it allows to collect resource usage from non-joined master nodes by having them listed in the masterIPs env/flag, is it right?

That's right, will add more clear description.

chemikadze:
@mm4tt I've updated PR description and addressed comments, can you please take a look?

wojtek-t:
/assign @mm4tt @mborsz 

@mm4tt @mborsz - can you please take a look?

PR #1477, opened 2020-09-18
Fix memory usage when listing pods/events

marseel:
**What type of PR is this?**
Limits memory usage after slo-monitor restart

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/slo-monitor/Makefile
+++ b/slo-monitor/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 PACKAGE = k8s.io/perf-tests/slo-monitor
-TAG = 0.14.0
+TAG = 0.15.0
 # Image should be pulled from k8s.gcr.io, which will auto-detect
 # region (us, eu, asia, ...) and pull from the closest.
 REPOSITORY?=k8s.gcr.io
--- a/slo-monitor/src/monitors/pod_monitor.go
+++ b/slo-monitor/src/monitors/pod_monitor.go
@@ -129,8 +129,20 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
 	controller := NewWatcherWithHandler(
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-				// This call is paginated by cache.Reflector.
-				return pm.kubeClient.CoreV1().Pods(v1.NamespaceAll).List(options)
+				// slo-monitor is not using the events returned from the list
+				// (it uses NoStoreQueue implementation of Store which discards them),
+				// only the resourceVersion is used to instantiate the watch from this point.
+				// This trick allows us to reduce memory usage on startup and further relists.
+				o := metav1.ListOptions{
+					Limit: 1,
+				}
+				result, err := pm.kubeClient.CoreV1().Pods(v1.NamespaceAll).List(o)
+				if err != nil {
+					return nil, err
+				}
+				result.Continue = ""
+				result.Items = nil
+				return result, nil
 			},
 			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
 				return pm.kubeClient.CoreV1().Pods(v1.NamespaceAll).Watch(options)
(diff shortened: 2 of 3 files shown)


slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -130,7 +130,15 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
 				// This call is paginated by cache.Reflector.
-				return pm.kubeClient.CoreV1().Pods(v1.NamespaceAll).List(options)
+				o := metav1.ListOptions{
```
wojtek-t, line 133: Add a comment explaining why we're doing this strange thing (also remove the comment above, it's no longer really relevant).

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -130,7 +130,15 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
...
+					Limit: 1,
+				}
+				result, err := pm.kubeClient.CoreV1().Pods(v1.NamespaceAll).List(o)
+				if err != nil {
+					return nil, err
+				}
+				result.Continue = ""
```
wojtek-t, line 140: Let's also remove the Items of the list - we don't really list them - it will be easier to catch errors if someone will change other parts.

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -175,9 +183,16 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
 	eventController := NewWatcherWithHandler(
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-				options.FieldSelector = eventSelector
-				// This call is paginated by cache.Reflector.
-				return pm.kubeClient.CoreV1().Events(v1.NamespaceAll).List(options)
+				o := metav1.ListOptions{
```
wojtek-t, line 186: Same comments as above...

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -175,9 +183,16 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
...
+					Limit: 1,
+					FieldSelector: eventSelector,
```
wojtek-t, line 188: Actually - you don't want this FieldSelector.

slo-monitor/src/monitors/watcher.go
```diff
@@ -28,7 +28,6 @@ import (
 
 const (
 	resyncPeriod      = time.Duration(0)
-	watchListPageSize = 5000
```
wojtek-t, line 31: @mborsz - given we don't need this, do we want to revert the changes you were doing?

mborsz: Yes, once we validate this approach, we can submit follow up PR that will remove this, fork of client-go and other ugly stuff I added.

slo-monitor/src/monitors/pod_monitor.go
marseel: Done.

slo-monitor/src/monitors/pod_monitor.go
marseel: Done.

slo-monitor/Makefile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 PACKAGE = k8s.io/perf-tests/slo-monitor
-TAG = 0.14.0
+TAG = 0.14.1
```
wojtek-t, line 16: let's make 0.15 - it's a bit non-negligible change

slo-monitor/src/monitors/pod_monitor.go
marseel: Done.

slo-monitor/src/monitors/pod_monitor.go
marseel: Done.

slo-monitor/src/monitors/pod_monitor.go
marseel: Done.

slo-monitor/Makefile
marseel: Done.

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -175,9 +186,19 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
 	eventController := NewWatcherWithHandler(
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-				options.FieldSelector = eventSelector
-				// This call is paginated by cache.Reflector.
-				return pm.kubeClient.CoreV1().Events(v1.NamespaceAll).List(options)
+				// slo monitor is not using list of events.
```
mborsz, line 189: I would add more context, like: NewWatcherWithHandler internally uses NoStore store that ignores all relist results, so here we do List call only to get the current resource version.

slo-monitor/src/monitors/pod_monitor.go
```diff
@@ -129,8 +129,19 @@ func (pm *PodStartupLatencyDataMonitor) Run(stopCh chan struct{}) error {
 	controller := NewWatcherWithHandler(
 		&cache.ListWatch{
 			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-				// This call is paginated by cache.Reflector.
-				return pm.kubeClient.CoreV1().Pods(v1.NamespaceAll).List(options)
+				// slo monitor is not using list of pods.
+				// It's only used to get resourceVersion for Watch.
+				// This implementation allows us to reduce memory usage significantly for large clusters.
```
wojtek-t, line 134: nit: 
``` 
slo-monitor is not using the pods returned from the list
(it uses NoStore implementation of Store which discards them),
only the resourceVersion is used to instantiate the watch from this point.
This trick allows us to reduce memory usage on startup and further relists.
```

wojtek-t:
/ok-to-test

/approve

Will let @mborsz to lgtm

mborsz:
/lgtm

Thanks!

PR #1478, opened 2020-09-18
Append ordinal numbers to test steps names

tosi3k:
**What type of PR is this?**
/kind bug

If test configs have steps with duplicate names, they will be erroneously shown as one in the Prow GCS viewer. The change will prevent from such situations as we add ordinal numbers at the end of the test steps' names.

No issue is opened for that.

/assign wojtek-t

```release-note
NONE
```


clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -119,9 +119,11 @@ func (ste *simpleExecutor) ExecuteTestSteps(ctx Context, conf *api.Config) *erro
 
 	errList := errors.NewErrorList()
 	for i := range conf.Steps {
-		if conf.Steps[i].Name == "" {
-			conf.Steps[i].Name = fmt.Sprintf("Step number %d [autogenerated, please name your step in the test config]", i+1)
+		namePrefix := conf.Steps[i].Name
+		if namePrefix == "" {
+			namePrefix = "[autogenerated, please name your step in the test config]"
 		}
+		conf.Steps[i].Name = fmt.Sprintf("%s (step no. %d)", namePrefix, i+1)
```
wojtek-t, line 126: nit: [step: %d]

tosi3k: Done.

mm4tt:
Nice, but I'd prefer if the step number was the prefix of the step name. By doing so, we'd have the tests displayed nicely in the test_grid. Currently they are ordered alphabetically, which is odd 
8EzZbHYPbsqCHJK

PR #1480, opened 2020-09-21
Slo-monitor cleanup

mborsz:
Removing client-go's fork that is no longer used.
/assign @wojtek-t 
/cc @marseel


wojtek-t:
/lgtm
/approve

Thanks!

PR #1485, opened 2020-09-22
Add density test to load test

marseel:
**What type of PR is this?**
/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -19,6 +19,7 @@
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
+{{$ENABLE_DENSITY_TEST := DefaultParam .CL2_ENABLE_DENSITY_TEST false}}
 {{$ENABLE_SYSTEM_POD_METRICS:= DefaultParam .ENABLE_SYSTEM_POD_METRICS true}}
 {{$ENABLE_CLUSTER_OOMS_TRACKER := DefaultParam .CL2_ENABLE_CLUSTER_OOMS_TRACKER false}}
 {{$CLUSTER_OOMS_IGNORED_PROCESSES := DefaultParam .CL2_CLUSTER_OOMS_IGNORED_PROCESSES ""}}
@@ -48,6 +49,27 @@
 {{$mediumDeploymentsPerNamespace := SubtractInt $mediumDeploymentsPerNamespace 1}}
 {{$bigDeploymentsPerNamespace := SubtractInt $bigDeploymentsPerNamespace 1}}
 
+# The minimal number of pods to be used to measure various things like
+# pod-startup-latency or scheduler-throughput. The purpose of it is to avoid
+# problems in small clusters where we wouldn't have enough samples (pods) to
+# measure things accurately.
+{{$MIN_PODS_IN_SMALL_CLUSTERS := 500}}
+
+# BEGIN scheduler-throughput section
+# TODO( https://github.com/kubernetes/perf-tests/issues/1027): Lower the number of "min-pods" once we fix the scheduler throughput measurement.
+{{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) (MultiplyInt 2 .Nodes)}}
+{{$schedulerThroughputPodsPerDeployment := .Nodes}}
+{{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
+{{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 0}}
+# END scheduler-throughput section
+
+# TODO(https://github.com/kubernetes/perf-tests/issues/1024): Investigate and get rid of this section.
+# BEGIN pod-startup-latency section
+{{$totalLatencyPods := MaxInt $MIN_PODS_IN_SMALL_CLUSTERS .Nodes}}
+{{$latencyReplicas := DivideInt $totalLatencyPods $namespaces}}
+{{$podStartupLatencyThreshold := DefaultParam .CL2_POD_STARTUP_LATENCY_THRESHOLD "1h"}}
+# END pod-startup-latency section
+
 # Probe measurements shared parameter
 {{$PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT := DefaultParam .CL2_PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT "5m"}}
 
@@ -56,11 +78,20 @@
 
 name: load
 namespace:
-  number: {{$namespaces}}
+  number: {{AddInt $namespaces $schedulerThroughputNamespaces}}
 tuningSets:
 - name: Sequence
   parallelismLimitedLoad:
     parallelismLimit: 1
+# Dedicated tuningSet for SchedulerThroughput phases that results in fully
+# parallel creation of deployments.
+- name: SchedulerThroughputParallel
+  parallelismLimitedLoad:
+    parallelismLimit: {{$schedulerThroughputNamespaces}}
+# TODO(https://github.com/kubernetes/perf-tests/issues/1024): This TuningSet is used only for pod-startup-latency, get rid of it
+- name: Uniform5qps
+  qpsLoad:
+    qps: 5
 - name: RandomizedSaturationTimeLimited
   RandomizedTimeLimitedLoad:
     timeLimit: {{$saturationTime}}s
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -39,15 +39,35 @@
 {{$mediumDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 4 $MEDIUM_GROUP_SIZE)}}
 # smallDeployments - 1/2 of namespace pods should be in small Deployments.
 {{$smallDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 2 $SMALL_GROUP_SIZE)}}
-# Reduce the number of small and medium deployments per namespace
+# If StatefulSets are enabled reduce the number of small and medium deployments per namespace
 {{$smallDeploymentsPerNamespace := SubtractInt $smallDeploymentsPerNamespace $SMALL_STATEFUL_SETS_PER_NAMESPACE}}
 {{$mediumDeploymentsPerNamespace := SubtractInt $mediumDeploymentsPerNamespace $MEDIUM_STATEFUL_SETS_PER_NAMESPACE}}
 
-# Reduce the number of small, medium, big deployments per namespace.
+# If Jobs are enabled reduce the number of small, medium, big deployments per namespace.
```
mborsz, line 46: Revert?

clusterloader2/testing/load/config.yaml
```diff
@@ -39,15 +39,35 @@
 {{$mediumDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 4 $MEDIUM_GROUP_SIZE)}}
 # smallDeployments - 1/2 of namespace pods should be in small Deployments.
 {{$smallDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 2 $SMALL_GROUP_SIZE)}}
-# Reduce the number of small and medium deployments per namespace
+# If StatefulSets are enabled reduce the number of small and medium deployments per namespace
```
mborsz, line 42: This doesn't seem to be density-related. Please revert?

clusterloader2/testing/load/config.yaml
```diff
@@ -56,11 +76,20 @@
 
 name: load
 namespace:
-  number: {{$namespaces}}
+  number: {{AddInt $namespaces $schedulerThroughputNamespaces}}
```
mborsz, line 79: Technically, we should wrap this into if to make this PR noop without setting the additional env variable.

But in practice, we can leave this without additional if I think. The expected value for schedulerThroughputNamespaces is approx 2, so I don't think creating two more empty namespaces in all tests will be an issue.

clusterloader2/testing/load/config.yaml
```diff
@@ -74,8 +103,8 @@ tuningSets:
 chaosMonkey:
   nodeFailure:
     failureRate: 0.01
-    interval: 5m
-    jitterFactor: 2.0
+    interval: 1m
```
mborsz, line 106: Revert?

clusterloader2/testing/load/config.yaml
```diff
@@ -216,6 +245,7 @@ steps:
         ReplicasMin: {{$BIG_GROUP_SIZE}}
         ReplicasMax: {{$BIG_GROUP_SIZE}}
         SvcName: big-service
+        CpuRequest: 5m
```
mborsz, line 248: revert?

clusterloader2/testing/load/config.yaml
```diff
@@ -236,6 +266,7 @@ steps:
         ReplicasMin: {{$MEDIUM_GROUP_SIZE}}
         ReplicasMax: {{$MEDIUM_GROUP_SIZE}}
         SvcName: medium-service
+        CpuRequest: 5m
```
mborsz, line 269: revert?

clusterloader2/testing/load/config.yaml
```diff
@@ -339,6 +371,148 @@ steps:
       {{end}}
 {{end}}
 
+# BEGIN scheduler throughput
```
mborsz, line 374: Wrap this section in if?

clusterloader2/testing/load/config.yaml
```diff
@@ -89,12 +118,12 @@ steps:
     Method: APIResponsivenessPrometheus
     Params:
       action: start
-  - Identifier: PodStartupLatency
+  - Identifier: CreatePhasePodStartupLatency
```
mborsz, line 121: In fact this is a good change that can be left without if

clusterloader2/testing/load/config.yaml
wojtek-t, line 374: @mm4tt @mborsz - I started wondering about this.
Given we look at max from the samples, Actually wouldn't it make sense to start it at the beginning and gather at the end? [Not necessary in this PR, but it would allow us to join this with the SLO measurement once we have that done.]

clusterloader2/testing/load/config.yaml
wojtek-t: Yes - same above.

clusterloader2/testing/load/config.yaml
marseel: Done

clusterloader2/testing/load/config.yaml
```diff
@@ -19,6 +19,7 @@
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
+{{$ENABLE_DENSITY_TEST := DefaultParam .CL2_DENSITY_TEST false}}
```
mborsz, line 22: I think CL2_ENABLE_DENSITY_TEST is better variable name: it avoids unnecessary renaming.

clusterloader2/testing/load/config.yaml
```diff
@@ -339,6 +369,150 @@ steps:
...
+- name: Waiting for scheduler throughput pods to be deleted
+  measurements:
+  - Identifier: WaitForSchedulerThroughputDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
+# END scheduler throughput
+{{end}}
+
+# TODO(https://github.com/kubernetes/perf-tests/issues/1024): Ideally, we wouldn't need this section.
```
mborsz, line 449: This section should be also in if :)

In general, this PR shouldn't change anything unless ENABLE_DENSITY_TEST is set.

marseel: typo, thanks for noticing.

clusterloader2/testing/load/config.yaml
marseel: typo, thanks for noticing.

clusterloader2/testing/load/config.yaml
```diff
@@ -89,12 +119,12 @@ steps:
     Method: APIResponsivenessPrometheus
     Params:
       action: start
-  - Identifier: PodStartupLatency
+  - Identifier: CreatePhasePodStartupLatency
```
jprzychodzen, line 122: Please remember to update https://github.com/kubernetes/perf-tests/blob/master/perfdash/config.go

clusterloader2/testing/load/config.yaml
mborsz: Let's move the renaming to a separate PR (so this line and the one at the end of the file)

marseel: Done.

wojtek-t:
This failure suggests that something is broken:
```
- PodStartupLatency error: metric PodStartupLatency has not been started
```

marseel:
@wojtek-t yes, I saw it. I will check it ;)

clusterloader2/testing/load/config.yaml
mborsz: I think we may want to do something like that in future, but before that we need to make sure that we create in test large enough deployments to make the scheduler be able to achieve full throughput. AFAIR currently we create max 250 pods per deployment, which should give ~2.5s of max scheduler throughput. I'm not sure if we are able to detect this with any measurement we have now (the current scheduler throughput uses 5s window and even if we switch to prometheus there, I expect similar window size).

If we start creating ~1k pods at the same time (to have > measurement windows to make it reliable), I think we can easily get rid of this part.

wojtek-t: What I had on my mind was basically leaving those additional deployments in this phase for now (since as you mentioned we need them), just moving scheduler throughput measurement (as it only looks at max 5s window).
So if there will be good window here, there will also be a good window across the whole test :)

But that's not for this particular PR - we can do that later.

mborsz:
@marseel, LGTM once you flip "ENABLE_DENSITY_TEST" back to "false".

wojtek-t:
@marseel - something strange happened to this PR - it has a lot of lines.

Also - please squash the commits.

marseel:
@wojtek-t yea, I was rebasing it. 
I've squash commits.

clusterloader2/testing/load/config.yaml
```diff
@@ -339,6 +379,151 @@ steps:
...
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
+- name: Collecting pod startup latency
+  measurements:
+  - Identifier: PodStartupLatency
+    Method: PodStartupLatency
+    Params:
+      action: gather
+      threshold: {{$podStartupLatencyThreshold}}
```
mborsz, line 523: threshold is a parameter for action:start call

PR #1505, opened 2020-09-29
Reimplement scheduling throughput to prometheus metric

marseel:
**What type of PR is this?**
/kind cleanup

Before SchedulingThroughput measurement was doing expensive calls to apiserver. Currently, it will only use prometheus metric to calculate scheduling throughput. 

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/scheduling_throughput_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/scheduling_throughput_prometheus.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	schedulingThroughputPrometheusMeasurementName = "SchedulingThroughputPrometheus"
+
+	maxSchedulingThroughputQuery = `max_over_time(sum(irate(apiserver_request_total{verb="POST", resource="pods", subresource="binding",code="201"}[1m]))[%v:5s])`
+)
+
+func init() {
+	create := func() measurement.Measurement { return CreatePrometheusMeasurement(&schedulingThroughputGatherer{}) }
+	if err := measurement.Register(schedulingThroughputPrometheusMeasurementName, create); err != nil {
+		klog.Fatalf("Cannot register %s: %v", schedulingThroughputMeasurementName, err)
+	}
+}
+
+type schedulingThroughputGatherer struct{}
+
+func (a *schedulingThroughputGatherer) Gather(executor QueryExecutor, startTime time.Time, config *measurement.Config) ([]measurement.Summary, error) {
+	throughputSummary, err := a.getThroughputSummary(executor, startTime, config)
+	if err != nil {
+		return nil, err
+	}
+
+	content, err := util.PrettyPrintJSON(throughputSummary)
+	if err != nil {
+		return nil, err
+	}
+
+	threshold, err := util.GetFloat64OrDefault(config.Params, "threshold", 0)
+	if threshold > 0 && throughputSummary.Max < threshold {
+		err = errors.NewMetricViolationError(
+			"scheduler throughput_prometheus",
+			fmt.Sprintf("actual throughput %f lower than threshold %f", throughputSummary.Max, threshold))
+	}
+
+	summaries := []measurement.Summary{
+		measurement.CreateSummary(a.String(), "json", content),
+	}
+
+	return summaries, err
+}
+
+func (a *schedulingThroughputGatherer) getThroughputSummary(executor QueryExecutor, startTime time.Time, config *measurement.Config) (*schedulingThroughputPrometheus, error) {
+	measurementEnd := time.Now()
+	measurementDuration := measurementEnd.Sub(startTime)
+	promDuration := measurementutil.ToPrometheusTime(measurementDuration)
+	query := fmt.Sprintf(maxSchedulingThroughputQuery, promDuration)
+
+	samples, err := executor.Query(query, measurementEnd)
+	if err != nil {
+		return nil, err
+	}
+	if len(samples) != 1 {
+		return nil, fmt.Errorf("got unexpected number of samples: %d", len(samples))
+	}
+
+	maxSchedulingThroughput := samples[0].Value
+	throughputSummary := &schedulingThroughputPrometheus{
+		Max: float64(maxSchedulingThroughput),
+	}
+
+	return throughputSummary, nil
+}
+
+func (a *schedulingThroughputGatherer) String() string {
+	return schedulingThroughputPrometheusMeasurementName
+}
+
+func (a *schedulingThroughputGatherer) IsEnabled(config *measurement.Config) bool {
+	return true
+}
+
+type schedulingThroughputPrometheus struct {
+	Max float64 `json:"max"`
+}


clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
 
 const (
 	schedulingThroughputMeasurementName = "SchedulingThroughput"
-	defaultSchedulingThroughputInterval = 5 * time.Second
+
```
wojtek-t, line 32: It might be safer to instead of changing the existing measurement, create a new one next to the existing ones and keep two of them at the same time running for a bit of time.

In case of issues, the rollback is easy and you don't need to rollback everything - just disable it in tests.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
...
+	return summaries, err
 }
 
-func (s *schedulingThroughputMeasurement) stop() {
-	if s.isRunning {
-		close(s.stopCh)
-		s.isRunning = false
+func (a *schedulingThroughputGatherer) getThroughputSummary(executor QueryExecutor, startTime time.Time, config *measurement.Config) (*schedulingThroughput, error) {
+
```
wojtek-t, line 79: nit: remove this empty line

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
...
+	measurementEnd := time.Now()
+	measurementDuration := measurementEnd.Sub(startTime)
+	promDuration := measurementutil.ToPrometheusTime(measurementDuration)
+
+	query := fmt.Sprintf(maxSchedulingThroughputQuery, promDuration)
+	samples, err := executor.Query(query, measurementEnd)
+
```
wojtek-t, line 86: nit: remove this empty line

Generally, you should split the conceptual blocks with empty lines - error handling is clearly the same block as where the err was computed.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
```
wojtek-t, line 90: same here

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
...
+func (a *schedulingThroughputGatherer) IsEnabled(config *measurement.Config) bool {
+	return true
 }
 
 type schedulingThroughput struct {
-	Perc50 float64 `json:"perc50"`
-	Perc90 float64 `json:"perc90"`
-	Perc99 float64 `json:"perc99"`
-	Max    float64 `json:"max"`
+	Max float64 `json:"max"`
```
wojtek-t, line 113: @mborsz - will removing the fields here require changes in perf-dash?

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
...
-}
-
-// String returns a string representation of the measurement.
-func (*schedulingThroughputMeasurement) String() string {
-	return schedulingThroughputMeasurementName
-}
 
-func (s *schedulingThroughputMeasurement) start(clientSet clientset.Interface, selector *measurementutil.ObjectSelector, measurmentInterval time.Duration) error {
-	ps, err := measurementutil.NewPodStore(clientSet, selector)
+	summaryName, err := util.GetStringOrDefault(config.Params, "summaryName", a.String())
```
wojtek-t, line 52: Why can we need a param for this?

wojtek-t: I vote for removing this and using "schedulingThroughputMeasurementName" as previously.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
```diff
@@ -32,153 +29,86 @@ import (
...
+		measurement.CreateSummary(summaryName, "json", content),
+	}
+
+	threshold, err := util.GetFloat64OrDefault(config.Params, "threshold", 0)
+
 	if threshold > 0 && throughputSummary.Max < threshold {
 		err = errors.NewMetricViolationError(
 			"scheduler throughput",
 			fmt.Sprintf("actual throughput %f lower than threshold %f", throughputSummary.Max, threshold))
+		klog.Errorf("%s: %v", a, err)
```
wojtek-t, line 72: remove

clusterloader2/testing/density/config.yaml
```diff
@@ -115,7 +115,6 @@ steps:
     Method: SchedulingThroughput
     Params:
       action: start
-      labelSelector: group = saturation
```
wojtek-t, line 118: Going back to my first comment:
I actually vote for making it a new measurement, and simply:
- start it at the beginning of the test
- finish it at the end of the test
- not touch the existing measurement

Once we prove it's stable, we will simply get rid of this one (and we will be able to merge that with SLOMeasurement).

clusterloader2/pkg/measurement/common/scheduling_throughput.go
marseel: Removed. I was following other prometheus based slos, not sure why they have it.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
marseel: Removed

clusterloader2/pkg/measurement/common/scheduling_throughput.go
marseel: Actually that's exactly what I did before I merged them :D I moved it back to separate file.

wojtek-t:
/ok-to-test

This looks fine to me - will let Maciek to stamp it.

clusterloader2/pkg/measurement/common/scheduling_throughput_prometheus.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (a *schedulingThroughputGatherer) String() string {
+	return schedulingThroughputPrometheusMeasurementName
+}
+
+func (a *schedulingThroughputGatherer) IsEnabled(config *measurement.Config) bool {
+	return true
+}
+
+type schedulingThroughputPrometheus struct {
```
mborsz, line 100: This requires changes to perfdash: https://github.com/kubernetes/perf-tests/blob/28d4b32e0d2cb9b5b94a8769286bfd0f0ca771d8/perfdash/parser.go#L287-L309

I think we can submit this, and fix perfdash in next iteration.

clusterloader2/pkg/measurement/common/scheduling_throughput.go
marseel: @wojtek-t After discussing with @mborsz I will revert it and add summaryName parameter. It will make it easier to process results in perf-dash.

wojtek-t:
> @wojtek-t After discussing with @mborsz I will revert it and add summaryName parameter. It will make it easier to process results in perf-dash.

Why? Can't we explicitly mark is "scheduling_throughput_prometheus" or sth like that?
What would ability to parametrize it buy us?

mborsz:
> > Wojciech Tyczynski After discussing with Maciej Borsz I will revert it and add summaryName parameter. It will make it easier to process results in perf-dash.
> 
> Why? Can't we explicitly mark is "scheduling_throughput_prometheus" or sth like that?
> What would ability to parametrize it buy us?

We will be able to use this measurement as a drop-in replacement for the old scheduling throughput, without changes to perf-dash

mborsz:
Even without this change, we have 5 different SchedulingThroughputs: https://github.com/kubernetes/perf-tests/blob/master/perfdash/config.go#L205

If we make this scheduling throughput use a different summary name, we will need to add next 5 entries there (for SchedulingThroughputPrometheus). This is something I would like to avoid.

We can also hardcode SchedulingThroughput summary here, but this is something I wanted to avoid to improve readability of test configs (we need to explicitly override summaryName here so for someone reading the test config it will be clear that it's generated by this measurement).

wojtek-t:
> If we make this scheduling throughput use a different summary name, we will need to add next 5 entries there (for SchedulingThroughputPrometheus). This is something I would like to avoid.

> We can also hardcode SchedulingThroughput summary here, but this is something I wanted to avoid to improve readability of test configs (we need to explicitly override summaryName here so for someone reading the test config it will be clear that it's generated by this measurement).

I actually really prefer having a bit more boilerplate in perf-dash code instead of having additional things in test configs (passing them as params). Our test configs are already way too complicated.

jprzychodzen:
@wojtek-t @mborsz  Could you transfer you lgtm/approval to #1512 ? I would like to keep this open for some time, this PR can be useful to get additional data for kubernetes/test-infra#19411. Thanks!

PR #1506, opened 2020-09-30
Config changes to add SLOMeasurement- initial commit

vamossagar12:
**What type of PR is this?**

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -132,8 +132,8 @@ steps:
       action: start
       checkProbesReadyTimeout: {{$PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT}}
       replicasPerProbe: {{AddInt 2 (DivideInt .Nodes 100)}}
-  - Identifier: DnsLookupLatency
-    Method: DnsLookupLatency
+  - Identifier: SLOMeasurement
+    Method: SLOMeasurement
     Params:
       action: start
       checkProbesReadyTimeout: {{$PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT}}
@@ -863,8 +863,8 @@ steps:
     Method: InClusterNetworkLatency
     Params:
       action: gather
-  - Identifier: DnsLookupLatency
-    Method: DnsLookupLatency
+  - Identifier: SLOMeasurement
+    Method: SLOMeasurement
     Params:
       action: gather
   {{if $PROMETHEUS_SCRAPE_KUBE_PROXY}}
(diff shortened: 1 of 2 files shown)


vamossagar12:
@wojtek-t This is the PR with the config change. One question that I had was that initially, NetworkProgrammingLatency was invoked only if the flag **PROMETHEUS_SCRAPE_KUBE_PROXY**  was set to true:

https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L141-L146

With the blanket SLOMeasurement, it gets invoked every time. Do you think we need some changes to handle this?

clusterloader2/testing/load/config.yaml
```diff
@@ -132,18 +132,10 @@ steps:
...
-      action: start
-      checkProbesReadyTimeout: {{$PROBE_MEASUREMENTS_CHECK_PROBES_READY_TIMEOUT}}
-      replicasPerProbe: {{AddInt 2 (DivideInt .Nodes 100)}}
-  {{if $PROMETHEUS_SCRAPE_KUBE_PROXY}}
-  - Identifier: NetworkProgrammingLatency
-    Method: NetworkProgrammingLatency
-    Params:
-      action: start
-  {{end}}
+      action:start
```
wojtek-t, line 138: The other params (checkProbesReadyTimeout etc. are still needed - they are use by the measurement so we need to pass them.

wojtek-t:
> @wojtek-t This is the PR with the config change. One question that I had was that initially, NetworkProgrammingLatency was invoked only if the flag PROMETHEUS_SCRAPE_KUBE_PROXY was set to true:

> https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L141-L146

We should ideally try to solve that first - let me look into that a bit later today/tomorrow.

clusterloader2/testing/load/config.yaml
```diff
@@ -854,16 +846,10 @@ steps:
...
     Params:
       action: gather
-  - Identifier: DnsLookupLatency
-    Method: DnsLookupLatency
-    Params:
-      action: gather
-  {{if $PROMETHEUS_SCRAPE_KUBE_PROXY}}
-  - Identifier: NetworkProgrammingLatency
-    Method: NetworkProgrammingLatency
+  - Identifier: SLOMeasurement
```
wojtek-t, line 849: Actually - to no block on addressing the kube-proxy scraping, let's start differently:
- let's temporarily remove the NetworkProgrammingLatency from the SLO measurement (leave only DNSLookupLatency for now)
- and we will proceed with this PR (and address NetworkProgramminlatency separately)

vamossagar12:
@wojtek-t , PR to remove network programming latency:
https://github.com/kubernetes/perf-tests/pull/1518

wojtek-t:
@vamossagar12 - there seem to be a bug in:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/slo_measurement.go#L65

it should be:
```
if !errList.IsEmpty() {
  return nil, errList
}
return summaries, nil
```

wojtek-t:
Please fix (can be this PR).

PR #1470, opened 2020-09-15
Migrate dns benchmarks from python2 to python3 

vamossagar12:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #1507, opened 2020-09-30
Mark 'verify-all' as deprecated, but leave it to make pass existing PRs

jprzychodzen:
/kind cleanup

Ref. #1470 

`verify-all` presubmit is already skipped for master branch. Changing test target to allow passing jobs after rebase if there will be pending presubmit statuses.

/cc @wojtek-t 

Also, fix `verify-all-python` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,16 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# TODO(jprzychodzen): Remove after all PRs with 'pull-perf-tests-verify-all' are closed
+# Deprecated, use verify-all-dev for local development
 .PHONY: verify-all
-verify-all: test verify-boilerplate verify-flags
+verify-all:
+	echo "Deprecated, use 'verify-all-dev' for local development"
 
 # verify-lint-dev is convenience target for developers to run all checks
 # at once locally
 .PHONY: verify-all-dev
-verify-all-dev: verify-all verify-dashboard verify-lint
+verify-all-dev: test verify-all-python verify-dashboard verify-lint
 
 .PHONY: verify-all-python
-verify-all: verify-boilerplate verify-flags
+verify-all-python: verify-boilerplate verify-flags
 
 # TODO(oxddr): go-build.sh doesn't work at the moment decide whether we need this at all
 # .PHONY: build


jprzychodzen:
I see that `pull-perf-tests-verify-all Expected — Waiting for status to be reported` is marked as requested here. I do not have an access to perf-tests repo setting. @wojtek-t, could you check that and disable it?

jprzychodzen:
I know that I can pass this by using `/test`, but it doesn't solve the problem for other PRs.

wojtek-t:
@jprzychodzen - We should also enable test for all PRs, right?

jprzychodzen:
Yes, this is one option (1), but it would leave not needed test that will be no-op.

The only problem is with PRs that are have `pull-perf-tests-verify-all` failed now and will not have automatic retest - after quick check it seems that it's a pretty small number of jobs.

The other option (2) is merge this PR, rebase affected PRs and run `/test pull-perf-tests-verify-all` in each affected PRs after this is merged (or leave instruction to do this) for future reviewers.

So I prefer option (2) - find affected PRs (not passing `pull-perf-tests-verify-all`) and leave instruction to rebase and run test.

wojtek-t:
That's not what I'm saying.

I'm saying that with this change we will stop running unit tests as part of presubmit completely. We don't want to do that.

jprzychodzen:
[pull-perf-tests-verify-test](https://github.com/jprzychodzen/test-infra/blob/c1ddfcb476a714251c84f761031ed57651c029d1/config/jobs/kubernetes/sig-scalability/sig-scalability-presubmit-jobs.yaml#L632) runs unit tests - we we running unit tests in `pull-perf-tests-verify-test` and `pull-perf-tests-verify-all`. We do not need `pull-perf-tests-verify-all` anymore.

wojtek-t:
I'm not saying we dont have such job. I'm saying we don't use it - see this PR :)

jprzychodzen:
Yes, I see that it's not running. Configuration for `pull-perf-tests-verify-test` runs on specific branches instead of skipping specific branches.

jprzychodzen:
Will be /unholded after merge of kubernetes/test-infra#19404

PR #1508, opened 2020-09-30
Fix cpu requests in load test when density test is enabled.

marseel:
**What type of PR is this?**
/kind bug

Fixes cpu requests in load test when density test is enabled.
Example failed test: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1310988110418415616

Fixes #

```release-note

```

```docs

```


wojtek-t:
Something doesn't work with retesting - @jprzychodzen 

I'm going to merge manually to fix the 5k test regression.

jprzychodzen:
Thanks, I was already checking logs of Prow to see if we need to act on this. It's possible that with all recent changes some strange state happened. There was an unexpected abort for this job after ~30 minutes of runtime. I will validate that presubmits work as expected.

PR #1511, opened 2020-09-30
Add mborsz to perfdash/OWNERS

mborsz:
Also reorders approvers to match reviewers (and to match lexicographic order)
/assign @wojtek-t

### DIFF
--- a/perfdash/OWNERS
+++ b/perfdash/OWNERS
@@ -1,11 +1,10 @@
 approvers:
 - krzysied
+- mborsz
 - shyamjvs
 - wojtek-t
-- oxddr
 reviewers:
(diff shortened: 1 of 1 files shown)


perfdash/OWNERS
```diff
@@ -1,8 +1,9 @@
 approvers:
 - krzysied
+- mborsz
+- oxddr
```
wojtek-t, line 4: Let's remove Janek from that list.

mborsz: DONE

mborsz:
PTAL

wojtek-t:
/lgtm
/approve

Thanks!

PR #1525, opened 2020-10-07
New method of counting slow queries [1.18]

jkaniuk:
Fix: #1406

Cherry-pick #1417, #1416

/cc @wojtek-t


wojtek-t:
/lgtm
/approve

Thanks!

PR #1526, opened 2020-10-07
Run a DNS workload as part of load tests.

prameshj:
This will run parallel DNS lookups via 'git' commands that do not retry.
So, a DNS error will result in a 'git clone' error.

The image has been built from some prototype code in https://github.com/prameshj/dns-stress-test, specifically the simple script - https://github.com/prameshj/dns-stress-test/blob/master/git-fetch.sh

This can be enhanced to send queries at a desired QPS and also expose latency metrics.

**What type of PR is this?**

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -20,6 +20,7 @@
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
+{{$ENABLE_DNSTESTS := DefaultParam .CL2_ENABLE_DNSTESTS false}}
 {{$ENABLE_DENSITY_TEST := DefaultParam .CL2_ENABLE_DENSITY_TEST false}}
 {{$ENABLE_SYSTEM_POD_METRICS:= DefaultParam .ENABLE_SYSTEM_POD_METRICS true}}
 {{$ENABLE_CLUSTER_OOMS_TRACKER := DefaultParam .CL2_ENABLE_CLUSTER_OOMS_TRACKER false}}
@@ -249,6 +250,8 @@ steps:
     - basename: big-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        # DNS Test clients are enabled only in the medium-size deployment.
+        EnableDNSTests: false
         ReplicasMin: {{$BIG_GROUP_SIZE}}
         ReplicasMax: {{$BIG_GROUP_SIZE}}
         SvcName: big-service
@@ -272,6 +275,7 @@ steps:
     - basename: medium-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        EnableDNSTests: {{$ENABLE_DNSTESTS}}
         ReplicasMin: {{$MEDIUM_GROUP_SIZE}}
         ReplicasMax: {{$MEDIUM_GROUP_SIZE}}
         SvcName: medium-service
@@ -295,6 +299,8 @@ steps:
     - basename: small-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        # DNS Test clients are enabled only in the medium-size deployment.
+        EnableDNSTests: false
         ReplicasMin: {{$SMALL_GROUP_SIZE}}
         ReplicasMax: {{$SMALL_GROUP_SIZE}}
         SvcName: small-service
@@ -541,6 +547,8 @@ steps:
     - basename: big-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        # DNS Test clients are enabled only in the medium-size deployment.
+        EnableDNSTests: false
         ReplicasMin: {{MultiplyInt $BIG_GROUP_SIZE 0.5}}
         ReplicasMax: {{MultiplyInt $BIG_GROUP_SIZE 1.5}}
         SvcName: big-service
@@ -556,6 +564,7 @@ steps:
     - basename: medium-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        EnableDNSTests: {{$ENABLE_DNSTESTS}}
         ReplicasMin: {{MultiplyInt $MEDIUM_GROUP_SIZE 0.5}}
         ReplicasMax: {{MultiplyInt $MEDIUM_GROUP_SIZE 1.5}}
         SvcName: medium-service
@@ -571,6 +580,8 @@ steps:
     - basename: small-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        # DNS Test clients are enabled only in the medium-size deployment.
+        EnableDNSTests: false
         ReplicasMin: {{MultiplyInt $SMALL_GROUP_SIZE 0.5}}
         ReplicasMax: {{MultiplyInt $SMALL_GROUP_SIZE 1.5}}
         SvcName: small-service
--- a/clusterloader2/testing/load/deployment.yaml
+++ b/clusterloader2/testing/load/deployment.yaml
@@ -23,8 +23,24 @@ spec:
     spec:
       hostNetwork: {{$HostNetworkMode}}
       containers:
+{{if .EnableDNSTests}}
+      - image: k8s.gcr.io/kube-dns-perf-client-amd64:1.1
+      # Fetches the dns server from /etc/resolv.conf and
+      # runs as 5 clients and 2 threads sending 10 queries per second.
+      # dnsperf has a client timeout of 5s. It sends queries for 60s,
+      # then sleeps for 10s, to mimic bursts of DNS queries.
+        command:
+        - sh
+        - -c
+        - server=$(cat /etc/resolv.conf | grep nameserver | cut -d ' ' -f 2); echo
+          "Using nameserver ${server}"; while true; do ./dnsperf -s $server -l 60
+          -d /var/configmap/all-queries -c 5 -T 2 -Q 10; sleep 10; done
+        name: {{.Name}}-dnsperf
+        app: dns-perf-client
+{{else}}
       - image: k8s.gcr.io/pause:3.1
         name: {{.Name}}
+{{end}}
         resources:
           requests:
             cpu: {{$CpuRequest}}


clusterloader2/testing/load/dns-deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+  template:
+    metadata:
+      labels:
+        group: load
+        name: {{.Name}}
+        svc: {{.SvcName}}-{{.Index}}
+    spec:
+      hostNetwork: {{$HostNetworkMode}}
+      containers:
+      - image: gcr.io/pavithrar-k8s-dev/dnstest:latest
```
wojtek-t, line 26: If we decide to go with this one - I would like the source code of this to be merged in perf-tests repo too.

clusterloader2/testing/load/experimental-config.yaml
```diff
@@ -21,6 +21,7 @@
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
+{{$ENABLE_DNSTESTS := DefaultParam .CL2_ENABLE_DNSTESTS false}}
```
wojtek-t, line 24: Let's ignore this file - we will remove that fairly soon (probably within a week or two).

clusterloader2/testing/load/config.yaml
```diff
@@ -270,7 +271,11 @@ steps:
       objectTemplatePath: networkpolicy.yaml
     {{end}}
     - basename: medium-deployment
+    {{if $ENABLE_DNSTESTS}}
+      objectTemplatePath: dns-deployment.yaml
```
wojtek-t, line 275: I'm fine with that conceptually, but I would like to clarify one thing - there are basically O(7.5*#nodes) pods in the cluster coming from this deployment during the test (so ~35000 in 5k-node cluster).

If we can handle that and this will work - I'm fine with that. I'm just ensuring that we will be able to handle that.

prameshj: Each pod will do 10 queries per second. We might run into dnsmasq max connections issue perhaps.  Is there a different config to change, so we have fewer pods to start with - like maybe 5k pods in a 5k node cluster?

clusterloader2/testing/load/experimental-config.yaml
prameshj: done

clusterloader2/testing/load/dns-deployment.yaml
prameshj: Changed the deployment file to use dnsperf which already has the source code in the perf-tests repo.

clusterloader2/testing/load/config.yaml
```diff
@@ -677,8 +692,16 @@ steps:
     replicasPerNamespace: 0
     tuningSet: RandomizedDeletionTimeLimited
     objectBundle:
+    - basename: medium-deployment
```
wojtek-t, line 695: seems like type - should be removed

clusterloader2/testing/load/dnsperf-deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+  template:
+    metadata:
+      labels:
+        group: load
+        name: {{.Name}}
+        svc: {{.SvcName}}-{{.Index}}
+    spec:
+      hostNetwork: {{$HostNetworkMode}}
+      containers:
+      # Fetches the dns server from /etc/resolv.conf and runs as 5 clients and 2 threads sending 10 queries per second.
```
wojtek-t, line 26: nit: can you please shorten those comment lines?

mborsz:
```
W1021 04:33:31.207] I1021 04:33:31.201698   14849 simple_test_executor.go:165] Step "[step: 06] Waiting for pods to be running" ended
W1021 04:33:31.208] W1021 04:33:31.201706   14849 simple_test_executor.go:167] Got errors during step execution: [measurement call WaitForControlledPodsRunning -  error: [measurement call WaitForControlledPodsRunning - WaitForRunningDeployments error: 23 objects timed out: Deployments: test-0rbffm-1/medium-deployment-2, test-0rbffm-1/medium-deployment-0, test-0rbffm-1/medium-deployment-1, test-0rbffm-1/medium-deployment-11, test-0rbffm-1/medium-deployment-19, test-0rbffm-1/medium-deployment-17, test-0rbffm-1/medium-deployment-21, test-0rbffm-1/medium-deployment-4, test-0rbffm-1/medium-deployment-13, test-0rbffm-1/medium-deployment-22, test-0rbffm-1/medium-deployment-10, test-0rbffm-1/medium-deployment-15, test-0rbffm-1/medium-deployment-6, test-0rbffm-1/medium-deployment-18, test-0rbffm-1/medium-deployment-20, test-0rbffm-1/medium-deployment-3, test-0rbffm-1/medium-deployment-7, test-0rbffm-1/medium-deployment-9, test-0rbffm-1/medium-deployment-5, test-0rbffm-1/medium-deployment-8, test-0rbffm-1/medium-deployment-12, test-0rbffm-1/medium-deployment-14, test-0rbffm-1/medium-deployment-16]]
```

sample failing deployment:
```
W1021 04:33:31.203] I1021 04:33:31.201325   14849 wait_for_pods.go:75] WaitForControlledPodsRunning: namespace(test-0rbffm-1), labelSelector(name=medium-deployment-12): pods status: [{medium-deployment-12-ff6c87bb7-2wf8s Pending e2e-1526-62db2-minion-group-fbkp} {medium-deployment-12-ff6c87bb7-vj8k2 Pending e2e-1526-62db2-minion-gr
oup-k3h1} {medium-deployment-12-ff6c87bb7-8jrrh Pending e2e-1526-62db2-minion-group-ckdn} {medium-deployment-12-ff6c87bb7-fv5tn Pending e2e-1526-62db2-minion-group-gnpk} {medium-deployment-12-ff6c87bb7-gmj6z Pending e2e-1526-62db2-minion-group-hqqn} {medium-deployment-12-ff6c87bb7-r7q55 Pending e2e-1526-62db2-minion-group-fdfg} {me
dium-deployment-12-ff6c87bb7-rlhfn Pending e2e-1526-62db2-minion-group-4n8s} {medium-deployment-12-ff6c87bb7-8wq9d Pending e2e-1526-62db2-minion-group-8n31} {medium-deployment-12-ff6c87bb7-bcwdk Pending e2e-1526-62db2-minion-group-mrbb} {medium-deployment-12-ff6c87bb7-4wb52 Pending e2e-1526-62db2-minion-group-v3jr} {medium-deployme
nt-12-ff6c87bb7-82xh6 Pending e2e-1526-62db2-minion-group-t2mh} {medium-deployment-12-ff6c87bb7-qwjgt Pending e2e-1526-62db2-minion-group-3q18} {medium-deployment-12-ff6c87bb7-4jtxp Pending e2e-1526-62db2-minion-group-91gq} {medium-deployment-12-ff6c87bb7-sv4jj Pending e2e-1526-62db2-minion-group-9q2j} {medium-deployment-12-ff6c87b
b7-6gxxn Pending e2e-1526-62db2-minion-group-4hf5} {medium-deployment-12-ff6c87bb7-cx84q Pending e2e-1526-62db2-minion-group-4250} {medium-deployment-12-ff6c87bb7-xgfb9 Pending e2e-1526-62db2-minion-group-njdq} {medium-deployment-12-ff6c87bb7-psc6g Pending e2e-1526-62db2-minion-group-3x30} {medium-deployment-12-ff6c87bb7-mb22h Pend
ing e2e-1526-62db2-minion-group-76ch} {medium-deployment-12-ff6c87bb7-zmzp5 Pending e2e-1526-62db2-minion-group-v00p} {medium-deployment-12-ff6c87bb7-pvn5t Pending e2e-1526-62db2-minion-group-2mph} {medium-deployment-12-ff6c87bb7-6vtwk Pending e2e-1526-62db2-minion-group-rpgq} {medium-deployment-12-ff6c87bb7-j24j7 Pending e2e-1526-
62db2-minion-group-vxzp} {medium-deployment-12-ff6c87bb7-cgzr9 Pending e2e-1526-62db2-minion-group-cfh4} {medium-deployment-12-ff6c87bb7-bggst Pending e2e-1526-62db2-minion-group-k813} {medium-deployment-12-ff6c87bb7-8jvqw Pending e2e-1526-62db2-minion-group-khw6} {medium-deployment-12-ff6c87bb7-w8jr7 Pending e2e-1526-62db2-minion-
group-rc8b} {medium-deployment-12-ff6c87bb7-z7l7p Pending e2e-1526-62db2-minion-group-03f5} {medium-deployment-12-ff6c87bb7-shsn8 Pending e2e-1526-62db2-minion-group-zv70} {medium-deployment-12-ff6c87bb7-zclwl Pending e2e-1526-62db2-minion-group-gz4q}], expected 30
```
sample pod:
```
medium-deployment-12-ff6c87bb7-2wf8s Pending e2e-1526-62db2-minion-group-fbkp
```


e2e-1526-62db2-minion-group-fbkp's kubelet logs:
```
E1021 04:18:31.964533    1274 configmap.go:200] Couldn't get configMap test-0rbffm-1/dns-perf-queries: configmap "dns-perf-queries" not found
E1021 04:18:31.964692    1274 nestedpendingoperations.go:301] Operation for "{volumeName:kubernetes.io/configmap/dbf4b144-049b-4ad1-82ef-b2bc23050a91-config-volume podName:dbf4b144-049b-4ad1-82ef-b2bc23050a91 nodeName:}" failed. No retries permitted until 2020-10-21 04:18:32.464669615 +0000 UTC m=+484.956168899 (durationBeforeRetry
 500ms). Error: "MountVolume.SetUp failed for volume \"config-volume\" (UniqueName: \"kubernetes.io/configmap/dbf4b144-049b-4ad1-82ef-b2bc23050a91-config-volume\") pod \"medium-deployment-12-ff6c87bb7-2wf8s\" (UID: \"dbf4b144-049b-4ad1-82ef-b2bc23050a91\") : configmap \"dns-perf-queries\" not found"
```

so the pods seem to be pending as they try to mount 'dns-perf-queries' configmap that doesn't exist.
The dns-perf-queries configmap doesn't exist as clusterloader ignores the 'metadata.name' field in yaml file and '{$basename}-{$index}' is used instead, where "{$basename}" is a basename param (medium-deployment in your case) and index iterates between 0 and replicasPerNamespace. So clusterloader creates your configmaps with names like 'medium-deployment-7'.

I suggest migrating to the approach, where you create a single configmap per namespace and mount it later in all deployments: You can add a new step BEFORE the step you currently create your deployments.
```
- namespaceRange:
      min: 1
      max: {{$namespaces}}
      replicasPerNamespace: 1
      tuningSet: Sequence
      objectBundle:
      - basename: dns-perf-queries
         objectTemplatePath: dnsperf-configmap.yaml
(...)

```
Then modify dnsperf-deployment.yaml to use "dns-perf-queries-0" configmap (so basename="dns-perf-queries", index=0).

This way, you will create a single configmap per namespace and then use it everywhere.

Please remember to add a step that will remove all configmaps (simply the same step as above, but with repliacsPerNamespace set to 0) after the step that deletes your deployment.

clusterloader2/testing/load/dnsperf-deployment.yaml
```diff
@@ -0,0 +1,76 @@
...
+      volumes:
+        - name: configmap
+          configMap:
+            name: {{.BaseName}}-{{.Index}}
+        - name: secret
+          secret:
+            secretName: {{.BaseName}}-{{.Index}}
+        - name: config-volume
+          configMap:
+            name: dns-perf-queries
```
mborsz, line 72: This line must be changed. It tries to mount nonexisting configmap.

mborsz:
There is one thing I do not understand in the clusterloader's logs:

```
:0
[namespace test-0rbffm-1 object medium-deployment-2 updating error: ConfigMap "medium-deployment-2" is invalid: data: Forbidden: field is immutable when `immutable` is set
```
this error suggests that you are trying to change 'data' field in configmap, but as you can see it's clearly constant (the file yaml file doesn't include any variable there). This shouldn't block your tests, as you should remove your dns configmaps from "Scaling and updating objects" phase, but still quite interesting.

mborsz:
> There is one thing I do not understand in the clusterloader's logs:
> 
> ```
> :0
> [namespace test-0rbffm-1 object medium-deployment-2 updating error: ConfigMap "medium-deployment-2" is invalid: data: Forbidden: field is immutable when `immutable` is set
> ```
> 
> this error suggests that you are trying to change 'data' field in configmap, but as you can see it's clearly constant (the file yaml file doesn't include any variable there). This shouldn't block your tests, as you should remove your dns configmaps from "Scaling and updating objects" phase, but still quite interesting.

I understood this. You are creating two different configmaps with the same name: there is medium-deployment configmap created from configmap.yaml and medium-deployment configmap created from 'dnsperf-configmap.yaml'. For that reason, apiserver rejects updates to the 'medium-deployment-X' configmaps as they change content from one yaml file to another

mborsz:
@prameshj I'm not aware of the context here, but few questions:
* Why are you making this a part of load test? Why not a separate test approach?
* Where are you planning to enable this? Is it going to be a part of our continuous load test? Or are you planning to create a new fork of load test?

Depending on the answer, some other approach may be reasonable:

If this is going to be a part of load test and will be enabled in our release qualification pipeline, then you may want to use yet another approach:
* do not add a step to create a new configmap in each namespace (as suggested by me above)
* instead, change content of the existing medium-deployment configmap to use your yaml when ENABLE_DNSTESTS is set
* change your deployment yaml to avoid mounting two configmaps (medium-deployment-X and dns-perf-queries) and load queries from medium-deployment-X instead.

This way you will generate the same load (number for configmaps and mounts) on kube-apiserver as our load test and it will be quite easy from coding perspective (no need for additional steps).

prameshj:
> @prameshj I'm not aware of the context here, but few questions:
> 
> * Why are you making this a part of load test? Why not a separate test approach?
Making this part of load tests based on conversation with @wojtek-t . We want a quick test to verify DNS at scale, before pulling it out into its own test. Using existing 5k runs will help with this and give us some alerting/regressions with a little bit of extra work.

> * Where are you planning to enable this? Is it going to be a part of our continuous load test? Or are you planning to create a new fork of load test?
> 
Yes, maybe a different workflow or part of default once it is stable.

> Depending on the answer, some other approach may be reasonable:
> 
> If this is going to be a part of load test and will be enabled in our release qualification pipeline, then you may want to use yet another approach:
> 
> * do not add a step to create a new configmap in each namespace (as suggested by me above)
> * instead, change content of the existing medium-deployment configmap to use your yaml when ENABLE_DNSTESTS is set

Interesting.. I wasn't sure why the existing configmap and secrets were needed. I can replace them if they are not required for the pod to run successfully.

> * change your deployment yaml to avoid mounting two configmaps (medium-deployment-X and dns-perf-queries) and load queries from medium-deployment-X instead.

Sounds good.
> 
> This way you will generate the same load (number for configmaps and mounts) on kube-apiserver as our load test and it will be quite easy from coding perspective (no need for additional steps).

clusterloader2/testing/load/configmap.yaml
```diff
@@ -6,7 +6,36 @@ metadata:
 immutable: true
 {{end}}
 data:
+{{if $ENABLE_DNSTESTS}}
```
mborsz, line 9: You don't have access to any variable that is not listed explicitly in templateFillMap in config.yaml

mborsz:
> > @prameshj I'm not aware of the context here, but few questions:
> > 
> > * Why are you making this a part of load test? Why not a separate test approach?
> >   Making this part of load tests based on conversation with Wojciech Tyczynski . We want a quick test to verify DNS at scale, before pulling it out into its own test. Using existing 5k runs will help with this and give us some alerting/regressions with a little bit of extra work.
> 
> > * Where are you planning to enable this? Is it going to be a part of our continuous load test? Or are you planning to create a new fork of load test?
> 
> Yes, maybe a different workflow or part of default once it is stable.
> 
> > Depending on the answer, some other approach may be reasonable:
> > If this is going to be a part of load test and will be enabled in our release qualification pipeline, then you may want to use yet another approach:
> > 
> > * do not add a step to create a new configmap in each namespace (as suggested by me above)
> > * instead, change content of the existing medium-deployment configmap to use your yaml when ENABLE_DNSTESTS is set
> 
> Interesting.. I wasn't sure why the existing configmap and secrets were needed. I can replace them if they are not required for the pod to run successfully.
> 

The only purpose of mounting configmaps and secrets is to generate some load on kube-apiserver (each unique secrets/configmaps mounted on a given node adds some load on kube-apiserver). Technically, you can change them.

> > * change your deployment yaml to avoid mounting two configmaps (medium-deployment-X and dns-perf-queries) and load queries from medium-deployment-X instead.
> 
> Sounds good.
> 
> > This way you will generate the same load (number for configmaps and mounts) on kube-apiserver as our load test and it will be quite easy from coding perspective (no need for additional steps).

I briefly checked the current state. I think the only technical issue I'm seeing is that you are using $ENABLE_DNSTESTS in configmap.yaml without rewriting value in templateFillMap. You should add something like:
```
templateFillMap:
   ENABLE_DNSTESTS: {{$ENABLE_DNSTESTS}}
```
to make it available in configmap.yaml as ".ENABLE_DNSTESTS" (not $ENABLE_DNSTESTS).

After this change, it should work.

I'm still not sure if this is the best approach to take to make this readable. I will think about this and review this PR properly tomorrow morning. I understand that you want to submit this quickly so I will do my best to avoid unnecessary review iterations.

clusterloader2/testing/load/config.yaml
```diff
@@ -269,9 +270,15 @@ steps:
...
     {{end}}
+    {{if $ENABLE_DNSTESTS}}
+    - basename: medium-deployment
+      objectTemplatePath: dnsperf-deployment.yaml
+    {{else}}
     - basename: medium-deployment
       objectTemplatePath: deployment.yaml
+    {{end}}
       templateFillMap:
+        ENABLE_DNSTESTS: {{$ENABLE_DNSTESTS}}
```
mborsz, line 281: This templateFillMap section is for 'deployment.yaml' or 'dnsperf-deployment.yaml' file.

Please move it from here to newly created templateFillMap section in:
" - basename: medium-deployment
     objectTemplatePath: configmap.yaml
     templateFillMap:
        EnableDNSTests: {{$ENABLE_DNSTESTS}}"

Also please try to be consistent with the naming style in templateFillMaps (so CamelCase).

clusterloader2/testing/load/config.yaml
```diff
@@ -20,6 +20,7 @@
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
+{{$ENABLE_DNSTESTS := DefaultParam .CL2_ENABLE_DNSTESTS true}}
```
mborsz, line 23: Before submitting please change this to false by default. We want to enable this in our tests gradually.

clusterloader2/testing/load/dnsperf-deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+  replicas: {{RandIntRange .ReplicasMin .ReplicasMax}}
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        group: load
+        name: {{.Name}}
+        svc: {{.SvcName}}-{{.Index}}
```
mborsz, line 22: If you are planing to expose any metrics in dns-perf-client, please consider adding some new label that can be used in future for metrics gathering like: "app: dns-perf-client".

clusterloader2/testing/load/config.yaml
prameshj: Yes, i plan to change it before submitting. This was only to get a presubmit test to verify that the config is correct.

clusterloader2/testing/load/config.yaml
prameshj: Thanks for catching this, will fix.

clusterloader2/testing/load/dnsperf-deployment.yaml
prameshj: Good idea.

clusterloader2/testing/load/dnsperf-deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+      # dnsperf has a client timeout of 5s. It sends queries for 60s,
+      # then sleeps for 10s, to mimic bursts of DNS queries.
+      - command:
+        - sh
+        - -c
+        - server=$(cat /etc/resolv.conf | grep nameserver | cut -d ' ' -f 2); echo
+          "Using nameserver ${server}"; while true; do ./dnsperf -s $server -l 60
+          -d /etc/queries/all-queries -c 5 -T 2 -Q 10; sleep 10; done
+        image: k8s.gcr.io/kube-dns-perf-client-amd64:1.1
+        imagePullPolicy: Always
```
mborsz, line 37: Is it really needed? I suggest changing imagePullPolicy to something that will reduce load on nodes.

clusterloader2/testing/load/dnsperf-deployment.yaml
```diff
@@ -0,0 +1,71 @@
+{{$HostNetworkMode := DefaultParam .CL2_USE_HOST_NETWORK_PODS false}}
+{{$CpuRequest := DefaultParam .CpuRequest "10m"}}
```
mborsz, line 2: This looks like copy paste of deployment.yaml file from https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/deployment.yaml

I we expect this file to be drop-in replacement for deployment.yaml it must be compatible with it (so have the same labels, resource requests, taints etc). I'm afraid that keeping two files with very similar content isn't the best approach and sooner or later we will forget updating one of them.

How about merging those files together? I think the content looks quite similar, only container and configmaps sections are different.

clusterloader2/testing/load/config.yaml
prameshj: Also, would I need to fill this in everytime for configmap, or only when I set it to true?

clusterloader2/testing/load/dnsperf-deployment.yaml
prameshj: Good point, let me try to merge them.

clusterloader2/testing/load/config.yaml
mborsz: Not sure if I understand the question, but you need to set EnableDNSTests field in templateFillMap when configmap.yaml will read this field (so basically all the time).

clusterloader2/testing/load/config.yaml
mborsz: What is expected number of pods running dns-perf? Is it 5k? Less or more?

We can modify only some fraction of medium-deployments (so that we can control by variable a number of pods from 0 to 37k with steps of size 1500 pods).

Or if the desired number of pods is really small (so like resources on nodes are not a problem), we can simply run separate deployment (not a part of small,medium or big deployments).

clusterloader2/testing/load/config.yaml
```diff
@@ -249,6 +252,7 @@ steps:
     - basename: big-deployment
       objectTemplatePath: deployment.yaml
       templateFillMap:
+        EnableDNSTests: {{$ENABLE_DNSTESTS}}
```
mborsz, line 255: If you set EnableDNSTests to {{$ENABLE_DNSTESTS}} in big-deployment, medium-deployment and small-deployment you will get ~150k pods running queries (all pods will be running queries).

As we agreed on chat, let's start with 37k first (so only medium-deployment) and we will see if this works.

In order to do that, you need to set EnableDNSTests to false in both big-deployment and small-deployment in deployment.yaml and configmap.yaml.
Please also add a comment like "We are setting this to false in big and small to enable this only in medium", as it's not obvious for random reader why we are copying this value only for medium".

mborsz:
One comment (you enabled this in all 150k pods, which IIUC is not what we want), other than that LGTM.

clusterloader2/testing/load/config.yaml
```diff
@@ -20,6 +20,7 @@
 {{$ENABLE_PROMETHEUS_API_RESPONSIVENESS := DefaultParam .ENABLE_PROMETHEUS_API_RESPONSIVENESS false}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
+{{$ENABLE_DNSTESTS := DefaultParam .CL2_ENABLE_DNSTESTS false}}
```
wojtek-t, line 23: Can you change that to true for now to test that behavior (and once done job with the test in the PR itself).

clusterloader2/testing/load/configmap.yaml
```diff
@@ -6,7 +6,36 @@ metadata:
...
+   1-2-3-4.default.pod.cluster.local A
+   metrics-server A
+   metrics-server AAAA
+   metrics-server.cluster.local A
+   metrics-server.cluster.local AAAA
+   metrics-server.svc.cluster.local A
+   metrics-server.svc.cluster.local AAAA
+   metrics-server.kube-system.svc.cluster.local A
+   metrics-server.kube-system.svc.cluster.local AAAA
+{{else}}
```
wojtek-t, line 36: I'm wondering if we should be "else-ing" here.
The current data was simply something artifical just to have something. If we have usecase for something else, which is clearly here, I'm with with simply changing this to your data.

clusterloader2/testing/load/deployment.yaml
```diff
@@ -51,6 +67,11 @@ spec:
         - name: configmap
           configMap:
             name: {{.BaseName}}-{{.Index}}
+{{if .ENABLE_DNSTESTS}}
```
wojtek-t, line 70: If we apply my above comment in configmap.yaml, then we can do this also unconditionally.

[In other words - we should probably split this update and configmap update to a sepraate PR (and done the unconditionally) and leave the rest in this PR.]
@mborsz - thoughts?

clusterloader2/testing/load/config.yaml
mborsz: It used to be true in this PR, but I asked to revert before submitting this PR.

Given that we are going to change this PR anyway I suggest to:
1. Applying comments to this PR and changing this to true
2. Waiting for prow to trigger presubmit
3. Changing this to false

This way we will test this PR with the changes needed to apply comments and ultimately submit this without enabling everywhere.

clusterloader2/testing/load/configmap.yaml
mborsz: In fact I like the clear distinction between the case when we expect this configmap to contain some useful data and case when this can be a random stuff, as we have it with changing this conditionally.

I understand that removing this else will make this configmap easier to understand from coding perspective, but this hides dependency between deployment usage and configmap content. The current state is more complicated from coding perspective, but the dependency between configmap content and deployment is clear.

E.g if in future someone wants to implement a change where some percentile of deployments will need yet another configmap content, it will be much simpler to reason about when we can change the random data to something useful for that feature (it's clear that random data cannot be used anywhere).

So I recommend keeping the current state.

clusterloader2/testing/load/deployment.yaml
mborsz: See answer above.

clusterloader2/testing/load/config.yaml
wojtek-t: That's exactly what I was suggesting.

clusterloader2/testing/load/configmap.yaml
wojtek-t: I see what you're saying, but my counterargument for that is that, we shouldn't be changing the contents in that case, but instead:
- add new items to the map
- use those other items (mount just these other items at the pod template space) in this case

In other words, the configmap should be contatenation of everything that is potentially needed and we should solve what to mount at the pod template spec level.

mborsz: Wojtek volunteered to do that in https://github.com/kubernetes/perf-tests/pull/1546. Once it's merged, please rebase this PR and remove all "EnableDnsTests" in all configmap sections.

clusterloader2/testing/load/deployment.yaml
mborsz: As discussed @wojtek-t . I think it makes sense to simply always mount all keys and using one of them, so let's remove this "items" section from here.

The config should be available under /var/configmap/all-queries anyway

clusterloader2/testing/load/configmap.yaml
wojtek-t: That PR has just merged.

clusterloader2/testing/load/config.yaml
prameshj: Thanks! Done.

clusterloader2/testing/load/config.yaml
prameshj: I had the bool set to true and one round of presubmit run over the weekend. I have set it to true again after addressing all the other comments.

clusterloader2/testing/load/deployment.yaml
prameshj: sounds good.. thanks for submitting the other PR!

clusterloader2/testing/load/configmap.yaml
prameshj: Thanks, this simplifies the config quite a bit.

clusterloader2/testing/load/deployment.yaml
wojtek-t: Let's remove it - you no longer need it.

We discussed that and we belive it makes sense to always mount everything - it will be even simpler.

wojtek-t (commented):
One more comment - please update, ensure that presubmits are still passing and then remove change the default to false and we would be able to proceed (sorry it took so long).

prameshj:
> One more comment - please update, ensure that presubmits are still passing and then remove change the default to false and we would be able to proceed (sorry it took so long).

Presubmits passed, reverted the flag default to false.

clusterloader2/testing/load/deployment.yaml
```diff
@@ -51,6 +67,9 @@ spec:
         - name: configmap
           configMap:
             name: {{.BaseName}}-{{.Index}}
+            items:
```
wojtek-t, line 70: Let's also remove this "items" completely.
If you don't pass items, we will simply mount everything, so given we now have only all-queries, it's simply noop.

wojtek-t (commented):
One more comment.

clusterloader2/testing/load/deployment.yaml
prameshj: good to know, done.

PR #1527, opened 2020-10-08
Turn on alerting in load test for pod latency and scheduling throughput

marseel:
…ing throughput below 100

**What type of PR is this?**

/kind feature

We are in a process of merging density test into load test. Based on data gathered in perfdash for 100 and 5k nodes runs, we are comfortable with turning on alerting for SchedulingThroughput and PodStartupLatency.
Metrics:
100 Nodes:
Pod startup latency:
http://perf-dash.k8s.io/#/?jobname=gce-100Nodes-master&metriccategoryname=E2E&metricname=LoadPodStartup&Metric=pod_startup
Scheduling throughput:
http://perf-dash.k8s.io/#/?jobname=gce-100Nodes-master&metriccategoryname=Scheduler&metricname=SchedulingThroughput&TestName=load

5k Nodes:
Pod startup latency:
http://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=E2E&metricname=LoadPodStartup&Metric=pod_startup
Scheduling throughput:
http://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=Scheduler&metricname=SchedulingThroughput&TestName=load

The difference between scheduling throughput in density test and load test comes from interval (1s vs 5s) and burstQPS.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -62,14 +62,14 @@
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) (MultiplyInt 2 .Nodes)}}
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
-{{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 0}}
+{{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 100}}
 # END scheduler-throughput section
 
 # TODO(https://github.com/kubernetes/perf-tests/issues/1024): Investigate and get rid of this section.
 # BEGIN pod-startup-latency section
 {{$totalLatencyPods := MaxInt $MIN_PODS_IN_SMALL_CLUSTERS .Nodes}}
 {{$latencyReplicas := DivideInt $totalLatencyPods $namespaces}}
-{{$podStartupLatencyThreshold := DefaultParam .CL2_POD_STARTUP_LATENCY_THRESHOLD "1h"}}
+{{$podStartupLatencyThreshold := DefaultParam .CL2_POD_STARTUP_LATENCY_THRESHOLD "5s"}}
 # END pod-startup-latency section
 
 # Probe measurements shared parameter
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -441,6 +441,7 @@ steps:
     Method: SchedulingThroughput
     Params:
       action: gather
+      enableViolations: true
```
mborsz, line 444: FTR: this is no-op when threshold=0 (it is safe to rollout this without guarding everywhere and control rollout with $schedulerThroughputThreshold).

clusterloader2/testing/load/config.yaml
```diff
@@ -526,6 +527,7 @@ steps:
     Method: PodStartupLatency
     Params:
       action: gather
+      enableViolations: true
```
mborsz, line 530: I don't think PodStartupLatency has enableViolations field: https://github.com/kubernetes/perf-tests/blob/c8098dfa0d3c09becd3f669b9ba2b49e413c5bda/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go#L73-L95

Please remove

clusterloader2/testing/load/config.yaml
```diff
@@ -62,14 +62,14 @@
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) (MultiplyInt 2 .Nodes)}}
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
-{{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 0}}
+{{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 100}}
```
mborsz, line 65: We shouldn't override this in test config. This will rollout to all places including kubemarks and presubmits. 

We should add override in test-infra repo for two jobs only: 100nodes and 5k nods instead, like in https://github.com/kubernetes/test-infra/pull/19377

clusterloader2/testing/load/config.yaml
marseel: ah true, I though it uses prometheus metric...

clusterloader2/testing/load/config.yaml
marseel: Why it will rollout to kubemarks and presubmits? I think it makes sense to have some default thresholds when CL2_ENABLE_DENSITY_TEST is true.

mborsz: Yes, you are right. I forgot that we already guard this by CL2_ENABLE_DENSITY_TEST. Makes sense.

PR #1529, opened 2020-10-12
Add overrides and SchedulingThroughputPrometheus to SLOMeasurement

marseel:
**What type of PR is this?**

/kind feature

Adds support for overrides in SLOMeasurement and includes SchedulingThroughputPrometheus in SLOMeasurement.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/slo_measurement.go
+++ b/clusterloader2/pkg/measurement/common/slos/slo_measurement.go
@@ -20,13 +20,14 @@ import (
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
 
 const (
 	sloMeasurementName = "SLOMeasurement"
 )
 
-var sloMeasurementsNames = []string{"DnsLookupLatency"}
+var sloMeasurementsNames = []string{"DnsLookupLatency", "SchedulingThroughputPrometheus"}
 
 func init() {
 	if err := measurement.Register(sloMeasurementName, createSLOMeasurements); err != nil {
@@ -54,7 +55,13 @@ type sloMeasurement struct {
 func (s sloMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
 	errList := errors.NewErrorList()
 	for _, m := range s.measurements {
-		summaries, err := m.Execute(config)
+		measurementConfig, err := getMeasurementConfig(config, m.String())
+		if err != nil {
+			errList.Append(err)
+			continue
+		}
+
+		summaries, err := m.Execute(measurementConfig)
 		if err != nil {
 			errList.Append(err)
 			continue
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/slos/slo_measurement.go
```diff
@@ -68,6 +72,23 @@ func (s sloMeasurement) Execute(config *measurement.Config) ([]measurement.Summa
 	return s.summaries, nil
 }
 
+func getMeasurementConfig(config *measurement.Config, measurement_name string) (*measurement.Config) {
+	measurementConfig := *config
+	measurementConfig.Params = util.CloneMap(config.Params)
+
+	overrides, err := util.GetMap(config.Params, "overrides")
+	if err == nil {
```
wojtek-t, line 80: You shouldn't silently ignoring the error.

wojtek-t: In other words, this function should return (Config, error)

clusterloader2/pkg/measurement/common/slos/slo_measurement.go
```diff
@@ -54,7 +55,10 @@ type sloMeasurement struct {
 func (s sloMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
 	errList := errors.NewErrorList()
 	for _, m := range s.measurements {
-		summaries, err := m.Execute(config)
+
```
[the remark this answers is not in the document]
marseel: Done.

clusterloader2/pkg/measurement/common/slos/slo_measurement.go
marseel: Done.

wojtek-t:
/hold

I would like to see if enabling scheduling throughput in our test configs won't break anything.

marseel:
@wojtek-t looks good.

wojtek-t:
Scheduling throughput is 73:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1529/pull-perf-tests-clusterloader2/1315962499295612928/artifacts/SchedulingThroughputPrometheus_load_2020-10-13T10:56:08Z.json

Is that expected?

marseel:
Yes, the difference comes from the fact that prometheus metric is scraped every 5s (previous scheduling throughput interval in load test was 1s). For 100 nodes it takes around 7 seconds to create pods used for measuring scheduling throughput. Also, even with old SchedulingThroughput, this measurement was turned off (alerting) for 100 Nodes clusters.

marseel:
I think we can expect results like this. SchedulingThroughput on 100 nodes is quite short, previous SchedulingThroughput was measuring with interval 1 s, this one based on prometheus metric is measuring with interval 5s. Also, it's worth to mention that for 100 nodes, alerting is turned off for SchedulingThroughput.

wojtek-t:
OK - we should probably document this somewhere in the subsequent PR.
Another thing to do will be to get rid of the old scheduling throughput measurement.

/lgtm
/hold cancel

PR #1530, opened 2020-10-13
PVC creation test

pohly:
**What type of PR is this?**

/kind feature

I started using the experimental pod-startup test to measure how quickly provisioning of volumes via PVC can be done. Some tweaks to the test were necessary - see individual commit messages for details.

```release-note
NONE
```

### DIFF
--- a/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
+++ b/clusterloader2/testing/experimental/storage/pod-startup/config.yaml
@@ -6,14 +6,33 @@
   {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 
 # Test Variales
-  {{$PODS_PER_NODE := .PODS_PER_NODE}}
+  {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 1}}
   {{$DEPLOYMENT_TEMPLATE_PATH := .DEPLOYMENT_TEMPLATE_PATH }}
-  {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
+  {{$VOLUMES_PER_POD := DefaultParam .VOLUMES_PER_POD 1}}
   {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
+  # Set this to false if metrics data is not needed.
+  {{$GATHER_METRICS := DefaultParam .GATHER_METRICS true}}
+  {{$START_PODS := DefaultParam .START_PODS true}}
   {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
-  {{$USE_CSI := DefaultParam .USE_CSI false}}
+  {{$STORAGE_CLASS := DefaultParam .STORAGE_CLASS "csi-gce-pd"}}
+  # Shortcut for provisioning volumes with some external provisioner. Without it,
+  # the PV controller must update each PVC before the external provisioner
+  # starts to work on it, which limits the volume creation rate.
+  # With it, volume creation starts immediately. This is best
+  # used together with WAIT_FOR_PVS_CREATED=true and
+  # WAIT_FOR_PVS_BOUND=false because then test does not to wait for the
+  # PV controller during volume creation at all.
+  {{$PROVISIONER := DefaultParam .PROVISIONER ""}}
+  # When checking PVs, the test by default considers *all* PVs in the cluster,
+  # whether they were created by the test or not. In other words, the test
+  # only works in clusters with no PVs. To make it a bit more flexible,
+  # the value of the expected provisioner can be set here and then only
+  # PVs with that annotation will be counted.
+  {{$EXPECTED_PROVISIONER := DefaultParam .EXPECTED_PROVISIONER ""}}
   {{$VOL_SIZE := DefaultParam .VOL_SIZE "8Gi"}}
-  {{$WAIT_FOR_PVS := DefaultParam .WAIT_FOR_PVS false}}
+  {{$WAIT_FOR_PVS_CREATED := DefaultParam .WAIT_FOR_PVS_CREATED false}}
+  {{$WAIT_FOR_PVS_BOUND := DefaultParam .WAIT_FOR_PVS_BOUND false}}
+  {{$WAIT_FOR_PVS_DELETED := DefaultParam .WAIT_FOR_PVS_DELETED false}}
   {{$POD_THROUGHPUT := DefaultParam .POD_THROUGHPUT 10}}
   # TODO(hantaowang): remove knob after deciding on right values
   {{$POD_STARTUP_TIMEOUT := DefaultParam .POD_STARTUP_TIMEOUT "15m"}}
@@ -24,7 +43,10 @@
   {{$podsPerNamespace := MultiplyInt $NODES_PER_NAMESPACE $PODS_PER_NODE}}
   {{$volumesPerNamespace := MultiplyInt $podsPerNamespace $VOLUMES_PER_POD}}
   {{$totalVols := MultiplyInt $volumesPerNamespace $namespaces}}
-  {{$PVCBoundTime := MultiplyInt $totalVols 2 | MaxInt 60}}
+  {{$guessedStepTime := MultiplyInt $totalVols 2 | MaxInt 60}}
+
+# This is a very conservative estimate of 2 seconds per volume.
+  {{$StepTimeSeconds := DefaultParam .STEP_TIME_SECONDS $guessedStepTime}}
 
 
 name: storage
@@ -35,6 +57,7 @@ tuningSets:
   qpsLoad:
     qps: {{$POD_THROUGHPUT}}
 steps:
+{{ if $GATHER_METRICS }}
 # Start measurements
 - name: Starting measurement for the entire test
   measurements:
@@ -56,6 +79,7 @@ steps:
       action: start
       labelSelector: group = volume-test
       threshold: {{$POD_STARTUP_SLO}}s
+{{ end }}
 {{ if $PROVISION_VOLUME }}
 # Provision volumes
 - name: Provisioning volumes
@@ -71,11 +95,22 @@ steps:
       templateFillMap:
         Group: volume-test
         VolSize: {{$VOL_SIZE}}
-        UseCSI: {{$USE_CSI}}
-        CSIStorageClass: {{$CSI_STORAGE_CLASS}}
+        StorageClass: {{$STORAGE_CLASS}}
+        Provisioner: {{$PROVISIONER}}
 {{ end }}
-{{ if $WAIT_FOR_PVS }}
-- name: Waiting for PVs
+{{if $WAIT_FOR_PVS_CREATED }}
+- name: Waiting for volume creation
+  measurements:
+    - Identifier: WaitForPVsToBeCreated
+      Method: WaitForAvailablePVs
+      Params:
+        desiredPVCount: {{$totalVols}}
+        apiVersion: v1
+        provisioner: {{$EXPECTED_PROVISIONER}}
+        timeout: {{$StepTimeSeconds}}s
+{{ end }}
+{{ if $WAIT_FOR_PVS_BOUND }}
+- name: Waiting for PVs to be bound
   measurements:
   - Identifier: WaitForPVCsToBeBound
     Method: WaitForBoundPVCs
@@ -83,8 +118,9 @@ steps:
       desiredPVCCount: {{$totalVols}}
       apiVersion: v1
       labelSelector: group = volume-test
-      timeout: {{$PVCBoundTime}}s
+      timeout: {{$StepTimeSeconds}}s
 {{ end }}
+{{ if $START_PODS }}
 - name: Starting measurement for waiting for deployments
   measurements:
   - Identifier: WaitForRunningDeployments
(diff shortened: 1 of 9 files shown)


clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -7,12 +7,22 @@
   {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 
 # Test Variales
-  {{$PODS_PER_NODE := .PODS_PER_NODE}}
+  {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 1}}
   {{$DEPLOYMENT_TEMPLATE_PATH := .DEPLOYMENT_TEMPLATE_PATH }}
-  {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
+  {{$VOLUMES_PER_POD := DefaultParam .VOLUMES_PER_POD 1}}
   {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
+  # Set this to false if apiserver, kube-controller-manager and kube-scheduler
```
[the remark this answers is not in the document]
pohly: But you do for the other components?

I need to double-check whether I can drop the "insecure-port: '8080'" in my local cluster config. It might be that I added it pro-actively.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -7,12 +7,22 @@
...
+  # do not export metrics data via the insecure http ports.
```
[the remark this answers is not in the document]
pohly: I ran against a cluster where "kubeadm" locked down access to metrics via http and then gathering those metrics in clusterload2 failed because it couldn't connect to the ports.

The code which pulls metrics uses the insecure ports: https://github.com/kubernetes/kubernetes/blob/e799c852fb337be414fa2dae982fed69a67806fe/test/e2e/framework/metrics/metrics_grabber.go#L35-L44

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -57,9 +69,11 @@ steps:
       action: start
       labelSelector: group = volume-test
       threshold: {{$POD_STARTUP_SLO}}s
+{{ end }}
 {{ if $PROVISION_VOLUME }}
 # Provision volumes
-- phases:
+- name: Provisioning volumes
```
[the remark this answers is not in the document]
pohly: Sure.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
```diff
@@ -5,11 +5,13 @@ metadata:
   labels:
     app: {{.Name}}
     group: {{.Group}}
+  annotations:
+    volume.beta.kubernetes.io/storage-provisioner: {{.Provisioner}}
```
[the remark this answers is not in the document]
pohly: Gotcha :sob: I assumed that the actual value would still get added, but I didn't test that. Now I tested and the volume provisioning gets stuck because it remains empty. I'll revise this.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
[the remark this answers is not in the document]
pohly: kube-apiserver indeed works without the insecure port.

But a missing insecure port for kube-controller-manager causes the test to fail:
```
...
I1013 17:00:22.539742  220515 resource_gather_worker.go:69] Get container kube-controller-manager-pmem-csi-pmem-govm-master/kube-controller-manager usage on node pmem-csi-pmem-govm-master. CPUUsageInCores: 0.011748654, MemoryUsageInBytes: 0, MemoryWorkingSetInBytes: 44888064
I1013 17:00:22.539758  220515 resource_gather_worker.go:69] Get container kube-proxy-7474b/kube-proxy usage on node pmem-csi-pmem-govm-master. CPUUsageInCores: 0.000631058, MemoryUsageInBytes: 0, MemoryWorkingSetInBytes: 13516800
E1013 17:00:30.651327  220515 metrics_for_e2e.go:79] MetricsForE2E: metricsGrabber failed to grab some of the metrics: Errors while grabbing metrics: [error waiting for controller manager pod to expose metrics: timed out waiting for the condition; an error on the server ("unknown") has prevented the request from succeeding (get pods kube-controller-manager-pmem-csi-pmem-govm-master:10252)]
...
```

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -7,12 +7,13 @@
   {{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
 
 # Test Variales
-  {{$PODS_PER_NODE := .PODS_PER_NODE}}
+  {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 1}}
```
pohly, line 10: I actually wanted to ask about this: `DefaultParam` is supposed to make it possible to provide those values via environment variables instead of overrides, right?

I'm following the pattern from `NODES_PER_NAMESPACE` here, but when I tested it, I found that the env variables are ignored. Is that perhaps because they don't start with `CL2_`?

That's broken then also for the other variables, something that I should fix before repeating the same mistake.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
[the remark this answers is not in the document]
wojtek-t: OK - looked into the code and I know what is happening (I was wondering why this works for GKE where we don't even expose access to pods)

And apparently - we're grabbing those metrics via listing pods and if we don't find pods, we simply ignore them:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/metrics/metrics_grabber.go#L98

This code will have to be rewritten anyway (it e.g. doesn't support HA setups).

But in the meantime, the correct way to solve this would be to:
- introducing new features similar to SupportGrabMetricsFromKubelets in provider package:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/provider/provider.go#L52
- using this in the measurement similarly as we do for kubelets:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/metrics_for_e2e.go#L63

That should address your problem and allow you to run all the other measurements that are actually useful.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: Yeah - only the ones prefixed with CL2_ are propagated down the stack (I can't remember the exact motivation for that, but there was a good reason to not propagate everything).

pohly: I decided to stay with the current naming, i.e. env variables continue to not work...

pohly:
@wojtek-t : can you have another look? I rebased and cleaned up the commits.

I kept the "disable metrics gather" option. It's not needed anymore to avoid the test failure, but I found it useful nonetheless when I just wanted to see whether provisioning works.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -7,12 +7,20 @@
...
   {{$DEPLOYMENT_TEMPLATE_PATH := .DEPLOYMENT_TEMPLATE_PATH }}
-  {{$VOLUMES_PER_POD := .VOLUMES_PER_POD}}
+  {{$VOLUMES_PER_POD := DefaultParam .VOLUMES_PER_POD 1}}
   {{$VOLUME_TEMPLATE_PATH := .VOLUME_TEMPLATE_PATH}}
+  # Set this to false if metrics data is not needed.
+  {{$GATHER_METRICS := DefaultParam .GATHER_METRICS true}}
```
wojtek-t, line 15: I still believe this should be removed due to reasons discussed in the other thread (in particular api-call latencies SLO is measured behind this param and it should be measured).

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -7,12 +7,20 @@
...
+  {{$START_PODS := DefaultParam .START_PODS true}}
   {{$PROVISION_VOLUME := DefaultParam .PROVISION_VOLUME false}}
-  {{$USE_CSI := DefaultParam .USE_CSI false}}
+  {{$STORAGE_CLASS := DefaultParam .STORAGE_CLASS "csi-gce-pd"}}
+  # Used for provisioning volumes via CSI. Without it,
+  # the PV controller must update each PVC before external-provisioner
+  # starts to work on it, which limits the volume creation rate.
+  # With it, volume creation starts immediately.
+  {{$PROVISIONER := DefaultParam .PROVISIONER ""}}
```
wojtek-t, line 23: What happens when this is set to empty string?

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
```diff
@@ -5,11 +5,15 @@ metadata:
   labels:
     app: {{.Name}}
     group: {{.Group}}
+{{ if .Provisioner }}
```
wojtek-t, line 8: Hmm - how does "ifs" for string work?

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
pohly: This depends on why you run the test. If it is to measure performance, then yes. But I was also using it to trigger a race condition (https://github.com/kubernetes-csi/external-provisioner/issues/507) and then metrics gathering is just unnecessary overhead. In particular in a large cluster it can take time as it walks through various pods.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
pohly: Due to the if check, the annotation is not set and the volume controller must add it after the PVC got created.

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
pohly: Apparently it checks for non-empty, i.e. empty string = false.

pohly:
/hold

Skipping the PV controller for provisioning but then still waiting for it to bind volumes is inconsistent.

I'll change it so that it's either used for setting the annotation and binding or not relied upon at all.

pohly:
/hold cancel

Now volume creation can be tested with and without dependency on the PV controller.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
```diff
@@ -71,20 +95,32 @@ steps:
...
-- name: Waiting for PVs
+{{if $WAIT_FOR_PVS_CREATED }}
+- name: Waiting for volume creation
+  measurements:
+    - Identifier: WaitForPVsToBeCreated
+      Method: WaitForAvailablePVs
+      Params:
+        desiredPVCount: {{$totalVols}}
+        apiVersion: v1
+        provisioner: {{$EXPECTED_PROVISIONER}}
```
wojtek-t, line 109: I think we want to pass labelSelector too (as in other waits).

clusterloader2/testing/experimental/storage/pod-startup/volume-types/persistentvolume/pvc.yaml
wojtek-t: Interesting - but that's fine :)

wojtek-t (commented):
Couple small-ish comments.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
pohly: Hmm, I wonder whether `WaitForPVOptions` should even have a `Selector` field. That was added in 21d99e9d0 but I don't see what it is good for.

pohly (commented):
@wojtek-t I've pushed an update (all changes in the last commit), please take another look.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: I think there shouldn't be an option for that, but ok - let's not block this PR on it.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: Can we remove it then?

wojtek-t (commented):
Please also squash commits.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
pohly: I'm not sure - perhaps it has some purpose with other provisioners that I just don't know about. Let's keep it.

pohly:
> Please also squash commits.

Each commit is self-contained, but I can also squash - done.

clusterloader2/testing/experimental/storage/pod-startup/config.yaml
wojtek-t: It seems suspicious to have it set in some "waits" and not in others.

But OK - these tests are marked as experimental anyway for now ...

PR #1539, opened 2020-10-20
Fix OOM tracker in clusters with large number of events

### LINKED ISSUE #1536: OOM tracker timing out in clusters with very large number of events
With > 500k events, OOM tracker have problems starting watching for events with narrow selector for OOMs
```
[measurement call TestMetrics - TestMetrics error: [
  action start failed for ClusterOOMsTracker measurement: 
    starting cluster OOMs measurement problem: 
      problem with OOM events informer starting: 
        timed out waiting for caches to sync]
```
As there are O(0) OOM events, initial list before watch start needs to process all events before returning the result. apiserver call times out, as apiserver tries retrieve results from etcd paginating over all events with default page size of 500 results.

tosi3k:
Fixes #1536.

### DIFF
--- a/clusterloader2/pkg/measurement/common/ooms_tracker.go
+++ b/clusterloader2/pkg/measurement/common/ooms_tracker.go
@@ -17,6 +17,7 @@ limitations under the License.
 package common
 
 import (
+	"context"
 	"fmt"
 	"regexp"
 	"strings"
@@ -26,6 +27,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/pager"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
@@ -39,6 +45,7 @@ const (
 	clusterOOMsIgnoredProcessesParamName = "clusterOOMsIgnoredProcesses"
 	informerTimeout                      = time.Minute
 	oomEventReason                       = "OOMKilling"
+	initialListPageSize                  = 10000
 )
 
 var (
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -115,6 +123,57 @@ func (m *clusterOOMsTrackerMeasurement) String() string {
...
+	watchFunc := func(options metav1.ListOptions) (watch.Interface, error) {
+		options.FieldSelector = m.selector.FieldSelector
+		return client.CoreV1().Events(metav1.NamespaceAll).Watch(ctx, options)
+	}
+	i := cache.NewSharedInformer(&cache.ListWatch{ListFunc: listFunc, WatchFunc: watchFunc}, nil, 0)
+	i.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			m.handleOOMEvent(obj)
+		},
+		UpdateFunc: func(_, _ interface{}) {},
```
wojtek-t, line 148: You want UpdateFunc to be set too - subsequent OOMs may be reported as even updates.

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -115,6 +123,57 @@ func (m *clusterOOMsTrackerMeasurement) String() string {
...
+}
+
+func (m *clusterOOMsTrackerMeasurement) handlePriorOOMs(ctx context.Context, client clientset.Interface) error {
+	pg := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
+		return client.CoreV1().Events(metav1.NamespaceAll).List(ctx, opts)
+	}))
+	pg.PageSize = initialListPageSize
+
+	options := metav1.ListOptions{
+		FieldSelector: m.selector.FieldSelector,
```
jkaniuk, line 161: I would not use selector during initial list. That way we can paginate  over all events client side issuing more lighter calls to apiserver instead of single heavy one.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Done.

jkaniuk:
/lgtm

As discussed with @tosi3k - it was e2e tested.

@wojtek-t - PTAL for final review

jkaniuk:
/hold
for @wojtek-t review

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -115,6 +123,56 @@ func (m *clusterOOMsTrackerMeasurement) String() string {
...
+	return i
+}
+
+func (m *clusterOOMsTrackerMeasurement) handlePriorOOMs(ctx context.Context, client clientset.Interface) error {
+	pg := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
+		return client.CoreV1().Events(metav1.NamespaceAll).List(ctx, opts)
+	}))
+	pg.PageSize = initialListPageSize
+
+	obj, _, err := pg.List(ctx, metav1.ListOptions{})
```
wojtek-t, line 162: Instead of Listing everything - you want to use EachListItem:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/pager/pager.go#L156

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -124,13 +182,16 @@ func (m *clusterOOMsTrackerMeasurement) start(config *measurement.Config) error
 	if err := m.initFields(config); err != nil {
 		return fmt.Errorf("problem with OOMs tracking measurement fields initialization: %w", err)
 	}
+	ctx := context.Background()
+	client := config.ClusterFramework.GetClientSets().GetClient()
+
+	// Searching for OOM events that happened before the measurement start.
+	if err := m.handlePriorOOMs(ctx, client); err != nil {
+		return fmt.Errorf("problem with handling prior OOMs: %w", err)
+	}
```
wojtek-t, line 191: Technically you may miss some events between this list and starting informer.

I know it's not that big deal, but you can probably relatively easily reduce the probability by changing the order:
- start the informer
- then call handlePriorOOMs

You can then count something twice potentially, but that's probably not big deal too.

I don't have strong opinion which one is better - but if you want to leave as it, please add a comment that we know that some events might be missed in the meantime and we're fine with that.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/ooms_tracker.go
tosi3k: Good catch. I went with your suggestion of changing the sequence of the function calls.

>You can then count something twice potentially, but that's probably not big deal too.

We already dedupe OOM events by checking the value of the object's RV in the `handleOOMEvent` method.

tosi3k:
@wojtek-t PTAL. Thanks!

PR #1546, opened 2020-10-26
Update configmap in load test for preparation to DNS tests

wojtek-t:
Ref https://github.com/kubernetes/perf-tests/pull/1526/files#r511885934

### DIFF
--- a/clusterloader2/testing/load/configmap.yaml
+++ b/clusterloader2/testing/load/configmap.yaml
@@ -5,8 +5,33 @@ metadata:
 {{if not (eq (Mod .Index 20) 0 19) }} # .Index % 20 in {0,19} - only 10% deployments will have non-immutable ConfigMap.
 immutable: true
 {{end}}
+# Every pod that needs its own configmap entry should be unconditionally
+# added below. That allows us to avoid complicating it with ifs.
 data:
-  data.yaml: |-
-    a: 1
-    b: 2
-    c: 3
+  # all-queries is used by DNS tests.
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/configmap.yaml
```diff
@@ -6,7 +6,29 @@ metadata:
 immutable: true
 {{end}}
 data:
-  data.yaml: |-
-    a: 1
-    b: 2
-    c: 3
+  all-queries: |
```
mborsz, line 9: Please add comment that this is for dns tests + add a comment that other use cases can be added here next to the all-queries file.

clusterloader2/testing/load/deployment.yaml
```diff
@@ -51,6 +51,11 @@ spec:
         - name: configmap
           configMap:
             name: {{.BaseName}}-{{.Index}}
+{{if false}}
```
mborsz, line 54: Do we need to mount only selected items? I think it makes sense to mount all and using only /var/configmap/feature-name if needed.

clusterloader2/testing/load/configmap.yaml
wojtek-t: done

clusterloader2/testing/load/deployment.yaml
wojtek-t: I wanted to have minimal changes, but yeah - we can go with the eventual solution from the beginning.

PR #1549, opened 2020-10-28
Define resource tracking mode based on cluster size

wojtek-t:
This will allow us to remove some overrides and thanks that simplify our test job definitions.

/assign @mborsz @jkaniuk

### DIFF
--- a/clusterloader2/pkg/measurement/common/resource_usage.go
+++ b/clusterloader2/pkg/measurement/common/resource_usage.go
@@ -32,6 +32,9 @@ import (
 
 const (
 	resourceUsageMetricName = "ResourceUsageSummary"
+	// maxNodeCountForAllNodes defines the threshold for cluster size above which
+	// we no longer gather resource usage from all system components on all nodes.
+	maxNodeCountForAllNodes = 1000
 )
 
 func init() {
@@ -67,10 +70,6 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Config) ([]
 		if err != nil {
 			return nil, err
 		}
-		nodeMode, err := util.GetStringOrDefault(config.Params, "nodeMode", "")
-		if err != nil {
-			return nil, err
-		}
 		namespace, err := util.GetStringOrDefault(config.Params, "namespace", "kube-system")
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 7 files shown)


clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
```diff
@@ -42,10 +42,8 @@ const (
 	AllNodes NodesSet = 0
 	// MasterNodes - all containers on Master nodes only
 	MasterNodes NodesSet = 1
```
jkaniuk, line 44: where do we set MasterNodes?

clusterloader2/pkg/measurement/common/resource_usage.go
```diff
@@ -94,16 +93,12 @@ func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Config) ([]
...
+		// Compute the node based on the cluster size.
+		nodeCount := config.ClusterFramework.GetClusterConfig().Nodes
+		nodesSet := gatherers.AllNodes
+		if nodeCount > maxNodeCountForAllNodes {
 			nodesSet = gatherers.MasterAndNonDaemons
-		default:
-			nodesSet = gatherers.AllNodes
 		}
 
 		klog.V(2).Infof("%s: starting resource usage collecting...", e)
```
jkaniuk, line 104: we should log selected mode

mm4tt:
Looks ok modulo Jacek's comments.

However, given that there is a chance it may break something it shouldn't be merged without explicit OK from the scalability oncall - @mborsz

clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
wojtek-t: I think we don't set it anywhere now. We used to set it for larger clusters when we didn't have MasterAnd* modes...

clusterloader2/pkg/measurement/common/resource_usage.go
wojtek-t: done

wojtek-t (commented):
PTAL

clusterloader2/pkg/measurement/util/gatherers/container_resource_gatherer.go
jkaniuk: If that is not used anymore, please remove all the code related.

wojtek-t: I actually think this one is useful - at some point we may not be able to gather even from system components.
I wouldn't remove this one actually.

jkaniuk: Looks like YAGNI.
I'm fine with either removing this or restoring a flag that could enable it, but not leaving a dead code.

wojtek-t: ok - good point that it's no longer posisble to set.

Removed

jkaniuk: Thanks

jkaniuk:
LGTM

Leaving for final @mborsz approval as an oncall

PR #1553, opened 2020-11-02
Allow to wait and exit after command execution

marseel:
**What type of PR is this?**

/kind feature

This PR allows us to stop test after execution of command.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -77,6 +77,8 @@
 
 # Command to be executed
 {{$EXEC_COMMAND := DefaultParam .CL2_EXEC_COMMAND nil}}
+{{$EXIT_AFTER_EXEC := DefaultParam .CL2_EXIT_AFTER_EXEC false}}
+{{$WAIT_AFTER_EXEC_DURATION := DefaultParam .CL2_WAIT_AFTER_EXEC_DURATION "0s"}}
 
 name: load
 namespace:
@@ -387,7 +389,12 @@ steps:
       {{range $EXEC_COMMAND}}
       - {{.}}
       {{end}}
+  - Identifier: WaitAfterExec
+    Method: Sleep
+    Params:
+        duration: {{$WAIT_AFTER_EXEC_DURATION}}
 {{end}}
+{{if not $EXIT_AFTER_EXEC}}
 
 {{if $ENABLE_DENSITY_TEST}}
 # BEGIN scheduler throughput
@@ -851,6 +858,7 @@ steps:
     objectBundle:
     - basename: small-service
       objectTemplatePath: service.yaml
+{{end}} # not EXIT_AFTER_EXEC
 
 - name: Collecting measurements
   measurements:


clusterloader2/testing/load/config.yaml
```diff
@@ -897,3 +904,4 @@ steps:
       clusterOOMsTrackerEnabled: {{$ENABLE_CLUSTER_OOMS_TRACKER}}
       restartCountThresholdOverrides: {{YamlQuote $RESTART_COUNT_THRESHOLD_OVERRIDES 4}}
       enableRestartCountCheck: {{$ENABLE_RESTART_COUNT_CHECK}}
+{{end}}
```
mborsz, line 907: I'm not sure if this is a good idea to skip gathering measurements. We may want to use those at some point and they should have little to no impact on cluster, so it should be ok to run them after exec command.

marseel: Done. I've also added comment that indicates which if ends.

clusterloader2/testing/load/config.yaml
```diff
@@ -77,6 +77,8 @@
 
 # Command to be executed
 {{$EXEC_COMMAND := DefaultParam .CL2_EXEC_COMMAND nil}}
+{{$EXIT_AFTER_EXEC := DefaultParam .CL2_EXIT_AFTER_EXEC false}}
+{{$WAIT_AFTER_EXEC_DURATION := DefaultParam .CL2_WAIT_AFTER_EXEC_DURATION "0s"}}
```
wojtek-t, line 81: WAIT_AFTER_EXEC_DURATION is misleading, because don't wait for anything we just sleep.

should probably be SLEEP_AFTER_EXEC_DURATION

clusterloader2/testing/load/config.yaml
```diff
@@ -387,7 +389,12 @@ steps:
       {{range $EXEC_COMMAND}}
       - {{.}}
       {{end}}
+  - Identifier: WaitAfterExec
+    Method: Sleep
+    Params:
+        duration: {{$WAIT_AFTER_EXEC_DURATION}}
 {{end}}
+{{if not $EXIT_AFTER_EXEC}}
```
wojtek-t, line 397: having if spanning 500 lines seem like a bad idea TBH...

wojtek-t: @mm4tt

mborsz: Yeah, I agree it's far from perfection, but it seems to be the better than using skip-steps feature.

The goal is to reuse first part of the test to generate workload on cluster before running command and NOT running anything after the command that to remove potential impact of clusterloader on the operation started by the command.

An alternative I can see is to use --skip-steps feature for this, but this seems to be an error prone to me (we need to hardcopy list of all steps in the prowjob and adding any step to load test will require reflecting this in all prowjobs using this config). In the current approach, there is no risk that we will accidentally start running some step in our tests using exec step.

When I'm thinking about this now, I think we can have another solution: we can implement a "skip-all-steps-after--step" feature that will skip all 'phases' steps (leaving measurements) or removing all steps after the step with the given name.

clusterloader2/testing/load/config.yaml
marseel: Sure, I will fix it.

clusterloader2/testing/load/config.yaml
marseel: I am wondering if we shouldn't rethink format of our tests and maybe come up with some better idea than hardcoding steps/huge ifs/skipping some steps after.

mm4tt: The issue with `--skip-all-steps-after-step` is that you need to skip some measurements (e.g. `WaitForControlledPods`) but you really want to execute some (e.g. collect on api-call-latency or resource usages). I agree with Marcel that solving that properly would probably require rethinking cl2 api and the structure of the test file. We could even kill two birds with a single stone and at the same time introduce "modularization" and refactor single huge config.yaml into multiple files. It would make the test easier to maintain / understand. But let's discuss this, it'd be a big change. I'd like to see a full design first before we dig into implementation.

wojtek-t: @mborsz @mm4tt @marseel - did you discuss that offline?

Being able to reference the other files (e.g. a single file defining a single step or a set of steps) would be useful. But that's a big thing to do, so I'm wondering if we don't need something more shorter-term.

Re -skip-all-steps-after-step would work - I like the idea modulo the fact that Matt described. That would work though if we would introduce the concept of deferred step, e.g. you do:
- step Start Metrics
- deferred-step Collect Metrics

wojtek-t: Briefly discussed offline - we will discuss that a bit more in the upcoming weeks and see how to attack this in a better way.

PR #1555, opened 2020-11-04
[Golang CI jobs] Print most recent Golang tip commits

tosi3k:
This will facilitate establishing which of the recently committed Golang changes has introduced a regression.

/sig scalability
/assign @mm4tt

### DIFF
--- a/golang/build-go.sh
+++ b/golang/build-go.sh
@@ -29,5 +29,13 @@ function build_go {
   ./make.bash
 }
 
+function print_recent_commits {
+  local -r hours=5
+  local -r dashline=$(printf %100s | tr " " "-")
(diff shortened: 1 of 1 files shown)


mm4tt:
Nice, but as discussed, I think we'd benefit more from solving a different problem. I.e., we want to have a way to directly connect a golang test run with golang commit

tosi3k:
> Nice, but as discussed, I think we'd benefit more from solving a different problem. I.e., we want to have a way to directly connect a golang test run with golang commit

Opened a PR with an approach both of us agreed would be more convenient: https://github.com/kubernetes/perf-tests/pull/1556.

PR #1556, opened 2020-11-05
Add Golang job script printing Golang repo commit hash

tosi3k:
Small facilitation utility for debugging Golang-related regressions.

/sig scalability

### DIFF
--- a/golang/run-e2e.sh
+++ b/golang/run-e2e.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


golang/run-e2e.sh
```diff
@@ -0,0 +1,37 @@
...
+function print_golang_commit {
+  # Running "go version" on any of the extracted K8s binaries will give us
+  # the version of Go that's been used to build the binaries.
+  # In case of a non-standard, customly built Go, this will return the hash of
+  # the commit that was used for that Go build (plus some unnecessary junk that
+  # we remove below).
+  local -r go_version=$(go version $GOPATH/src/k8s.io/perf-tests/kubernetes/client/bin/kubectl)
+  local -r go_commit_hash=$(echo -n $go_version | awk '{ print $3 }' | cut -c2-)
+  local -r dashline=$(printf %100s | tr " " "-")
+  echo "${dashline}"
```
mm4tt, line 30: Can we also print the go_version? It wouldn't hurt, but will save us time if in the future the go version format changes and go_commit_hash will become gibberish

tosi3k: Good suggestion. Done.

PR #1563, opened 2020-11-09
Replace references to old topology labels with the new ones

tosi3k:
Output of grepping `failure-domain.beta.kubernetes.io` at the top-level perf-tests repo directory, `master` branch:
```
./clusterloader2/vendor/k8s.io/api/core/v1/well_known_labels.go:22:	LabelZoneFailureDomain       = "failure-domain.beta.kubernetes.io/zone"
./clusterloader2/vendor/k8s.io/api/core/v1/well_known_labels.go:23:	LabelZoneRegion              = "failure-domain.beta.kubernetes.io/region"
./clusterloader2/pkg/util/ssh.go:39:	zone, ok := node.Labels["failure-domain.beta.kubernetes.io/zone"]
./clusterloader2/pkg/util/ssh.go:41:		return fmt.Errorf("unknown zone for %q node: no failure-domain.beta.kubernetes.io/zone label", node.Name)
./clusterloader2/pkg/prometheus/experimental.go:89:		zone = pv.ObjectMeta.Labels["failure-domain.beta.kubernetes.io/zone"]
./compare/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/well_known_labels.go:23:	LabelZoneFailureDomain = "failure-domain.beta.kubernetes.io/zone"
./compare/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/well_known_labels.go:24:	LabelZoneRegion        = "failure-domain.beta.kubernetes.io/region"
```

Not seeing any references to `LabelZoneFailureDomain` nor `LabelZoneRegion` in the non-vendored code.

Fixes #1554.

/sig scalability
/assign @mm4tt @wojtek-t

PR #1566, opened 2020-11-12
Revert "Replace references to old topology labels with the new ones"

mborsz:
Reverts kubernetes/perf-tests#1563

This seems to be causing https://github.com/kubernetes/kubernetes/issues/96509

/assign @wojtek-t


RobertKielty:
Possible root cause of #96431

PR #1579, opened 2020-11-19
Adapt the event message regex in the OOM tracker

tosi3k:
Get rid of `Kill process <PID> (<PROCESS_NAME>) score 0 or sacrifice child\n` part of the OOM event message regex as it's no longer printed out in the kernel logs.

/sig scalability
/assign @jkaniuk

### DIFF
--- a/clusterloader2/pkg/measurement/common/ooms_tracker.go
+++ b/clusterloader2/pkg/measurement/common/ooms_tracker.go
@@ -42,7 +42,7 @@ const (
 )
 
 var (
-	oomEventMsgRegex = regexp.MustCompile(`Kill process (\d+) \((.+)\) score \d+ or sacrifice child\nKilled process \d+ .+ total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`)
+	oomEventMsgRegex = regexp.MustCompile(`Killed process (\d+) \((.+)\) total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`)
 )
 
 func init() {


clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -42,7 +42,7 @@ const (
 )
 
 var (
-	oomEventMsgRegex = regexp.MustCompile(`Kill process (\d+) \((.+)\) score \d+ or sacrifice child\nKilled process \d+ .+ total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`)
+	oomEventMsgRegex = regexp.MustCompile(`Killed process (\d+) \((.+)\) total-vm:(\d+kB), anon-rss:\d+kB, file-rss:\d+kB.*`)
```
wojtek-t, line 45: Shouldn't we try to have both?

tosi3k: I'm not sure I follow but here's the context: https://github.com/kubernetes/node-problem-detector/issues/480#issuecomment-714467335.

Basically, I'm adapting the regex to handle OOM events emitted by both the old and the new NPD here.

mm4tt:
IIUC, the new method will be backward compatible, i.e. it will also work for old NPD releases.
In that a case, it looks reasonable. Adding hold to confirm. Feel free to void it yourself
/hold

/lgtm
/approve

tosi3k:
> IIUC, the new method will be backward compatible, i.e. it will also work for old NPD releases.

Exactly. OOM tracker searches for the substring match of the regex, not for an exact match, so both formats will be covered.

clusterloader2/pkg/measurement/common/ooms_tracker.go
jkaniuk: If you are replacing `Kill` with `Killed` it would not work with old NPD.

tosi3k: It will - I'm simply removing the prefix of the message we observe in the NPD's kernel monitor.

jkaniuk: You are right, I did not notice the second part.

jkaniuk: Could you add some unit tests? It could be in another commit so it would be easier to backport.

tosi3k: Sure, I will open a separate PR for that.

jkaniuk: As backports already have merged, ok.

Issue #1586, opened 2020-11-19
nodelocaldns perf tests failing after migration to python3


prameshj:
**Which jobs are failing**: 
NodelocalDNS benchmark/scale test.
Results used to be http://perf-dash.k8s.io/#/?jobname=node-local-dns%20benchmark&metriccategoryname=dns&metricname=Queries, but this is empty since tests arent running.

**Which test(s) are failing**: nodelocaldns scale test, no periodic CI

**Since when has it been failing**:
Since https://github.com/kubernetes/perf-tests/pull/1470 merged

**Testgrid link**:
No periodic OSS tests are being run

**Reason for failure**:
most likely due to migrating to python3, it worked fine when using python2.
```
W1119 01:09:33.272] 2020/11/19 01:09:33 process.go:153: Running: /go/src/k8s.io/perf-tests/run-e2e.sh node-local-dns /workspace/_artifacts/out /workspace/_artifacts
I1119 01:09:34.084] Collecting numpy
I1119 01:09:34.151]   Downloading numpy-1.19.4-cp37-cp37m-manylinux2010_x86_64.whl (14.5 MB)
I1119 01:09:34.862] Installing collected packages: numpy
I1119 01:09:36.211] Successfully installed numpy-1.19.4
W1119 01:09:36.728] INFO 11-19 01:09:36 runner.py:81] Using existing node-local-dns for tests
W1119 01:09:36.729] INFO 11-19 01:09:36 runner.py:86] DNS service IP is 169.254.20.10
W1119 01:09:37.572] INFO 11-19 01:09:37 runner.py:470] Created rundir /workspace/_artifacts/out/run-1605748177
W1119 01:09:37.572] INFO 11-19 01:09:37 runner.py:476] Updated symlink /workspace/_artifacts/out/latest
W1119 01:09:37.573] INFO 11-19 01:09:37 runner.py:416] Starting client teardown
W1119 01:09:37.819] INFO 11-19 01:09:37 runner.py:426] Client teardown complete
W1119 01:09:37.823] INFO 11-19 01:09:37 runner.py:340] Starting server teardown
W1119 01:09:38.369] INFO 11-19 01:09:38 runner.py:439] Waiting for server to be deleted (0 pods active)
W1119 01:09:38.369] INFO 11-19 01:09:38 runner.py:348] Server teardown ok
W1119 01:09:38.444] INFO 11-19 01:09:38 runner.py:416] Starting client teardown
[...]

PR #1587, opened 2020-11-20
Fix dns benchmark failing after python 2->3 migration

mm4tt:
Ref. https://github.com/kubernetes/perf-tests/issues/1586

### DIFF
--- a/dns/py/runner.py
+++ b/dns/py/runner.py
@@ -41,7 +41,8 @@ _remove_query_pattern=["setting3[.]yeahost[.]com"]
 MAX_TEST_SVC = 20
 
 def add_prefix(prefix, text):
-  return '\n'.join([prefix + l for l in text.decode().split('\n')])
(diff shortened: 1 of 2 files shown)


dns/py/runner.py
```diff
@@ -161,8 +162,6 @@ def _kubectl(self, stdin, *args):
     """
     cmdline = [self.args.kubectl_exec] + list(args)
     _log.debug('kubectl %s', cmdline)
-    if stdin:
-      _log.debug('kubectl stdin\n%s', add_prefix('in  | ', stdin))
```
mm4tt, line 165: I removed this logging as it was causing more harm than good. We're logging stdin here, which can be anything, e.g. a tar archive...

wojtek-t:
/lgtm

It seems the tests are passing now, which is great.

/hold
Waiting for Pavithra.

prameshj:
/hold cancel
/lgtm

Thanks Matt!

PR #1588, opened 2020-11-20
Add missing bash options to dns/run.sh

mm4tt:
This is, among others, to properly propagate error status.

Ref. https://github.com/kubernetes/perf-tests/issues/1586

### DIFF
--- a/dns/run
+++ b/dns/run
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -o errexit
+set -o nounset
+
 # $1 - Test type.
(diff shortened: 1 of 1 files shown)


mm4tt:
The `pull-perf-tests-benchmark-kube-dns` test failed, but it's exactly what we wanted to achieve. We'll fix it later.
It's not a required presubmit so it won't block merging.

mm4tt:
Looks like the pull-perf-tests-benchmark-kube-dns is not optional :/

I could make it optional then merge this and then make it back required. However, I'm not sure if it's worth the hassle. Let me try merging this PR with the test failing.

PR #1498, opened 2020-09-28
Enhancing Network Performance Benchmarking Testcases

hanamantagoudvk:

PR #1592, opened 2020-11-24
Adding network performance measurement testcases. (Test script addition)

hanamantagoudvk:
**What type of PR is this?**

/kind feature
/kind regression

**What this PR does / why we need it:**

This review is one of the changeset done as part of the enhancement/feature (Reference #1498)

This changeset has testcase yaml files added for clusterloader framework.

It also has PerfDash changes for displaying the results in dashboard.


clusterloader2/testing/network/1_1_ratio_override.yaml
```diff
@@ -0,0 +1,2 @@
+Ratio: "1:1"
```
wojtek-t, line 1: I missed that during the original review - can you please capitalize those?

We're using the convention that all env-vars that we're passing are capitalized, e.g.:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/experimental/storage/pod-startup/cluster_load_scale_by_nodes/override.yaml

wojtek-t, line 1: BTW - I think we moved from ratio and now have number of servers/clients, right?

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,40 @@
+  {{$Replicas := .Replicas}}
```
wojtek-t, line 1: remove initial whiltespaces

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+  {{$Ratio := .Ratio}}
+  {{$Protocol := .Protocol}}
+name: network
```
wojtek-t, line 4: add empty line before

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+    parallelismLimit: 1
+steps:
+- name: start network metric measurement
+  measurements:
+    - Identifier: NetworkPerformanceMetrics
+      Method: NetworkPerformanceMetrics
+      Params:
+        action: start 
+        podReplicas: {{$Replicas}}
+        duration: 10
```
wojtek-t, line 18: what is duration? isn't it of "Duration type"? (i.e. shouldn't it have units?"

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+        ratio: {{$Ratio}} 
+        protocol: {{$Protocol}}
+- name: Wait for 15 secs controller to be up
```
wojtek-t, line 21: That's not needed anymore, right?

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+      Method: Sleep
+      Params:
+        duration: 15s
+- name: Gather
+  measurements:
+    - Identifier: NetworkPerformanceMetrics
+      Method: NetworkPerformanceMetrics
+      Params:
+        action: gather 
+- name: Wait 0min
```
wojtek-t, line 33: This also isn't needed.

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+  measurements:
+    - Identifier: Wait
+      Method: Sleep
+      Params:
+        duration: 0m
+
```
wojtek-t, line 39: and remove empty lines at the end

clusterloader2/testing/network/1_1_ratio_override.yaml
wojtek-t: Same for all variables

perfdash/config.go
```diff
@@ -340,6 +340,45 @@ var (
 					Parser:           parsePerfData,
 				}},
 		},
+		"Network_Baseline": {
```
wojtek-t, line 343: Please split perf-dash changes to a separate PR.

clusterloader2/testing/network/1_1_ratio_override.yaml
hanamantagoudvk: yes i have capitalized the variables.

hanamantagoudvk: yes modified

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes i have modified it to duration type here , as well as in code. But while sending it to worker pod via CR , i convert it into seconds.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: Its requred. After calling start() , if we dont have wait in-between , gather() is called immediately. In gather() we read the result and return immediately. We dont wait or do retry in gather() call.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified

perfdash/config.go
hanamantagoudvk: yes created another PR.

clusterloader2/testing/network/50_50_ratio_override.yaml
```diff
@@ -0,0 +1,2 @@
+NUMBER_OF_SERVERS: 50
+NUMBER_OF_CLIENTS: 50
```
wojtek-t, line 2: nit: add EOL

clusterloader2/testing/network/1_1_ratio_override.yaml
```diff
@@ -0,0 +1,2 @@
+NUMBER_OF_SERVERS: 1
```
wojtek-t, line 1: If you look into our tests, we're moving towards prefixing those vars with CL2_ (then you basically can parametrize tests without even having overrides).

So please prefix those variables with CL2_
(here and in other override files).

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
+{{$PROTOCOL := .PROTOCOL}}
```
wojtek-t, line 1: And here do:
{{$PROTOCOL := .CL2_PROTOCOL}}

same below

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+{{$NUMBER_OF_SERVERS := .NUMBER_OF_SERVERS}}
+{{$NUMBER_OF_CLIENTS := .NUMBER_OF_CLIENTS}}
+
+name: network
```
wojtek-t, line 5: Let's maybe come up with a more meaningful name.
Also capitalize.

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+automanagedNamespaces: 1
+tuningSets:
+- name: Sequence
```
wojtek-t, line 8: This isn't used anywhere - please remove tuningSets completely.

clusterloader2/testing/network/config.yaml
wojtek-t: Maybe "Network Performance"

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  parallelismLimitedLoad:
+    parallelismLimit: 1
+steps:
+- name: start network metric measurement
```
wojtek-t, line 12: Start network performance measurement.

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  measurements:
+    - Identifier: NetworkPerformanceMetrics
+      Method: NetworkPerformanceMetrics
+      Params:
+        action: start
+        duration: 10s
```
wojtek-t, line 18: Only 10s?

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+        protocol: {{$PROTOCOL}}
+        numberOfServers: {{$NUMBER_OF_SERVERS}}
+        numberOfClients: {{$NUMBER_OF_CLIENTS}}
+- name: Wait for 25 secs controller to be up
```
wojtek-t, line 22: This isn't needed, right? The gather will wait until everything will finish anyway, right?

clusterloader2/testing/network/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  measurements:
+    - Identifier: Wait
+      Method: Sleep
+      Params:
+        duration: 25s
+- name: Gather
```
wojtek-t, line 28: Please add better name Gather network performance metrics maybe?

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes its needed. There is no wait in gather implementation. When gather is executed , it will read the results stored in local cache and return.

clusterloader2/testing/network/50_50_ratio_override.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/1_1_ratio_override.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes we just started with this value. Once the entire code is ready and it starts running in prow job , we can decide the runtime of each of these testcases. It has an impact on total runtime.

clusterloader2/testing/network/config.yaml
hanamantagoudvk: yes modified.

clusterloader2/testing/network/config.yaml
wojtek-t: I see - I missed that during the review. We should fix that.

Basically gather should instead do:
- wait (with timeout) until all objects that we created have already been updated with the result

hanamantagoudvk: There are certain aspects we need to think through:
a) Should we have wait based on number of pods configured ? or have a fixed timeout value ? If we have fixed timeout, at lower case runtime will be high. 
b) When timeout happens , it possible that say some pod didnot update result (due to error in pod or could be delay etc). Should we still go ahead with our percentile calculation and return the summary with whatever numbers we have or return error ? my opinion is , we should return whatever result is available.

wojtek-t: > Should we have wait based on number of pods configured ? or have a fixed timeout value ? If we have fixed timeout, at lower case runtime will be high.

You know what the expected "test duration" is (because you pass it to the workers).
So basically you should wait until (a) you receive results from all pods (b) but no longer than '"initial test delay" + "test duration" + headroom (let's say headroom being max(A (e.g. 10s), number of pods * B (e.g. 0.1s)))

> When timeout happens , it possible that say some pod didnot update result (due to error in pod or could be delay etc). Should we still go ahead with our percentile calculation and return the summary with whatever numbers we have or return error ? my opinion is , we should return whatever result is available.

yes - let's return whathever is available. But also log the pods from which we didn't got results.

hanamantagoudvk: modified the gather() to wait for the response. Regarding logging , i think it will be hard to log about pods which didnot update the metrics. We may have to maintain an additional info about the pairs. and then display which have not sent the response.

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -396,3 +402,21 @@ func (npm *networkPerformanceMeasurement) getMetricData(data *testResultSummary,
 	data.DataItems = append(data.DataItems, metricDataItem)
 	klog.V(3).Infof("TestResultSummary: %v", data)
 }
+
+func (npm *networkPerformanceMeasurement) waitForMetricsFromPods() {
+	bufferTime := time.Duration(math.Max(10, float64(npm.numberOfClients)*podFactor)) * time.Second
+	timeout := initialDelayForTestExecution + npm.testDuration + bufferTime
+	interval := time.Duration(2) * time.Second
```
wojtek-t, line 409: interval := 2 * time.Second

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -396,3 +402,21 @@ func (npm *networkPerformanceMeasurement) getMetricData(data *testResultSummary,
 	data.DataItems = append(data.DataItems, metricDataItem)
 	klog.V(3).Infof("TestResultSummary: %v", data)
 }
+
+func (npm *networkPerformanceMeasurement) waitForMetricsFromPods() {
+	bufferTime := time.Duration(math.Max(10, float64(npm.numberOfClients)*podFactor)) * time.Second
```
wojtek-t, line 407: Let's just inline the "0.1" value here (the podFactor name isn't meaningful.

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -396,3 +402,21 @@ func (npm *networkPerformanceMeasurement) getMetricData(data *testResultSummary,
...
+	timeout := initialDelayForTestExecution + npm.testDuration + bufferTime
+	interval := time.Duration(2) * time.Second
+	if err := wait.Poll(interval, timeout, npm.checkResponseReceivedFromPods); err != nil {
+		klog.Errorf("Failed to receive response from all pods: %v", err)
```
wojtek-t, line 411: Add a TODO to log which pods didn't receive success.

Also - log here how many pods didn't return results.

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
hanamantagoudvk: modified

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
hanamantagoudvk: modified

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
hanamantagoudvk: modified

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -396,3 +399,23 @@ func (npm *networkPerformanceMeasurement) getMetricData(data *testResultSummary,
 	data.DataItems = append(data.DataItems, metricDataItem)
 	klog.V(3).Infof("TestResultSummary: %v", data)
 }
+
+func (npm *networkPerformanceMeasurement) waitForMetricsFromPods() {
+	bufferTime := time.Duration(math.Max(10, float64(npm.numberOfClients)*0.1)) * time.Second
+	timeout := initialDelayForTestExecution + npm.testDuration + bufferTime
+	interval := 2 * time.Second
+	if err := wait.Poll(interval, timeout, npm.checkResponseReceivedFromPods); err != nil {
+		//TODO:log pod names from which response is not received.
```
wojtek-t, line 408: // TODO: Log pod names ...

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -396,3 +399,23 @@ func (npm *networkPerformanceMeasurement) getMetricData(data *testResultSummary,
...
+		klog.Errorf("Failed to receive response from all pods: %v", err)
+		klog.Errorf("Number of pods failed to return results: %v", npm.numberOfClients-len(npm.metricVal))
+	}
+}
+
+func (npm *networkPerformanceMeasurement) checkResponseReceivedFromPods() (bool, error) {
+	npm.metricLock.Lock()
+	defer npm.metricLock.Unlock()
+	if len(npm.metricVal) == npm.numberOfClients {
```
wojtek-t, line 417: return len(npm.metricVal) == npm.numberOfClients, nil

wojtek-t (commented):
Just last two nits - other than that LGTM.

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
hanamantagoudvk: done

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
hanamantagoudvk: done

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -396,3 +399,20 @@ func (npm *networkPerformanceMeasurement) getMetricData(data *testResultSummary,
...
 	klog.V(3).Infof("TestResultSummary: %v", data)
 }
+
+func (npm *networkPerformanceMeasurement) waitForMetricsFromPods() {
+	bufferTime := time.Duration(math.Max(10, float64(npm.numberOfClients)*0.1)) * time.Second
+	timeout := initialDelayForTestExecution + npm.testDuration + bufferTime
+	interval := 2 * time.Second
+	if err := wait.Poll(interval, timeout, npm.checkResponseReceivedFromPods); err != nil {
+		// TODO: Log pod names from which response is not received.
+		klog.Errorf("Failed to receive response from all pods: %v", err)
```
wojtek-t, line 409: Given the check function never returns the error - the err will always be the timeout.

So it's enough to have:

klog.Errorf("Failed to receive response from %v pods", <value>)

hanamantagoudvk: done

wojtek-t:
@hanamantagoudvk - This isn't regression (you should stop adding "kind regression" label to your PRs).
Regression is about fixing regressions.

/lgtm
/approve

PR #1594, opened 2020-11-24
Upgrade prometheus-operator to 0.31.0 and improve requests/limits

marseel:
**What type of PR is this?**
/kind feature

Upgrades prometheus-operator to 0.31.0.
This allows us to use PodMonitor instead of ServiceMonitor.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
@@ -9,18 +9,26 @@ metadata:
   name: k8s
   namespace: monitoring
 spec:
+  logLevel: debug
   baseImage: gcr.io/k8s-testimages/quay.io/prometheus/prometheus
   nodeSelector:
     kubernetes.io/os: linux
   replicas: 1
   resources:
     requests:
+      cpu: {{AddInt 200 (MultiplyInt 500 (DivideInt .Nodes 1000))}}m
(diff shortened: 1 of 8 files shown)


clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -9,19 +9,29 @@ metadata:
...
   namespace: monitoring
 spec:
+  logLevel: debug
   baseImage: gcr.io/k8s-testimages/quay.io/prometheus/prometheus
   nodeSelector:
     kubernetes.io/os: linux
   replicas: 1
   resources:
     requests:
+      cpu: 200m
```
wojtek-t, line 19: Let's set it to the same value as limit.

mm4tt: I'd say:

1. Let's set memory request and limit the same
2. Let's set cpu request depending on number of nodes but don't set the limit. Why? Quoting @thockin: 
    > Unlike memory, CPU is “compressible”. when you put a CPU limit, you artificially restrict access to otherwise idle CPU cycles, forcing things to go slow when they don't need to and causing weird issues like quota exhaustion when using more threads than cores”

marseel: I've set memory requests/limits to same value and set only cpu requests.

PR #1596, opened 2020-11-25
When kubectl exec dig, the dns perf client selector podname is not  correct

yuzhiquan:
**What type of PR is this?**

/kind bug

when kubectl exec dig, the dns perf client selector podname is not  correct

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/dns/py/runner.py
+++ b/dns/py/runner.py
@@ -31,7 +31,7 @@ from params import ATTRIBUTE_CLUSTER_DNS, ATTRIBUTE_NODELOCAL_DNS, Inputs, TestC
 
 _log = logging.getLogger(__name__)
 _app_label = 'app=dns-perf-server'
-_client_podname = 'dns-perf-client'
+_client_label = 'app=dns-perf-client'
 _test_svc_label = 'app=test-svc'
 _dnsperf_qfile_name='queryfile-example-current'
 _dnsperf_qfile_path='ftp://ftp.nominum.com/pub/nominum/dnsperf/data/queryfile-example-current.gz'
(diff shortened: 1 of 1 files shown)


dns/py/runner.py
```diff
@@ -444,13 +439,22 @@ def _wait_for_status(self, active):
       time.sleep(1)
 
     if active:
+      break_loop = False
       while True:
-        code, out, err = self._kubectl(
+        client_pods = self._get_pod('app=dns-perf-client')
+        if client_pods == 0:
```
prameshj, line 445: len(client_pods)?

dns/py/runner.py
```diff
@@ -459,6 +463,16 @@ def _wait_for_status(self, active):
 
       _log.info('DNS is up')
 
+  def _get_pod(self, label, output='custom-columns=:metadata.name'):
+    code, client_pods, err = self._kubectl(None, 'get', 'pods', '-l', label, '--no-headers', '-o',
```
prameshj, line 467: The label selector looks identical to what exists in the code already. "app=dns-perf-client" What is this fixing?

dns/py/runner.py
```diff
@@ -444,13 +439,22 @@ def _wait_for_status(self, active):
...
+      break_loop = False
       while True:
-        code, out, err = self._kubectl(
+        client_pods = self._get_pod('app=dns-perf-client')
+        if client_pods == 0:
+          continue
+        for podname in client_pods:
+          code, out, err = self._kubectl(
             None,
-            'exec', _client_podname, '--',
```
prameshj, line 450: I see the issue now. This part was harcoding client podname as "dns-perf-client". Can you also remove the definition in line 34?

dns/py/runner.py
prameshj: ignore this.. I found the issue in one exec call which was hardcoding client name.

dns/py/runner.py
yuzhiquan: Ops, fix it.

dns/py/runner.py
yuzhiquan: Sure, and define "_client_label".

yuzhiquan:
Gently ping @prameshj

PR #1599, opened 2020-11-26
Test of the adhoc framework

mm4tt:
TEST TEST TEST
DO NOT SUBMIT

/hold


mm4tt:
The test failed because cluster didn't go up due to misconfiguration. Fixed in https://github.com/kubernetes/test-infra/pull/20048

mm4tt:
Looks like it's working:

- test results - https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1599/pull-perf-tests-100-adhoc/1331922270544728064/
- the test dumped `kubectl get nodes` into https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1599/pull-perf-tests-100-adhoc/1331922270544728064/artifacts/node-list.txt

PR #1600, opened 2020-11-26
Fix slow count for api responsiveness for cluster list

marseel:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Slow count for cluster scope lists was calculated incorrectly.
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #1602, opened 2020-11-26
Backport slow count fix to 1.18

marseel:
**What type of PR is this?**
/kind bug

Cherry-pick #1600 
Also, fixes slow count for events in api responsiveness prometheus

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -63,11 +63,11 @@ const (
 	countFastQuery = "sum(increase(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
 
 	// exclude all buckets of 1s and shorter
-	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT|LIST", le="1"`
+	filterGetAndMutating = `resource!="events", verb!~"WATCH|WATCHLIST|PROXY|CONNECT|LIST", le="1"`
 	// exclude all buckets below or equal 5s
-	filterNamespaceList = `scope!="cluster", verb="LIST", le="5"`
+	filterNamespaceList = `resource!="events", scope!="cluster", verb="LIST", le="5"`
 	// exclude all buckets below or equal 30s
-	filterClusterList = `scope="cluster", verb="LIST", le="30"`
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -63,11 +63,11 @@ const (
 	countFastQuery = "sum(increase(apiserver_request_duration_seconds_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
 
 	// exclude all buckets of 1s and shorter
-	filterGetAndMutating = `verb!~"WATCH|WATCHLIST|PROXY|CONNECT|LIST", le="1"`
+	filterGetAndMutating = `resource!="events", verb!~"WATCH|WATCHLIST|PROXY|CONNECT|LIST", le="1"`
```
wojtek-t, line 66: Why you're readding events?

marseel: readding?
I've removed events, because they were not counted in 1.18. Check line https://github.com/kubernetes/perf-tests/pull/1602/files#diff-d4933bdf320dd42179e6f8ef4c50c4ff7c2aa5a5b35f179278e5adbd345ae8ebR47

marseel: This was the reason why SlowCount was negative in our tests. SLO for events was enabled only in 1.19.

wojtek-t: I mean readding excluding events.

But my question is - why this can reasult in negative slo count? events should be orthogonal

mm4tt: This is because line 47:
```
filters = `resource!="events", verb!~"WATCH|WATCHLIST|PROXY|CONNECT"`
``` 
needs to match this line. 

We were not counting events there but counting here, hence negative count

marseel: Slow count is computed as WholeCount - FastCount. for WholeCount (line I posted above) we exclude events. So when object in golang is created default value is 0. Then we subtract event from FastCount and that's the reason why it's negative.

jkaniuk:
100 and 5k node tests are green on master.

/ok-to-test
/lgtm
/approve

PR #1623, opened 2020-11-30
Make run-e2e.sh print commit history

mm4tt:


mm4tt:
/hold

Wait for https://github.com/kubernetes/test-infra/pull/20062

mborsz:
I guess this should happen as part of bootstrap.py where we fetch repos and write infra-commit and other fields in finished.json

mm4tt:
> I guess this should happen as part of bootstrap.py where we fetch repos and write infra-commit and other fields in finished.json

Yeah, I considered that, but bootstrap is deprecated and we'll have a pushback making changes there. We should migrate to pod-utils, but that's a bigger project (e.g. we'll have to rewrite log-exporter). This is a hacky solution to fix the problem for now. I'll add a TODO to get rid of this once we migrate to pod-utils.

Issue #1619, opened 2020-11-27
CL2 image-preload feature doesn't work with containerd CRI


hanamantagoudvk:
**What happened**:

node-preload-images option fails to pre-load the images to all the nodes in the cluster.

**What you expected to happen**:

When node-preload-images option is used , it should pre-load the images before starting the testcases.

**How to reproduce it (as minimally and precisely as possible)**:

There are two ways to reproduce this problem:

1) By directly applying daemonset yaml file (with any image-name hardcoded inside).

kubectl apply -f /home/k8s/go/src/k8s.io/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml

2) When we run clusterloader2 testcases as shown below

a) 

/home/k8s/go/src/k8s.io/perf-tests/clusterloader2/run-e2e.sh --provider=local --kubeconfig=$HOME/.kube/config --testsuite=/home/k8s/go/src/k8s.io/perf-tests/clusterloader2/testing/network/suite.yaml 
--node-preload-images=vivekthrivikraman/netperfbenchmark:0.01

b) vivekthrivikraman/netperfbenchmark:0.01 image is present in docker hub and we expect it to get downloaded using "docker pull vivekthrivikraman/netperfbenchmark:0.01" command

c) But we see the following error with preload pods 

root@k8s:~# kubectl get pods --all-namespaces
preload              preload-d4nqt                                0/1     Init:CrashLoopBackOff   3          64s
preload              preload-hgvzc                                0/1     Init:CrashLoopBackOff   3          64s
preload              preload-ngkt5                                0/1     Init:CrashLoopBackOff   3          64s
root@k8s:~#

root@k8s:~# kubectl describe pod preload-d4nqt -n preload
Name:         preload-d4nqt
Namespace:    preload
Priority:     0
Node:         kind-worker2/172.18.0.5
Start Time:   Fri, 27 Nov 2020 19:50:24 +0530
Labels:       controller-revision-hash=6c7bc9d988
              name=preload
              pod-template-generation=1
Annotations:  cni.projectcalico.org/podIP: 10.168.110.147/32
              cni.projectcalico.org/podIPs: 10.168.110.147/32
              k8s.v1.cni.cncf.io/network-status:
                [{
                    "name": "",
                    "ips": [
                        "10.168.110.147"
                    ],
                    "default": true,
                    "dns": {}
                }]
              k8s.v1.cni.cncf.io/networks-status:
                [{
                    "name": "",
[...]

PR #1625, opened 2020-12-02
Image preload test

mm4tt:
Repro for https://github.com/kubernetes/perf-tests/issues/1619

DO NOT SUBMIT
/hold

### DIFF
--- a/clusterloader2/run-e2e.sh
+++ b/clusterloader2/run-e2e.sh
@@ -34,4 +34,4 @@ if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
 fi
 
 cd ${CLUSTERLOADER_ROOT}/ && go build -o clusterloader './cmd/'
-./clusterloader --alsologtostderr --v=2 "$@"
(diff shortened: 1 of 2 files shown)


clusterloader2/run-e2e.sh
```diff
@@ -34,4 +34,4 @@ if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
 fi
 
 cd ${CLUSTERLOADER_ROOT}/ && go build -o clusterloader './cmd/'
-./clusterloader --alsologtostderr --v=2 "$@"
+./clusterloader --alsologtostderr --v=2 --node-preload-images=gcr.io/k8s-testimages/probes:v0.0.4 "$@"
```
wojtek-t, line 37: Can we make the version of the image from somewhere else (so that it won't get outdated with the nearest bump of it)?

mm4tt: Wojtek, this is a test PR just to trigger and debug the image-preload. I'm not going to submit this.

PR #1626, opened 2020-12-02
Clean unused permissions for prometheus

marseel:
**What type of PR is this?**
/kind cleanup

Cleans up role and role bindings for prometheus. Currently prometheus already has cluster scope permissions, so we don't need to worry about namespaced permissions.

Fixes #

```release-note

```

```docs

```


marseel:
Looks like tests are passing, also I haven't seen any errors in prometheus logs so I think it's good to merge. 
FRT here are cluster level permissions for prometheus, which are giving more permissions than namespaced roles.
Here is list of cluster level permissions to compare:  https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/manifests/0prometheus-operator-clusterRole.yaml

PR #1628, opened 2020-12-07
Measure pod startup for stateless pods

wojtek-t:
This is a step towards getting rid of latency phase from our test.
Pod startup from "CreatePhasePodStartupLatency":
```
    {
      "data": {
        "Perc50": 2203.834508,
        "Perc90": 2990.742296,
        "Perc99": 12677.203517
      },
      "unit": "ms",
      "labels": {
        "Metric": "pod_startup"
      }
    },
```
Pod startup from "Stateles CreatePhasePodStartupLatency":
```
    {
      "data": {
        "Perc50": 2199.278227,
        "Perc90": 2953.810973,
        "Perc99": 4251.361823
      },
      "unit": "ms",
      "labels": {
        "Metric": "pod_startup"
      }
    }
```

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -54,6 +54,7 @@ func createPodStartupLatencyMeasurement() measurement.Measurement {
 	return &podStartupLatencyMeasurement{
 		selector:          measurementutil.NewObjectSelector(),
 		podStartupEntries: measurementutil.NewObjectTransitionTimes(podStartupLatencyMeasurementName),
+		podMetadata:       measurementutil.NewPodsMetadata(podStartupLatencyMeasurementName),
 	}
 }
 
@@ -62,6 +63,7 @@ type podStartupLatencyMeasurement struct {
 	isRunning         bool
 	stopCh            chan struct{}
 	podStartupEntries *measurementutil.ObjectTransitionTimes
+	podMetadata       *measurementutil.PodsMetadata
 	threshold         time.Duration
 }
 
@@ -162,6 +164,11 @@ func podStartupTransitionsWithThreshold(threshold time.Duration) map[string]meas
 	return result
 }
 
+type podStartupLatencyCheck struct {
+	namePrefix string
+	filter     measurementutil.KeyFilterFunc
+}
+
 func (p *podStartupLatencyMeasurement) gather(c clientset.Interface, identifier string) ([]measurement.Summary, error) {
 	klog.V(2).Infof("%s: gathering pod startup latency measurement...", p)
 	if !p.isRunning {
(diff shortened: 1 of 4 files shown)


mborsz:
I see two potential approaches for this:
1. Implementing distinction between stateless and stateful pods on measurement layer (like in this PR)
2. Marking all stateful pods with some label on config level and starting two separate measurements: for stateful pods and for stateless pods

I can see arguments for both approaches. @wojtek-t could you comment why this approach vs the other one?

wojtek-t:
> I can see arguments for both approaches. @wojtek-t could you comment why this approach vs the other one?

1. We want this to be generic - I don't want to force every test owner to think about adding some labels (or something else). SLO measurements should eventually be enabled everywhere and work out-of-the-box for every test.

2. The definition of stateless/stateful pods is actually part of the SLO:
https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md
 So given we need to define it there, it's no reason to require the user to pass any additional information.

@mborsz - PTAL

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -140,42 +181,36 @@ func (p *podStartupLatencyMeasurement) gather(c clientset.Interface, identifier
...
 	var err error
-	if slosErr := podStartupLatency["pod_startup"].VerifyThreshold(p.threshold); slosErr != nil {
-		err = errors.NewMetricViolationError("pod startup", slosErr.Error())
-		klog.Errorf("%s: %v", p, err)
-	}
+	for _, check := range checks {
+		transitions := podStartupTransitionsWithThreshold(p.threshold)
+		podStartupLatency := p.podStartupEntries.CalculateTransitionsLatency(transitions, check.filter)
+
+		if slosErr := podStartupLatency["pod_startup"].VerifyThreshold(p.threshold); slosErr != nil {
```
mborsz, line 201: I think that stateful pods should use other threshold (and this validation should be disabled by default everywhere to make it possible to rollout it gradually)

mborsz:
> > I can see arguments for both approaches. Wojciech Tyczynski could you comment why this approach vs the other one?
> 
> 1. We want this to be generic - I don't want to force every test owner to think about adding some labels (or something else). SLO measurements should eventually be enabled everywhere and work out-of-the-box for every test.
> 2. The definition of stateless/stateful pods is actually part of the SLO:
>    https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md
>    So given we need to define it there, it's no reason to require the user to pass any additional information.
> 
> Maciej Borsz - PTAL

Makes sense to me.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
wojtek-t: Note that in this PR I'm not really executing the stateful pods at all.
What I'm doing is:
- I'm leaving the existing "PodStartupLatency" as it was (includes all pods)
- I'm creating a new "StatelessPodsStartupLatency" (to include only stateless)

They both still use the same threshold at this point.


Once this soaks for a moment, we will create a dedicated StatefulPodsStartupLatency (yes - with a different treshold) and then sunset the original "PodStartupLatency". But both of these are further steps. I first want to make it more generic and see what we have for stateless explicitly.

Does that make sense?

mborsz: The new StatelessPodsStartupLatency can still generate metric violation and fail the test, but given that this is a subset of podstartuplatency pods which have the same property, we should be fine (unless we have some super fast stateful pods that reduces p99 of pod startup latency which is rather unlikely)

clusterloader2/pkg/measurement/util/pods_metadata.go
```diff
@@ -0,0 +1,63 @@
...
+		meta.stateless = stateless
+		return
+	}
+	o.metadata[key] = &podMetadata{
+		stateless: stateless,
+	}
+}
+
+func (o *PodsMetadata) FilterStateless() KeyFilterFunc {
+	return func(key string) bool {
```
mborsz, line 57: Style nit: why do you wrap this in func()?

Can't you simply do: p.podMetadata.FilterStateless?
func (o *PodsMetadata) FilterStateless(key string) bool {
		o.lock.Lock()
		defer o.lock.Unlock()
		meta, ok := o.metadata[key]
		return ok && meta.stateless
}

And then, use it as '

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
wojtek-t: OK yeah - there is an assumption that stateful pods don't start super fast. They barely ever fit into 5s,  so the stateless one shouldn't be higher in any of our existing tests.

Do you suggest setting it artificially higher now? Or are you ok as is for now?
[Note that we use 5s now only for the latency phase, where all pods are stateless anyway :) ]

mborsz (commented):
One nit, other than that LGTM

clusterloader2/pkg/measurement/util/pods_metadata.go
wojtek-t: Done

wojtek-t (commented):
@mborsz - comment applied; PTAL

PR #1632, opened 2020-12-11
WIP PoC of CL2 modules

mm4tt:

PR #1634, opened 2020-12-14
ClusterLoader2 Module API

mm4tt:
If you want to see it in action please take a look at this PoC PR  - https://github.com/kubernetes/perf-tests/pull/1632


clusterloader2/api/extensions.go
```diff
@@ -55,3 +55,13 @@ func (d *Duration) ToTimeDuration() time.Duration {
...
 }
+
+// IsPhase returns true whether a step is a phase-step.
+func (s *Step) IsPhase() bool {
+	return len(s.Phases) > 0
+}
+
+// IsModule returns true whether a step is a module.
+func (s *Step) IsModule() bool {
+	return len(s.Module.Steps) > 0 || s.Module.Path != ""
```
tosi3k, line 66: I'm slightly baffled by this condition. Why not simply `s.Module.Path != ""`?

In addition, this contradicts the logic included in the proposed `loadModule` function where we currently run a following `if` statement:
```
if len(module.Steps) > 0 {
	return nil, fmt.Errorf("module %v has both path and steps set", module)
}
```

clusterloader2/api/types.go
```diff
@@ -54,20 +54,41 @@ type Config struct {
...
+// Exactly one field (Phases or Measurements or Module) should be non-empty.
 type Step struct {
 	// Phases is a collection of declarative definitions of objects.
 	// Phases will be executed in parallel.
 	Phases []Phase `json:"phases"`
 	// Measurements is a collection of parallel measurement calls.
 	Measurements []Measurement `json:"measurements"`
-	// Name is an optional name for given step. If name is set,
-	// timer will be run for the step execution.
+	// Module is a group of steps, usually defined in a separate file.
```
tosi3k, line 68: Nit: `sequence` would probably be a better word here than `group`.

clusterloader2/api/types.go
```diff
@@ -54,20 +54,41 @@ type Config struct {
...
+	Module Module `json:"module"`
+	// Name is an optional name for given step. If name is set the step execution
+	// time will be measured and the step will be reported at the end of the test.
+	// The name is ignored if the step is of type 'Module'.
 	Name string `json:"name"`
 }
 
+// Module is a group of steps, usually defined in a separate file.
```
tosi3k, line 76: Ditto about wording.

clusterloader2/api/types.go
```diff
@@ -54,20 +54,41 @@ type Config struct {
...
+type Module struct {
+	// Path is the path to the filename with the module template.
+	// If the path is set the the Steps field will get overridden with the content
```
tosi3k, line 79: Nit: double `the`.

tosi3k (changes requested):

clusterloader2/api/types.go
```diff
@@ -54,20 +54,41 @@ type Config struct {
...
 }
 
-// Step represents encapsulation of some actions. These actions could be
-// object declarations or measurement usages.
-// Exactly one field (Phases or Measurements) should be non-empty.
+// Step represents a unit of work in ClusterLoader2. It can be either:
+// - a collection of measurements,
+// - a collection of phases,
+// - a group of other steps (module).
+// Exactly one field (Phases or Measurements or Module) should be non-empty.
```
marseel, line 61: Correct me If I'm wrong, but I don't see a place where this behaviour is validated (Making sure that only one of 3 is non-empty).  What I propose is replacing IsMeasurement/IsPhase/IsModule with enum + method GetType that also validates if this requirement is fulfilled (or some similar solution). What do you think about it?

clusterloader2/api/types.go
```diff
@@ -54,20 +54,41 @@ type Config struct {
...
+	// of the file.
+	Path string `json:"path"`
+	// Params specifies template parameters to be substituted inside the template.
+	Params map[string]interface{} `json:"params"`
+
+	// Steps is the list of steps composing the module. Steps are executed
+	// serially. While it's possible to set the steps directly via this field,
+	// usually a cleaner approach is to have the template definition inside a
+	// separate file and instantiate it via setting the Path/Params fields.
+	Steps []Step `json:"steps"`
```
wojtek-t, line 89: I don't understand this one (and I'm actually against having Steps here).
If you want to define Steps, you can simply define them at the higher level and you don't need module.

Having steps here make it possible to express the same thing in many different ways, which will always be misleading to users.

Let's stick to just a path and params.

wojtek-t (commented):
/hold

clusterloader2/api/types.go
mm4tt: From a physical point of view Module is a sequence of steps. That's why we have []Steps here. What do you not understand? 

I think the question is whether we should allow initializing modules via setting the Steps directly or just via Path/Params. I don't have a strong opinion here. Current code allows that and it's explained in the comment. 
There might be a value in allowing that in the future. E.g. imagine having conditional modules that would allow to group bunch of steps and disable/enable them depending on some condition. 

So I think it boils down to whether we want to support inline modules (now or in the future). Thoughts?

mborsz: So I understand that 'Steps' field here is an implementation detail and should not be set by the end user?

I can't see any value from 'inline modules', I wouldn't make it possible to set this.

How about removing this field completely?

clusterloader2/README.md
```diff
@@ -44,6 +44,16 @@ Templates for test definition come with one predefined value - ```{{.Nodes}}```,
 which represents the number of schedulable nodes in the cluster. \
 Example of a test definition can be found here: [load test].
 
+### Modules
+
+ClusterLoader2 supports modularization of the test configs via the [Module API](https://github.com/kubernetes/perf-tests/blob/1bbb8bd493e5ce6370b0e18f3deaf821f3f28fd0/clusterloader2/api/types.go#L77).
+With the Module API, you can divide a single test config file into multiple 
+module files. A module can be parametrized and used multiple times by
```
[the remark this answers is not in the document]
mm4tt: Done.

clusterloader2/api/extensions.go
mm4tt: Obsolete.

clusterloader2/api/types.go
mm4tt: Done.

clusterloader2/api/types.go
mm4tt: Done.

clusterloader2/api/types.go
mm4tt: Obsolete.

clusterloader2/api/types.go
mm4tt: Ok, if two people are baffled by this inline modules then it's clearly not a good idea. I changed the code to not allow inline modules. PTAL

clusterloader2/api/types.go
mm4tt: We don't validate it directly rather check during execution. Here - https://github.com/kubernetes/perf-tests/blob/f0c628601e7721063a28dda50df422a90cabb446/clusterloader2/pkg/test/simple_test_executor.go#L169 and in the code that flattens module steps where we process all "module" steps.

I think it'd be better if we validated the config before we start executing it. It will result in a cleaner code and better user experience (test will fail fast, not in the middle of Xh run). But it's a separate issue. 

Introducing an enum would be a backward incompatible change, so probably not a good way to go.

mm4tt: @marseel, would you mind opening an issue for this (cl2 config validation) ?

clusterloader2/pkg/test/simple_test_executor.go
```diff
@@ -117,14 +117,19 @@ func (ste *simpleExecutor) ExecuteTestSteps(ctx Context, conf *api.Config) *erro
 		return errors.NewErrorList(fmt.Errorf("automanaged namespaces creation failed: %v", err))
 	}
 
+	steps, err := flattenModuleSteps(ctx, conf.Steps)
```
mborsz, line 120: Nit: For the random code reader (like me), it's not clear why conf.Steps up to this point had type '[]api.Step', but after this line we start working on '[]*api.Step'.

I think it's error-prone that for loop below expects pointers, while half of this function uses pointers and half of this function uses objects directly.

Can we keep []api.Step everywhere (i.e. change return type for flattenModuleSteps)?

I guess, some day we can remove the flattenModuleSteps, without modifying loop below and we will have a bug in code.

clusterloader2/api/types.go
```diff
@@ -54,20 +54,41 @@ type Config struct {
...
+type ModuleRef struct {
+	// Path is the path to the filename with the module template.
+	Path string `json:"path"`
+	// Params specifies template parameters to be substituted inside the template.
+	Params map[string]interface{} `json:"params"`
+}
+
+// Module is a structure with the definition of a CL2 module. Conceptually, a
+// module is a sequence of steps.
+type Module struct {
```
wojtek-t, line 86: This shouldn't be part of the API - if you really need that, we can consider making it an internal structure somewhere.

mborsz: This is the format of yaml file with module. The user must know what format is expected. For me it looks like definitely part of API, not some internal structure.

wojtek-t: Ahh ok - for module definition. Yeah - that makes sense.

clusterloader2/pkg/test/modules.go
```diff
@@ -0,0 +1,74 @@
...
+// the input is:
+// - step 1:
+// - step 2:
+//   module:
+//     - step 3:
+//     - step 4:
+//       module:
+//         -step 5
+// - step 6
+// the method will return [step1, step3, step5, step6].
```
wojtek-t, line 38: I don't understand this:
- where is step2 ?
- where is step4 ?

BTW - it would be good to accomodate the comment to the new API.

tosi3k: `step 2` and `step 4` are modules and therefore they're formally steps that are not executable but wrap some other steps (at least that's my understanding here).

clusterloader2/api/types.go
marseel: Sure, I've opened it. It's exactly what I had in mind - cleaner code and better user experience.

mm4tt: Thanks!

clusterloader2/pkg/test/simple_test_executor.go
mm4tt: Would it change anything in terms of readability? We're operating on pointers in the methods below the call stack (e.g. executeStep). I'm just making the conversion step before. It's reasonable to operate on pointers from the beginning as we're avoiding unnecessary copying of steps (which we'd have if flattenModuleSteps returned []api.Step).

I don't buy the argument about the bug. If you made that change the code wouldn't compile.

clusterloader2/pkg/test/modules.go
mm4tt: Yes, Antek explained it well. 

Changed the comment.

wojtek-t:
/hold cancel

I'm fine with the direction - I will let others review details.

clusterloader2/pkg/test/simple_test_executor.go
mborsz: Makes sense. I haven't noticed that ExecuteStep expects pointer. In this case it will stop compiling. Otherwise, the code below would compile perfectly well, but will work incorrectly.

mborsz:
/lgtm
/hold

LGTM for me

jkaniuk:
LGTM for me

PR #1643, opened 2020-12-17
Allow to turn off logging for probes to file.

marseel:
**What type of PR is this?**
/kind feature

This PR allows us to turn off logging for probes to host path and instead print it to stderr.

Fixes #

```release-note

```

```docs

```


mborsz:
As discussed offline, I think that your purpose, disabling the probes completely will work better than adjusting it to your use case (there is no point in maintaining multiple probe configuration, given that you don't need this measurement at all in your use case).

wojtek-t:
Closing - I think we decided for a different approach.

PR #1648, opened 2021-01-07
Snapshot prometheus disk even if we failed to tear down prometheus

mborsz:
It's not clear what would be a value of the disk in case when we do not stop prometheus cleanly before taking snapshot, but hopefully this will give us at least partial data.

Without this change, we do not take snapshot of the disk when kube-apiserver is down, e.g. in https://github.com/kubernetes/kubernetes/issues/97801#issuecomment-756080299


mborsz:
/hold
I want to manually double check that it works in presubmit.

mborsz:
/hold cancel
Looks good

PR #1651, opened 2021-01-12
Add node local dns latency to load test

marseel:
**What type of PR is this?**
/kind feature
/kind bug

Adds measuring of NodeLocalDNSLatency to load test.
Fixes metric name used in NodeLocalDNSLatency.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go
@@ -29,7 +29,7 @@ import (
 
 const (
 	nodelocaldnsLatencyPrometheusMeasurementName = "NodeLocalDNSLatencyPrometheus"
-	percLatencyQueryTemplate                     = `histogram_quantile(%v, sum(rate(http_request_duration_seconds_bucket[%v])) by (le))`
+	percLatencyQueryTemplate                     = `histogram_quantile(%v, sum(rate(coredns_dns_request_duration_seconds_bucket[%v])) by (le))`
 	defaultLatencyUpperBoundSeconds              = 5.0
 )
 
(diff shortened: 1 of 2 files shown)


wojtek-t:
/lgtm
/approve

Thanks!

/hold
@prameshj - please cancel hold if that looks fine to you

clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go
```diff
@@ -29,7 +29,7 @@ import (
 
 const (
 	nodelocaldnsLatencyPrometheusMeasurementName = "NodeLocalDNSLatencyPrometheus"
-	percLatencyQueryTemplate                     = `histogram_quantile(%v, sum(rate(http_request_duration_seconds_bucket[%v])) by (le))`
+	percLatencyQueryTemplate                     = `histogram_quantile(%v, sum(rate(coredns_health_request_duration_seconds_bucket[%v])) by (le))`
```
prameshj, line 32: The metric name is "kubernetes.io/internal/nodes/coredns/dns_request_duration_seconds". Should we use "coredns_dns_request_duration_seconds"?

marseel: I've changed it to "coredns_dns_request_duration_seconds"
Yesterday, we discovered that it was not working due to #1652 
I am running one more test to make sure it works with current setup.

marseel: I think we can merge as it is right now, but from my initial test I can see that some requests end up in bucket inf+, which results in 99 percentile to be ~8s (two highest buckets are 8s and +inf) so this test fails.

prameshj: okay, thanks for trying it out. The threshold value can be overridden by commandline if needed and we can make tweaks to kube-dns replica count to measure how much of a scale-up is needed to achieve 5s. Do you know what the 50th percentile and 90th percentile values were in your test?

wojtek-t:
@marseel - please squash commits and it seems we're good to go

marseel:
I've squashed it, although I think two commits made sense before - one with bugfix, one with feature ;)

wojtek-t:
/lgtm
/hold cancel

> I've squashed it, although I think two commits made sense before - one with bugfix, one with feature ;)

Yeah - that's true. That made sense too (didn't think about that)

PR #1652, opened 2021-01-12
Use ClusterFirst dnsPolicy in workloads.

prameshj:
"ClusterFirst" is the default dns option. "Default" inherits the node config.
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

We could put this in an if block in case Default dnsPolicy is needed in other testcases.

 For the dns tests, we want ClusterFirst to be used in order to test cluster DNS solution.

**What type of PR is this?**

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/deployment.yaml
+++ b/clusterloader2/testing/load/deployment.yaml
@@ -50,7 +50,6 @@ spec:
             mountPath: /var/configmap
           - name: secret
             mountPath: /var/secret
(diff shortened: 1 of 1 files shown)


wojtek-t:
We didn't rely on default anywhere according to my knowledge.

/lgtm
/approve

thanks!

PR #1654, opened 2021-01-13
Use containerd in image preloader

lyzs90:
**What type of PR is this?**

/kind bug

Details in https://github.com/kubernetes/perf-tests/issues/1619

Part of https://github.com/kubernetes/perf-tests/issues/1619

Retained the docker volume mount but gating logic to fallback to docker is not implemented yet. Should be doable by updating framework / image preload.

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml
+++ b/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml
@@ -15,16 +15,21 @@ spec:
       initContainers:
       {{range $idx, $image := .Images}}
       - name: preload-{{$idx}}
-        image: docker
-        command: ["docker", "pull", "{{$image}}"]
+        image: gcr.io/k8s-testimages/perf-tests-util/containerd:v0.0.1
+        command: ["sh", "-c", "ctr -n=k8s.io image pull {{$image}} 2>&1 | tee /var/log/cl2-image-preload-{{$idx}}.log"]
         volumeMounts:
-        - name: docker
-          mountPath: /var/run
+        - name: containerd
+          mountPath: /run/containerd
+        - name: logs-volume
+          mountPath: /var/log
       {{end}}
       volumes:
-      - name: docker
+      - name: containerd
         hostPath:
-          path: /var/run
+          path: /run/containerd
+      - name: logs-volume
+        hostPath:
+          path: /var/log
       containers:
       - name: pause
         image: gcr.io/google_containers/pause


util-images/containerd/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM docker:latest
```
mm4tt, line 1: Do we need docker in this image?

lyzs90: Oh didn't we want to maintain backward compatibility with docker?

mm4tt: Nah, let's make it work only for containerd. It's already a default and docker is being deprecated. If people want to use this feautre with docker they can use older Cl2 release. In case there is a need to support both docker and containerd we can always add some gating logic in the daemonset to use either docker image & docker command or the new containerd image and ctr command. An image should support only one solution, either ctr or docker.

mm4tt (commented):
This is great! One ask, can we move the dockerfile to a separate PR? We'll submit it first and I'll push the image to the registry. Then, we'll be able to test the second part

util-images/containerd/Dockerfile
lyzs90: Got it, if that's the case I'll use a minimal base image and create a separate PR for that

mm4tt:
Please rebase and then let's test it by making the changes in the run-e2e.sh like these - https://github.com/kubernetes/perf-tests/pull/1625

lyzs90:
@mm4tt done, i set it to latest tag. let me know if you want to pass an env var instead

lyzs90:
@mm4tt do you have a way to check the logs on those nodes?

mm4tt:
Oh, right. Could you modify the command to also dump logs to var/logs as I did in the repro - https://github.com/kubernetes/perf-tests/pull/1625/files ?

mm4tt:
It fails with - https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1654/pull-perf-tests-clusterloader2/1349722543606468608/artifacts/e2e-1654-62db2-minion-group-00ml/cl2-image-preload-0.log
```
ctr: failed to resolve reference "gcr.io/k8s-testimages/probes": object required
```

lyzs90:
> It fails with - https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1654/pull-perf-tests-clusterloader2/1349722543606468608/artifacts/e2e-1654-62db2-minion-group-00ml/cl2-image-preload-0.log
> 
> ```
> ctr: failed to resolve reference "gcr.io/k8s-testimages/probes": object required
> ```

not sure if it's because it's a private registry? i'll try to set up a minimal repro on a VM

mm4tt:
> > It fails with - https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1654/pull-perf-tests-clusterloader2/1349722543606468608/artifacts/e2e-1654-62db2-minion-group-00ml/cl2-image-preload-0.log
> > ```
> > ctr: failed to resolve reference "gcr.io/k8s-testimages/probes": object required
> > ```
> 
> not sure if it's because it's a private registry? i'll try to set up a minimal repro on a VM

It looks like the `ctr image pull` requires both name of the image and tag. Repro:
```
mmatejczyk@gke-containerd-test-default-pool-181818c8-4bg6 ~ $ ctr image pull gcr.io/k8s-testimages/probes
ctr: failed to resolve reference "gcr.io/k8s-testimages/probes": object required
mmatejczyk@gke-containerd-test-default-pool-181818c8-4bg6 ~ $ ctr image pull gcr.io/k8s-testimages/probes:latest
gcr.io/k8s-testimages/probes:latest:                                              resolved       |++++++++++++++++++++++++++++++++++++++| 
manifest-sha256:bf80b8c9bbf4fa986012b48dad0bde8252e10f87c412e0af2e89272d720f3ad6: done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:e2bc2b9f5790da665d3f19ab886da7b2def5ea19ce2ad228d7f33284155ef39b:    done           |++++++++++++++++++++++++++++++++++++++| 
config-sha256:e49e62e89c2475a39c6dc40e96c14853f52dee1648237db663c2b598989caeb2:   done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:8e402f1a9c577ded051c1ef10e9fe4492890459522089959988a4852dee8ab2c:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:ce7779d8bfe3415e27ec3bf5950b2ab67a854c608595f0f2e49066fb5806fd12:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:de1a1e452942df2228b914d2ce9be43f18b137f4ebc3dce9491bc08c2630a2ea:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:a7ffcdf77d94827be5c9826e80d73ce8ddc9c6fa619e70abd99e6aebcd6551ab:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:e7348c6471e5ee4e6f419ee029a40c7352c65f64e76a7885380436d66149420e:    done           |++++++++++++++++++++++++++++++++++++++| 
layer-sha256:69f2c21b6694e6080c1ee40d4aefd49461a5859c0c2f47924953249cdc23046c:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 6.6 s                                                                    total:  128.3  (19.4 MiB/s)                                      
unpacking linux/amd64 sha256:bf80b8c9bbf4fa986012b48dad0bde8252e10f87c412e0af2e89272d720f3ad6...
done
```

lyzs90:
ah 👍  let me slap on the latest tag then

clusterloader2/pkg/imagepreload/manifests/daemonset.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       initContainers:
       {{range $idx, $image := .Images}}
       - name: preload-{{$idx}}
-        image: gcr.io/k8s-testimages/perf-tests-util/containerd
+        image: gcr.io/k8s-testimages/perf-tests-util/containerd:latest
```
wojtek-t, line 18: Let's don't use latest - this means that it can change behind the scenes and change behavior of our tests without us knowing anything potentially.

Please use a sepecific tag (IIRC Matt pushed 0.1)

clusterloader2/run-e2e.sh
```diff
@@ -34,4 +34,4 @@ if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
 fi
 
 cd ${CLUSTERLOADER_ROOT}/ && go build -o clusterloader './cmd/'
-./clusterloader --alsologtostderr --v=2 "$@"
+./clusterloader --alsologtostderr --v=2 --node-preload-images=gcr.io/k8s-testimages/probes:latest "$@"
```
mm4tt, line 37: Let's bump verbosity to `--v=4` to debug what fails now. It looks that ctr pulls images on the nodes, but CL2 time-outs.

lyzs90: `W0115 02:29:06.644] I0115 02:29:06.644125   15153 imagepreload.go:133] 0 out of 100 nodes have pulled images`

did the node logs show that the images were pulled successfully?

mm4tt: Yes, e.g. node - https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1654/pull-perf-tests-clusterloader2/1349902397828239360/artifacts/e2e-1654-62db2-minion-group-00c3/cl2-image-preload-0.log

It's either that everything works but kubelet needs more time to realize the new image has been pulled and to update the node image or that something doesn't work as it should, e.g. the image is persisted on host or something like that. Let's bump this timeout to 20min to check whether it;s a timeout issue - https://github.com/kubernetes/perf-tests/blob/203cd2740c9eaf8d563998a972d842b1cd0b560f/clusterloader2/pkg/imagepreload/imagepreload.go#L45

lyzs90: same issue. shall we add some v4 logs in `hasPreloadedImages`? i'm keen to see what's in `nodeImages`

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -165,6 +164,7 @@ func (c *controller) hasPreloadedImages(node *v1.Node) bool {
 	for _, nodeImg := range node.Status.Images {
 		nodeImages = append(nodeImages, nodeImg.Names...)
 	}
+	klog.V(3).Infof("Found the following images on node: %s", nodeImages)
```
mm4tt, line 167: Hmmm, I don't see any logs for this line. That means we didn't get any node update events. We can exclude the situation where there were no node updates as the node-problem-detectors updates the nodes periodically. The only explanation is we set up the watch/informer incorrectly. Let me dig into this. 

In general, I'm not sure if this code was ever tested, so the bug can be anywhere.

mm4tt: Oh, the logs ARE there - https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1654/pull-perf-tests-clusterloader2/1350026427449217024/

I was looking at the old run. That means the ctr pull image in the container is not updating the images on the host. Let me play a bit with that.

clusterloader2/pkg/imagepreload/manifests/daemonset.yaml
```diff
@@ -15,16 +15,21 @@ spec:
       initContainers:
       {{range $idx, $image := .Images}}
       - name: preload-{{$idx}}
-        image: docker
-        command: ["docker", "pull", "{{$image}}"]
+        image: gcr.io/k8s-testimages/perf-tests-util/containerd:v0.0.1
+        command: ["sh", "-c", "ctr image pull {{$image}} 2>&1 | tee /var/log/cl2-image-preload-{{$idx}}.log"]
```
mm4tt, line 19: I think I got it. We need to pull the image in the k8s.io namespace for the kubelet to update the node's image list. So, the command should be
```
ctr -n=k8s.io image pull ...
```

Could you change it?

lyzs90: done. are containerd namespaces equivalent to k8s namespaces?

mm4tt: Depending on what you mean by equivalent :)
The idea is the same - provide logical grouping of objects. But there is no mapping between k8s namespace and containerd namespaces. All images used in kubernetes are simply put into k8s.io containerd namespace.

lyzs90: > All images used in kubernetes are simply put into k8s.io containerd namespace.

ah, saw that in the main repo containerd package. it's clear to me now 👍

lyzs90:
```
Found the following images on node: [gcr.io/k8s-testimages/probes:latest gcr.io/k8s-testimages/perf-tests-util/containerd@sha256:5ea41e1e29801ff8d07865c862f13a380a838f6e94b90f529aca7343e087e705 gcr.io/k8s-testimages/perf-tests-util/containerd:v0.0.1 k8s.gcr.io/kube-proxy-amd64:v1.21.0-alpha.1.63_083290d996cd07 k8s.gcr.io/prometheus-to-sd@sha256:14666989f40bb7c896c3e775a93c6873e2b791d65bc65579f58a078b7f9a764e k8s.gcr.io/prometheus-to-sd:v0.5.0 k8s.gcr.io/metadata-proxy@sha256:e914645f22e946bce5165737e1b244e0a296ad1f0f81a9531adc57af2780978a k8s.gcr.io/metadata-proxy:v0.1.12 k8s.gcr.io/pause@sha256:927d98197ec1141a368550822d18fa1c60bdae27b78b0c004f705f548c07814f k8s.gcr.io/pause:3.2]
```
🎉

mm4tt: Nice! Let's revert then all the changes except daemonset.yaml and squash the commits into one and submit :)

lyzs90: cool its done. when / where will we actually use the image preloader?

mm4tt: We'll use it as soon as it's implemented to improve our scale tests. One place is kubemark which currently has no node-images which makes the size of the node object in kubemark completely not realistic (<<1KB). Another place are the regular scale tests where there are so node images but very few. In the end, the node object size there is ~1KB while in production clusters it's usually >10KB. It has an impact of the system performance and it's a gap in our e2e tests that we should close.

mm4tt:
This is great! Thank you so much!

/lgtm
/approve

wojtek-t:
This is great to see how simple code change that is in the end - thanks!

PR #1656, opened 2021-01-13
Add containerd util image

lyzs90:
**What type of PR is this?**

/kind bug

Refer to https://github.com/kubernetes/perf-tests/pull/1654

Part of https://github.com/kubernetes/perf-tests/issues/1619

```release-note
NONE
```

```docs

```

### DIFF
--- a/util-images/containerd/Dockerfile
+++ b/util-images/containerd/Dockerfile
@@ -1,0 +1,4 @@
+FROM docker:latest
+
+RUN wget https://github.com/containerd/containerd/releases/download/v1.4.3/containerd-1.4.3-linux-amd64.tar.gz
+RUN tar xvf containerd-1.4.3-linux-amd64.tar.gz
--- a/util-images/containerd/Dockerfile
+++ b/util-images/containerd/Dockerfile
@@ -1,4 +1,5 @@
-FROM docker:latest
+FROM ubuntu:xenial
 
+RUN apt-get update && apt-get install -y wget
 RUN wget https://github.com/containerd/containerd/releases/download/v1.4.3/containerd-1.4.3-linux-amd64.tar.gz
 RUN tar xvf containerd-1.4.3-linux-amd64.tar.gz
--- a/util-images/containerd/Makefile
+++ b/util-images/containerd/Makefile
@@ -4,8 +4,7 @@ TAG = v0.0.1
 
 all: push
 
-.PHONY: build
-build: cmd
+build: Dockerfile
 	docker build --pull -t $(IMG):$(TAG) .
 	docker tag $(IMG):$(TAG) $(IMG):latest
 	@echo Built $(IMG):$(TAG) and tagged with latest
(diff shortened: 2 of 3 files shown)


util-images/containerd/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM docker:latest
```
mm4tt, line 1: I think we agreed that we don't need to base it on the docker img, right?

lyzs90: ah, i tried alpine but i couldn't get ctr running on it. any suggestions there?

mm4tt: What about golang-alpine? I believe ctr is written in golang, isn't it?

lyzs90: i tried golang:1.12.1-alpine, same issue there. its rather odd, let me dig into it a bit more

lyzs90: because i believe the docker base image was also based off alpine

lyzs90: i can make it work with ubuntu:xenial, its even smaller in size than the previous alpine, golang-alpine, docker images

mm4tt: What is the exact error you're getting for golang-alpine?

lyzs90: i get ctr command not found. so if you `ls -l` in the `/bin` folder of an alpine container, all the utilities are symlinked to `/bin/busybox`, i suspect it has something to do with that.. because all my non-busybox utils do not work

mm4tt: I see. I guess there must be some missing runtime dependency in the golang-alpine image. The regular golang seems to be working: 
```
FROM golang:1.13.5

RUN wget https://github.com/containerd/containerd/releases/download/v1.4.3/containerd-1.4.3-linux-amd64.tar.gz
RUN tar xvf containerd-1.4.3-linux-amd64.tar.gz
ENTRYPOINT bin/ctr
```

What's the difference in size of the image when using ubuntu and golang?

lyzs90: im getting 
ubuntu 299mb
golang 934mb

util-images/containerd/Makefile
```diff
@@ -0,0 +1,17 @@
+PROJECT ?= k8s-testimages
+IMG = gcr.io/$(PROJECT)/perf-tests-util/containerd
+TAG = v0.0.1
+
+all: push
+
+.PHONY: build
```
mm4tt, line 7: Remove this line

util-images/containerd/Makefile
```diff
@@ -0,0 +1,17 @@
...
+build: cmd
```
mm4tt, line 8: There is no cmd dir here, let's make it depend on "Dockerfile"

mm4tt (commented):
2 minor comments, otherwise LGTM

util-images/containerd/Makefile
lyzs90: 👍 done for both

mm4tt:
Looks great! Thanks!

Let's submit and then I'll push the image

mm4tt:
Pushed - gcr.io/k8s-testimages/perf-tests-util/containerd:v0.0.1

PR #1657, opened 2021-01-14
Fix measuring latency of nodelocaldns

marseel:
**What type of PR is this?**
/kind bug

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go
@@ -99,7 +99,7 @@ func (n *nodelocaldnsLatencyGatherer) getPercentileLatencies(executor QueryExecu
 			errList.Append(fmt.Errorf("got unexpected number of samples: %d for query %q", len(samples), query))
 			continue
 		}
-		result.SetQuantile(percVal, time.Duration(float64(samples[0].Value))*time.Second)
+		result.SetQuantile(percVal, time.Duration(float64(samples[0].Value)*float64(time.Second)))
 	}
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go
```diff
@@ -99,7 +99,7 @@ func (n *nodelocaldnsLatencyGatherer) getPercentileLatencies(executor QueryExecu
 			errList.Append(fmt.Errorf("got unexpected number of samples: %d for query %q", len(samples), query))
 			continue
 		}
-		result.SetQuantile(percVal, time.Duration(float64(samples[0].Value))*time.Second)
+		result.SetQuantile(percVal, time.Duration(float64(samples[0].Value)*float64(time.Second)))
```
wojtek-t, line 102: hm - why the previous didn't work? the original one should also work (the first duration is in nanoseconds than then we multiple by second...

marseel: Look at parentheses. In previous case, time.Duration(float64) is executed. If float64 is small number like 0.001 it will be treated as 0 nanoseconds. 
Example program:
`package main

import (
	"fmt"
	"time"
)

func main() {
	input := 1.0
	input = input / 10
	duration := time.Duration(input)*time.Second
	fmt.Println(input)
	fmt.Println(duration)
}
`
Output:
`
0.1
0s
`

wojtek-t: OK - the fractional part is what matter here. Makes sense.

PR #1660, opened 2021-01-18
Revamp API availability measurement

tosi3k:
This builds on https://github.com/kubernetes/perf-tests/pull/1162 where the measurement was introduced.

Changes in regard to the measurement:
* measuring master-level availability:
  * relies on `/readyz` instead of deprecated `/healthz`,
  * doesn't SSH to master - instead, it leverages the [exec service](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/pkg/execservice) to ping master's internal IPs,
* added optional `pause` and `unpause` actions to temporarily prevent the measurement from probing,
* code cleanup.

Tested the measurement manually.

### DIFF
--- a/clusterloader2/pkg/measurement/common/api_availability_measurement.go
+++ b/clusterloader2/pkg/measurement/common/api_availability_measurement.go
@@ -19,26 +19,26 @@ package common
 import (
 	"context"
 	"fmt"
-	"k8s.io/perf-tests/clusterloader2/pkg/errors"
-	"k8s.io/perf-tests/clusterloader2/pkg/provider"
+	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
-	goerrors "github.com/go-errors/errors"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/execservice"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
-	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/provider"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
 
 const (
-	apiAvailabilityName = "APIAvailability"
+	apiAvailabilityMeasurementName = "APIAvailability"
 )
 
 func init() {
-	if err := measurement.Register(apiAvailabilityName, createAPIAvailabilityMeasurement); err != nil {
-		klog.Fatalf("Cannot register %s: %v", apiAvailabilityName, err)
+	if err := measurement.Register(apiAvailabilityMeasurementName, createAPIAvailabilityMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", apiAvailabilityMeasurementName, err)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -93,81 +135,114 @@ func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMast
...
 }
 
-func (a *apiAvailabilityMeasurement) stopAndSummarize(SSHToMasterSupported bool, probeFrequency int) error {
+func (a *apiAvailabilityMeasurement) gather() ([]measurement.Summary, error) {
 	if !a.isRunning {
-		return nil
+		return nil, nil
+	}
+	if a.isHeld {
+		a.unhold()
```
mm4tt, line 220: Why do we need to unhold on gather?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,43 +48,85 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
+func (a *apiAvailabilityMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
 	}
 
+	switch action {
+	case "start":
+		return nil, a.start(config)
+	case "hold":
```
mm4tt, line 107: Maybe pause? And unpause? Same for the variable names.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -93,81 +135,114 @@ func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMast
...
+
+func (a *apiAvailabilityMeasurement) hold() {
+	if !a.isRunning {
+		klog.V(2).Infof("%s: measurement is not running", a)
+		return
 	}
-	probeFrequency, err := util.GetIntOrDefault(config.Params, "frequency", 1)
-	if err != nil {
-		return nil, err
+	if a.isHeld {
```
mm4tt, line 193: There is a potential race here. If I have a measurement step with two measurements that hold (measurements are executed in parallel) I may either
1. Hold the measurement and get "already held" error <- OK
2. Hold and "unhold" the measurement. The a.isHeld will be true, but measurement will continue running <- not OK, program will be in a malformed state.

This is because you allow two or more goroutines to modify the a.isHeld. You could fix that with either locks or (more golang idiomatic) having the a.IsHeld being modified only inside a single go-routine. You're already have this go-routine in the start() method. So the code could be something like:

```
func (a *apiAvailabilityMeasurement) hold() {
  a.hold <- struct{}{}
}

func (a *apiAvailabilityMeasurement) unhold() {
  a.unhold <- struct{}{}
}

func (a *apiAvailabilityMeasurement) start() {
  ...
  go func() {
    for {
      if (a.isHeld) {
        select {
        case <- a.unholdCh:
          a.isHeld = false
        case <- a.holdCh:
           klog.Warning("Meaasurement is already held!")
        case <- a.stopCh:
          return
        }

         
      }
      select {
        ...
        case <- a.holdCh:
          a.isHeld = true
          continue
        case <- a.unHoldCh:
          klog.Warning("Meaasurement is already unheld!")
      }
    }
  }
}
```

I think it's safer and also easier to read.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,43 +48,85 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
+	isHeld              bool
+	holdCh              chan struct{}
 	stopCh              chan struct{}
-	hosts               []string
+	frequency           time.Duration
+	masterIPs           []string
 	summaries           []measurement.Summary
 	clusterLevelMetrics *apiAvailabilityMetrics
-	hostLevelMetrics    map[string]*apiAvailabilityMetrics
+	masterLevelMetrics  map[string]*apiAvailabilityMetrics
```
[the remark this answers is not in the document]
wojtek-t: we're trying to avoid using a "master" word whenever possible - so I actually prefer reverting that change if possible.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,43 +48,85 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
+	if err != nil {
+		return "", fmt.Errorf("problem with RunCommand(): %w", err)
+	}
+	return output, nil
+}
+
 func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
-	// Check the availability of the cluster by issuing a REST call to /healthz end point
-	result := c.CoreV1().RESTClient().Get().AbsPath("/healthz").Do(context.TODO())
+	result := c.CoreV1().RESTClient().Get().AbsPath("/livez").Do(context.TODO())
```
wojtek-t, line 88: Let's use readyz

clusterloader2/pkg/measurement/common/api_availability_measurement.go
wojtek-t: Agree - pause sounds more intuitive

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -93,81 +135,114 @@ func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMast
...
 
-// Execute starts the api-server healthz probe end point from start action and
-// collects availability metrics in gather.
-func (a *apiAvailabilityMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
-	SSHToMasterSupported := !config.ClusterFramework.GetClusterConfig().Provider.Features().SupportSSHToMaster
-
-	if !SSHToMasterSupported {
-		klog.Infof("ssh to master nodes not supported. Measurement would have only Cluster Level Metrics")
+func (a *apiAvailabilityMeasurement) initFields(config *measurement.Config) error {
+	a.isRunning = true
```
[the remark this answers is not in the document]
wojtek-t: subjective :) - I actually prefer how it is now

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -28,64 +28,51 @@ type apiAvailabilityMetrics struct {
...
+	AvailabilityPercentage   float32 `json:"availabilityPercentage"`
+	LongestUnavailablePeriod string  `json:"longestUnavailablePeriod"`
 }
 
 type apiAvailabilityOutput struct {
-	ClusterMetrics *apiAvailabilitySummary   `json:"clusterMetrics"`
-	MasterMetrics  []*apiAvailabilitySummary `json:"masterMetrics"`
+	ClusterSummary apiAvailabilitySummary   `json:"clusterMetrics"`
+	MasterSummary  []apiAvailabilitySummary `json:"masterMetrics"`
+	Frequency      string                   `json:"frequency"`
```
wojtek-t, line 39: Why do you need frequency here?

clusterloader2/pkg/measurement/common/api_availability_metrics.go
```diff
@@ -28,64 +28,51 @@ type apiAvailabilityMetrics struct {
...
-	metrics.buildAPIAvailabilityMetricsSummary(probeFrequency, true, "")
-}
-
-func (output apiAvailabilityOutput) createMastersSummary(metrics map[string]*apiAvailabilityMetrics, hosts []string, probeFrequency int) {
-	for _, host := range hosts {
-		masterHostAvailabilitySummary := metrics[host].buildAPIAvailabilityMetricsSummary(probeFrequency, false, host)
-		output.MasterMetrics = append(output.MasterMetrics, masterHostAvailabilitySummary)
+func (a *apiAvailabilityMetrics) update(availability bool) {
+	if availability {
+		a.numSuccesses++
```
wojtek-t, line 44: return

and everything else in the main flow of function - to reduce the nesting

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,43 +48,85 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
-		sshResult, err := measurementutil.SSH(command, host+":22", provider)
-		availability := err != nil || sshResult.Code != 0
-		a.updateAvailabilityMetrics(availability, a.hostLevelMetrics[host])
+func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, provider provider.Provider) {
+	for _, ip := range a.masterIPs {
+		statusCode, err := a.pingMaster(ip)
+		availability := statusCode == strconv.Itoa(http.StatusOK)
+		if err != nil {
+			klog.Warningf("execservice issue: %w", err)
+			availability = false
```
lyzs90, line 68: you probably don't need this line because if err is returned, `"" == strconv.Itoa(http.StatusOK)` evaluates to false

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -93,81 +135,114 @@ func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMast
...
+	a.stopCh = make(chan struct{})
+	a.holdCh = make(chan struct{})
+	frequency, err := util.GetDuration(config.Params, "frequency")
+	if err != nil {
+		return err
+	}
+	a.frequency = frequency
+
+	a.clusterLevelMetrics = &apiAvailabilityMetrics{}
+	if config.ClusterLoaderConfig.EnableExecService {
```
lyzs90, line 167: exec service is [disabled by default](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/cmd/clusterloader.go#L116), just fyi

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -93,81 +135,114 @@ func (a *apiAvailabilityMeasurement) start(config *measurement.Config, SSHToMast
...
+				case <-a.stopCh:
+					return
+				case <-a.holdCh:
+					continue
+				}
+			case <-time.After(a.frequency):
 				a.updateClusterAvailabilityMetrics(k8sClient)
-				if SSHToMasterSupported {
-					a.updateMasterAvailabilityMetrics(k8sClient, config, provider)
+				if a.masterLevelAvailabilityEnabled() {
```
lyzs90, line 147: would checking `config.ClusterLoaderConfig.EnableExecService` be more appropriate here? `masterLevelAvailabilityEnabled` is about whether master level metrics have been collected and should probably be renamed

tosi3k: Apart from exec service being enabled we also require the internal IPs of masters for the per-VM availability, so it wouldn't suffice. `masterLevelAvailabilityEnabled` boolean function is true iff we are running per-VM polling in the measurement as we fill the `a.hostIPs` field after we check that both exec service is enabled and the internal IPs are provided.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
tosi3k: I used the lock to prevent two actions of the same measurement to be run in the same step, so we don't run into any race by accidental putting two actions of the same measurement into one step.

I really like the idea of separating the pausing/unpausing channel into two different ones, though, as it indeed increases the readability of the code. Thanks for that remark.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,126 +48,208 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
+		a.hostLevelMetrics[ip].update(availability)
+	}
+}
+
+func (a *apiAvailabilityMeasurement) pingHost(hostIP string) (string, error) {
+	pod, err := execservice.GetPod()
+	if err != nil {
+		return "", fmt.Errorf("problem with GetPod(): %w", err)
 	}
+	cmd := fmt.Sprintf("curl --connect-timeout 5 -s -k -w \"%%{http_code}\" -o /dev/null https://%s:443/readyz", hostIP)
```
mm4tt, line 84: Can we parametrize the timeout?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,126 +48,208 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
-func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.Config, provider provider.Provider) {
-	for _, host := range a.hosts {
-		// SSH and check the health of the host
-		command := fmt.Sprintf("curl -f -s -k %slocalhost:%v/healthz", "https://", 443)
-		sshResult, err := measurementutil.SSH(command, host+":22", provider)
-		availability := err != nil || sshResult.Code != 0
-		a.updateAvailabilityMetrics(availability, a.hostLevelMetrics[host])
+func (a *apiAvailabilityMeasurement) updateHostAvailabilityMetrics(c clientset.Interface, provider provider.Provider) {
+	for _, ip := range a.hostIPs {
+		statusCode, err := a.pingHost(ip)
```
mm4tt, line 67: Can we do that in parallel? If so, could you add a TODO, I don't want to complicate this PR any further.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,126 +48,208 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
+	if err != nil {
+		return "", fmt.Errorf("problem with RunCommand(): %w", err)
+	}
+	return output, nil
 }
 
 func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
-	// Check the availability of the cluster by issuing a REST call to /healthz end point
-	result := c.CoreV1().RESTClient().Get().AbsPath("/healthz").Do(context.TODO())
+	result := c.CoreV1().RESTClient().Get().AbsPath("/readyz").Do(context.TODO())
```
mm4tt, line 93: NIT: Not sure if context.TODO is a good choice here. I don't think we have plans to bring the context into measurements, so more appropriate would be context.Background()

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,126 +48,208 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
 	status := 0
 	result.StatusCode(&status)
-	a.updateAvailabilityMetrics(status == 200, a.clusterLevelMetrics)
+	availability := status == http.StatusOK
+	if !availability {
+		klog.Warning("cluster not available")
```
mm4tt, line 98: Can we log the http code?

mm4tt (commented):
Looks good, just some minor comments. 

Can we enable it in the Cl2 perf-test presubmit so we can test this PR before we merge?

clusterloader2/pkg/measurement/common/api_availability_measurement.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
tosi3k: Added an appropriate TODO comment above. Done.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -48,126 +48,215 @@ func createAPIAvailabilityMeasurement() measurement.Measurement {
...
 
-func (a *apiAvailabilityMeasurement) updateMasterAvailabilityMetrics(c clientset.Interface, config *measurement.Config, provider provider.Provider) {
-	for _, host := range a.hosts {
-		// SSH and check the health of the host
-		command := fmt.Sprintf("curl -f -s -k %slocalhost:%v/healthz", "https://", 443)
-		sshResult, err := measurementutil.SSH(command, host+":22", provider)
-		availability := err != nil || sshResult.Code != 0
-		a.updateAvailabilityMetrics(availability, a.hostLevelMetrics[host])
+func (a *apiAvailabilityMeasurement) updateHostAvailabilityMetrics(c clientset.Interface, provider provider.Provider) {
+	// TODO: Parallelize polling individual hosts.
```
mm4tt, line 67: Ownerless / issueless TODO. Please either assign to someone, or preferable create an oss issue for this and mark as /help wanted

mm4tt (commented):
LGTM, except this one ownerless TODO :)

clusterloader2/pkg/measurement/common/api_availability_measurement.go
tosi3k: Done: https://github.com/kubernetes/perf-tests/issues/1683.

mm4tt:
/lgtm
/approve

Thanks!

Issue #1027, opened 2020-02-06
Re-implement scheduling-throughput measurement


mm4tt:
Currently, we have a very crude logic for computing [scheduling-throughput](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/scheduling_throughput.go). We list all pods every 5s and see how many new pod very scheduled in that time. It doesn't work well in small tests where number of pods is low as we only get 1-2 representative 5s windows. 

We should rewrite it to make it more reliable and accurate, e.g. use prometheus query to monitor rate of pod bindings

wojtek-t:
/remove-lifecycle stale

/assign @marseel 

Since this is already happening - @marseel can you please update the issue with info what work is remaining to do?

marseel:
Currently, new prometheus metric is already implemented and data is collected for all tests in master branch.
It's worth to mention that previous SchedulingThroughput allowed us to gather metric with interval 1s, based on Prometheus we are only able to gather with interval 5s. 
The only thing remaining is turning on alerting for specific tests and removing old SchedulingThroughput. I've postponed this step, because it might have been conflicting with merging density with load test.

wojtek-t:
> only able to gather with interval 5s.

Unless we start scraping metrics more/less often. We may consider scraping more often potentially too.

> I've postponed this step, because it might have been conflicting with merging density with load test.

Which is the right decision :)
Though we're almost done with this merging, so we will be able to get back to it.

dmatch01:
Hi @marseel Do you know if the old SchedulingThroughput was ever removed or is it still active?  I still the source code in the main branch.  

> The only thing remaining is turning on alerting for specific tests and removing old SchedulingThroughput. I've postponed this step, because it might have been conflicting with merging density with load test.

PR #1667, opened 2021-01-19
Make Cl2 tests to work in <100 node clusters

lyzs90:
**What type of PR is this?**

/kind bug

Details: https://github.com/kubernetes/perf-tests/issues/887

Fixes https://github.com/kubernetes/perf-tests/issues/887

The type [conversion](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/config/template_functions.go#L133) was rounding fractions to 0. Hence when `.Nodes` passed in is less than `NODES_PER_NAMESPACE (defaults to 100)`, `namespaces = DivideInt .Nodes  NODES_PER_NAMESPACE = 0`. thereafter `namespace` is used as the denominator for several computations eg. https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L7

```release-note
Users can now run load tests with < 100 nodes
```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -5,7 +5,10 @@
 # - Only half of Deployments will be assigned 1-1 to existing SVCs.
 
 #Constants
-{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+# Cater for the case where the number of nodes is less than nodes per namespace. See https://github.com/kubernetes/perf-tests/issues/887
+{{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .NODES_PER_NAMESPACE 100)}}
+# See https://github.com/kubernetes/perf-tests/pull/1667#issuecomment-769642266
+{{$IS_SMALL_CLUSTER := lt .Nodes 100}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$DELETE_TEST_THROUGHPUT := DefaultParam .CL2_DELETE_TEST_THROUGHPUT $LOAD_TEST_THROUGHPUT}}
@@ -50,14 +53,21 @@
 {{$mediumDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 4 $MEDIUM_GROUP_SIZE)}}
 # smallDeployments - 1/2 of namespace pods should be in small Deployments.
 {{$smallDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 2 $SMALL_GROUP_SIZE)}}
-# Reduce the number of small and medium deployments per namespace
-{{$smallDeploymentsPerNamespace := SubtractInt $smallDeploymentsPerNamespace $SMALL_STATEFUL_SETS_PER_NAMESPACE}}
-{{$mediumDeploymentsPerNamespace := SubtractInt $mediumDeploymentsPerNamespace $MEDIUM_STATEFUL_SETS_PER_NAMESPACE}}
 
-# Reduce the number of small, medium, big deployments per namespace.
-{{$smallDeploymentsPerNamespace := SubtractInt $smallDeploymentsPerNamespace 1}}
-{{$mediumDeploymentsPerNamespace := SubtractInt $mediumDeploymentsPerNamespace 1}}
-{{$bigDeploymentsPerNamespace := SubtractInt $bigDeploymentsPerNamespace 1}}
+# Stateful sets are enabled. Reduce the number of small and medium deployments per namespace
+# See https://github.com/kubernetes/perf-tests/issues/1036#issuecomment-607631768
+# Ensure non zero or negative after subtraction.
+{{$smallDeploymentsPerNamespace := MaxInt 0 (SubtractInt $smallDeploymentsPerNamespace $SMALL_STATEFUL_SETS_PER_NAMESPACE)}}
+{{$mediumDeploymentsPerNamespace := MaxInt 0 (SubtractInt $mediumDeploymentsPerNamespace $MEDIUM_STATEFUL_SETS_PER_NAMESPACE)}}
+
+# Jobs are enabled. Reduce the number of small, medium, big deployments per namespace.
+# Ensure non zero or negative after subtraction.
+{{$smallDeploymentsPerNamespace := MaxInt 0 (SubtractInt $smallDeploymentsPerNamespace 1)}}
+{{$mediumDeploymentsPerNamespace := MaxInt 0 (SubtractInt $mediumDeploymentsPerNamespace 1)}}
+{{$bigDeploymentsPerNamespace := MaxInt 0 (SubtractInt $bigDeploymentsPerNamespace 1)}}
+
+# Disable big jobs on small clusters.
+{{$bigJobsPerNamespace := IfThenElse $IS_SMALL_CLUSTER 0 1}}
 
 # The minimal number of pods to be used to measure various things like
 # pod-startup-latency or scheduler-throughput. The purpose of it is to avoid
@@ -72,6 +82,10 @@
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
 {{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 100}}
 # END scheduler-throughput section
+
+# Set schedulerThroughputNamespaces to 0 on small clusters otherwise it will result
+# in an unnecessary number of namespaces.
+{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 0 $schedulerThroughputNamespaces}}
 
 # TODO(https://github.com/kubernetes/perf-tests/issues/1024): Investigate and get rid of this section.
 # BEGIN pod-startup-latency section
@@ -348,7 +362,7 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: 1
+    replicasPerNamespace: {{$bigJobsPerNamespace}}
     tuningSet: RandomizedSaturationTimeLimited
     objectBundle:
       - basename: big-job
@@ -388,6 +402,7 @@ steps:
 {{end}}
 {{if not $EXIT_AFTER_EXEC}}
 
+{{if not $IS_SMALL_CLUSTER}}
 # BEGIN scheduler throughput
 - name: Creating scheduler throughput measurements
   measurements:
@@ -463,7 +478,9 @@ steps:
     Params:
       action: gather
 # END scheduler throughput
+{{end}}
 
+{{if not $IS_SMALL_CLUSTER}}
 # TODO(https://github.com/kubernetes/perf-tests/issues/1024): Ideally, we wouldn't need this section.
 # BEGIN pod-startup-latency
 - name: Starting latency pod measurements
@@ -526,6 +543,7 @@ steps:
     Params:
       action: gather
 # END pod-startup-latency
+{{end}}
 
 - name: Scaling and updating objects
   phases:
@@ -630,7 +648,7 @@ steps:
   - namespaceRange:
       min: 1
       max: {{$namespaces}}
-    replicasPerNamespace: 1
+    replicasPerNamespace: {{$bigJobsPerNamespace}}
     tuningSet: RandomizedScalingTimeLimited
     objectBundle:
       - basename: big-job
--- a/clusterloader2/pkg/tuningset/randomized_time_limited.go
+++ b/clusterloader2/pkg/tuningset/randomized_time_limited.go
@@ -40,7 +40,10 @@ func (r *randomizedTimeLimitedLoad) Execute(actions []func()) {
 		index := i
 		wg.Start(func() {
 			// Sleeps for random duration in [0, TimeLimit].
-			time.Sleep(time.Duration(rand.Int63n(r.params.TimeLimit.ToTimeDuration().Nanoseconds())))
+			timeLimit := r.params.TimeLimit.ToTimeDuration().Nanoseconds()
+			if timeLimit > 0 {
+				time.Sleep(time.Duration(rand.Int63n(timeLimit)))
+			}
 			actions[index]()
 		})
 	}
--- a/clusterloader2/testing/density/config.yaml
+++ b/clusterloader2/testing/density/config.yaml
@@ -4,7 +4,8 @@
 
 #Constants
 {{$DENSITY_RESOURCE_CONSTRAINTS_FILE := DefaultParam .DENSITY_RESOURCE_CONSTRAINTS_FILE ""}}
-{{$NODES_PER_NAMESPACE := DefaultParam .NODES_PER_NAMESPACE 100}}
+# Cater for the case where the number of nodes is less than nodes per namespace. See https://github.com/kubernetes/perf-tests/issues/887
+{{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .NODES_PER_NAMESPACE 100)}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 {{$DENSITY_TEST_THROUGHPUT := DefaultParam .DENSITY_TEST_THROUGHPUT 20}}
 {{$SCHEDULER_THROUGHPUT_THRESHOLD := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 0}}
--- a/clusterloader2/testing/density/high-density-config.yaml
+++ b/clusterloader2/testing/density/high-density-config.yaml
@@ -5,7 +5,8 @@
 
 #Constants
 {{$DENSITY_RESOURCE_CONSTRAINTS_FILE := DefaultParam .DENSITY_RESOURCE_CONSTRAINTS_FILE ""}}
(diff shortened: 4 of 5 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -39,7 +39,8 @@
 {{$ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS := DefaultParam .CL2_ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS false}}
 {{$ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS_SIMPLE := DefaultParam .CL2_ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS_SIMPLE true}}
 #Variables
-{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
+# set min namespaces to 1 to avoid divide by zero errors. See https://github.com/kubernetes/perf-tests/issues/887
+{{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE | MaxInt 1}}
 {{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
```
mm4tt, line 44: We should also adjust this value. Otherwise, if I run this test on a 1 node cluster, the total pods will be calculated as if there was 100 nodes (and it will be equal to 3000pods). Obviously, it won't work

lyzs90: ah my bad, i didn't look beyond that line. the real culprit seems to be the `NODES_PER_NAMESPACE` constant.  is it ok to set `NODES_PER_NAMESPACE` to be the minimum of num nodes and 100? then we can do away with the rounding fix on namespaces

mm4tt: Yeah, that should work.

lyzs90:
@mm4tt Noticed a few other places where this fix might not work for small clusters eg. https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L67 and https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/density/config.yaml#L17. These may have to be scaled down also - the goal is so that anyone can run load / density tests on a 1 node cluster?

lyzs90:
Instead of trying to patch the load and density tests so that they will work on small clusters, wdyt about defining a separate, minimal test config that will run on a 1 node cluster?

mm4tt:
I'd really like to make the load test to work at any scale. All the hard-coded values we have there are actually a technical debt and we should aim to get rid of them. A few notes:

- Please disregard the density test, we're aiming to deprecate it soon
- Is there anything else other than `MIN_PODS_IN_SMALL_CLUSTERS` that's blocking us?
- For `MIN_PODS_IN_SMALL_CLUSTERS`, I believe we have some options:
   1. Ideally, we'd just fix the https://github.com/kubernetes/perf-tests/issues/1027, so we don't need this constant. Unfortunately, that's not that simple, so let's think of something else
   2. Instead of having the `MIN_PODS_IN_SMALL_CLUSTERS` as a constant, hard-coded in the config.yaml, it could be a test param that is set to 500 in the test-infra presets only for our 100+ node tests
   3. Given that this part of the test doesn't really work if we have less than 100 nodes (if we have less then we're not able to accurately measure the scheduling throughput), we should probably conditionally disable this part if # nodes < 100.

I believe 3 is the best option currently. Thoughts?

lyzs90:
Yes I think approach 3 sounds good. Will also need to handle `BIG_GROUP_SIZE` and `MEDIUM_GROUP_SIZE` in the same manner I suppose?

mm4tt:
> Will also need to handle BIG_GROUP_SIZE and MEDIUM_GROUP_SIZE in the same manner I suppose?

Hmm, wouldn't it work if we just round down everywhere? E.g. if we have just 10 nodes, the total number of pods will be 300 (30*10). Then we'll calculate how many small/medium/big deployments we can fit and e.g. for big we'll do: 300 / 4 / 250 = 0 (assuming we round down). For medium it'll be 300/4/30 = 2. For small it'll be 300/2/5 = 30. Does it make sense?

lyzs90:
> > Will also need to handle BIG_GROUP_SIZE and MEDIUM_GROUP_SIZE in the same manner I suppose?
> 
> Hmm, wouldn't it work if we just round down everywhere? E.g. if we have just 10 nodes, the total number of pods will be 300 (30*10). Then we'll calculate how many small/medium/big deployments we can fit and e.g. for big we'll do: 300 / 4 / 250 = 0 (assuming we round down). For medium it'll be 300/4/30 = 2. For small it'll be 300/2/5 = 30. Does it make sense?

But it's also used in number of replicas https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#257. Seems like `BIG_GROUP_SIZE`, `MEDIUM_GROUP_SIZE` needs to be a factor of number of nodes?

mm4tt:
It will be fine if we have 250 there even if there is no place for so many pods. This is because this object will be never created. The replicasPerNamespace will be 0 - https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L244

lyzs90:
Ah yes. I'm curious if that will all fit on 1 node 😅

lyzs90:
Doing a local run, noting down some issues:
- [ ] Creates ~1000 namespaces due to https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L94

lyzs90:
@mm4tt Added defaults and disabled heavier workloads when running on clusters < 100 nodes (I tested with a single node cluster)

clusterloader2/testing/load/config.yaml
```diff
@@ -90,7 +97,7 @@
 
 name: load
 namespace:
-  number: {{AddInt $namespaces $schedulerThroughputNamespaces}}
+  number: {{MinInt .Nodes (AddInt $namespaces $schedulerThroughputNamespaces)}}
```
mm4tt, line 100: Not sure If I understand this line. Wouldn't be better to make sure the `$namespaces` and `schedulerThroughputNamespaces` are just set properly in all the cases?

clusterloader2/testing/load/config.yaml
```diff
@@ -229,6 +236,7 @@ steps:
       objectTemplatePath: daemonset.yaml
       templateFillMap:
         Image: k8s.gcr.io/pause:3.0
+  {{if not $IS_SMALL_CLUSTER}}
```
mm4tt, line 239: Is this if block needed here? I believe the `$bigDeploymentsPerNamespace` will be 0 if `.Nodes` < 100. In that case no object is going to be created

clusterloader2/testing/load/config.yaml
```diff
@@ -345,6 +354,7 @@ steps:
         templateFillMap:
           ReplicasMin: {{$MEDIUM_GROUP_SIZE}}
           ReplicasMax: {{$MEDIUM_GROUP_SIZE}}
+  {{if not $IS_SMALL_CLUSTER}}
```
mm4tt, line 357: Instead of havign this if block, we could introduce a variable`$bigJobsPerNamespace` and make sure it will be set to 0 if #Nodes < 100

clusterloader2/testing/load/config.yaml
```diff
@@ -526,9 +540,11 @@ steps:
     Params:
       action: gather
 # END pod-startup-latency
+{{end}}
 
 - name: Scaling and updating objects
   phases:
+  {{if not $IS_SMALL_CLUSTER}}
```
mm4tt, line 547: Same as above, I believe it's not needed

clusterloader2/testing/load/config.yaml
```diff
@@ -652,6 +671,7 @@ steps:
 
 - name: Deleting objects
   phases:
+  {{if not $IS_SMALL_CLUSTER}}
```
mm4tt, line 674: Same here

clusterloader2/testing/load/config.yaml
```diff
@@ -627,6 +644,7 @@ steps:
         templateFillMap:
           ReplicasMin: {{MultiplyInt $MEDIUM_GROUP_SIZE 0.5}}
           ReplicasMax: {{MultiplyInt $MEDIUM_GROUP_SIZE 1.5}}
+  {{if not $IS_SMALL_CLUSTER}}
```
mm4tt, line 647: Same here, one we introduce a variable it shoudln't be needed anymore

mm4tt (commented):
Thanks, this is great! I know how painful it's to write in yaml and go-templates, so I really appreciate it!

clusterloader2/testing/load/config.yaml
lyzs90: Will try again, I may have mixed this up with the error coming from a negative `$bigDeploymentsPerNamespace`

clusterloader2/testing/load/config.yaml
lyzs90: I added some calculations below so it's easier to make sense of the change.

Without MinInt:
```
# 100 nodes
$MIN_PODS_IN_SMALL_CLUSTERS := 500
$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) (MultiplyInt 2 .Nodes) # 1000
$schedulerThroughputPodsPerDeployment := .Nodes # 100
$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment # 10
number: AddInt $namespaces $schedulerThroughputNamespaces # 11
```

```
# 1 node
$MIN_PODS_IN_SMALL_CLUSTERS := 500
$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) (MultiplyInt 2 .Nodes) # 1000
$schedulerThroughputPodsPerDeployment := .Nodes # 1
$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment # 1000
number: AddInt $namespaces $schedulerThroughputNamespaces # 1001 <- !!!
```

With MinInt:
```
# 100 nodes
same as above
number: MinInt .Nodes (AddInt $namespaces $schedulerThroughputNamespaces) # 11
```

```
# 1 node
same as above
number: MinInt .Nodes (AddInt $namespaces $schedulerThroughputNamespaces) # 1
```

clusterloader2/testing/load/config.yaml
lyzs90: Big deployment objects are still being created even though `bigDeploymentsPerNamespace` is 0. Seems like ReplicasMin and ReplicasMin have the final say. So maybe targeting the `BIG_GROUP` size variable would be better as it will affect jobs created at the same time.

image

mm4tt: Nah, we're missing something. If the `replicasPerNamespace` is set to 0 then no new object should be created - https://github.com/kubernetes/perf-tests/blob/8c85b1d608511e76d8c8180b3bc6163b736e20ec/clusterloader2/pkg/test/simple_test_executor.go#L261

The only explanation is that the $bigDeploymentsPerNamespace` is not set to 0.  Isn't it because in line 67 we're doing:
```
{{$bigDeploymentsPerNamespace := MaxInt 1 (SubtractInt $bigDeploymentsPerNamespace 1)}}
```
?

lyzs90: Yes! This is it, addressed in my latest commit https://github.com/kubernetes/perf-tests/pull/1667/commits/99c29b498be48a07791d7b92e500bba406af45f9

clusterloader2/testing/load/config.yaml
```diff
@@ -50,14 +53,22 @@
...
+# See https://github.com/kubernetes/perf-tests/issues/1036#issuecomment-607631768
+# Ensure non zero or negative after subtraction.
+{{$smallDeploymentsPerNamespace := MaxInt 1 (SubtractInt $smallDeploymentsPerNamespace $SMALL_STATEFUL_SETS_PER_NAMESPACE)}}
+{{$mediumDeploymentsPerNamespace := MaxInt 1 (SubtractInt $mediumDeploymentsPerNamespace $MEDIUM_STATEFUL_SETS_PER_NAMESPACE)}}
+
+# Jobs are enabled. Reduce the number of small, medium, big deployments per namespace.
+# Ensure non zero or negative after subtraction.
+{{$smallDeploymentsPerNamespace := MaxInt 1 (SubtractInt $smallDeploymentsPerNamespace 1)}}
+{{$mediumDeploymentsPerNamespace := MaxInt 1 (SubtractInt $mediumDeploymentsPerNamespace 1)}}
+{{$bigDeploymentsPerNamespace := MaxInt 1 (SubtractInt $bigDeploymentsPerNamespace 1)}}
```
mm4tt, line 67: We should have: `MaxInt 0 ...` instead of `MaxInt 1 ...`. Here and above. Then we won't need lines 69-71

lyzs90: I added :69-71 because there is nothing limiting `$bigDeploymentsPerNamespace` above any more, it will be `250 * 4 - 1 = 999` on this line

mm4tt: Hmm, the `bigDeploymentsPerNamespace` is calculated as
```
{{$bigDeploymentsPerNamespace := DivideInt $podsPerNamespace (MultiplyInt 4 $BIG_GROUP_SIZE)}}
```
And if we have `$namespaces=1` then `$podsPerNamespace = 30 * #Nodes`

So, e.g. in a 10 node cluster
```
bigDeploymentsPerNamespace := 30* 10 / 4 / 250 = 0
```

I believe it should work, we don't need any custom logic there

lyzs90: Ah did I miss something. Let me try again later. I vaguely recall it wasn't working as expected when I set it to  MaxInt 0

lyzs90: Ok it works! Made the change accordingly

clusterloader2/testing/load/config.yaml
```diff
@@ -112,7 +122,7 @@ tuningSets:
     # The expected number of created/deleted pods is totalPods/4 when scaling,
     # as each RS changes its size from X to a uniform random value in [X/2, 3X/2].
     # To match 10 [pods/s] requirement, we need to divide saturationTime by 4.
-    timeLimit: {{DivideInt $saturationTime 4}}s
+    timeLimit: {{MaxInt 1 (DivideInt $saturationTime 4)}}s
```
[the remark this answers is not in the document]
lyzs90: The RNG panics if timeLimit is 0:

```
panic: invalid argument to Int63n

goroutine 321 [running]:
math/rand.(*Rand).Int63n(0xc0000a8420, 0x0, 0xc0008b67c0)
	/usr/local/go/src/math/rand/rand.go:111 +0x11d
math/rand.Int63n(...)
	/usr/local/go/src/math/rand/rand.go:327
k8s.io/perf-tests/clusterloader2/pkg/tuningset.(*randomizedTimeLimitedLoad).Execute.func1()
	/Users/shemleong/go/src/perf-tests/clusterloader2/pkg/tuningset/randomized_time_limited.go:43 +0x4a
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1(0xc000b00420, 0xc000b7a750)
	/Users/shemleong/go/pkg/mod/k8s.io/apimachinery@v0.18.0/pkg/util/wait/wait.go:73 +0x51
created by k8s.io/apimachinery/pkg/util/wait.(*Group).Start
	/Users/shemleong/go/pkg/mod/k8s.io/apimachinery@v0.18.0/pkg/util/wait/wait.go:71 +0x65
exit status 2
```

lyzs90:
Other than the timeLimit one, I've addressed the other comments!

clusterloader2/testing/load/config.yaml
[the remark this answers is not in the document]
mm4tt: Oh, so let's fix that here - https://github.com/kubernetes/perf-tests/blob/0cef7332080a00091ab8d310ee0d7cb00cff8974/clusterloader2/pkg/tuningset/randomized_time_limited.go#L43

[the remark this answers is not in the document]
mm4tt: Basically, call it only if timeLimit > 0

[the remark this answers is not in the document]
lyzs90: Cool, added the check there

clusterloader2/testing/load/config.yaml
```diff
@@ -5,12 +5,15 @@
...
+# Cater for the case where the number of nodes is less than nodes per namespace. See https://github.com/kubernetes/perf-tests/issues/887
+{{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .NODES_PER_NAMESPACE 100)}}
+# See https://github.com/kubernetes/perf-tests/pull/1667#issuecomment-769642266
+{{$IS_SMALL_CLUSTER := lt .Nodes 100}}
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$DELETE_TEST_THROUGHPUT := DefaultParam .CL2_DELETE_TEST_THROUGHPUT $LOAD_TEST_THROUGHPUT}}
 {{$BIG_GROUP_SIZE := DefaultParam .BIG_GROUP_SIZE 250}}
-{{$MEDIUM_GROUP_SIZE := DefaultParam .MEDIUM_GROUP_SIZE 30}}
+{{$MEDIUM_GROUP_SIZE := DefaultParam .MEDIUM_GROUP_SIZE 15}}
```
mm4tt, line 16: Revert

mm4tt (commented):
LGTM with one comment

clusterloader2/testing/load/config.yaml
lyzs90: Oops my bad, forgot to undo the change after testing

lyzs90:
> LGTM with one comment

Cool :) will squash commits

PR #1668, opened 2021-01-20
Set name in services module

mborsz:
It should use name 'name', not 'actionName': https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/config.yaml#L181

/assign @mm4tt

### DIFF
--- a/clusterloader2/testing/load/modules/services.yaml
+++ b/clusterloader2/testing/load/modules/services.yaml
@@ -1,10 +1,10 @@
 ## Services modules provides a module for creating / deleting services.
 
 ## Input params
-{{$name := .actionName}}
+{{$name := .name}}
 {{$namespaces := .namespaces}}
 {{$smallServicesPerNamespace := .smallServicesPerNamespace}}
-{{$mediumServicesPerNamespace := .smallServicesPerNamespace}}
+{{$mediumServicesPerNamespace := .mediumServicesPerNamespace}}
 {{$bigServicesPerNamespace := .bigServicesPerNamespace}}
 
 steps:


mm4tt:
Nice catch. I wonder why my differ hasn't caught this, let me check...

/lgtm

mm4tt:
Ok, there was a bug in my differ. I fixed it. It it worked correctly the diff would be:

```
Diff is:
70c70
<     replicaspernamespace: 149
---
>     replicaspernamespace: 12
92c92
<   name: <no value>
---
>   name: Creating SVCs
913c913
<   name: <no value>
---
>   name: Deleting SVCs
```

Looks like I made another mistake, I increased the number of medium services

clusterloader2/testing/load/modules/services.yaml
```diff
@@ -1,7 +1,7 @@
 ## Services modules provides a module for creating / deleting services.
 
 ## Input params
-{{$name := .actionName}}
+{{$name := .name}}
 {{$namespaces := .namespaces}}
 {{$smallServicesPerNamespace := .smallServicesPerNamespace}}
 {{$mediumServicesPerNamespace := .smallServicesPerNamespace}}
```
mm4tt, line 7: Could you also fix this line?

mm4tt (commented):
/hold

mborsz:
Fixed. PTAL

clusterloader2/testing/load/modules/services.yaml
mborsz: Done

mm4tt:
/lgtm
/unhold

Thanks!

PR #1670, opened 2021-01-20
Add custom thresholds param to simple APIResponsiveness measurement

tosi3k:
/assign @mm4tt

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -831,6 +831,7 @@ steps:
       useSimpleLatencyQuery: true
       summaryName: APIResponsivenessPrometheus_simple
       allowedSlowCalls: {{$ALLOWED_SLOW_API_CALLS}}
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -831,6 +831,7 @@ steps:
       useSimpleLatencyQuery: true
       summaryName: APIResponsivenessPrometheus_simple
       allowedSlowCalls: {{$ALLOWED_SLOW_API_CALLS}}
+      customThresholds: {{YamlQuote $CUSTOM_API_CALL_THRESHOLDS 4}}
```
wojtek-t, line 834: I don't understand this change - isn't customThreshold supposed to be a map?

tosi3k: Explained offline but for posterity - it is `YamlQuote`, not `DefaultParam` ;).

PR #1678, opened 2021-01-25
Dump instanceIDs for master nodes in clusterloader

mborsz:
Idea:

If we start dumping instanceIDs for master nodes in gce tests and write it to 'finished.json' this information will be then available in `k8s-gubernator.build.all` and can be used for programmatically analyzing runs.

Usage: this PR writes 'cl2-metadata.json' file. If we include this file in --metadata-sources, it should be merged into finished.json which is what we need.

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/perf-tests/clusterloader2/pkg/flags"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/imagepreload"
+	"k8s.io/perf-tests/clusterloader2/pkg/metadata"
 	"k8s.io/perf-tests/clusterloader2/pkg/modifier"
 	"k8s.io/perf-tests/clusterloader2/pkg/prometheus"
 	"k8s.io/perf-tests/clusterloader2/pkg/provider"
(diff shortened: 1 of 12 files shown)


clusterloader2/pkg/metadata/metadata.go
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
mm4tt, line 2: 2021 :)

clusterloader2/cmd/clusterloader.go
```diff
@@ -300,6 +301,11 @@ func main() {
 		klog.Exitf("Error while preloading images: %v", err)
 	}
 
+	if err := metadata.Dump(f, path.Join(clusterLoaderConfig.ReportDir, "cl2-metadata.json")); err != nil {
+		// Do not Exitf on metadata.Dump error which is not critical.
```
mm4tt, line 305: I think the comment is redundant. The klog.Error is non critical as opposed to klog.Fatal

clusterloader2/pkg/metadata/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+
+	for k, v := range providerMetadata {
+		output[k] = v
+	}
+
+	return write(output, outputFile)
+}
```
mm4tt, line 45: NIT: Empty line before func write(...

clusterloader2/pkg/provider/aws.go
```diff
@@ -61,3 +62,8 @@ func (p *AWSProvider) RunSSHCommand(cmd, host string) (string, string, int, erro
 	user := defaultSSHUser()
 	return sshutil.RunSSHCommand(cmd, user, host, signer)
 }
+
+// Metadata returns provider-specific test run metadata.
```
mm4tt, line 66: These comments don't bring any value IMHO. I'd leave the comment in the interface but remove them from implementations.

mm4tt (commented):
Looks good, just some minor comments.

/lgtm
/hold

mborsz:
Fixed, ptal

Issue #1636, opened 2020-12-15
Validate config before running test


marseel:
**What would you like to be added**:
Validate if config is correct before running test.
**Why is this needed**:
Currently, step can be either measurement step or phase step (modules are being added in #1634)
https://github.com/kubernetes/perf-tests/blob/18e90fc65c6b95c0bff96938458a77d74e19b27a/clusterloader2/api/types.go#L60
Unfortunately, this behaviour is not well implemented:
https://github.com/kubernetes/perf-tests/blob/14afe3828ba8d9eaa98aa8a9f5ddd0cdc5f8eebf/clusterloader2/api/extensions.go#L54
https://github.com/kubernetes/perf-tests/blob/14afe3828ba8d9eaa98aa8a9f5ddd0cdc5f8eebf/clusterloader2/pkg/test/simple_test_executor.go#L143
Current implementation does not validate if config is correct before running test.
From user experience perspective it would be much better to return error right after test starts. 
Also, If I'm not wrong, if user specifies both measurements and phases, only measurements will be executed.

lyzs90:
@marseel could I help out with this one?

marseel:
@lyzs90  Yes sure, I would appreciate it :) 
If you will have some PR I can help with reviewing.

wojtek-t:
@lyzs90 - this long-standing PR is relevant to this one:
https://github.com/kubernetes/perf-tests/pull/142
It shows how should probably approach this problem - feel free to pick it up and continue working on that.

lyzs90:
Thanks @wojtek-t I'll check it out

/assign

lyzs90:
@wojtek-t I had a look at https://github.com/kubernetes/perf-tests/pull/142 and had some thoughts:
- Currently Modules are being recursively compiled into steps at [runtime](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/test/simple_test_executor.go#L125). Correct me if I am wrong but the mapping passed to the Module seems static and there is no use case where runtime variables are passed in (unlike Objects). Hence to support validation of Modules, may I suggest to move loading of modules before test execution - after https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/test/test.go#L66, so that validation of the entire Config (less Objects) can be validated ahead of time
[...]

PR #1680, opened 2021-01-25
Extract module compilation from test execution

lyzs90:
**What type of PR is this?**

/kind cleanup

This is to allow validation of the test config prior to running the test. The validation will be done in a follow up PR.

Part of https://github.com/kubernetes/perf-tests/issues/1636

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -31,9 +31,6 @@ type TestScenario struct {
 	Identifier string `json:"identifier"`
 	// ConfigPath defines path to the file containing a single Config definition.
 	ConfigPath string `json:"configPath"`
-	// OverridePaths defines what override files should be applied
-	// to the config specified by the ConfigPath.
-	OverridePaths []string `json:"overridePaths"`
 }
 
 // Config is a structure that represents configuration
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -283,8 +283,7 @@ func main() {
 	var prometheusFramework *framework.Framework
 	if clusterLoaderConfig.PrometheusConfig.EnableServer {
 		// Pass overrides to prometheus controller
-		clusterLoaderConfig.TestScenario.OverridePaths = testOverridePaths
-		if prometheusController, err = prometheus.NewController(&clusterLoaderConfig); err != nil {
+		if prometheusController, err = prometheus.NewController(&clusterLoaderConfig, testOverridePaths); err != nil {
 			klog.Exitf("Error while creating Prometheus Controller: %v", err)
 		}
 		prometheusFramework = prometheusController.GetFramework()
@@ -297,7 +296,7 @@ func main() {
 			klog.Exitf("Error while setting up exec service: %v", err)
 		}
 	}
-	if err := imagepreload.Setup(&clusterLoaderConfig, f); err != nil {
+	if err := imagepreload.Setup(&clusterLoaderConfig, f, testOverridePaths); err != nil {
 		klog.Exitf("Error while preloading images: %v", err)
 	}
 
@@ -318,8 +317,8 @@ func main() {
 		}
 
 		for i := range testSuite {
-			clusterLoaderConfig.TestScenario = testSuite[i]
-			ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter)
+			testScenario := testSuite[i]
+			ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter, &testScenario)
 			if !errList.IsEmpty() {
 				klog.Exitf("Test context creation failed: %s", errList.String())
 			}
@@ -337,9 +336,11 @@ func main() {
 		}
 	} else {
 		for i := range testConfigPaths {
-			clusterLoaderConfig.TestScenario.ConfigPath = testConfigPaths[i]
-			clusterLoaderConfig.TestScenario.OverridePaths = testOverridePaths
-			ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter)
+			testScenario := api.TestScenario{
+				ConfigPath:    testConfigPaths[i],
+				OverridePaths: testOverridePaths,
+			}
+			ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter, &testScenario)
 			if !errList.IsEmpty() {
 				klog.Exitf("Test context creation failed: %s", errList.String())
 			}
@@ -377,8 +378,7 @@ func runSingleTest(
 	ctx test.Context,
 	testConfig *api.Config,
 ) {
-	clusterLoaderConfig := ctx.GetClusterLoaderConfig()
-	testID := getTestID(clusterLoaderConfig.TestScenario)
+	testID := getTestID(*ctx.GetTestScenario())
 	testStart := time.Now()
 	printTestStart(testID)
 	errList := test.RunTest(ctx, testConfig)
@@ -387,7 +387,7 @@ func runSingleTest(
 	} else {
 		printTestResult(testID, "Success", "")
 	}
-	testConfigPath := clusterLoaderConfig.TestScenario.ConfigPath
+	testConfigPath := ctx.GetTestScenario().ConfigPath
 	ctx.GetTestReporter().ReportTestFinish(time.Since(testStart), testConfigPath, errList)
 }
 
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -55,7 +55,6 @@ var (
 	clusterLoaderConfig config.ClusterLoaderConfig
 	providerInitOptions provider.InitOptions
 	testConfigPaths     []string
-	testOverridePaths   []string
 	testSuiteConfigPath string
 )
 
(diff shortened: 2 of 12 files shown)


clusterloader2/pkg/test/modules.go
```diff
@@ -36,8 +36,8 @@ import (
 //         -step 5
 // - step 6
 // the method will return [step1, step3, step5, step6].
-func flattenModuleSteps(ctx Context, unprocessedSteps []api.Step) ([]*api.Step, error) {
-	var processedSteps []*api.Step
+func flattenModuleSteps(ctx Context, unprocessedSteps []api.Step) ([]api.Step, error) {
```
wojtek-t, line 39: Let's not change this - passing the whole objects around is a performance hit.

lyzs90: Updated, the ExecuteTest func signature will be slightly uglier now because we need to pass the slice of pointers down

mm4tt: I wonder whether we can change the CL2 API to use slices of pointers everywhere instead of slices of structs. E.g. [here](https://github.com/kubernetes/perf-tests/blob/039c642e4d11d8c4c1a438f5e8bc2b56c1fc7eef/clusterloader2/api/types.go#L50) we would change
```
Steps []Step `json:"steps"`
```
to 
```
Steps []*Step `json:"steps"`
```

IIUC, it should be transparent to the users (no changes in the yaml files would be needed), serialization code would work without any changes and we wouldn't have to conversion from `[]A` to '[]*A` in the codebase. Thoughts?

lyzs90: It'll make the Config struct less awkward to work with and I can make this PR a lot cleaner

lyzs90:
@mm4tt I split `RunTest` into three functions. In the clusterloader main function, `CreateTestContext`, `CompileTestConfig` are invoked on all tests ahead of time. I also made the api change to slice of pointers.

mm4tt:
Nice! I think the api change is something we want to do anyway as it simplifies the code and makes it more performant (less object copying). Could you extract that to a separate PR and send it to me? In that PR, can we change not only steps but other slices we have in the api as well?  E.g. phases, measurements, etc. There shouldn't be a lot of them and the needed code changes should be mechanical.

lyzs90:
rebased and squashed

mm4tt:
So sorry, it fell off my radar. Will take a look shortly

clusterloader2/cmd/clusterloader.go
```diff
@@ -307,20 +307,53 @@ func main() {
...
+			testConfigs = append(testConfigs, testConfig)
+		}
+
+		for i := range testSuite {
+			runSingleTest(contexts[i], testConfigs[i])
 		}
 	} else {
 		for i := range testConfigPaths {
 			clusterLoaderConfig.TestScenario.ConfigPath = testConfigPaths[i]
 			clusterLoaderConfig.TestScenario.OverridePaths = testOverridePaths
```
mm4tt, line 341: Hmm, I'm not sure whether changing the clusterLoaderConfig and then passing it by reference (pointer) is correct. I think it's not, as the second iteration will modify the context we created in the first iteration (because it'll modify the clusterLoaderConfig which is referenced in the testContext). Do I miss something? 

If I'm right, I believe we should get rid of the TestScenario field from the ClusterLoaderConfig and move it to the TestContext, thoughts?

lyzs90: Good catch, I think you are right. I should be able to move it to test context (the alternative is to make copies of ClusterLoaderConfig which we are trying to avoid)

lyzs90: I've decoupled TestScenario from ClusterLoaderConfig here https://github.com/kubernetes/perf-tests/pull/1680/commits/b793c9e39c2ca7433520d629aed578c06b3c1e3e

lyzs90: Did a bit of digging to understand why it still worked for the previous [run](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/perf-tests/1680/pull-perf-tests-clusterloader2/1354805480483983360). It was because the testConfigs were [compiled](https://github.com/kubernetes/perf-tests/pull/1680/files#diff-d6057f57c97c93f50c6e310f19970261bbd5d145d2f9682473e3749afe7abb97R349) and passed directly to the test executor. So leaving TestScenario in ClusterLoaderConfig indeed doesn't make sense any more.

clusterloader2/cmd/clusterloader.go
```diff
@@ -341,21 +375,20 @@ func main() {
...
 
 func runSingleTest(
-	f *framework.Framework,
-	prometheusFramework *framework.Framework,
-	testReporter test.Reporter,
+	ctx test.Context,
+	testConfig *api.Config,
 ) {
-	testID := getTestID(clusterLoaderConfig.TestScenario)
+	testID := getTestID(*ctx.GetTestScenario())
```
mm4tt, line 381: Can we change getTestId to accept pointer ?

clusterloader2/cmd/clusterloader.go
```diff
@@ -283,8 +283,7 @@ func main() {
 	var prometheusFramework *framework.Framework
 	if clusterLoaderConfig.PrometheusConfig.EnableServer {
 		// Pass overrides to prometheus controller
-		clusterLoaderConfig.TestScenario.OverridePaths = testOverridePaths
-		if prometheusController, err = prometheus.NewController(&clusterLoaderConfig); err != nil {
+		if prometheusController, err = prometheus.NewController(&clusterLoaderConfig, testOverridePaths); err != nil {
```
mm4tt, line 286: Actually the `testOverridePaths` is global and it'd be totally reasonable to have it as a field in the ClusterLoaderConfig. The benefit is that we won't have to modify these signatures to accept the `testOverridePaths` in addition to `clusterLoaderConfig`

clusterloader2/pkg/test/test.go
```diff
@@ -37,41 +38,77 @@ var (
...
+
+	mapping, errList := config.GetMapping(clusterLoaderConfig, testScenario.OverridePaths)
+	if errList != nil {
+		return nil, errList
 	}
+
+	return CreateContext(clusterLoaderConfig, clusterFramework, prometheusFramework, state.NewState(), testReporter, mapping, testScenario), errors.NewErrorList()
+}
+
+// CompileTestConfig loads the test configuration and nested modules
```
mm4tt, line 67: Nit: Dot at the end of doc comment.

clusterloader2/cmd/clusterloader.go
```diff
@@ -307,20 +306,55 @@ func main() {
...
+			contexts = append(contexts, ctx)
+
+			testConfig, errList := test.CompileTestConfig(ctx)
+			if !errList.IsEmpty() {
+				klog.Exitf("Test compliation failed: %s", errList.String())
+			}
+			testConfigs = append(testConfigs, testConfig)
+		}
+
+		for i := range testConfigPaths {
```
mm4tt, line 356: This is duplicated, we could move it outside if-else and have it only once:
```
for i := range contexts {
   runSingleTest(...)
}
```

clusterloader2/cmd/clusterloader.go
```diff
@@ -307,20 +306,55 @@ func main() {
 
 	testReporter := test.CreateSimpleReporter(path.Join(clusterLoaderConfig.ReportDir, "junit.xml"), "ClusterLoaderV2")
 	testReporter.BeginTestSuite()
+
+	var contexts []test.Context
+	var testConfigs []*api.Config
```
mm4tt, line 311: It looks like there is a 1-1 relationship between a `testContext` and its `testConfig`. What's more, a `testConfig` can be created from a `testContext`. Given that, I'm not sure whether it makes sense to keep these two objects separately. We could either introduce a `ConfigWithContext` object or just add `testConfig` to the `Context`.  Thoughts?

mm4tt (commented):
Nice!

clusterloader2/cmd/clusterloader.go
lyzs90: Adding `testConfig` to the `Context` makes more sense to me given that we have already moved `TestScenarios` in there. Or perhaps even added to TestScenario?

mm4tt: Yeah, so let's do that (move testConfig to the context). Other than that, this PR looks great!

mm4tt (commented):
/lgtm
/approve

clusterloader2/api/types.go
```diff
@@ -31,9 +31,6 @@ type TestScenario struct {
 	Identifier string `json:"identifier"`
 	// ConfigPath defines path to the file containing a single Config definition.
 	ConfigPath string `json:"configPath"`
-	// OverridePaths defines what override files should be applied
-	// to the config specified by the ConfigPath.
-	OverridePaths []string `json:"overridePaths"`
```
wojtek-t, line 36: Wait - why this was removed. How can I pass it now?
What happens with this one:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/experimental/storage/pod-startup/suite.yaml#L3

mm4tt: Good point, we shouldn't have removed OverridePaths from TestScenario. Let me revert the change, we can roll-forward with the fix, which will be simply leaving this field as is.

lyzs90: Oh I thought it would be passed as a global https://github.com/kubernetes/perf-tests/pull/1680/commits/4cf2c18e872d6bf18122717b66ac1787f9320744#diff-d6057f57c97c93f50c6e310f19970261bbd5d145d2f9682473e3749afe7abb97R119

But I didn't realize it could be scenario specific for test suites

mm4tt: Yeah, my bad, I didn't explain it clear enough. The one passed via a flag is global, but it's also possible to do it per test scenario. I wanted to move only the global one

lyzs90: Ah okay then I'll raise another PR later to revert the revert and apply a fix for suites

PR #1681, opened 2021-01-26
Add huge services to load test

tosi3k:
This is to add a way of testing the scalability of EndpointSlice objects.

/sig scalability


clusterloader2/testing/load/config.yaml
```diff
@@ -428,6 +441,9 @@ steps:
         Group: scheduler-throughput
         CpuRequest: 1m
         MemoryRequest: 10M
+{{if $ENABLE_HUGE_SERVICES}}
+        SvcName: huge-service
```
wojtek-t, line 445: What is this used for?

tosi3k: Needed [here](https://github.com/kubernetes/perf-tests/pull/1681/files#diff-c83abe3d6dd61012c9df0cdc289fad5c771e1f8c634f622e4285b07563870068R19-R21) as label based on which the huge service [selects](https://github.com/kubernetes/perf-tests/blob/e34912ed01fda93e0e054a544e107f443498ab9e/clusterloader2/testing/load/service.yaml#L13) pods to front.

PR #1684, opened 2021-01-27
Add kube-state-metrics latency measurement 

lilic:
Hello, as discussed and asked on the SIG scalability [mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-scale/UHx_ISEHT24/TEeAyVn7CAAJ), we need a performance test for kube-state-metrics. I followed the suggestions there, and hopefully, this is correct!

**What type of PR is this?**
/kind feature

This is needed as the last step of https://github.com/kubernetes/kube-state-metrics 2.0 release. As mentioned in the mailing list, we want to run performance and scale tests to ensure we can run on a certain amount of nodes, but also to give our users some expectations on the number of resources needed, but also the latency they can expect.

kube-state-metrics has two /metrics endpoints, one /metrics that serve metrics about Kubernetes objects and one that is a self metrics about kube-state-metrics itself. In this test we first query the main /metrics endpoint and after that get the latency metrics on the self /metrics endpoint. We do this every 10 seconds, and until the gather action is called. When gather is called we create a summary report based on the collected latency metrics quantiles.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1341.

- I deployed kube-state-metrics as part of exports, this was just initial thought, open to whatever you suggest here, thanks!
- I can move the exposed functions I reused from other tests to util package, any thoughts on this? Or leave as is?
- Finally how do we test these? I tested it locally against my cluster by importing the newly created `CreateKSMLatencyMeasurement ` function and running start and gather actions and got summary correctly. Is that enough?

```release-note
NONE
```

```docs

```

Thank you!! 🎉 

cc @brancz as mantainer of kube-state-metrics, as discussed a while ago here are the perf tests.

### DIFF
--- a/clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
+++ b/clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
@@ -1,0 +1,213 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+
+	"github.com/prometheus/common/model"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+const (
+	ksmLatencyName               = "KubeStateMetricsLatency"
+	ksmRequestDurationMetricName = model.LabelValue("http_request_duration_seconds_bucket")
+	probeIntervalDefault         = 30 * time.Second
+	ksmNamespace                 = "kube-state-metrics-perf-test"
+	ksmServiceName               = "kube-state-metrics"
+	ksmSelfPort                  = 8081
+	ksmMetricsPort               = 8080
+)
+
+type ksmLatencyMeasurement struct {
+	ctx            context.Context
+	cancel         func()
+	isRunning      bool
+	namespace      string
+	serviceName    string
+	metricsPort    int
+	selfPort       int
+	initialLatency *measurementutil.Histogram
+	wg             sync.WaitGroup
+}
+
+func init() {
+	if err := measurement.Register(ksmLatencyName, CreateKSMLatencyMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", ksmLatencyName, err)
+	}
+}
+
+// CreateKSMLatencyMeasurement creates a new Kube State
+// Metrics Measurement.
+func CreateKSMLatencyMeasurement() measurement.Measurement {
+	ctx, cancel := context.WithCancel(context.Background())
+	return &ksmLatencyMeasurement{
+		namespace:   ksmNamespace,
+		serviceName: ksmServiceName,
+		selfPort:    ksmSelfPort,
+		metricsPort: ksmMetricsPort,
+		ctx:         ctx,
+		cancel:      cancel,
+	}
+}
+
+// Execute supports two actions:
+// - start - starts goroutine and queries /metrics every probeIntervalDefault interval,
+// it also collects initial latency metrics.
+// - gather - gathers latency metrics and creates a latency summary.
+func (m *ksmLatencyMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	client := config.ClusterFramework.GetClientSets().GetClient()
+	switch action {
+	case "start":
+		if m.isRunning {
+			klog.V(2).Infof("%s: measurement already running", m)
+			return nil, nil
+		}
+		// Start executing calls towards the kube-state-metrics /metrics endpoint
+		// every probeIntervalDefault until gather is called.
+		// probeIntervalDefault equals the scrapping interval we suggest.
+		// If we cannot get metrics for two minutes we are already going over
+		// the scrape interval so we should cancel.
+		m.startQuerying(m.ctx, client, probeIntervalDefault)
+		// Retrieve initial latency when first call is done.
+		m.initialLatency, err = m.retrieveKSMLatencyMetrics(m.ctx, client)
+		return nil, err
+	case "gather":
+		defer m.cancel()
+		return m.createKSMLatencySummary(m.ctx, client)
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (m *ksmLatencyMeasurement) stop() error {
+	if !m.isRunning {
+		return fmt.Errorf("%s: measurement was not running", m)
+	}
+	m.cancel()
+	m.wg.Wait()
+	return nil
+}
+
+// createKSMLatencyReport gathers the latency one last time and creates the summary based on the Quantile from the sub histograms.
+// Afterwards it creates the Summary Report.
+func (m *ksmLatencyMeasurement) createKSMLatencySummary(ctx context.Context, client clientset.Interface) ([]measurement.Summary, error) {
+	latestLatency, err := m.retrieveKSMLatencyMetrics(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+	if err = m.stop(); err != nil {
+		return nil, err
+	}
+	// We want to subtract the latest histogram from the first one we collect.
+	finalLatency := HistogramSub(latestLatency, m.initialLatency)
+	// Pretty Print the report.
+	result := &measurementutil.LatencyMetric{}
+	if err = SetQuantileFromHistogram(result, finalLatency); err != nil {
+		return nil, err
+	}
+	content, err := util.PrettyPrintJSON(result)
+	if err != nil {
+		return nil, err
+	}
+	// Create Summary.
+	return []measurement.Summary{measurement.CreateSummary(ksmLatencyName, "json", content)}, nil
+}
+
+// startQuerying queries /metrics endpoint of kube-state-metrics kube_ metrics every interval
+// and stops when stop is called.
+func (m *ksmLatencyMeasurement) startQuerying(ctx context.Context, client clientset.Interface, interval time.Duration) {
+	m.isRunning = true
+	m.wg.Add(1)
+	go m.queryLoop(ctx, client, interval)
+}
+
+func (m *ksmLatencyMeasurement) queryLoop(ctx context.Context, client clientset.Interface, interval time.Duration) {
+	defer m.wg.Done()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(interval):
+			var output string
+			output, err := m.getMetricsFromService(ctx, client, m.metricsPort)
+			if err != nil {
+				klog.V(2).Infof("error during fetching metrics from service: %v", err)
+			}
+			if output == "" {
+				klog.V(2).Infof("/metrics endpoint of kube-state-metrics returned no data in namespace: %s from service: %s port: %d", m.namespace, m.serviceName, m.metricsPort)
+			}
+
+		}
+	}
+}
+
+func (m *ksmLatencyMeasurement) retrieveKSMLatencyMetrics(ctx context.Context, c clientset.Interface) (*measurementutil.Histogram, error) {
+	ksmHist := measurementutil.NewHistogram(nil)
+	output, err := m.getMetricsFromService(ctx, c, m.selfPort)
+	if err != nil {
+		return ksmHist, err
+	}
+	samples, err := measurementutil.ExtractMetricSamples(output)
+	if err != nil {
+		return ksmHist, err
+	}
+	for _, sample := range samples {
+		switch sample.Metric[model.MetricNameLabel] {
+		case ksmRequestDurationMetricName:
+			measurementutil.ConvertSampleToHistogram(sample, ksmHist)
+		}
+	}
+	return ksmHist, nil
+}
+
+func (m *ksmLatencyMeasurement) getMetricsFromService(ctx context.Context, client clientset.Interface, port int) (string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+	out, err := client.CoreV1().RESTClient().Get().
+		Resource("services").
+		SubResource("proxy").
+		Namespace(m.namespace).
+		Name(fmt.Sprintf("%v:%v", m.serviceName, port)).
+		Suffix("metrics").
+		Do(ctx).Raw()
+	return string(out), err
+}
+
+// Dispose cleans up after the measurement.
+func (m *ksmLatencyMeasurement) Dispose() {
+	if err := m.stop(); err != nil {
+		klog.V(2).Infof("error during dispose call: %v", err)
+	}
+}
+
+// String returns string representation of this measurement.
+func (m *ksmLatencyMeasurement) String() string {
+	return ksmLatencyName
+}
--- a/clusterloader2/pkg/measurement/common/scheduler_latency.go
+++ b/clusterloader2/pkg/measurement/common/scheduler_latency.go
@@ -144,8 +144,8 @@ func (*schedulerLatencyMeasurement) String() string {
 	return schedulerLatencyMetricName
 }
 
-// histogramSub is a helper function to substract two histograms
-func histogramSub(finalHist, initialHist *measurementutil.Histogram) *measurementutil.Histogram {
+// HistogramSub is a helper function to substract two histograms
+func HistogramSub(finalHist, initialHist *measurementutil.Histogram) *measurementutil.Histogram {
 	for k := range finalHist.Buckets {
 		finalHist.Buckets[k] = finalHist.Buckets[k] - initialHist.Buckets[k]
 	}
@@ -154,17 +154,17 @@ func histogramSub(finalHist, initialHist *measurementutil.Histogram) *measuremen
 
 func (m *schedulerLatencyMetrics) substract(sub schedulerLatencyMetrics) {
 	if sub.preemptionEvaluationHist != nil {
-		m.preemptionEvaluationHist = histogramSub(m.preemptionEvaluationHist, sub.preemptionEvaluationHist)
+		m.preemptionEvaluationHist = HistogramSub(m.preemptionEvaluationHist, sub.preemptionEvaluationHist)
 	}
 	if sub.schedulingAlgorithmDurationHist != nil {
-		m.schedulingAlgorithmDurationHist = histogramSub(m.schedulingAlgorithmDurationHist, sub.schedulingAlgorithmDurationHist)
+		m.schedulingAlgorithmDurationHist = HistogramSub(m.schedulingAlgorithmDurationHist, sub.schedulingAlgorithmDurationHist)
 	}
 	if sub.e2eSchedulingDurationHist != nil {
-		m.e2eSchedulingDurationHist = histogramSub(m.e2eSchedulingDurationHist, sub.e2eSchedulingDurationHist)
+		m.e2eSchedulingDurationHist = HistogramSub(m.e2eSchedulingDurationHist, sub.e2eSchedulingDurationHist)
 	}
 	for _, ep := range extentionsPoints {
 		if sub.frameworkExtensionPointDurationHist[ep] != nil {
-			m.frameworkExtensionPointDurationHist[ep] = histogramSub(m.frameworkExtensionPointDurationHist[ep], sub.frameworkExtensionPointDurationHist[ep])
+			m.frameworkExtensionPointDurationHist[ep] = HistogramSub(m.frameworkExtensionPointDurationHist[ep], sub.frameworkExtensionPointDurationHist[ep])
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


brancz:
This is really awesome. Thank you so much for taking care of this!

/lgtm

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,221 @@
...
+	return &ksmLatencyMeasurement{
+		namespace:   "kube-state-metrics-perf-test",
+		serviceName: "kube-state-metrics",
+		selfPort:    8081,
+		metricsPort: 8080,
+	}
+}
+
+// Execute supports two actions:
+// - start -  query on /metrics and gathers initial latency data.
```
wojtek-t, line 72: nit: it's not exactly accurate - it's just starts a goroutine that is peridically calling /metrics on KSM, right?

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,221 @@
...
+// - gather - gets the latency data again and creates a latency summary.
+func (m *ksmLatencyMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	probeDuration := 10 * time.Second
```
wojtek-t, line 79: make it a class-level constant (probeDurationDefault or sth like that)

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,221 @@
...
+
+	switch action {
+	case "start":
+		if m.isRunning {
+			klog.V(2).Infof("%s: measurement already running", m)
+			return nil, nil
+		}
+		m.stopCh = make(chan struct{})
+		err = m.start(client, probeDuration)
+		if err != nil {
```
wojtek-t, line 90: just replace these 5 lines with:
```
return nil, m.start(client, probeDuration)
```

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,221 @@
...
+	var latHist *measurementutil.Histogram
+	go func() {
+		defer m.wg.Done()
+		for {
+			select {
+			case <-m.stopCh:
+				return
+			case <-time.After(probeDuration):
+				latHist, err = m.retrieveKSMLatencyMetrics(client, probeDuration)
+				m.latency = append(m.latency, ksmLatencyMetrics{
```
wojtek-t, line 138: Why we're storing every single call, given that we seem to be computing the result from the first and last call only?

In other words, based on the getKSMLatency() method, I'm wondering, why the pattern isn't:
- start: "get KSM metrics"
- gather: "get KSM metric, substract those two results"
- ?

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,221 @@
...
+			Resource("services").
+			SubResource("proxy").
+			Namespace(m.namespace).
+			Name(fmt.Sprintf("%v:%v", m.serviceName, m.selfPort)).
+			Suffix("metrics").
+			Do(context.TODO()).Raw()
+		finished <- struct{}{}
+	}()
+	select {
+	case <-time.After(probeDuration):
```
wojtek-t, line 205: Hmm - this handling seems a bit risky, because you're effectively not killing that goroutine.
So you may stack forever it it (and accumulate many of those in theory).

Why is this even needed?

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Currently, we are storing all calls yes, this was mainly as in the future I also want to add a few other things which we need storing for. So I might have prematurely optimized for that in mind. But yes just getting two calls, first and last is enough, can change and in the future when I need to can revert to this?

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: yes, that is more accurate indeed

mm4tt:
Before we get into a detailed review, I'd like to understand the motivation here and make sure the changes you're making align with this motivation and sig-scalability goals for CI K8s scale tests.

1. Do you want to run the tests continuously or do you just want to test the v2 release (for which it might be enough to run a series of one-off tests)?
1. If a one-off testing is enough for you, do you really need to implement a new measurement? Wouldn't you be able to verify whether kube-state-metrics v2 scales by looking at metrics in grafana? Do you plan to maintain / extend this measurement in the future?
1. Aside from testing the new release kube-state-metrics, do we know how safe it is to enable it in a 5k node cluster? I'd like to avoid a situation where enabling it breaks our expensive CI tests.
1. Are you planning to run the tests yourself or do you need our help / resources (e.g. test projects)?

If there are more things to discuss, we have the sig-scale meeting this week - https://github.com/kubernetes/community/tree/master/sig-scalability#meetings. I'm more than happy to discuss it there.

lilic:
Thanks @mm4tt I answered the questions below:

1. We would need to at least run this test on each release of kube-state-metrics. We release at least when Kubernetes versions are released, so that cadence would be fine. How often do you run the scale tests?
2. That was my question in the initial mailing list and the suggestion was to add a measurement hence I added one. Either works, but since we already put in the work for this, having these automatic reports is useful going forward for next releases!
3. Enabling should be safe, I can't think of any problems, don't think it can blow up any nodes, or at least it has not yet. To be safe we can skip ingesting the metrics into Prometheus, to prevent any cardinality explosions. I can simply remove the ServiceMonitor if this is the fear?
4. As mentioned in the [initial mailing list](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-scale/UHx_ISEHT24/TEeAyVn7CAAJ), we would need the resources to run these, the project is under the k8s org and one of the core projects of SIG instrumentation. I ran this test against my very small cluster, so I can run against larger clusters if given access, if that is what is meant here?

Just a side note: as mentioned in one the comments, in the future we want to add other tests to this measurement, this is currently enough to pass the initial 2.0 release but we have some ideas for further performance metrics to further optimize.

Thank you!

mm4tt:
Thanks @lilic. If you plan to run the test periodically then having the measurement makes sense to automate the verification. 


I'm worried that KSM may be causing issues in our 5k node tests. IUUC, it LISTs and WATCHes a lot of k8s resources. I imagine it can eat a lot of resources (cpu/mem) and if it's restarted for some reason (e.g. OOM) the initialization logic (re-issuing all the LIST calls) may cause temporarily overloading of the k8s control-plane, which in turn may affect the results of our scale tests. Because of that I propose the following:

1. Make all the changes you're making flag/config gated and disabled by default
1. Ideally, split this PR into a few smaller. For example, one PR for implementing the measurement, second for enabling (in a gated way) KSM in our prometheus stack. Third for extending the load test to enable the measurement (also gated).
1. We won't make the KSM test part of our daily load test. There are multiple reasons for that, I'm happy to discuss this in detail during tomorrow's SIG Scale meeting.  Instead we'll create a dedicated test, based on our load test for KSM testing. In fact, it will be the regular load test just configured to enable KMS and KMS measurement. IIUC, you need 1000 nodes and running the test 1-2 per month should be enough for you, right?

Let me know if this sounds reasonable.

lilic:
Yes we do watch all the native resources indeed. 1k nodes and 1-2 times per month should be plenty yes if we need it more frequently or higher loads for some reason we will just come back to this at another point. That makes perfect sense to me sounds like a good plan to move forward, thanks @mm4tt! 

Will take the manifests out of this PR, do you have an example of doing gated tests via flag for any existing tests somewhere?

mm4tt:
Yes, some examples:

- changing manifest conditionally - https://github.com/mm4tt/perf-tests/blob/7aadff231121cfbc9cb26279f47db8ee8f662f53/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L20
   - Search for `PROMETHEUS_SCRAPE_KUBELETS` in the golang files to see how this variable is propagated into the templates
- applying (or not) a whole manifest conditionally - https://github.com/kubernetes/perf-tests/blob/d0680bd97db6a08d42b9e94abc0becedbacde0e1/clusterloader2/pkg/prometheus/prometheus.go#L187-L194

Similarly in the load test, enabling some measurement conditionally - https://github.com/kubernetes/perf-tests/blob/ece6960810e96260f6b99443c0280de4d384b13b/clusterloader2/testing/load/config.yaml#L863

lilic:
Ah nice, thanks! But the measurement can still stay in the same directory as I have it now? e.g. won't the `measurement.Register(ksmLatencyName, CreateKSMLatencyMeasurement)` be called? Or is this done via the load/config.yaml step?

mm4tt:
The measurement implementation doesn't have to (and shoudln't) be gated. It can be registered, it will be a no-op as long as it's not used inside a test config (yaml file).

lilic:
Yes makes sense, will address the existing comments and make the changes, thanks!

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Yes I copied this over from another file that gather metrics from this directory, but you are right, I see two options here:
1. Pass context with a timeout and cancel the context, this should fix this issue.
2. Or if you think we should switch to collecting only at the start and end of measurement as per your previous comment, we then won't have this problem.

Let me know which option you would prefer here, thank you!

lilic:
I addressed the majority of the comments that @wojtek-t had, thanks for those! I have one major thing left, after we decide in which direction we want to take it, I will finish up addressing those.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
wojtek-t: I would prefere (1), but we would probably need to update client library to allow passing context to the calls.

If you have time to do that - it would be amazing.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
wojtek-t: Yeah - we can always extend and start gathering more information if needed.
I wouldn't over-engineer for things that "we may potentially need in the future".

lilic:
@mm4tt @wojtek-t @brancz please have another look. I changed that now we collect the latency metrics twice, once on start and once on gather, so the flow is the following:
On start action we do:
1. we start querying the /metrics endpoint of kube-state-metrics kube_ metrics (this is not self metrics), we do this every x interval (30 seconds) to gather enough data as nothing has been scrapping kube-state-metrics yet.
2. we get initial latency metrics
On gather:
1. We stop querying /metrics endpoint.
2. We gather latency metrics one more time and create a summary.

Let me know what you think, thank you!!

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+// Execute supports two actions:
+// - start - starts goroutine and queries /metrics every probeIntervalDefault interval,
+// it also collects initial latency metrics.
+// - gather - gathers latency metrics and creates a latency summary.
+func (m *ksmLatencyMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	if m.ctx == nil {
+		m.ctx, m.cancel = context.WithCancel(context.Background())
```
wojtek-t, line 76: There is no way to pass it here.

So given that, please simply initialize this in CreateKSMLatencyMeasurement

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+type ksmLatencyMeasurement struct {
+	ctx         context.Context
+	cancel      func()
+	isRunning   bool
+	stopCh      chan struct{}
+	namespace   string
+	serviceName string
+	metricsPort int
+	selfPort    int
+	latency     *measurementutil.Histogram
```
wojtek-t, line 49: Let's maybe call it "initialLatency" or "initialiLatencies" or "initialHistogram" or sth like that.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := measurement.Register(ksmLatencyName, CreateKSMLatencyMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", ksmLatencyName, err)
+	}
+}
+
+// CreateKSMLatencyMeasurement creates a new Kube State
+// Metrics Measurement.
+func CreateKSMLatencyMeasurement() measurement.Measurement {
+	return &ksmLatencyMeasurement{
+		namespace:   "kube-state-metrics-perf-test",
```
wojtek-t, line 63: please define constants for those

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+	err = SetQuantileFromHistogram(result, finalLatency)
+	if err != nil {
+		return nil, err
+	}
+	content, err := util.PrettyPrintJSON(result)
+	if err != nil {
+		return nil, err
+	}
+	// Create Summary.
+	return []measurement.Summary{measurement.CreateSummary(schedulerLatencyMetricName, "json", content)}, nil
```
wojtek-t, line 136: not sure what the name should be, but definitely not "schedulerLatencyMetricName" :)

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,201 @@
...
+
+// startQuerying queries /metrics endpoint of kube-state-metrics kube_ metrics every interval
+// and stops when stop is called.
+func (m *ksmLatencyMeasurement) startQuerying(ctx context.Context, client clientset.Interface, interval time.Duration) error {
+	m.isRunning = true
+	m.wg.Add(1)
+	var err error
+	go func() {
+		defer m.wg.Done()
+		for {
```
wojtek-t, line 147: Can you please move L147-L158 to a separate function (for easier readability)

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: hah oops good catch :D

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Do you mean the entire go func or just the inner for loop?

wojtek-t: Either would work :)
Moving the whole function is probably cleaner.

lilic: @wojtek-t Thanks, I addressed my comments, please have another look.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+	"sync"
+	"time"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+
+	"github.com/prometheus/common/model"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
```
wojtek-t, line 31: nit: can you please sort the non-builtin imports alphabetically?

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil, err
+	}
+	client := config.ClusterFramework.GetClientSets().GetClient()
+	switch action {
+	case "start":
+		if m.isRunning {
+			klog.V(2).Infof("%s: measurement already running", m)
+			return nil, nil
+		}
+		m.stopCh = make(chan struct{})
```
wojtek-t, line 93: Do you even need stopCh?

I think instead of introducing stopCh, you should simply call m.cancel() and just wait on that whereever needed.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+	m.wg.Wait()
+	latestLatency, err := m.retrieveKSMLatencyMetrics(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+	// We want to subtract the latest histogram from the first one we collect.
+	finalLatency := HistogramSub(latestLatency, m.initialLatency)
+	// Pretty Print the report.
+	result := &measurementutil.LatencyMetric{}
+	err = SetQuantileFromHistogram(result, finalLatency)
```
wojtek-t, line 128: nit:
if err := SetQu...(..); err != nil {

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+		SubResource("proxy").
+		Namespace(m.namespace).
+		Name(fmt.Sprintf("%v:%v", m.serviceName, port)).
+		Suffix("metrics").
+		Do(ctx).Raw()
+	return string(out), err
+}
+
+// Dispose cleans up after the measurement.
+func (m *ksmLatencyMeasurement) Dispose() {}
```
wojtek-t, line 204: Dispose should stop the measurement if it's still running.

So basically, you should probably:
- extract lines 114-119 to a separate stop() method
- call stop() there
- call stop here in Dispose method

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,209 @@
...
+		switch sample.Metric[model.MetricNameLabel] {
+		case ksmRequestDurationMetricName:
+			measurementutil.ConvertSampleToHistogram(sample, ksmHist)
+		}
+	}
+	return ksmHist, nil
+}
+
+func (m *ksmLatencyMeasurement) getMetricsFromService(ctx context.Context, client clientset.Interface, port int) (string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
```
wojtek-t, line 191: Isn't 2 minutes too high? Can we align this with the 30s interval period?

wojtek-t (commented):
This looks great - I just added a couple smaller comments.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Yes 2 minutes is above the 30 second scrape interval we recommend, but timeout should be a bit above 30 seconds no? So 40 seconds? Not sure if that makes sense.

cc @brancz any suggestions?

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Thanks good to know! Will do. FYI I copied from other files, might be good to address it there as well.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: I can try, but vim-go sorts it automatically this way for me.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
brancz: If we want to align it with the interval, then I'd actually prefer to make the interval 2m as well. The point of this is kind of to see how it behaves and if it's over 30s then we'd want to understand how much over 30s it is (this is effectively a bulk endpoint so 30s is not unthinkable in extremely huge clusters). 2m is so high that if it reaches that value it should legitimately just fail.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
wojtek-t: Many measurements don't start anything - they just do something on start and something on gather but they don't start any goroutines or anything like that (in which case nothing more is needed there).
There are certainly measurements that do that, e.g.
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go#L152

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Update: I ran both gofmt as well golang.org/x/tools/cmd/goimports and both did not change the ordering here. If we switch the ordering, next time someone with gofmt plugin like mine modifies we will have changes.

Curious to know what do you use?

wojtek-t: gofmt doesn't change order of important (I don't know about goimports).

I'm just using my eyes :)

lilic: From what I can tell, it orders it by org, then repo, think that makes sense here. vim-go has both goftm and goimports.

https://goinbigdata.com/goimports-vs-gofmt/

lilic:
@wojtek-t thanks so much for your comments, I believe I addressed all but the imports as formatting seems to be correct judging by my investigations. Can you please have another look, thanks!

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
```diff
@@ -0,0 +1,214 @@
...
+
+// createKSMLatencyReport gathers the latency one last time and creates the summary based on the Quantile from the sub histograms.
+// Afterwards it creates the Summary Report.
+func (m *ksmLatencyMeasurement) createKSMLatencySummary(ctx context.Context, client clientset.Interface) ([]measurement.Summary, error) {
+	latestLatency, err := m.retrieveKSMLatencyMetrics(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+	if err = m.stop(); err != nil {
+		klog.V(2).Info(err)
```
wojtek-t, line 126: Let's remove this log (the error will be logged on higher levels anyway - it's propagated out of measurement).

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
wojtek-t: OK - we can tune that later too.

wojtek-t (commented):
One last minor nit. Other than that LGTM.

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Done, makes sense!

clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
lilic: Sounds good to me!

lilic:
@wojtek-t done, please have another look, thanks!

wojtek-t:
/lgtm
/approve

thanks!

PR #1690, opened 2021-01-28
modify the error url of override

timyinshi:
Signed-off-by: timyinshi <shiguangyin@inspur.com>

/kind bug
/sig docs

modify the error url of override
the error url is : https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/density/5000_nodes/override.yaml
the right url is : https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/density/scheduler/pod-affinity/overrides.yaml


mm4tt:
/lgtm
/approve

Thanks!

PR #1692, opened 2021-01-28
Customize enabling of immutable secrets and configmaps in CL2

tosi3k:
/assign @wojtek-t


mm4tt:
I'm afraid this is not enough. To make it work we'll also need to take into account changes made in the deployment.yaml in the https://github.com/kubernetes/perf-tests/commit/688f020031489a6ecdf2afea92323a8f2d99d80b

tosi3k:
> I'm afraid this is not enough. To make it work we'll also need to take into account changes made in the deployment.yaml in the [688f020](https://github.com/kubernetes/perf-tests/commit/688f020031489a6ecdf2afea92323a8f2d99d80b)

Good point, thanks. Had to perform larger config gymnastics because of that :grinning:.

wojtek-t:
TBH - because of that, I'm not the fan of this change...

Issue #1693, opened 2021-01-30
kind provider for CL2


lyzs90:
I got really close to running CL2 on a [kind](https://kind.sigs.k8s.io/) cluster with the local provider. The first issue I faced was having to wrap the poll eg. `docker exec -it <containerID> curl -X GET -k https://localhost:10259/metrics` during the metric collection. The second was because of unauthenticated access to the scheduler's metrics endpoint (somewhat related to https://github.com/kubernetes/kubernetes/issues/93748).

If we can get these out of the way, I'm wondering if we could consider supporting a kind provider. It would make it easy for anyone to start working on CL2

mm4tt:
It'd be great even to have kind working without those features. If we need to disable them only for kind we can do that using the [provider](https://github.com/kubernetes/perf-tests/blob/fd7ccf5b51ae0b54e6c2152f54c3c51e56147c83/clusterloader2/pkg/provider/provider.go) interface. After that we can open an issues to make them work for kind and work on them independently

lyzs90:
Closing as it was added in https://github.com/kubernetes/perf-tests/pull/1701.

PR #1701, opened 2021-02-04
Add CL2 kind provider

lyzs90:
**What type of PR is this?**

/kind feature

Adds a kind provider for CL2. More details here: https://github.com/kubernetes/perf-tests/issues/1693

Part of https://github.com/kubernetes/perf-tests/issues/1693

SSH to master to be tackled in a separate PR

```release-note
Support kind as a CL2 provider
```

```docs

```

### DIFF
--- a/clusterloader2/README.md
+++ b/clusterloader2/README.md
@@ -4,13 +4,18 @@
 
 To run ClusterLoader type:
 ```
-go run cmd/clusterloader.go --kubeconfig=kubeConfig.yaml --testconfig=config.yaml
+go run cmd/clusterloader.go --kubeconfig=kubeConfig.yaml --testconfig=config.yaml --provider=local
 ```
 OR
 ```
-./run-e2e.sh --testconfig=config.yaml
+./run-e2e.sh --testconfig=config.yaml --provider=local
 ```
-Flags kubeconfig and testconfig are necessary.
+
+The simplest way to get acquainted with ClusterLoader is using [kind](https://kind.sigs.k8s.io/). Provision a cluster and ensure you can SSH to localhost. Eg. running the load test:
+
+```
+go run cmd/clusterloader.go --testconfig=testing/load/config.yaml --nodes=1 --provider=kind --kubeconfig=<path-to-kubeconfig> --masterip=127.0.0.1 --mastername=kind-control-plane --master-internal-ip=127.0.0.1
+```
 
 ### Flags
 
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/provider/kind.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (p *KindProvider) GetConfig() Config {
+	return Config{}
+}
+
+func (p *KindProvider) RunSSHCommand(cmd, host string) (string, string, int, error) {
+	// TODO(#1693): To maintain compatibility with the use of SSH to scrape measurements,
+	// we can SSH to localhost then run `docker exec -t <masterNodeContainerID> <cmd>`.
+	signer, err := sshSignerFromKeyFile("LOCAL_SSH_KEY", "id_rsa")
```
mm4tt, line 60: IIUC, the code here is copied from other provider and it wouldn't work in kind provider, right? 
In such case let's leave the TODO above, but return an error. E.g. 

```
return "", "", 0, fmt.Errorf("kind: ssh not yet implemented")
```

lyzs90: This code is still relevant I think, but we need to retrieve the container ID of the master node, then prepend the command with `docker exec -t <containerID>`

mm4tt: I see. Nevertheless, I'd still ask you to remove this code and return a "not implemented" error.  We shouldn't check in code that isn't complete (and doesn't work).

lyzs90: Okay, done

mm4tt:
This is amazing. Thanks so much, @lyzs90!

/lgtm
/approve

clusterloader2/README.md
```diff
@@ -4,13 +4,18 @@
 
 To run ClusterLoader type:
 ```
-go run cmd/clusterloader.go --kubeconfig=kubeConfig.yaml --testconfig=config.yaml
+go run cmd/clusterloader.go --kubeconfig=kubeConfig.yaml --testconfig=config.yaml --provider=local
```
wojtek-t, line 7: provider=kind

right?

[same below]

lyzs90: This section may not be relevant any more, I could get rid of it since we already have the kind example below? run-e2e.sh has some gce related functionality so local provider may not be applicable

mm4tt: I think local here is actually fine. This section is not about kind, but in general about running clusterloader2. You need to specify a provider (it's a required flag) and `local` seems to be the most generic choice.  

On the other hand, we should clean up and document the providers. For example, I have no idea what is the difference between `local` and `skeleton` and what is the purpose of them.

PR #1711, opened 2021-02-08
Revert "Merge pull request #1680 from lyzs90/extract-module-compilation"

mm4tt:
This reverts commit f2ec68631780e2c8c4ba6e114c03a068e5549af4, reversing changes made to 11eae442ad488ec5bd99ddcbb1a0123f33af43e3.

PR #1712, opened 2021-02-08
Revert #1711 and fix test suite overrides

lyzs90:
**What type of PR is this?**

/kind cleanup

This PR reverts #1711 and applies the changes introduced in #1680. On top of that, a fix has been made to bring back TestSuite overrides while keeping the global overrides within ClusterLoaderConfig.

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -281,7 +281,6 @@ func main() {
 	var prometheusController *prometheus.Controller
 	var prometheusFramework *framework.Framework
 	if clusterLoaderConfig.PrometheusConfig.EnableServer {
-		// Pass overrides to prometheus controller
 		if prometheusController, err = prometheus.NewController(&clusterLoaderConfig); err != nil {
 			klog.Exitf("Error while creating Prometheus Controller: %v", err)
 		}
@@ -306,43 +305,35 @@ func main() {
 	testReporter := test.CreateSimpleReporter(path.Join(clusterLoaderConfig.ReportDir, "junit.xml"), "ClusterLoaderV2")
 	testReporter.BeginTestSuite()
 
-	var contexts []test.Context
-
+	var testScenarios []api.TestScenario
 	if testSuiteConfigPath != "" {
 		testSuite, err := config.LoadTestSuite(testSuiteConfigPath)
 		if err != nil {
 			klog.Exitf("Error while reading test suite: %v", err)
 		}
-
-		for i := range testSuite {
-			testScenario := testSuite[i]
-			ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter, &testScenario)
-			if !errList.IsEmpty() {
-				klog.Exitf("Test context creation failed: %s", errList.String())
-			}
-			testConfig, errList := test.CompileTestConfig(ctx)
-			if !errList.IsEmpty() {
-				klog.Exitf("Test compliation failed: %s", errList.String())
-			}
-			ctx.SetTestConfig(testConfig)
-			contexts = append(contexts, ctx)
-		}
+		testScenarios = []api.TestScenario(testSuite)
 	} else {
 		for i := range testConfigPaths {
 			testScenario := api.TestScenario{
-				ConfigPath: testConfigPaths[i],
+				ConfigPath:    testConfigPaths[i],
+				OverridePaths: []string{},
 			}
-			ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter, &testScenario)
-			if !errList.IsEmpty() {
-				klog.Exitf("Test context creation failed: %s", errList.String())
-			}
-			testConfig, errList := test.CompileTestConfig(ctx)
-			if !errList.IsEmpty() {
-				klog.Exitf("Test compliation failed: %s", errList.String())
-			}
-			ctx.SetTestConfig(testConfig)
-			contexts = append(contexts, ctx)
+			testScenarios = append(testScenarios, testScenario)
 		}
+	}
+
+	var contexts []test.Context
+	for i := range testScenarios {
+		ctx, errList := test.CreateTestContext(f, prometheusFramework, &clusterLoaderConfig, testReporter, &testScenarios[i])
+		if !errList.IsEmpty() {
+			klog.Exitf("Test context creation failed: %s", errList.String())
+		}
+		testConfig, errList := test.CompileTestConfig(ctx)
+		if !errList.IsEmpty() {
+			klog.Exitf("Test compliation failed: %s", errList.String())
+		}
+		ctx.SetTestConfig(testConfig)
+		contexts = append(contexts, ctx)
 	}
 
 	for i := range contexts {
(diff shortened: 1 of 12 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -283,7 +282,6 @@ func main() {
 	var prometheusFramework *framework.Framework
 	if clusterLoaderConfig.PrometheusConfig.EnableServer {
 		// Pass overrides to prometheus controller
```
mm4tt, line 284: Remove the comment as well

clusterloader2/cmd/clusterloader.go
```diff
@@ -307,22 +305,51 @@ func main() {
...
+			contexts = append(contexts, ctx)
 		}
 	} else {
 		for i := range testConfigPaths {
-			clusterLoaderConfig.TestScenario.ConfigPath = testConfigPaths[i]
-			clusterLoaderConfig.TestScenario.OverridePaths = testOverridePaths
-			runSingleTest(f, prometheusFramework, testReporter)
+			testScenario := api.TestScenario{
+				ConfigPath:    testConfigPaths[i],
+				OverridePaths: []string{},
```
mm4tt, line 334: We should set it to the `clusterLoaderConfig.OverridePaths`

clusterloader2/pkg/config/template_provider.go
```diff
@@ -194,29 +194,45 @@ func validateTestSuite(suite api.TestSuite) error {
 	return nil
 }
 
+func getMappingFromFile(mapping map[string]interface{}, path string) error {
```
mm4tt, line 197: I don't like the name of this function. Gets suggests that it returns a new map, but this function rather modifies the map in place

clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -67,7 +67,7 @@ type controller struct {
 //
 // Preloading is skipped in kubemark or if no images have been specified.
 func Setup(conf *config.ClusterLoaderConfig, f *framework.Framework) error {
-	mapping, err := config.GetMapping(conf)
+	mapping, err := config.GetMapping(conf, []string{})
```
mm4tt, line 70: Pass `conf.OverridePaths` instead of `[]string{}`

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -108,7 +108,7 @@ func NewController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *Control
 		return nil, err
 	}
 
-	mapping, errList := config.GetMapping(clusterLoaderConfig)
+	mapping, errList := config.GetMapping(clusterLoaderConfig, []string{})
```
mm4tt, line 111: Same here

clusterloader2/pkg/test/test.go
```diff
@@ -37,41 +38,77 @@ var (
...
 	}
 
-	mapping, errList := config.GetMapping(clusterLoaderConfig)
-	if errList != nil {
-		return errList
+	clusterLoaderConfig := ctx.GetClusterLoaderConfig()
+	testConfigFilename := filepath.Base(ctx.GetTestScenario().ConfigPath)
+	testConfig, err := ctx.GetTemplateProvider().TemplateToConfig(testConfigFilename, ctx.GetTemplateMappingCopy())
+	if err != nil {
+		return &api.Config{}, errors.NewErrorList(fmt.Errorf("config reading error: %v", err))
```
mm4tt, line 77: return nil *api.Config on errors, here and in the other places

clusterloader2/pkg/test/test.go
```diff
@@ -37,41 +38,77 @@ var (
...
+	mapping, errList := config.GetMapping(clusterLoaderConfig, testScenario.OverridePaths)
+	if errList != nil {
+		return nil, errList
 	}
+
+	return CreateContext(clusterLoaderConfig, clusterFramework, prometheusFramework, state.NewState(), testReporter, mapping, testScenario), errors.NewErrorList()
+}
+
+// CompileTestConfig loads the test configuration and nested modules.
+func CompileTestConfig(ctx Context) (*api.Config, *errors.ErrorList) {
```
mm4tt, line 68: Let's make it return error instead of errors.ErrorList. There is only one caller of this method and it converts the returned errorList to a single error anyway.

clusterloader2/pkg/config/template_provider.go
lyzs90: Indeed, I can see how it can be misleading. How does `updateMappingFromFile` sound?

clusterloader2/cmd/clusterloader.go
lyzs90: Not sure if this would be confusing because later I end up doing:
```go
mapping, err := LoadTestOverrides(clusterLoaderConfig.OverridePaths, testOverridePaths)
```

clusterloader2/pkg/test/test.go
lyzs90: The error list could be returned by `testConfig.Validate()`, now it's a list of one because there is only one check, but potentially longer after the config validation PR.

Did you mean the `CreateTestContext` function instead? It does return single errors, I didn't see the use of errList.Append.

clusterloader2/pkg/test/test.go
lyzs90: Addressed

clusterloader2/cmd/clusterloader.go
lyzs90: Done

wojtek-t (commented):
Couple comments, but overall looks great.

lyzs90:
@wojtek-t Thanks, I have addressed the comments.

clusterloader2/cmd/clusterloader.go
```diff
@@ -307,22 +304,46 @@ func main() {
...
 		testSuite, err := config.LoadTestSuite(testSuiteConfigPath)
 		if err != nil {
 			klog.Exitf("Error while reading test suite: %v", err)
 		}
+
 		for i := range testSuite {
-			clusterLoaderConfig.TestScenario = testSuite[i]
-			runSingleTest(f, prometheusFramework, testReporter)
+			testScenario := testSuite[i]
+			testScenarios = append(testScenarios, testScenario)
```
wojtek-t, line 317: why do you need this?
why not simply:

testScenarios = ([]api.TestScenario)testSuite

clusterloader2/pkg/test/test.go
```diff
@@ -37,41 +38,77 @@ var (
...
+	steps, err := flattenModuleSteps(ctx, testConfig.Steps)
 	if err != nil {
-		return errors.NewErrorList(fmt.Errorf("config reading error: %v", err))
+		return nil, errors.NewErrorList(
+			fmt.Errorf("erorr when flattening module steps: %w", err))
 	}
-	testName := clusterLoaderConfig.TestScenario.Identifier
+	testConfig.Steps = steps
+
+	testConfig.SetDefaults()
```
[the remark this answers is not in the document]
wojtek-t: sorry for not being explicit - those should come before SetDefaults() (as they were before)

wojtek-t (commented):
two more small comment.

Please also squash commits.

lyzs90:
> sorry for not being explicit - those should come before SetDefaults() (as they were before)
My bad, it's fixed now.

Cool, I've squashed down to the revert plus the fix.

wojtek-t:
/lgtm

Let's not merge on Friday afternoon - I will approve (and let it merge) on Monday morning.

PR #1715, opened 2021-02-09
DO NOT SUBMIT - Test preloading of scratch images

lyzs90:
For testing image preloading prior to enabling on test-infra jobs. Ref https://github.com/kubernetes/test-infra/pull/20802

/cc @mm4tt

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -238,11 +238,11 @@ func main() {
 		klog.Exitf("Flag parse failed: %v", err)
 	}
 
-	provider, err := provider.NewProvider(&providerInitOptions)
+	pvd, err := provider.NewProvider(&providerInitOptions)
 	if err != nil {
 		klog.Exitf("Error init provider: %v", err)
 	}
-	clusterLoaderConfig.ClusterConfig.Provider = provider
+	clusterLoaderConfig.ClusterConfig.Provider = pvd
 
 	if errList := validateFlags(); !errList.IsEmpty() {
 		klog.Exitf("Parsing flags error: %v", errList.String())
@@ -297,7 +297,17 @@ func main() {
 			klog.Exitf("Error while setting up exec service: %v", err)
 		}
 	}
-	if err := imagepreload.Setup(&clusterLoaderConfig, f); err != nil {
+
+	var imagepreloadFramework *framework.Framework
+	if clusterLoaderConfig.ClusterConfig.Provider.Name() == provider.KubemarkName {
+		imagepreloadFramework, err = framework.NewRootFramework(&clusterLoaderConfig.ClusterConfig, imagepreload.NumK8sClients)
+		if err != nil {
+			klog.Exitf("Error while setting up image preloader: %v", err)
+		}
+	} else {
+		imagepreloadFramework = f
+	}
+	if err := imagepreload.Setup(&clusterLoaderConfig, imagepreloadFramework); err != nil {
 		klog.Exitf("Error while preloading images: %v", err)
 	}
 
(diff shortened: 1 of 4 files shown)


lyzs90:
Oops, preloading is disabled in kubemark provider. Will we be able to check node object sizes for pull-perf-tests-clusterloader2?

mm4tt:
> Oops, preloading is disabled in kubemark provider. Will we be able to check node object sizes for pull-perf-tests-clusterloader2?

Good catch, could you send a PR to enable it. It should work with your recent fixes. 

I don't think we'll be able, but that's fine. We'll be able to examine the node logs to check that the images were indeed downloaded there if we want to double check that it works as expected.

mm4tt:
We can also add this line at the end of the file (after cl2 invocation):
```
kubectl get nodes  -o custom-columns='NAME:.metadata.name,IMAGES:.status.images[*].names' | head -n 6
```

To check that the node objects were indeed modified

mm4tt:
/hold

To make sure we don't submit this

lyzs90:
If I'm understanding this correctly, CL2 is running against the kubemark cluster, so the image preload daemonset is actually being scheduled on the hollow nodes. 

Nevertheless, https://github.com/kubernetes/kubernetes/pull/97858 should have made it such that hollow nodes can also do the pulling. So either it can't pull or the informer isn't receiving any notifications.

After updating the log to v2: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1715/pull-perf-tests-clusterloader2-kubemark/1359531775872733184/#1:build-log.txt%3A1166

mm4tt:
This is because the image preload is using the framework, but it should be using root framework. This is exactly the same case as we have for prometheus stack. In Kubemark clusters, we install the prometheus stack in the main (root) cluster, because it wouldn't be possible to install it in the kubemark cluster (the promethues pods wouldn't be run).

So, we need to change the imagepreload to not use the framework provided in setup - https://github.com/kubernetes/perf-tests/blob/203cd2740c9eaf8d563998a972d842b1cd0b560f/clusterloader2/pkg/imagepreload/imagepreload.go#L77, but rather create a new one using the NewRootFramework func, exactly as it's done for prometheus - https://github.com/kubernetes/perf-tests/blob/cab9080bb7b510604e8d30b3d748582f055a553c/clusterloader2/pkg/prometheus/prometheus.go#L107

Would you mind sending a separate PR for that?

mm4tt:
Actually, it's more complicated. We need to use two different frameworks. One for creating the daemonset (root framework) and one for checking whether the nodes have been updated (regular framework).

lyzs90:
> Actually, it's more complicated. We need to use two different frameworks. One for creating the daemonset (root framework) and one for checking whether the nodes have been updated (regular framework).

Makes sense. I will raise a fix to do this if the provider is kubemark.

lyzs90:
On second thoughts, if we 'apply' using root framework but 'check' using regular framework, we wouldn't receive any new object updates for the nodes right?

lyzs90:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1715/pull-perf-tests-clusterloader2-kubemark/1359563592097075200/#1:build-log.txt%3A1163 ✨ Works on root framework, will raise a separate PR for that

mm4tt:
Actually I wanted to use rootFramework to create the daemonSet and the regular framework to wait for all hollow nodes to get the image list updated.  I think this is the most correct solution

lyzs90:
Is there a particular event to watch for on the hollow node side to achieve that? I was planning to wait for all daemonsets to be up then list hollow nodes in chunks. However if there's a way to watch for image list changes, that will be great

lyzs90:
Closing this one as the fix has been applied to #1719

Issue #1717, opened 2021-02-10
Error reading SSH key even AWS_SSH_KEY env is set


Jeffwan:
**What happened**:

This is a bug in 1.19 and previous branches. 

Test command:

```
ETCD_CERTIFICATE=/etc/kubernetes/pki/kube-apiserver/etcd-client.crt 
ETCD_KEY=/etc/kubernetes/pki/kube-apiserver/etcd-client.key 
KUBE_SSH_USER=ubuntu 
AWS_SSH_KEY=/Users/jiaxin/.ssh/id_rsa 
go run cmd/clusterloader.go --kubeconfig=/Users/jiaxin/.kube/config --testconfig=/Users/jiaxin/go/src/k8s.io/perf-tests/clusterloader2/testing/density/config.yaml --provider=aws --enable-prometheus-server --alsologtostderr --v=2 --report-dir=/tmp/pt/
```

Result: 

```
E0209 23:19:24.425522   29323 clusterloader.go:221]   Status: Fail
E0209 23:19:24.425525   29323 clusterloader.go:223]   Errors: [measurement call TestMetrics - TestMetrics error: [action gather failed for EtcdMetrics measurement: unexpected error (code: 0) in ssh connection to master: 
&errors.errorString{s:"error getting signer for provider aws: 'error reading SSH key /Users/jiaxin/.ssh: 'read /Users/jiaxin/.ssh: is a directory''"}]]
```

The problem is there's a bug that a new variable keyFile is recreated under case scope. It doesn't change value of `keyFile` defined earlier.
https://github.com/kubernetes/perf-tests/blob/5fb41ba7134696e897ad416289de0041141906b8/clusterloader2/pkg/measurement/util/ssh.go#L97

When it tries to check if argument is an absolute path, 
https://github.com/kubernetes/perf-tests/blob/5fb41ba7134696e897ad416289de0041141906b8/clusterloader2/pkg/measurement/util/ssh.go#L117

it reads value which is empty and then it create `$HOME/.ssh` + '' as the new keyFile.
https://github.com/kubernetes/perf-tests/blob/5fb41ba7134696e897ad416289de0041141906b8/clusterloader2/pkg/measurement/util/ssh.go#L84

**What you expected to happen**:

keyFile assignment should work properly. 

**How to reproduce it (as minimally and precisely as possible)**:

Follow above command I attached. 

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
[...]

PR #1718, opened 2021-02-10
Fix keyFile assignment for AWS provider

Jeffwan:
**What type of PR is this?**

/kind bug

Original keyFile assignment is incorrect, it create a new variable under case scope and doesn't change keyFile defined earlier. 
This results in error reading SSH key even `AWS_SSH_KEY ` env is set.

Fixes https://github.com/kubernetes/perf-tests/issues/1717

I will cherry-pick to previous version. release-1.19 and earlier branches are affected. 

```release-note
NONE
```

```docs

```


mm4tt:
Good catch, thanks for fixing!

/lgtm
/approve

PR #1731, opened 2021-02-12
[WIP] Quick hack to unblock tests with tainted nodes

jkaniuk:
Quick hack to unblock NAP tests with tainted nodes

/cc @jkaniuk


tosi3k:
Proper fix in https://github.com/kubernetes/perf-tests/pull/1743 is already merged.

/close

PR #1738, opened 2021-02-13
[WIP] Decrease anet scrape interval to 15s

Weil0ng:
**What type of PR is this?**

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/default/prometheus-podMonitorAnet.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/default/prometheus-podMonitorAnet.yaml
@@ -10,7 +10,7 @@ metadata:
   namespace: monitoring
 spec:
   podMetricsEndpoints:
-    - interval: 10m
+    - interval: 30s
       port: advdpmetrics
   jobLabel: k8s-app
   selector:


wojtek-t:
/assign @marseel @mborsz @jprzychodzen 

There were reasons why this was set so high, so I'm not sure we can just lower this down.

[Also - if we decide to proceed with it, please first merge to master and only then backport.]

Weil0ng:
> Hi, sorry for long delay. I was wondering how many pods are you trying to scrape?
> If I remember correctly, 10m interval was tested with ~37k pods and a few metrics were exposed from each pod.

Hi, I tried this PR with up to 2k nodes/60k pods, it'll cause prometheus OOM in 5k node/150k pods test...

marseel:
You can find cpu/memory limits for prometheus here: https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L19

I would suggest giving a lot of resources for prometheus and checking how much it consumed during test execution. Also, it's worth to double check that all targets where scraped (for example if we are not lacking cpu too).

Weil0ng:
Thanks! I think at this point we are not dependent on fine-grained anet metrics as we were, let's leave this PR open in case we need to run with a modified interval. In the short term, we can decide either:
- close this PR directly without lowering the scrape interval if there's no compelling need from the tests
- find a proper interval that we can afford and adjust prom requests along with it (as you suggested)

marseel:
Sounds good ;)

marseel:
In case you need me to look at this PR, just assign it back to me.
/unassign @marseel

Weil0ng:
Seems like we no longer need anet metrics at this granularity for experiments, closing this PR, will reopen if needs resurface, thanks for all the feedbacks :)

PR #1739, opened 2021-02-15
Allow to disable daemonsets in load test

marseel:
**What type of PR is this?**

/kind feature
Currently, we don't have a way to disable daemonsets in load test. skip-steps does not help in this case, because there is one step that creates all objects. 

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -20,6 +20,7 @@
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$PROMETHEUS_SCRAPE_KUBE_PROXY := DefaultParam .PROMETHEUS_SCRAPE_KUBE_PROXY true}}
 {{$ENABLE_PVS := DefaultParam .CL2_ENABLE_PVS true}}
+{{$DISABLE_DAEMONSETS := DefaultParam .CL2_DISABLE_DAEMONSETS false}}
 {{$ENABLE_NETWORKPOLICIES := DefaultParam .CL2_ENABLE_NETWORKPOLICIES false}}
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -194,13 +195,15 @@ steps:
       mediumServicesPerNamespace: {{DivideInt (AddInt $mediumDeploymentsPerNamespace 1) 2}}
       bigServicesPerNamespace: {{DivideInt (AddInt $bigDeploymentsPerNamespace 1) 2}}
 
+{{if $DISABLE_DAEMONSETS}}
```
mborsz, line 198: Shouldn't it be {{if not $DISABLE_DAEMONSETS}}??

marseel: yea, my bad. Done.

clusterloader2/testing/load/config.yaml
```diff
@@ -373,7 +378,9 @@ steps:
     Instances:
      - Identifier: WaitForRunningDeployments
      - Identifier: WaitForRunningStatefulSets
+{{if not $DISABLE_DAEMONSETS}}
```
wojtek-t, line 381: is that needed? If there isn't any DS, it shouldn't matter at all, right?

marseel: Yes, we could skip this part. I personally don't see big difference between having it or not.

marseel: If you prefer to remove it, let me know and I will do it.

wojtek-t: The simpler the config is the better.
So if it works without this if too, I prefer removing it.

marseel: Let me double check to make sure it works without it

wojtek-t:
/hold cancel

thanks!

PR #1742, opened 2021-02-15
Do not take taints into account when checking for node readiness

tosi3k:
/sig scalability
/assign @jkaniuk


tosi3k:
Proper fix in https://github.com/kubernetes/perf-tests/pull/1743.

/close

PR #1694, opened 2021-01-31
CL2 config validation

lyzs90:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
Refinement of https://github.com/kubernetes/perf-tests/pull/142. Validates more aspects of the CL2 config. 
Steps validation will automatically apply to modules once compilation is moved before test execution in https://github.com/kubernetes/perf-tests/pull/1680.

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes/perf-tests/issues/1636

**Special notes for your reviewer**:
Copy pasted the logic from https://github.com/kubernetes/perf-tests/pull/142 but placed it in the existing api/validation.go file. Had to do a conversion back to CL2 error types.

**Does this PR introduce a user-facing change?**:

```release-note
Added CL2 config validation for namespaces, tuning sets and steps.
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #1744, opened 2021-02-18
Check for absolute path also while validating object files

shyamjvs:
When I was trying to run clusterloader from my laptop using a command like this:
```
./clusterloader \
    --alsologtostderr \
    --kubeconfig /home/ec2-user/.kube/config \
    --testconfig /home/ec2-user/perf-tests/clusterloader2/testing/load/config.yaml \
    --testoverrides /home/ec2-user/perf-tests/clusterloader2/build/overrides.yaml 
...
```

It failed with errors like these:
```
F0218 09:30:18.245345   64780 clusterloader.go:334] Test compliation failed: [steps[1].phases[0].objectBundle[0].objectTemplatePath: Invalid value: "service.yaml": file must exist
steps[1].phases[1].objectBundle[0].objectTemplatePath: Invalid value: "service.yaml": file must exist
steps[1].phases[2].objectBundle[0].objectTemplatePath: Invalid value: "service.yaml": file must exist
steps[2].phases[0].objectBundle[0].objectTemplatePath: Invalid value: "daemonset-priorityclass.yaml": file must exist
steps[4].phases[0].objectBundle[0].objectTemplatePath: Invalid value: "daemonset.yaml": file must exist
steps[4].phases[1].objectBundle[0].objectTemplatePath: Invalid value: "configmap.yaml": file must exist
...
```

Seems like this is happening after some recent validation added in https://github.com/kubernetes/perf-tests/pull/1694. Specifically, this piece of code is only checking for relative path (which will be broken if configDir is an absolute path):
```
func (v *ConfigValidator) fileExists(path string) bool {
        cwd, _ := os.Getwd()
        _, err := os.Stat(fmt.Sprintf("%s/%s/%s", cwd, v.configDir, path))
        return !os.IsNotExist(err)
}
```

Fixing it in this pr.

/kind bug

### DIFF
--- a/clusterloader2/api/validation.go
+++ b/clusterloader2/api/validation.go
@@ -243,5 +243,9 @@ func (v *ConfigValidator) validateGlobalQPSLoad(gl *GlobalQPSLoad, fldPath *fiel
 func (v *ConfigValidator) fileExists(path string) bool {
 	cwd, _ := os.Getwd()
 	_, err := os.Stat(fmt.Sprintf("%s/%s/%s", cwd, v.configDir, path))
+	if os.IsNotExist(err) {
+		// If relative path didn't work, we also try absolute path.
+		_, err = os.Stat(fmt.Sprintf("%s/%s", v.configDir, path))
+	}
 	return !os.IsNotExist(err)
 }


clusterloader2/api/validation.go
```diff
@@ -243,5 +243,9 @@ func (v *ConfigValidator) validateGlobalQPSLoad(gl *GlobalQPSLoad, fldPath *fiel
 func (v *ConfigValidator) fileExists(path string) bool {
 	cwd, _ := os.Getwd()
 	_, err := os.Stat(fmt.Sprintf("%s/%s/%s", cwd, v.configDir, path))
+	if os.IsNotExist(err) {
+			// If relative path didn't work, we also try absolute path.
```
lyzs90, line 247: The indentation looks a little off here

clusterloader2/api/validation.go
```diff
@@ -243,5 +243,9 @@ func (v *ConfigValidator) validateGlobalQPSLoad(gl *GlobalQPSLoad, fldPath *fiel
 func (v *ConfigValidator) fileExists(path string) bool {
 	cwd, _ := os.Getwd()
 	_, err := os.Stat(fmt.Sprintf("%s/%s/%s", cwd, v.configDir, path))
+	if os.IsNotExist(err) {
```
lyzs90, line 246: There's a function to check for abs path https://golang.org/pkg/path/filepath/#IsAbs

clusterloader2/api/validation.go
shyamjvs: fixed

clusterloader2/api/validation.go
shyamjvs: avoiding additional import since we're already using os.Stat() elsewhere in this file

wojtek-t:
thanks!

Issue #1659, opened 2021-01-18
Migrate CL2's load test to CL2 Modules


mm4tt:
The CL2 Modules have been introduced in https://github.com/kubernetes/perf-tests/pull/1634. They allow dividing a single test configs into multiple module files, which improves readability and maintainability. The primary target for this is the load test [config.yaml](https://github.com/kubernetes/perf-tests/blob/add2da270f0857e5495e6c4692d5b9c51adee0cb/clusterloader2/testing/load/config.yaml#L911-L910) which has over 900 lines of code! 

I started the migration of the load test in https://github.com/kubernetes/perf-tests/pull/1640 where I migrated the k8s service management (creation/deletion) to a separate module. Now is the time to migrate the rest of the test.

In the PoC PR, https://github.com/kubernetes/perf-tests/pull/1632, I migrated almost the whole load test, but it requires cleaning up and verification. Nevertheless, it can be used a baseline for creating actual PRs.

I've also created a tool to simplify verification of the module migration. It expands modules from a provided PR to executable steps and compare them against master. The tool is here - https://github.com/mm4tt/k8s-util/tree/master/experimental/cl2-compare-config. The example usage can be found here - https://github.com/kubernetes/perf-tests/pull/1640#issuecomment-762120698

To summarize the remaining work:

1. Start with https://github.com/kubernetes/perf-tests/pull/1632 and use it as a example of what we want to achieve
2. Take a single module from https://github.com/kubernetes/perf-tests/pull/1632 and extract it to a separate PR (e.g. extract a module that manages measurements).
3.  Once the PR is created, run the https://github.com/mm4tt/k8s-util/tree/master/experimental/cl2-compare-config tool to verify the changes. Share the result of the tool in the PR
4. Send it for review (I'll be happy to review all the changes)
5. Repeat until the load test has been "fully migrated to modules". The exact stop condition is a bit vague, it's an open question how many modules we should have. We'll see once we create a few modules whether it makes sense to create more.

/good-first-issue
/help-wanted

mm4tt:
/assign @jkaniuk 

Assigning Jacek to review @petersveter108`s work while I'm away.

marseel:
I think we are pretty much done with this issue. If we want to change something I think it will be reasonable to open new issue from now on. 
/close

PR #1698, opened 2021-02-03
Load test - Starting and Gathering measurements in a separate module

petersveter108:
Started migration of the load test, Starting and Collecting (Gathering) measurements into a separate module - measurements.yml

/kind feature
/hold

Part of #1659

PR #1745, opened 2021-02-19
Load test - scheduler throughput in a separate module

petersveter108:
Scheduler throughput moved from config.yaml to a separate module in scheduler-throughput.yaml

/kind cleanup
/hold

Part of #1659

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -77,7 +77,6 @@
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
 {{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 100}}
-{{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
 # END scheduler-throughput section
 
 # Set schedulerThroughputNamespaces to 0 on small clusters otherwise it will result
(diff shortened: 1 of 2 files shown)


lyzs90:
Looks like you are missing a newline at the end of both files:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/perf-tests/1745/pull-perf-tests-clusterloader2-kubemark/1363148030093561856/#1:build-log.txt%3A1133

petersveter108:
@lyzs90 fixed, thank you :)

The https://github.com/mm4tt/k8s-util/tree/master/experimental/cl2-compare-config reports no diff:

$GOPATH/src/github.com/mm4tt/k8s-util/experimental/cl2-compare-config/run.sh 1709
...
Diff is:

@jkaniuk Hello Jacek, another one here fyi

clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,117 @@
+## Input Params
+{{$namespaces := .namespaces}}
```
jkaniuk, line 2: why not use .namespaces directly?
similarly for other variables

jkaniuk:
LGTM apart from one comment

clusterloader2/testing/load/modules/scheduler-throughput.yaml
jkaniuk: I see that this is convention started by Matt, probably to state variables explicitly, so I retract my reservations.

petersveter108: that was my motivation behind that, thanks Jacek

clusterloader2/testing/load/config.yaml
```diff
@@ -77,7 +77,6 @@
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
```
jkaniuk, line 77: Whole scheduler-throughput section could be moved to the module. We do not plan to modify those parameters.
The end goal is to simplify main config as much as possible.

jkaniuk:
cl2-diff output:
```
Diff is:                
6a7,9          
>       allowedSlowCalls: 0                                                                          
>       customThresholds: ""                                                                         
>       enableViolations: false                                                                      
11a15,19                                                                                                                                                                                                  
>       allowedSlowCalls: 0                                                                          
>       customThresholds: ""        
>       enableViolations: true
>       summaryName: APIResponsivenessPrometheus_simple
>       useSimpleLatencyQuery: true                                                                  
53c61
<   name: Starting measurements                                                                                                                                                                           
---
>   name: starting measurements
922,925c930,931
<       enableViolations: true
<       summaryName: APIResponsivenessPrometheus_simple
<       useSimpleLatencyQuery: true
<     identifier: APIResponsivenessPrometheusSimple
---
>       enableViolations: false
>     identifier: APIResponsivenessPrometheus
932,933c938,941
<       enableViolations: false
<     identifier: APIResponsivenessPrometheus
---
>       enableViolations: true
>       summaryName: APIResponsivenessPrometheus_simple
>       useSimpleLatencyQuery: true
>     identifier: APIResponsivenessPrometheusSimple
937a946,947
>       labelSelector: group = load
>       threshold: 1h
942a953,954
>       checkProbesReadyTimeout: 15m
>       replicasPerProbe: 3
947a960,961
>       checkProbesReadyTimeout: 15m
>       replicasPerProbe: 3
957a972
>       clusterOOMsIgnoredProcesses: null
967c982
<   name: Collecting measurements
---
>   name: gathering measurements
```

I think that CL2_* env variables are not handled properly in modules.
Peter - could you confirm?

petersveter108:
Wow this is strange, just double checked again how it looks on my side and I'm getting an empty diff there:
....
Master config dumped in /var/folders/7h/3wyxj_bj5ns7ns14jwtw0w_c0000gn/T/cl2-compare-XXXXXXXXXX.Prni6Ubz/master-steps.yaml
PR 1745 config dumped in /var/folders/7h/3wyxj_bj5ns7ns14jwtw0w_c0000gn/T/cl2-compare-XXXXXXXXXX.Prni6Ubz/pr1745-steps.yaml
Diff stored in /var/folders/7h/3wyxj_bj5ns7ns14jwtw0w_c0000gn/T/cl2-compare-XXXXXXXXXX.Prni6Ubz/diff
Diff is:

petersveter108:
@jkaniuk This one is for the #1698 as I see from the content, the thing is with this one that those 2 sections, Starting and Collecting measurements have a slightly different content inside, I did it that way based on the Matt's #1632. Also talked with him in the e-mail about slight differences like that and he said some of them are expected and unavoidable. However when it comes to the extra parameters, I can't judge there.

jkaniuk:
You are right.
Apart from the comment about variables, LGTM

clusterloader2/testing/load/config.yaml
petersveter108: So I kept this one due to the 'load' section that starts on the line 94, which also uses these parameters, , in case we want to move it to the beginning of the scheduler-throughput model, do we want to leave something at around the line 94 where the config starts or it's not necessary and so after variable declarations, config will start with Starting measurements?

clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,117 @@
...
+{{$minPodsInSmallCluster := .minPodsInSmallCluster}}
```
jkaniuk, line 3: minPodsInSmallCluster is not used

clusterloader2/testing/load/config.yaml
```diff
@@ -77,7 +77,6 @@
...
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
 {{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 100}}
```
jkaniuk, line 79: schedulerThroughputThreshold could be moved to the module

clusterloader2/testing/load/config.yaml
jkaniuk: I think that creating ($namespace + 10) namespaces  in large clusters would solve that issue, at the cost of not explicit interdependence. Let's leave this for now as it is.

PR #1747, opened 2021-02-22
Print mapping before the test starts

mborsz:
When the test starts, print test overrides to make debugging easier.

/assign @jkaniuk

### DIFF
--- a/clusterloader2/pkg/test/test.go
+++ b/clusterloader2/pkg/test/test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/api"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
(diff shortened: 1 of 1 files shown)


jkaniuk:
Why do we have this line duplicated:
```

W0223 08:10:09.344] I0223 08:10:09.344093   14496 test.go:78] Running "config.yaml" with mapping map[CL2_ALLOWED_SLOW_API_CALLS:1 CL2_ENABLE_API_AVAILABILITY_MEASUREMENT:true CL2_ENABLE_DNS_PROGRAMMING:true CL2_LOAD_TEST_THROUGHPUT:20 CL2_SCHEDULER_THROUGHPUT_THRESHOLD:0 ENABLE_RESTART_COUNT_CHECK:true Nodes:100 USE_SIMPLE_LATENCY_QUERY:true]
W0223 08:10:09.352] I0223 08:10:09.352525   14496 test.go:78] Running "config.yaml" with mapping map[CL2_ALLOWED_SLOW_API_CALLS:1 CL2_ENABLE_API_AVAILABILITY_MEASUREMENT:true CL2_ENABLE_DNS_PROGRAMMING:true CL2_LOAD_TEST_THROUGHPUT:20 CL2_SCHEDULER_THROUGHPUT_THRESHOLD:0 ENABLE_RESTART_COUNT_CHECK:true Nodes:100 USE_SIMPLE_LATENCY_QUERY:true]
```

tosi3k:
> Why do we have this line duplicated:

This happens because the compilation of all test configs happens before we even execute the first test and in the presubmit job's build log we printed the mapping for both `load` and `access-tokens` before the load test was executed.

To fix this, we might put the printing of mapping inside [this](https://github.com/kubernetes/perf-tests/blob/6e8f97efb1d2f57c7bd9102d3f060ca81913e701/clusterloader2/cmd/clusterloader.go#L340) loop, at the beginning of [runSingleTest](https://github.com/kubernetes/perf-tests/blob/6e8f97efb1d2f57c7bd9102d3f060ca81913e701/clusterloader2/cmd/clusterloader.go#L361) or inside [printTestStart](https://github.com/kubernetes/perf-tests/blob/6e8f97efb1d2f57c7bd9102d3f060ca81913e701/clusterloader2/cmd/clusterloader.go#L213) (probably most appropriate place to do it there).

Issue #1752, opened 2021-03-02
Cannot run load test at scale smaller than 100 nodes


tosi3k:
https://github.com/kubernetes/perf-tests/pull/1667 has made it possible to run the load test in clusters with less than 100 worked nodes (meaning having the `--nodes` CL2 flag set to a value lower than 100).

After the introduction of https://github.com/kubernetes/perf-tests/pull/1694, running the load test from the current master branch results in the following error, though:
```
Test compliation failed: [tuningSets[1].parallelismLimitedLoad.parallelismLimit: Invalid value: 0: must have positive value]
```

This fail is understandable - we cannot validate the `SchedulerThroughputParallel` tuning set as we have `parallelismLimit` being set to 0:
```
{{$IS_SMALL_CLUSTER := lt .Nodes 100}}
(...)
{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 0 $schedulerThroughputNamespaces}}
(...)
tuningSets:
(...)
- name: SchedulerThroughputParallel
  parallelismLimitedLoad:
    parallelismLimit: {{$schedulerThroughputNamespaces}}
```

/cc @lyzs90

lyzs90:
Ah my bad. In this case, it should be okay to add the not small cluster condition over the offending tuning set, since it's only being used within the scheduler throughput steps.

PR #1753, opened 2021-03-03
Remove scheduler throughput tuning set on small clusters

lyzs90:
**What type of PR is this?**

/kind bug

On small clusters, tuning set will fail validation because parallelismLimit = 0. This PR omits the offending tuning set on small clusters, similar to how it is done for the scheduler throughput steps.

Fixes https://github.com/kubernetes/perf-tests/issues/1752

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -80,9 +80,9 @@
 {{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
 # END scheduler-throughput section
 
-# Set schedulerThroughputNamespaces to 0 on small clusters otherwise it will result
+# Set schedulerThroughputNamespaces to 1 on small clusters otherwise it will result
 # in an unnecessary number of namespaces.
-{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 0 $schedulerThroughputNamespaces}}
+{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}
 
 # Probe measurements shared parameter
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -99,11 +99,13 @@ tuningSets:
 - name: Sequence
   parallelismLimitedLoad:
     parallelismLimit: 1
+{{if not $IS_SMALL_CLUSTER}}
```
wojtek-t, line 102: Hmm - actually - why this is failing validation?
Looking at stuff above:
```
totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) (MultiplyInt 2 .Nodes)}
```
So it's at minimum 2*500=1000
```
$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment
```

and we defide by numberOfNodes.

So it should be pretty big. So why it is failing?

tosi3k: Failure takes place because of [this](https://github.com/kubernetes/perf-tests/blob/309e9ff97cfa9b3707ff6c01b5a8b7056bebcf22/clusterloader2/testing/load/config.yaml#L85) line which happens later in the config:
```
{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 0 $schedulerThroughputNamespaces}}
```

wojtek-t: Thanks.

So actually - can we remove this line instead?
I don't see anything super bad in creating those namespaces (they will not be used later).
Alternatively, we can set it to 1 instead of 0 in this if.

lyzs90: Yup it's not such a big deal. It would bloat the cluster with namespaces and if the suite fails midway, all the namespaces remain and you'll have to clean them up. I think setting to 1 will be okay.

PR #1761, opened 2021-03-11
Add kube-state-metrics manifests and apply if flag is enabled

lilic:
This is part 2 of 3 PRs for kube-state-metrics perf and scale testing. As per the suggestion https://github.com/kubernetes/perf-tests/pull/1684#issuecomment-772362245 I added a new directory kube-state-metrics under the exporters directory. We cut a 2.0.0-rc.0 release, so this should be a good candidate to test! I deploy the manifests on my cluster and I had no errors and could see all the metrics. 

In PR part 3 as per suggestion, I will enable the scale test conditionally.

@mm4tt @wojtek-t please have a look, thank you! 🙇‍♀️


lilic:
cc @brancz @tariq1890 @mrueg FYI this is part 2 or 3 of scale/perf tests.

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/0namespace.yaml
```diff
@@ -0,0 +1,5 @@
+
```
mrueg, line 1: here's a blank line that probably can be removed.

mrueg (approved):

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/0namespace.yaml
lilic: Nice catch, thanks!

lilic:
I enabled the flag so it ran in CI and the manifests applied with no errors, I wonder how can we check Prometheus to see if kube-state-metrics targets was up? Did not find anything the CI run artifacts.

tariq1890 (approved):
Thanks @lilic !

tariq1890:
Just noticed this - ` clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/0namespace.yaml `

Was the `0namespace.yaml` intended?

lilic:
> Was the 0namespace.yaml intended?

Yes, so the namespace gets created first in order when the directory is applied, I noticed this strategy with their other manifests so went with that.

brancz:
I have limited knowledge of the framework, but the manifests look good! I assume I can't lgtm here, but hopefully that gives the approvers more confidence :)

lilic:
I also opened https://github.com/kubernetes/perf-tests/pull/1773 which builds on this PR.

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/0namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
wojtek-t, line 1: I wouldn't put that into pkg/prometheus.

Can you please move that to:
pkg/kube-state-metrics maybe?

[i.e. pkg/kube-state-metrics/manifests to be more specific]

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/clusterRole.yaml
```diff
@@ -0,0 +1,117 @@
...
+  - limitranges
+  - persistentvolumeclaims
+  - persistentvolumes
+  - namespaces
+  - endpoints
+  verbs:
+  - list
+  - watch
+- apiGroups:
+  - extensions
```
wojtek-t, line 28: Do we really need extensions?

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kube-state-metrics
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: kube-state-metrics
+        app.kubernetes.io/version: 2.0.0-rc.0
+    spec:
```
wojtek-t, line 19: we definitely need to set requests/limits (otherwise this can be easily starved by the tests themselves)

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/0namespace.yaml
wojtek-t: Actually - we put there a bunch of stuff, so maybe that's fine. Feel free to ignore (we can refactor that in the future).

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/clusterRole.yaml
lilic: Will double-check good catch, we have it in https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/cluster-role.yaml#L27-L31

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/deployment.yaml
lilic: Makes sense, its a chicken and egg problem I guess. We want to use these tests to suggest requests and limits to users but need to set some as well I understand. 

I took the example from the Prometheus you have deployed but also added CPU to the limits so we have a guaranteed QoS.

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/serviceMonitor.yaml
```diff
@@ -0,0 +1,21 @@
...
+  labels:
+    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/version: 2.0.0-rc.0
+  name: kube-state-metrics
+  namespace: kube-state-metrics-perf-test
+spec:
+  endpoints:
+  - honorLabels: true
+    interval: 30s
+    port: http-metrics
```
lilic, line 13: On second thought I am not sure we want to ingest all the kube-state-metrics metrics it produces to the Prometheus instance during tests, as I do not think that will work with the current resources set for that Prometheus as it would produce a lot of series. 

I might drop this, we already check in our perf tests that the endpoint is not empty. @tariq1890 @brancz @mrueg thoughts for or against this? We would keep the telemetry endpoint that produces self metrics.

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/0namespace.yaml
lilic: Yeah was not sure where to put it, so placed it under exporters, agreed its not ideal.

lilic:
@wojtek-t thank you for the review! This should be ready for another look.

wojtek-t:
/lgtm
/approve

Thanks!
[Re requests - that's fine if they are no ideal - we can tune them later, I just want to ensure that we won't be OOMing other stuff or failing tests because it scheduled on too small node or sth.]

clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/serviceMonitor.yaml
tariq1890: I am okay with ingesting all metrics honestly. The perf test is also meant to be a sort of stress test, yes? If not, I am not too opposed to the limiting the exposed ksm metrics

PR #1771, opened 2021-03-18
Perfdash changes to display network performance benchmark results

hanamantagoudvk:
What type of PR is this?

/kind feature

This review is one of the changeset done as part of the enhancement/feature (Reference #1498)

This changeset has PerfDash changes for displaying the results in dashboard.

### DIFF
--- a/clusterloader2/testing/network/config.yaml
+++ b/clusterloader2/testing/network/config.yaml
@@ -2,7 +2,7 @@
 {{$NUMBER_OF_SERVERS := .CL2_NUMBER_OF_SERVERS}}
 {{$NUMBER_OF_CLIENTS := .CL2_NUMBER_OF_CLIENTS}}
 
-name: Network Performance
+name: network_performance 
 automanagedNamespaces: 1
 steps:
 - name: Start network performance measurement
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -340,6 +340,40 @@ var (
 					Parser:           parsePerfData,
 				}},
 		},
+		"Network_Baseline": {
+			"TCP_P2P": []TestDescription{{
+				Name:             "network_performance_tcp-1:1",
+				OutputFilePrefix: "NetworkPerformanceMetrics_1:1_TCP_P2P",
+				Parser:           parsePerfData,
+			}},
+			"UDP_P2P": []TestDescription{{
+				Name:             "network_performance_udp-1:1",
+				OutputFilePrefix: "NetworkPerformanceMetrics_1:1_UDP_P2P",
+				Parser:           parsePerfData,
+			}},
+			"HTTP_P2P": []TestDescription{{
+				Name:             "network_performance_http-1:1",
+				OutputFilePrefix: "NetworkPerformanceMetrics_1:1_HTTP_P2P",
+				Parser:           parsePerfData,
+			}},
+		},
+		"Network_Scale": {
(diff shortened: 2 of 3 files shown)


perfdash/config.go
```diff
@@ -340,6 +340,136 @@ var (
 					Parser:           parsePerfData,
 				}},
 		},
+		"Network_Baseline": {
+			"TCP_P2P": []TestDescription{{
+				Name:             "standalone",
+				OutputFilePrefix: "NetworkPerformanceMetrics1:1TCPP2P",
```
wojtek-t, line 346: Can we have better names (it's hard to parse it), e.g.:

NetworkPerformanceMetrics_1:1_TCP_P2P

[Yes - it requires small changes to the code generating them]

perfdash/config.go
```diff
@@ -340,6 +340,136 @@ var (
...
+				Parser:           parsePerfData,
+			}},
+
+			"TCP_ClusterIP": []TestDescription{
+				{
+					// TODO(oxddr): remove this around Sep '19 when we stop showing old data
```
wojtek-t, line 352: Please remove those TODOs - those are completely unrelated (it was for migration file names, which you're obviously nt doing).

perfdash/config.go
wojtek-t: Also - converge on the order (the the name should be the ame as in the file prefix)

VivekThrivikraman-est: Modified

perfdash/config.go
VivekThrivikraman-est: It wasn't there in the final commit, something went wrong during splitting(old commit was pushed). Have reverted to the actual changes.

clusterloader2/testing/network/config.yaml
```diff
@@ -2,7 +2,7 @@
 {{$NUMBER_OF_SERVERS := .CL2_NUMBER_OF_SERVERS}}
 {{$NUMBER_OF_CLIENTS := .CL2_NUMBER_OF_CLIENTS}}
 
-name: Network Performance
+name: network_performance
```
wojtek-t, line 5: Why do you change this?

perfdash/config.go
```diff
@@ -340,6 +340,45 @@ var (
...
 		},
+		"Network_Baseline": {
+			"TCP_P2P": []TestDescription{{
+				Name:             "network_performance_tcp-1:1",
+				OutputFilePrefix: "NetworkPerformanceMetrics_1:1_TCP_P2P",
+				Parser:           parsePerfData,
+			}},
+
+			"UDP_P2P": []TestDescription{
+				{
```
wojtek-t, line 351: Please make pagination consistent - just use the style you use above [i.e. {{ in the previous line]

same below

clusterloader2/testing/network/config.yaml
VivekThrivikraman-est: 2 reasons : 1) Other tests in cluster loader where there are 2 words they are not using capital letters(eg node-throughput). But just noticed they use small letters and words connected by "-" rather than "_".Shall I change this also to have "-" instead of "_"?
2)The file generated would have this as part of it's name , and having space just for this would seem awkward.

wojtek-t: ok - that's fine

perfdash/config.go
VivekThrivikraman-est: Modified

PR #1774, opened 2021-03-23
clusterloader2/testing/load: Adjust kube-state-metrics latency tests

lilic:
Enable kube-state-metrics to be run in the presubmit load tests.
/hold

### DIFF
--- a/clusterloader2/pkg/provider/provider.go
+++ b/clusterloader2/pkg/provider/provider.go
@@ -52,6 +52,8 @@ type Features struct {
 	SupportNodeKiller bool
 	// SupportGrabMetricsFromKubelets determines whether getting metrics from kubelet is supported.
 	SupportGrabMetricsFromKubelets bool
+	// SupportKubeStateMetrics determines if running kube-state-metrics is supported.
+	SupportKubeStateMetrics bool
 
 	// ShouldPrometheusScrapeApiserverOnly determines if we should set PROMETHEUS_SCRAPE_APISERVER_ONLY by default.
 	ShouldPrometheusScrapeApiserverOnly bool
--- a/clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
+++ b/clusterloader2/pkg/measurement/common/kube_state_metrics_measurement.go
@@ -78,6 +78,10 @@ func CreateKSMLatencyMeasurement() measurement.Measurement {
 // it also collects initial latency metrics.
 // - gather - gathers latency metrics and creates a latency summary.
 func (m *ksmLatencyMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	if !config.CloudProvider.Features().SupportKubeStateMetrics {
+		klog.Infof("not executing KSMLatencyMeasurement: unsupported for provider, %s", config.ClusterFramework.GetClusterConfig().Provider.Name())
+		return nil, nil
+	}
 	action, err := util.GetString(config.Params, "action")
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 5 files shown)


lilic:
FYI @mrueg @brancz @tariq1890 this just enables it temporarily so we see the results in the 100 node presubmit tests. 🎉

lilic:
@wojtek-t, this is the PR as discussed earlier. Thanks!

lilic:
Update: Seems like its failing to find the service object for kube-state-metrics, looking into it.

mrueg:
@lilic ``` F0323 14:49:28.989150   14091 clusterloader.go:291] Error while setting up prometheus stack: parsing error: template: :34: unexpected {{end}} ``` could it be related to https://github.com/kubernetes/perf-tests/commit/5a6de856f824a0b2a0ad7e17b6663e02c04c74e5#diff-95ffc56dd3788da48793718c29a759d29cd9a8fa1242e9ea8bcb6dd672f0b6e4R34  ? There are two  {{end}} statements that might cause it

lilic:
@mrueg I took that from https://github.com/kubernetes/perf-tests/blob/5a6de856f824a0b2a0ad7e17b6663e02c04c74e5/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L18-L28 but you might be right, there is no need for end as we are not using the if. Good catch!

mrueg:
Is it valid in k8s to have a non-numerical targetPort? Is it possible, that could be the reason it can't find the service? https://github.com/kubernetes/perf-tests/commit/5a6de856f824a0b2a0ad7e17b6663e02c04c74e5#diff-38d60ed161d961ff44471bdbfbdfef4c3a3c78d2baaec3bc4ba21cfc9c2cc5a2R17

lilic:
@mrueg it worked correctly for 
> pull-perf-tests-clusterloader2 — Job succeeded.

We can see the results of the latency test https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1774/pull-perf-tests-clusterloader2/1374398775702851584/artifacts/KubeStateMetricsLatency_load_2021-03-23T16:58:30Z.json

mrueg:
> @mrueg it worked correctly for
> 
> > pull-perf-tests-clusterloader2 — Job succeeded.
> 
> We can see the results of the latency test https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/1774/pull-perf-tests-clusterloader2/1374398775702851584/artifacts/KubeStateMetricsLatency_load_2021-03-23T16:58:30Z.json
> 
> About the non-numerical ports -> https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services
> 
> > For example, the names 123-abc and web are valid, but 123_abc and -web are not.

Thanks! I didn't know this was possible :+1:

lilic:
Update on this, KSM uses the  "in-cluster" config, which is incompatible with kubemark this is why they are failing. I will remove KSM from running under the kubemark job.

lilic:
kubmark passed as it was skipped and the other load test passed and these are the results for kube-state-metrics latency loads results for the 100 node cluster:

First run:
>{
   "Perc50": 243181818,
  "Perc90": 456730769,
  "Perc99": 725000000
}

Second run:
>  {
  "Perc50": 259615384,
  "Perc90": 475000000,
  "Perc99": 906666666
}

lilic:
/hold cancel

@wojtek-t do you might having a look, thank you!

clusterloader2/pkg/provider/provider.go
```diff
@@ -33,6 +33,9 @@ type InitOptions struct {
 
 // Features represents all features supported by this provider.
 type Features struct {
+	// KubeStateMetricsTestDisabled determines if running kube-state-metrics is supported.
+	KubeStateMetricsTestDisabled bool
```
wojtek-t, line 37: We're trying to be positive in those (e.g. talk about features not about disabling something).

So instead of that, we should make it more like:

`SupportKubeStateMetrics`

Given that our tests run on GCE, it's probably enough to enable that just for gce cloud provider for now.

wojtek-t:
/lgtm
/approve

Thanks!

PR #1775, opened 2021-03-23
Delay the exec command step in the load test config

tosi3k:
Run the exec command after we additionally load the cluster with the scheduler throughput pods.


jkaniuk:
/lgtm

@mborsz for an approval

PR #1777, opened 2021-03-31
Implement watch-based daemonset replica counter.

mborsz:
Problem: waitForControlledPodsRunning for daemonsets calculates a number of expected replicas based on the number of nodes that are schedule. That number can vary over time (e.g. node becoming notready/ready).

We are seeing a tests runs where e.g. node X was unhealthy for a short period of time for whatever reason (physical machine failure). During that time, we calculated a number of replicas. Then node becomes healthy and daemonset controller creates more replicas than we expect making the test fail.

This code introduces a watch for node objects to track accurate number of expected replicas over time

### DIFF
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package runtimeobjects
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+)
+
+const (
+	informerSyncTimeout = time.Minute
+)
+
+// ReplicasWatcher is a struct that allows to check a number of replicas at a given time.
+// Usage:
+// var rw ReplicasWatcher = (...)
+// if err := rw.Start(stopCh); err != nil {
+//   panic(err);
+// }
+// // Get number of replicas as needed.
+// val = rw.Replicas()
+// ...
+// val = rw.Replicas()
+type ReplicasWatcher interface {
+	Replicas() int
+	// Start must block until Replicas() returns a correct value.
+	Start(stopCh chan struct{}) error
+}
+
+// ConstReplicas is a ReplicasWatcher implementation that returns a constant value.
+type ConstReplicas struct {
+	ReplicasCount int
+}
+
+func (c *ConstReplicas) Replicas() int {
+	return c.ReplicasCount
+}
+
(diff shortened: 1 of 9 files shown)


clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
```diff
@@ -0,0 +1,159 @@
...
+// if err := rw.Start(stopCh); err != nil {
+//   panic(err);
+// }
+// // Get number of replicas as needed.
+// val = rw.Replicas()
+// ...
+// val = rw.Replicas()
+type ReplicasWatcher interface {
+	Replicas() int
+	// Start must block until Replicas() returns a correect value.
```
wojtek-t, line 52: nit: correct

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
```diff
@@ -0,0 +1,159 @@
...
+	Start(stopCh chan struct{}) error
+}
+
+// ConstReplicas is a ReplicasWatcher implementation that returns a constant value.
+type ConstReplicas struct {
+	Const int
```
wojtek-t, line 58: I think "Replicas int" would be more intuitive :)

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
```diff
@@ -0,0 +1,159 @@
...
+	new, err := n.shouldRun(newObj)
+	if err != nil {
+		return err
+	}
+	if new == old {
+		return nil
+	}
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	if new && !old {
```
wojtek-t, line 128: nit: feel fre to ignore:

n.replicas += (new == true)
n.replicas -= (old == true)

[then you can skip the if new == old too]

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
```diff
@@ -0,0 +1,159 @@
...
+		n.replicas++
+	} else {
+		n.replicas--
+	}
+	klog.Infof("Updated expected replicas: %v", n.replicas)
+	return nil
+}
+
+func (n *NodeCounter) Replicas() int { return n.replicas }
```
wojtek-t, line 137: should be protected by lock

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
```diff
@@ -538,12 +538,17 @@ func TestGetReplicasFromRuntimeObject(t *testing.T) {
...
 			t.Fatalf("error converting controller to unstructured: %v", err)
 		}
-		replicas, err := runtimeobjects.GetReplicasFromRuntimeObject(fakeClient, unstructured)
+		replicasWatcher, err := runtimeobjects.GetReplicasFromRuntimeObject(fakeClient, unstructured)
 		if err != nil {
 			t.Fatalf("get replicas from runtime object failed: %v", err)
 		}
 
-		if expected[i] != replicas {
+		replicas, err := runtimeobjects.GetReplicasOnce(replicasWatcher)
```
wojtek-t, line 546: don't we have to call Start() before calling GetReplicasOnce?

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
mborsz: I tried that initially, but I got "field and method with the same name Replicas (see details)" compiler error :)

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
mborsz: GetReplicasOnce does it internally

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
mborsz: I was thinking on something like "n.repliacas += int(new) - int(old)" already, but in golang it's not possible to easily cast bool -> int.
I can change return type for "shouldRun" to int, but it will require implementing some if inside, so it's not clear if this is overall win.

See https://play.golang.org/p/qurKupUdp7b

mborsz:
This is ready for review.

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
mborsz: Done

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
mborsz: Done

PR #1782, opened 2021-04-13
Add parsers for common measurement matching any test name

mm4tt:


perfdash/config.go
```diff
@@ -397,6 +437,10 @@ var (
 				OutputFilePrefix: "SystemPodMetrics",
 				Parser:           parseSystemPodMetrics,
 			}},
+			"SystemPodMetrics": []TestDescription{{
```
jkaniuk, line 440: Do we really need this kind of repetition?
It is stating for the third time almost the same thing.
There was some rule of the thumb - up to two occurrences - don't bother, three and up - deduplicate.

mm4tt: Yeah, but in this case I don't think that work on refactoring will ever pay off, this third repetition that I added is the "catch-all" clause, we rather won't need another one. Also, this whole code should be refactored, but we made the decision to not invest into perf-dash more than absolutely necessary.

PR #1785, opened 2021-04-16
add load test for L4 LB nodesync

freehan:
**What type of PR is this?**

/kind feature

Add a generic load test for L4 LBs

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
+++ b/clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
@@ -1,0 +1,315 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+	"time"
+)
+
+const (
+	loadBalancerNodeSyncLatencyName = "LoadBalancerNodeSyncLatency"
+	defaultNodeSyncLatencyTimeout   = 30 * time.Minute
+
+	// excludeFromLoadBalancersLabel is the node label to exclude a node from being a LB backend
+	excludeFromLoadBalancersLabel = "node.kubernetes.io/exclude-from-external-load-balancers"
+	// nodeSyncEventReason is the event reason emitted by service controller when it completes node sync on the lb.
+	nodeSyncEventReason = "UpdatedLoadBalancer"
+
+	phaseNodeSyncStart    = "nodesync_triggered"
+	phaseNodeSyncComplete = "nodesync_complete"
+)
+
+var nodeSyncTransition = map[string]measurementutil.Transition{
+	"nodesync_start_to_complete": {
+		From: phaseNodeSyncStart,
+		To:   phaseNodeSyncComplete,
+	},
+}
+
+func init() {
+	if err := measurement.Register(loadBalancerNodeSyncLatencyName, createLoadBalancerNodeSyncMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", loadBalancerNodeSyncLatencyName, err)
+	}
+}
+
+func createLoadBalancerNodeSyncMeasurement() measurement.Measurement {
+	return &LoadBalancerNodeSyncMeasurement{
+		selector:                  measurementutil.NewObjectSelector(),
+		svcNodeSyncLatencyTracker: measurementutil.NewObjectTransitionTimes(loadBalancerNodeSyncLatencyName),
+	}
+}
+
+type LoadBalancerNodeSyncMeasurement struct {
+	client clientset.Interface
+	// selector used to select relevant load balancer type service used for measurement
+	selector *measurementutil.ObjectSelector
+	// waitTimeout specify for the timeout for node sync on all LBs to complete
+	waitTimeout time.Duration
+	// svcNodeSyncLatencyTracker tracks the nodesync latency
+	svcNodeSyncLatencyTracker *measurementutil.ObjectTransitionTimes
+	// excludedNodeName is the node name used to trigger LB nodesync
+	excludedNodeName string
+	// lbSvcMap is the map that contains load balancer type service with key (namespaced/name) and service
+	lbSvcMap map[string]v1.Service
+}
+
+// LoadBalancerNodeSyncMeasurement takes measurement of node sync latency for selected lb type services.
+// This measurement only works for K8s 1.19 as it depends on the ExcludeNodeForLoadbalancer label.
+// Services can be specified by field and/or label selectors.
+// If namespace is not passed by parameter, all LoadBalancer type service with all-namespace scope is assumed.
+// "measure" action triggers nodesync and observation of nodesync completion for selected LB services.
+// "gather" returns node sync latency summary.
+func (s *LoadBalancerNodeSyncMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	s.client = config.ClusterFramework.GetClientSets().GetClient()
+	action, err := util.GetString(config.Params, "action")
+	if err != nil {
+		return nil, err
+	}
+	switch action {
+	case "measure":
+		if err := s.selector.Parse(config.Params); err != nil {
+			return nil, err
+		}
+		s.waitTimeout, err = util.GetDurationOrDefault(config.Params, "waitTimeout", defaultNodeSyncLatencyTimeout)
+		if err != nil {
+			return nil, err
+		}
+		return nil, s.measureNodeSyncLatency()
+	case "gather":
+		if err := s.labelNodeForLBs(false); err != nil {
+			return nil, err
+		}
+		return s.gather(config.Identifier)
+	default:
+		return nil, fmt.Errorf("unknown action %v", action)
+	}
+}
+
+func (s *LoadBalancerNodeSyncMeasurement) Dispose() {}
+
+func (s *LoadBalancerNodeSyncMeasurement) String() string {
+	return loadBalancerNodeSyncLatencyName + ": " + s.selector.String()
+}
+
+func (s *LoadBalancerNodeSyncMeasurement) measureNodeSyncLatency() error {
+	ctx := context.Background()
+	options := metav1.ListOptions{}
+	s.selector.ApplySelectors(&options)
+	svcList, err := s.client.CoreV1().Services(s.selector.Namespace).List(ctx, options)
+	if err != nil {
+		return err
+	}
+
+	s.lbSvcMap = map[string]v1.Service{}
+	for _, svc := range svcList.Items {
+		if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
+			s.lbSvcMap[keyFunc(svc.Namespace, svc.Name)] = svc
+		}
+	}
+	totalLbSvc := len(s.lbSvcMap)
+
+	// Use event informer to keep track of nodeSync events.
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	eventInformer := s.getEventInformer()
+	go eventInformer.Run(stopCh)
+
+	// trigger node sync by picking a node and add exclude lb label
+	nodeList, err := s.client.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodeList.Items {
+		if isCandidateNode(node) {
+			s.excludedNodeName = node.Name
+			break
+		}
+	}
+
+	if s.excludedNodeName == "" {
+		return fmt.Errorf("failed to find a node candidate to trigger nodesync from node list: %v", nodeList.Items)
+	}
+
+	defer func() {
+		if err = s.labelNodeForLBs(false); err != nil {
+			klog.Errorf("Failed to label node %v: %v", s.excludedNodeName, err)
+		}
+
+	}()
+	if err = s.labelNodeForLBs(true); err != nil {
+		return err
+	}
+
+	now := time.Now()
+	for key := range s.lbSvcMap {
+		s.svcNodeSyncLatencyTracker.Set(key, phaseNodeSyncStart, now)
+	}
+
+	return wait.Poll(5*time.Second, s.waitTimeout, func() (done bool, err error) {
+		if s.svcNodeSyncLatencyTracker.Count(phaseNodeSyncComplete) == totalLbSvc {
+			return true, nil
+		}
+		klog.V(2).Infof("out of a total of %v LBs, %v LB type service has %q event", totalLbSvc, s.svcNodeSyncLatencyTracker.Count(phaseNodeSyncComplete), nodeSyncEventReason)
+		return false, nil
+	})
+}
+
+func (s *LoadBalancerNodeSyncMeasurement) getEventInformer() cache.Controller {
+	ctx := context.Background()
+	listFunc := func(options metav1.ListOptions) (runtime.Object, error) {
+		o := metav1.ListOptions{
+			Limit: 1,
+		}
+		result, err := s.client.CoreV1().Events(metav1.NamespaceAll).List(ctx, o)
+		if err != nil {
+			return nil, err
+		}
+		result.Continue = ""
+		result.Items = nil
+		return result, nil
+	}
+
+	watchFunc := func(options metav1.ListOptions) (watch.Interface, error) {
+		options.FieldSelector = fields.Set{"reason": nodeSyncEventReason}.AsSelector().String()
+		return s.client.CoreV1().Events(metav1.NamespaceAll).Watch(ctx, options)
+	}
+
+	_, eventInformer := cache.NewInformer(&cache.ListWatch{ListFunc: listFunc, WatchFunc: watchFunc}, nil, 0,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				s.processEvent(obj.(*v1.Event))
+			},
+		})
+	return eventInformer
+}
+
+func (s *LoadBalancerNodeSyncMeasurement) processEvent(event *v1.Event) {
+	if event.Reason != nodeSyncEventReason {
+		return
+	}
+
+	key := keyFunc(event.InvolvedObject.Namespace, event.InvolvedObject.Name)
+	_, ok := s.lbSvcMap[key]
+	if ok {
+		_, found := s.svcNodeSyncLatencyTracker.Get(key, phaseNodeSyncComplete)
+		if !found {
+			s.svcNodeSyncLatencyTracker.Set(key, phaseNodeSyncComplete, event.CreationTimestamp.Time)
+		}
+	}
+}
+
+// labelNodeForLBs manipulates candidate node to include or exclude it from being LB backends.
+func (s *LoadBalancerNodeSyncMeasurement) labelNodeForLBs(exclude bool) error {
+	ctx := context.Background()
+	node, err := s.client.CoreV1().Nodes().Get(ctx, s.excludedNodeName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	newNode := node.DeepCopy()
+
+	if exclude {
+		newNode.Labels[excludeFromLoadBalancersLabel] = "true"
+	} else {
+		delete(newNode.Labels, excludeFromLoadBalancersLabel)
+	}
+
+	patchBytes, err := preparePatchBytes(node, newNode, v1.Node{})
+	if err != nil {
+		return err
+	}
+
+	_, err = s.client.CoreV1().Nodes().Patch(ctx, s.excludedNodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *LoadBalancerNodeSyncMeasurement) gather(identifier string) ([]measurement.Summary, error) {
+	klog.V(2).Infof("%s: gathering nodesync latency measurement...", s)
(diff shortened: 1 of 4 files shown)


mm4tt:
/assign @jkaniuk 

Jacek, could you please review this?

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,280 @@
...
+			eventList, err := s.client.CoreV1().Events(s.selector.Namespace).Search(Scheme, &svc)
+			if err != nil {
+				klog.Errorf("failed to list events associated with svc %v: %v", svc, err)
+				continue
+			}
+
+			for _, event := range eventList.Items {
+				if event.Reason == nodeSyncEventReason {
+					s.svcNodeSyncLatencyTracker.Set(key, phaseNodeSyncCompelte, event.CreationTimestamp.Time)
+				}
```
prameshj, line 194: Should we break here? 

Also, will events be in reverse chronological order? (so the latest lb update shows up first?). Otherwise we might catch the event where the LB was updated for the first time right?

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (s *LoadBalancerNodeSyncMeasurement) gather(identifier string) ([]measurement.Summary, error) {
+	klog.V(2).Infof("%s: gathering nodesync latency measurement...", s)
+	nodeSyncLatency := s.svcNodeSyncLatencyTracker.CalculateTransitionsLatency(nodeSyncTransition, measurementutil.MatchAll)
+	content, err := util.PrettyPrintJSON(measurementutil.LatencyMapToPerfData(nodeSyncLatency))
+	if err != nil {
+		return nil, err
+	}
+
```
prameshj, line 239: Maybe add a TODO to return an error here if the latency is higher than an upper bound (say 30 mins)?

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,280 @@
...
+	switch action {
+	case "measure":
+		if err := s.selector.Parse(config.Params); err != nil {
+			return nil, err
+		}
+		s.waitTimeout, err = util.GetDurationOrDefault(config.Params, "waitTimeout", defaultNodeSyncLatencyTimeout)
+		if err != nil {
+			return nil, err
+		}
+		return nil, s.measureNodeSyncLatency()
```
prameshj, line 113: The measure step picks one candidate node, marks it to be excluded from LB and then measures the time taken for the update event to show up. 

This is a good test, but I think the most problematic scenarios are where all nodes are removed and replaced by a new set of nodes, causing discrepancy in the k8s and GCE layer until steady state. Can we trigger that scenario by doing kubectl delete of all nodes?

prameshj (commented):
Logic looks good to me. Added some follow-up questions.

clusterloader2/testing/l4lb/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.Name}}
+  labels:
+    test: l4lb-load
+  annotations:
+    cloud.google.com/load-balancer-type: {{.LoadBalancerType}}
```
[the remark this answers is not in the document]
jkaniuk: We do not have to support 1.16-

clusterloader2/testing/l4lb/service.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  externalTrafficPolicy: {{.ExternalTrafficPolicy}}
+  type: LoadBalancer
+  selector:
+    name: {{.DeploymentBaseName}}-{{.Index}}
+  ports:
+  - port: 8080
```
jkaniuk, line 15: I think that 8080 port on a node will conflict with NodeLocal DNSCache

clusterloader2/testing/l4lb/service.yaml
```diff
@@ -0,0 +1,16 @@
...
+    targetPort: 80
```
jkaniuk, line 16: Shouldn't it match `containerPort: 8080` ?

clusterloader2/testing/l4lb/config.yaml
```diff
@@ -0,0 +1,95 @@
...
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
+- name: Measure NodeSync latency
+  measurements:
+  - Identifier: NodeSyncLatency
+    Method: LoadBalancerNodeSyncLatency
+    Params:
+      action: measure
+      labelSelector: test =l4lb-load
```
jkaniuk, line 72: nit: space missing

clusterloader2/testing/l4lb/config.yaml
```diff
@@ -0,0 +1,95 @@
...
+# LOAD_BALANCER_REPLICAS specifies the number of Load balancer type service
+{{$LOAD_BALANCER_REPLICAS := DefaultParam .CL2_LOAD_BALANCER_REPLICAS 3}}
+# LOAD_BALANCER_TYPE specifies the type of L4 LB created. Valid values are "INTERNAL" and "EXTERNAL"
+{{$LOAD_BALANCER_TYPE := DefaultParam .CL2_LOAD_BALANCER_TYPE "EXTERNAL"}}
+# $EXTERNAL_TRAFFIC_POLICY specifies the externalTrafficPolicy on LB type service. Valid values are "Cluster" and "Local"
+{{$EXTERNAL_TRAFFIC_POLICY := DefaultParam .CL2_EXTERNAL_TRAFFIC_POLICY "Cluster"}}
+# $NODE_SYNC_TIMEOUT specifies the timeout to wait for nodesync to complete
+{{$NODE_SYNC_TIMEOUT := DefaultParam .CL2_NODE_SYNC_TIMEOUT "30m"}}
+# adding a fixed value for first version of the test, rate of pod creation not a concern yet.
+{{$lbQPS := 500}}
```
jkaniuk, line 12: The biggest clusters can handle 100 qps pod creation and 1/3 of that for deletion. Default controller-manager qps limit (for smaller clusters) is 20 (deletion also 1/3 of that)

jkaniuk (commented):
Partially reviewed, mostly config.

@tosi3k - could you look into that?

clusterloader2/testing/l4lb/service.yaml
prameshj: Good point, though it is only a problem if the pod is in "hostNetwork mode". This one is not.

jkaniuk: Isn't port and targetPort swapped then? service should be on 80 (port), pod exposes 8080 (targetPort)

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
freehan: Adding a break. 

The measurement only triggers one nodesync event, so there should be only one nodeSyncEvent occur.

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
freehan: The problem with that is kubelet will try to add the node back immediately. The reason why I did not pick a rolling update or node pool operation is because it might cause multiple nodesyncs. Causing the unnecessary noise for measuring latency.

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
freehan: Ack

clusterloader2/testing/l4lb/config.yaml
freehan: ack

clusterloader2/testing/l4lb/service.yaml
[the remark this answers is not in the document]
freehan: done

clusterloader2/testing/l4lb/service.yaml
freehan: good catch

clusterloader2/testing/l4lb/service.yaml
freehan: yes. Corrected the yaml. 

Now service and container port are both 80

freehan:
Ready for another round.

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,283 @@
...
+	// nodeSyncEventReason is the event reason emitted by service controller when it completes node sync on the lb.
+	nodeSyncEventReason = "UpdatedLoadBalancer"
+
+	phaseNodeSyncStart    = "nodesync_triggered"
+	phaseNodeSyncComplete = "nodesync_complete"
+)
+
+// Scheme is the default instance of runtime.Scheme to which types in the Kubernetes API are already registered.
+// This is needed for ConfigMap search.
+var Scheme = runtime.NewScheme()
```
wojtek-t, line 53: Can we avoid this?

I see this is used only in one place for Search in events.
Can we instead create an appropriate field selector outselves?

It should actually be as simple as:
```
selector := client.CoreV1().Events(s.selector.Namespace).GetFieldSelector(svc.Name, svc.Namespace, nil, nil)
```

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,283 @@
...
+	nodeList, err := s.client.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodeList.Items {
+		if !isCandidateNode(node) {
+			continue
+		}
+		s.excludedNodeName = node.Name
```
wojtek-t, line 164: break ?

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,283 @@
...
+		s.svcNodeSyncLatencyTracker.Set(key, phaseNodeSyncStart, now)
+	}
+
+	return wait.Poll(5*time.Second, s.waitTimeout, func() (bool, error) {
+		for key, svc := range lbSvcMap {
+			if _, found := s.svcNodeSyncLatencyTracker.Get(key, phaseNodeSyncComplete); found {
+				continue
+			}
+
+			eventList, err := s.client.CoreV1().Events(s.selector.Namespace).Search(Scheme, &svc)
```
wojtek-t, line 185: I'm not a fan of depending on events in a programatic way - events aren't guaranteed.

What if instead we would make the measurement GCE-specific and try to get TargetPool contents from GCE and detect if the node was added/removed based on this?

The event-based test is really implementation-specific - we should target something more blackbox.

freehan: Making a measurement using GCE API calls has a few downsides in this tests. The test has to loop thru a potential high number of LBs and make GCE API calls repeatedly until the expected state has reached. This would eat up the read RPS quota in GCE the project and potentially cause controllers to hit GCE RateLimiting errors. Having a higher interval would make the measurement less accurate when high number of LBs are used in the load test.


Yes. I agree depending on events are less straightforward. We brainstormed a few ways to make this measurement. One better alternative is to read a metrics exposed by kube-controller-manager directly on master, but cluster loader currently cannot reach master. While the event creation timestamp does provide an accurate latency measurement and it is more a blackbox test than checking TargetPool.

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
freehan: Good call.

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
freehan: Fixed

prameshj (approved):

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,274 @@
...
+	for key := range lbSvcMap {
+		s.svcNodeSyncLatencyTracker.Set(key, phaseNodeSyncStart, now)
+	}
+
+	return wait.Poll(5*time.Second, s.waitTimeout, func() (bool, error) {
+		for key, svc := range lbSvcMap {
+			if _, found := s.svcNodeSyncLatencyTracker.Get(key, phaseNodeSyncComplete); found {
+				continue
+			}
+			selector := s.client.CoreV1().Events(s.selector.Namespace).GetFieldSelector(&svc.Name, &svc.Namespace, nil, nil)
```
wojtek-t, line 174: Continuing the discussion about events.

> Consuming GCE API calls quota.

How many services you would eventually like to test? I would like to understand if simple bump of quota won't help here. 
Instead of getting target pools one by one - can we list (potentially in batches)?


The current event-based approach is also inefficient, because given that your `s.selector.Namespace` is effectively empty, we will be listing all events from the whole clusters for every service independently (i.e. the returned result will be small, but we will need to internally process and filter out all events for every call).

freehan: I want to eventually test up to 500 LBs. (probably manual one time) And setup CI for 100 LBs. I have seen customers have >100 LBs. 

Listing TargetPools could not tell what instances are included. It requires a Get call to investigate the content and compare the instances one by one. 

I think clusterloader automatically creates separate namespaces for the load test. So `s.selector.Namespace` is only pointing to the test namespace. As I am using the feildSelector, the event Get call should only return events associated with the service, no?

Moreover, the event-based approach does not need to efficient, the retry interval can be configured to very high. This is because the latency measurement is depending on the `creationTimestamp` instead of the time when the validation took place.

wojtek-t: > I want to eventually test up to 500 LBs. (probably manual one time) And setup CI for 100 LBs. I have seen customers have >100 LBs.

OK - that doesn't sound that huge. On what size of cluster would you like to to test it?


> Listing TargetPools could not tell what instances are included. It requires a Get call to investigate the content and compare the instances against nodes one by one.

Got it.
I'm still a bit reluctant to using events - this is too whitebox-y to me. It's not future-proof, the fact that particular event was generated doesn't meen that LB works - it depends on the current implementation how the event is generated...

How sure are we that GCE API quota will be a problem? Can we mitigate it by choosing N random services and checking it for them? What if we would bump quotas?
[How many API calls do we need for updating nodes in a TP?]


> I think clusterloader automatically creates separate namespaces for the load test. 

Yes - but the selector.namespace is set to constant when you're parsing the config:
https://github.com/kubernetes/perf-tests/pull/1785/files#diff-dd38242128260b4a9cad8419f28cfb7f518618cd2e49c809fa532cfb4df36f4bR96
And you don't specify namespaces there - so it will be empty.
I'm not saying it wont work - I'm saying that it will be underneath listing all events in the system and filtering appropriate one.

> Moreover, the event-based approach does not need to efficient, the retry interval can be configured to very high. This is because the latency measurement is depending on the creationTimestamp instead of the time when the validation took place.

If we really need events (which I really hope we don't), a single informer sounds like a better option anyway:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/util/informer/informer.go

freehan: > I'm still a bit reluctant to using events - this is too whitebox-y to me. It's not future-proof, the fact that particular event was generated doesn't mean that LB works - it depends on the current implementation how the event is generated...

But checking TargetPool is also very whitebox-y. It only works for NetLB. For the classic L4 ILB, the event based approach also works. 

> How sure are we that GCE API quota will be a problem? 

The GCE API read quota is per project and across all APIs. By default, the quota is 20 QPS. I think it can be bumped but I only have impression with it bumped to 40 QPS at most. For 500 TPs, it took at least (assuming 20 QPS), 25s to traverse all of them. 

> Can we mitigate it by choosing N random services and checking it for them? What if we would bump quotas? [How many API calls do we need for updating nodes in a TP?]

This particular measurement is measuring how long would it take for service controller to update all LBs in the cluster when a nodesync got triggered. So it has to scan thru all LBs to be accurate. 

For each TP update, it actually consists of multiple GCE API calls. 1 List Instance , 1 TargetPool Add instances, 1 TargetPool delete instances, 2 Get TargetPool. So it is quite a few calls already. 


> If we really need events (which I really hope we don't), a single informer sounds like a better option anyway:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/util/informer/informer.go

I can change it to use informer if efficiency is the concern here.

wojtek-t: > But checking TargetPool is also very whitebox-y. It only works for NetLB. For the classic L4 ILB, the event based approach also works.

Can we use something different for ILB?
Maybe pinging the service and checking when it starts to reach the new node (we would have to add node instead of deleting it instead)?
I would really like to avoid event dependency...

>  For 500 TPs, it took at least (assuming 20 QPS), 25s to traverse all of them.

FYI - We're bumping it to 200qps in our largest projects.

prameshj: For ILB(Subsetting), I am planning to add a new measurement similar to [exec](https://github.com/kubernetes/perf-tests/blob/05b63b33bd15916ca92f9c724cc8ed83eca6a056/clusterloader2/pkg/measurement/common/exec.go#L47) so we can list the NEGs and verify the node names to be what we expect.

The presence of event can be used to trigger the NEG measurement, so we do not check the contents until necessary. We could omit it too. 

For IG-based ILB, we would need to list contents of the single instance group, since it is shared.
Perhaps we need very similar measurements for each case - NEG, IG, TargetPool. For IG and TargetPool, the steps are - 1) List all cluster nodes 2) list all instances in GCE resource. Ensure they have the same entries.

NEG is trickier since it uses a subset.

freehan: Let me summarize and compare the 2 approaches:

1. Event based approach (this PR)
   - Simple to implement
   - Latency measurement is accurate (from Nodesync starts and service controller finishes update LB hosts and emits event)
   - Can be used for both classic ILB/NetLB. 
   - Implementation specific (depending on the OSS service controller emitting the event)

2. TargetPool based approach
  - Polling based
  - Measurement is dependent on the frequency of the observation. (needs high RPS to )
  - Implementation specific (Only fits classic NetLB with Target Pool)


There are essentially 2 ways to implement the TargetPool based measurement:

A. Use Exec measurement as building block to run gcloud commands, the rough steps will be: 
  - Use the NetLB naming scheme and generate TargetPool name
  - Find the region
  - Run `gcloud to run gcloud compute target-pool describe ${targetpool} --region ${region}` (There is no list-istance gcloud API on targetPool)
  - Then do string parsing on the return output to match up with the expected nodes.

B. Introduce GCE API go clients into clusterloader2. 
- Construct the parameters (Targetpool Name and region) and then execute 
- use GCE API go client to poll target pool instances and parse it via golang. 

A few more factors to consider:
1. The immediate goal of this load test is to measure the scalability improvements caused by the parallelized backend management. 
2. RegionalBackendService migration is going to kick start soon. New NetLBs will be based on RBS instead of target pools

Overall, I would suggest the event based approach is a good compromise before RBS is in place. And we can improve this measurement incrementally. WDYT?

Liu-Aohan:
Hi,

Your Github profile and the contributions are really impressive and the work in Kubernetes is exactly what we are looking for!

As Shopee is expanding our Engineering teams here in Singapore and looking for talents like you. Here's our current openings https://careers.shopee.sg/job-detail/2628/ and I would like to check would that be something interesting to you?

Let me know if you are open for a quick discussion and I'm more than happy to assist you. If you're not keen right now, I'd still love to connect and stay in touch for other exciting opportunities in future.

Here are some key stats of Shopee for your quick review:

- 6th year upon launching
- Global presence in 11 countries, 13 offices
- 1 billion total # plays for in-app games
- 429.8 million gross orders just for Q1 2020
- US$6.2 billion gross merchant value just for Q1 2020  

Please feel free to add my WeChat: 13733199078 and ask any questions. Hope to hear from you soon!

Best regards

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
wojtek-t: I'm definitely not  fan of relying on events, but I don't want to block this infinitely.

So for the sake of making progress, let's proceed with the event-based approach for now (to unblock your tests).
I would like to get rid of this eventually, but we may discuss that separately.

So let's just fix it to use informer instead of periodically listing and proceed with this.

freehan:
Changed to use Event informer for processing. Ready for another round

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
freehan: I added a waitTimeout in the measurement. The test would fail if it took more than the timeout to measure nodesync.

wojtek-t:
@jkaniuk @marseel - can you take another look (at high-level I'm mostly fine with it)

clusterloader2/testing/l4lb/service.yaml
prameshj: port:8080 and targetPort: 80 should be ok as well, since the ServiceCreationLatency measurement will reach out to VIP:8080 in that case - https://github.com/kubernetes/perf-tests/blob/bb0129087ee1ac79d1800cad913e7f8a252e059b/clusterloader2/pkg/measurement/common/service_creation_latency.go#L340

prameshj (approved):

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,285 @@
...
+	s.selector.ApplySelectors(&options)
+	svcList, err := s.client.CoreV1().Services(s.selector.Namespace).List(ctx, options)
+	if err != nil {
+		return err
+	}
+
+	// lbSvcMap contains key->service mapping for load balancer type services
+	lbSvcMap := map[string]v1.Service{}
+	for _, svc := range svcList.Items {
+		if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
```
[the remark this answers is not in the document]
freehan: Yes. For classic ILB, this measurement should also work. 
But for one test, it can either run only NetLB or ILB. The current test config cannot run a mix of ILB and NetLB

freehan:
Fixed.

clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go
```diff
@@ -0,0 +1,284 @@
...
+			lbSvcMap[keyFunc(svc.Namespace, svc.Name)] = svc
+		}
+	}
+	totalLbSvc := len(lbSvcMap)
+
+	// Use event informer to keep track of nodeSync events.
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	eventInformer := informerv1.NewEventInformer(s.client, s.selector.Namespace, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
```
jkaniuk, line 142: Periodic listing of all events could be problematic in the biggest clusters. Could you use approach from [ooms_tracker.go#L125](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/ooms_tracker.go#L125) to ignore relists?
It basically ignores initial listing (and listing on e.g. broken watch) and uses only watches with [field selector](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/ooms_tracker.go#L203)

freehan: Fixed by using this pattern.

jkaniuk:
/lgtm
/approve
/hold
I would recommend to confirm it actually works in the wild before merging (thus holding), but leaving that up to you.

freehan:
Yes. I have run it many times. THanks!

/hold cancel

PR #1788, opened 2021-04-27
Reduce $totalSchedulerThroughputPods in large clusters

mborsz:
In large clusters we create 2x more latency pods than #nodes. We don't think it's needed, so let's try to remove that.

I will run 5k test to see what happens in that scale.
Presubmit should validate 100 nodes scale.
/hold

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -75,7 +75,7 @@
 
 # BEGIN scheduler-throughput section
 # TODO( https://github.com/kubernetes/perf-tests/issues/1027): Lower the number of "min-pods" once we fix the scheduler throughput measurement.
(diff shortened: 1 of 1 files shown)


mm4tt:
TBH, I'm more worried about the 100 node tests than the 5k tests. This change may increase the flakiness of the scheduling throughput measurement and to detect it a single presubmit run won't be enough. That said, I think we should just submit this and keep an eye on the periodic tests

mborsz:
Matt, correct me if I'm wrong, but I don't think this affects 100 nodes scale at all:
* before this change we had totalSchedulerThroughputPods = max(1000, 2 * nodes)
* after this change we will have max(1000, nodes)

This means, that up to 500 nodes, 1000 value is used in both cases.

mm4tt:
@mborsz, you're right. In that case I don't see any reason for not merging this. I'd be surprised if it affects 5k tests in any way. Creating 5k pods seems more than enough to accurately measure scheduling throughput.

mborsz:
/hold cancel

Thanks!

PR #1794, opened 2021-04-30
Add scheduler-throughput module

jupblb:
**What type of PR is this?**
/kind cleanup

Move scheduler throughput config to another module.

See #1659 for details. This change produces 0 difference in the config steps.
/assign marseel

```release-note
N/A
```

```docs
N/A
```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -63,8 +63,6 @@
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) .Nodes}}
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
-{{$schedulerThroughputThreshold := DefaultParam .CL2_SCHEDULER_THROUGHPUT_THRESHOLD 100}}
-{{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
 # END scheduler-throughput section
 
 # Set schedulerThroughputNamespaces to 1 on small clusters otherwise it will result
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,160 @@
...
+    Params:
+      action: unpause
+  {{end}}
+- name: Sleep
+  measurements:
+  - Identifier: WaitAfterExec
+    Method: Sleep
+    Params:
+      duration: {{$SLEEP_AFTER_EXEC_DURATION}}
+{{end}}
```
marseel, line 121: This whole part that executes arbitrary command shouldn't be part of scheduler-throughput.

jupblb: Probably fixed in https://github.com/kubernetes/perf-tests/pull/1794/commits/cd7360f8e9551620005b1063e37b72499d63d831 but let me know how do you see it.

jupblb: The diff is still empty.

clusterloader2/testing/load/config.yaml
```diff
@@ -428,45 +350,15 @@ steps:
       duration: {{$SLEEP_AFTER_EXEC_DURATION}}
 {{end}}
 
-{{if not $EXIT_AFTER_EXEC}}
+- module:
```
marseel, line 353: This merges some parts that were executed before and after ExecCommand so it changes test behaviour. Could we make something similar that we did in reconcile-module and have two actions "create" and "delete" so we wouldn't modify test behaviour?

clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,123 @@
+{{$namespaces := .namespaces}}
+{{$schedulerThroughputNamespaces := .schedulerThroughputNamespaces}}
+{{$schedulerThroughputPodsPerDeployment := .schedulerThroughputPodsPerDeployment}}
+
+{{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
+{{$EXIT_AFTER_EXEC := DefaultParam .CL2_EXIT_AFTER_EXEC false}}
```
marseel, line 6: If you implement idea with actions then we could get rid of this flag and move it to config.yaml.

clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,123 @@
...
+{{$IS_SMALL_CLUSTER := lt .Nodes 100}}
```
marseel, line 7: I would prefer to pass this as parameter so we don't have duplicated logic in two places that specifies what small cluster is.

clusterloader2/testing/load/modules/scheduler-throughput.yaml
jupblb: Fixed

clusterloader2/testing/load/config.yaml
jupblb: Done

clusterloader2/testing/load/modules/scheduler-throughput.yaml
jupblb: Not necessary anymore

clusterloader2/testing/load/modules/scheduler-throughput.yaml
jupblb: Done

jupblb:
The PR has been reworked as advised. The new diff is:
```
355c355
<   name: Creating scheduler throughput measurements
---
>   name: create scheduler throughput measurements
376c376
<   name: Creating scheduler throughput pods
---
>   name: create scheduler throughput pods
422c422
<   name: Deleting scheduler throughput pods
---
>   name: delete scheduler throughput pods
```
The major difference is that deleting throughput pods and huge services got swapped.

jupblb:
Updated to match the order of steps with the original config. The diff is now:
```
376c376
<   name: Creating scheduler throughput pods
---
>   name: create scheduler throughput pods
422c422
<   name: Deleting scheduler throughput pods
---
>   name: delete scheduler throughput pods
```

clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,113 @@
...
+      # The operation timeout shouldn't be less than 20m to make sure that ~10m node
+      # failure won't fail the test. See https://github.com/kubernetes/kubernetes/issues/73461#issuecomment-467338711
+      operationTimeout: 20m
+  - Identifier: SchedulingThroughput
+    Method: SchedulingThroughput
+    Params:
+      action: start
+      labelSelector: group = scheduler-throughput
+      measurmentInterval: 1s
+  {{if $ENABLE_HUGE_SERVICES}}
```
mborsz, line 43: I'm not a big fan of using "if is_creating" pattern -- technically we can simply take old configs put them here with if condition, but it doesn't reduce code complexity.

I think we can avoid doing that for "Creating huge services" step which (correct me if I'm wrong) can be implemented in a ccommon way for both creation and deletion paths (the only diff is replicasPerNamespace and acction name).

You can use the same pattern as for "{{$action}} scheduler throughput pods" step

clusterloader2/testing/load/modules/scheduler-throughput.yaml
```diff
@@ -0,0 +1,113 @@
...
+  phases:
+  - namespaceRange:
+      min: {{AddInt $namespaces 1}}
+      max: {{AddInt $namespaces $schedulerThroughputNamespaces}}
+    replicasPerNamespace: {{$replicasPerNamespace}}
+    tuningSet: SchedulerThroughputParallel
+    objectBundle:
+    - basename: scheduler-throughput-deployment
+      objectTemplatePath: simple-deployment.yaml
+{{if $is_creating}}
```
mborsz, line 66: You can avoid this if by simply putting templateFillMap unconditionally. It is simply ignored for deletion (i.e. the TemplateProvider is using some hack to avoid passing templateFillMap here, see https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/config/template_provider.go#L90-L95

clusterloader2/testing/load/modules/scheduler-throughput.yaml
mborsz: Ohh, I see why it's needed -- the order in creation and deletion is different here.

mborsz (commented):
Let's discuss this offline

clusterloader2/testing/load/modules/scheduler-throughput.yaml
jupblb: Yeah, it was a bit upsetting for me too. :/

clusterloader2/testing/load/modules/scheduler-throughput.yaml
jupblb: Done. Also thanks for referencing the code! :)

mborsz:
/lgtm
/approve

I think it's better than it was. We may want to iterate on this later to make it more readable, but it's not clear to me what we can do in this area for now, so LGTM for now

PR #1797, opened 2021-05-07
Add metrics-server latency measurement: part 1

sanwishe:
Signed-off-by: sanwishe <jiang.mingzhi35@zte.com.cn>

**What type of PR is this?**
/kind feature

implement of issue: [#710](https://github.com/kubernetes-sigs/metrics-server/issues/710)

Fixes #
part of  issue: [#710](https://github.com/kubernetes-sigs/metrics-server/issues/710)

```release-note
 "NONE"
```

```docs
 "NONE"
```


clusterloader2/pkg/measurement/common/metrics_server_measurement.go
```diff
@@ -0,0 +1,345 @@
...
+		resource == metricsServerRequestDurationResource &&
+		scope == metricsServerScope {
+		return true
+	}
+
+	return false
+}
+
+func (m *metricsServerLatencyMeasurement) getMetricsFromMS(kubeConf *rest.Config, metricsServerPodName string) (string, error) {
+	out, err := proxyRequestToPod(kubeConf, m.namespace, metricsServerPodName, "https", 4443, "/metrics")
```
serathius, line 243: This code assumes that metrics-server exposes port 4443. Unfortunatelly this differs from distribution to distribution and from release to release. 

For scalability tests I would expect that it uses https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/metrics-server configuraiton. This configuration uses port 443. 

 For this PR I would propose to use the port that is used by scalability tests, but we should prepare to be distribution independent by adding code to detect port based on containerPort configuration.

clusterloader2/pkg/measurement/common/metrics_server_measurement.go
```diff
@@ -0,0 +1,345 @@
...
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: trans}, http.MethodPost, &url.URL{Scheme: "https", Path: fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", namespace, podName), Host: hostIP})
+
+	var berr, bout bytes.Buffer
+	buffErr := bufio.NewWriter(&berr)
+	buffOut := bufio.NewWriter(&bout)
+
+	stopCh := make(chan struct{})
+	readyCh := make(chan struct{})
+
+	fw, err := portforward.New(dialer, []string{"4443:4443"}, stopCh, readyCh, buffOut, buffErr)
```
serathius, line 294: This port also needs to be updated. Check this code https://github.com/kubernetes-sigs/metrics-server/pull/730/files#diff-9e762d0df1502ee31a1de1a3bc67be6c2707ec3e6117be3147907a60ef5fb854R300

clusterloader2/pkg/measurement/common/metrics_server_measurement.go
```diff
@@ -0,0 +1,345 @@
...
+func (m *metricsServerLatencyMeasurement) getMetricsFromMS(kubeConf *rest.Config, metricsServerPodName string) (string, error) {
+	out, err := proxyRequestToPod(kubeConf, m.namespace, metricsServerPodName, "https", 4443, "/metrics")
+	if err != nil {
+		return "", err
+	}
+
+	return string(out), nil
+}
+
+func proxyRequestToPod(config *rest.Config, namespace, podName, scheme string, port int, path string) ([]byte, error) {
```
serathius, line 251: This is basically an implementation of HTTPS proxy, I think it would be useful to move it to some utils package.

sanwishe:
@serathius Thanks for your comments, and I will apply them as soon as possible.

serathius:
/lgtm 
PTAL @marseel

sanwishe:
> Generally, PR looks good.
> But I would prefer to extract common part, because we are duplicating a lot of code from https://github.com/kubernetes/perf-tests/pull/1684/files
> What we could do is basically prepare mechanism that takes 2 functions as interface:
> 
> 1. Returning latency histogram
> 2. Calling some endpoint
> 
> And move common logic to separate place, which takes this interface as parameter.
> It looks like this mechanism of gathering metrics at the beginning and end, subtracting histograms etc + calling some endpoint every X seconds might be used in the future too.
> What do you think about it?
> 
> @mborsz @sanwishe

Thanks, and I'm try to refactor it as soon

marseel:
> Thanks, and I'm try to refactor it as soon

One quick question before you proceed, is there any reason why we are not using prometheus here to scrape these metrics? We already have mechanism that allows you to specify pod/pods which prometheus should be scraping and we could just query prometheus to get metric we are interested in.

mborsz:
Hi, Thanks for PR!

I was wondering why can't we migrate to the following pattern:
* Instead of gathering metrics directly in measurement (retrieveMetricsServerLatencyMetrics function), use prometheusMeasurement, start scraping metrics-server's metric and in "gather", simply perform prometheus query to get summary? This is a pattern which we are trying to use in CL2 (see SchedulingThroughputPrometheus for this pattern and the previous, deprecated version SchedulingThroughput)
* Instead of probing metrics-server directly from CL2, we can deploy a pod in a cluster that can probe metrics-server endpoint periodically (see e.g. DnsLookupLatency and proberConfig struct).

This way we would be more consistent with the common measurement pattern used in CL2, we avoid a lot of logic duplication we have here and in kube-state-metrics measurement and make the code simpler (see InClusterNetworkLatency + DnsLookupLatency measurement which each takes ~10 lines of go each + some manifests).

yangjunmyfm192085:
@mborsz @marseel @serathius , we are ready for review.
could you please take a look?
Thank you very much.

clusterloader2/pkg/measurement/common/metrics_server_prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+func init() {
+	create := func() measurement.Measurement { return CreatePrometheusMeasurement(&metricsServerGather{}) }
+	if err := measurement.Register(metricsServerPrometheusMeasurementName, create); err != nil {
+		klog.Fatalf("Cannot register %s: %v", metricsServerPrometheusMeasurementName, err)
+	}
+}
+
+type metricsServerGather struct{}
```
marseel, line 40: Usually we call it Gather**er**

clusterloader2/pkg/measurement/common/metrics_server_prometheus.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (g *metricsServerGather) IsEnabled(config *measurement.Config) bool {
+	return config.CloudProvider.Features().SupportMetricsServerMetrics
+}
+
+func (g *metricsServerGather) String() string {
+	return metricsServerPrometheusMeasurementName
+}
+
+func (g *metricsServerGather) gatherLatencySummary(executor QueryExecutor, startTime time.Time) (*measurementutil.Histogram, error) {
```
marseel, line 69: Unless I am missing something, I think you could ski this function.
Please take a look at https://github.com/kubernetes/perf-tests/blob/81c96c34e3c1f11c5fe91744ef7ce7bf44c2fe5c/clusterloader2/pkg/measurement/common/nodelocaldns_latency_prometheus.go#L32
Check how percLatencyQueryTemplate looks like and how it's used. It computes latencies for specific period of time (so you don't need to do this computation on histograms)

sanwishe:
@marseel I had applied your suggestion. Please take a look, thank you.

util-images/probes/pkg/kubeclient/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+package kubeclient
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/perf-tests/util-images/probes/pkg/common"
+	"k8s.io/perf-tests/util-images/probes/pkg/util"
+)
+
+var (
+	inClusterAPIServerRequestLatency = prometheus.NewHistogram(prometheus.HistogramOpts{
```
marseel, line 26: If I understand correctly, you are specifying 3 prometheus histograms here, but you are not using them anywhere? I don't see any ".Observe" calls. I guess this can be used if you also want some client-side metrics, but if not I think you can get rid of it.

marseel:
Except for last comment everything else looks good to me
/lgtm

marseel:
/lgtm
/cc @wojtek-t 
/cc @jkaniuk 
Could you please take a look and approve if everything is fine?

mborsz:
WOW! This PR has changed a lot since the last time I was looking on it.

In my opinion it looks much better now, more code is reusable, logic is more clear + it's consistent with what we try to do in CL2.

Thanks a lot for all improvements!

/approve

mborsz:
I will reapprove once https://github.com/kubernetes/perf-tests/pull/1819 is merged

PR #1799, opened 2021-05-07
Revert "Upgrade perfdash to 2.33"

mborsz:
Reverts kubernetes/perf-tests#1796

Even 32GiB of memory isn't sufficient.

/assign @mm4tt

### DIFF
--- a/perfdash/deployment.yaml
+++ b/perfdash/deployment.yaml
@@ -15,7 +15,9 @@ spec:
     spec:
       containers:
       - name: perfdash
-        image: gcr.io/k8s-testimages/perfdash:2.33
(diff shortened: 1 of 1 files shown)


perfdash/deployment.yaml
```diff
@@ -15,7 +15,8 @@ spec:
     spec:
       containers:
       - name: perfdash
-        image: gcr.io/k8s-testimages/perfdash:2.33
+        # TODO: Bump to 2.33 and increase the memory limit once https://github.com/kubernetes/k8s.io/pull/1604 gets merged
```
mm4tt, line 18: Let's update this and explain the current state of affairs

PR #1803, opened 2021-05-13
Network Performance tc's: changes for fetching worker pod logs

VivekThrivikraman-est:
**What type of PR is this?**
/kind feature

Currently the worker pod logs cannot be accessed, this change is for making worker pods to write logs directly to /var/log on host.

Fixes #

NONE
```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/network/manifests/worker-deployment.yaml
+++ b/clusterloader2/pkg/measurement/common/network/manifests/worker-deployment.yaml
@@ -20,6 +20,10 @@ spec:
       - image: gcr.io/k8s-testimages/netperfbenchmark:0.1
         name: worker
         imagePullPolicy: IfNotPresent
+        args:
+          - --log_file=/var/log/cl2-netperf-worker.log
+          - --logtostderr=false
+          - --v=4
         resources:
           limits:
             cpu: "500m"
@@ -30,4 +34,11 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        volumeMounts:
+            - name: logs-volume
+              mountPath: /var/log
+      volumes:
+        - name: logs-volume
+          hostPath:
+            path: /var/log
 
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/network/manifests/worker-deployment.yaml
```diff
@@ -20,6 +20,9 @@ spec:
       - image: gcr.io/k8s-testimages/netperfbenchmark:0.1
```
marseel, line 20: You will probably need to create new image and push it ;)

util-images/network/netperfbenchmark/cmd/main.go
```diff
@@ -17,9 +17,15 @@ limitations under the License.
...
 import (
+	"flag"
+
+	"k8s.io/klog"
 	"k8s.io/perf-tests/util-images/network/netperfbenchmark/pkg/worker"
 )
 
 func main() {
+	klog.InitFlags(flag.CommandLine)
+	flag.Set("logtostderr", "false")
```
marseel, line 28: I would prefer to move it to args, that's how we usually do it.  Example
https://github.com/kubernetes/perf-tests/blob/81c96c34e3c1f11c5fe91744ef7ce7bf44c2fe5c/clusterloader2/pkg/measurement/common/probes/manifests/ping-server-deployment.yaml#L25

clusterloader2/pkg/measurement/common/network/manifests/worker-deployment.yaml
VivekThrivikraman-est: @wojtek-t could you please help in creating new image after this is committed?

util-images/network/netperfbenchmark/cmd/main.go
VivekThrivikraman-est: Modified.

clusterloader2/pkg/measurement/common/network/manifests/worker-deployment.yaml
marseel: I can help you with it ;)

VivekThrivikraman-est: Ok thanks!:)

PR #1809, opened 2021-05-20
Ignore exec subresource in APIResponsivenessPrometheus

jupblb:
**What type of PR is this?**
/kind bug

See this comment: https://github.com/kubernetes/perf-tests/issues/1096#issuecomment-842099659

N/A

See issue: https://github.com/kubernetes/perf-tests/issues/1096
See PR: https://github.com/kubernetes/test-infra/pull/21694

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -45,7 +45,7 @@ const (
 
 	currentAPICallMetricsVersion = "v1"
 
-	filters = `verb!="WATCH", subresource!="proxy"`
+	filters = `verb!="WATCH", subresource!="exec|proxy"`
 
 	// latencyQuery matches description of the API call latency SLI and measure 99th percentaile over 5m windows
 	//


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -45,7 +45,7 @@ const (
 
 	currentAPICallMetricsVersion = "v1"
 
-	filters = `verb!="WATCH", subresource!="proxy"`
+	filters = `verb!="WATCH", subresource!="exec|proxy"`
```
mborsz, line 48: Shouldn't it be subresource!~"exec|proxy"?

Instead it's just a simple string comparison with "exec|proxy" string, not regexp.

marseel: Definitely It should be

jupblb: Fixed in https://github.com/kubernetes/perf-tests/pull/1814

PR #1816, opened 2021-05-27
Change load test job to be more resilient

jprzychodzen:
Change to 'Parallel Jobs with a work queue' to 'Parallel Jobs with a fixed completion count'
Setting `.spec.completions` to double of possible replicas to make test more resilient for node restart.

https://kubernetes.io/docs/concepts/workloads/controllers/job/#parallel-jobs

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -20,6 +20,7 @@
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$ENABLE_API_AVAILABILITY_MEASUREMENT := DefaultParam .CL2_ENABLE_API_AVAILABILITY_MEASUREMENT false}}
 {{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
+{{$RANDOM_SCALE_FACTOR := 0.5}}
 #Variables
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
 {{$totalPods := MultiplyInt $namespaces $NODES_PER_NAMESPACE $PODS_PER_NODE}}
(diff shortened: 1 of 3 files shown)


clusterloader2/testing/load/job.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 spec:
   manualSelector: true
   parallelism: {{RandIntRange .ReplicasMin .ReplicasMax}}
+  completions: {{.Completions}}
```
mborsz, line 12: Personally I would simply write here:
```
# Set completions to value high enough, that we shouldn't never reach that
completions: 1000000
```

instead of doing this math and trying to calculate "right" value -- it's hard to say which one is right :)

mborsz: Alternatively write `{{MultiplyInt .ReplicasMax 100}}` here

clusterloader2/testing/load/config.yaml
```diff
@@ -255,7 +257,8 @@ steps:
       actionName: "scale and update"
       namespaces: {{$namespaces}}
       tuningSet: RandomizedScalingTimeLimited
-      randomScaleFactor: 0.5
+      randomScaleFactor: {{$RANDOM_SCALE_FACTOR}}
+      testMaxReplicaFactor: {{$RANDOM_SCALE_FACTOR}}
```
mborsz, line 261: It's very hard to tell what it does based on the name. Do we really need to expose this to top level of the config?

mborsz (commented):
I think it's a good change: It should remove a failure mode when a single pod succeeds for whatever reason and then we stop e.g. scaling those pods.

Just few comments re implementation

clusterloader2/testing/load/job.yaml
jprzychodzen: I was thing about it to write big enough number, however this could generate strange errors when huge clusters with jobs only workload would be tested.

Cannot use `{{MultiplyInt .ReplicasMax 100}}` as this needs to be constant over the test. - ReplicasMax is derivered from `randomScaleFactor`, which changes for different actions.

clusterloader2/testing/load/config.yaml
jprzychodzen: Not sure what other alternative we have - we could use constant, however I do not like the idea that this constant will silently fail based on some other change.

mborsz: Ohh,.. Now I understand what it means.

How about using "testMaxRandomScaleFactor" name or something like that? I initially didn't even now that it refers to maximum allowable value for randomScaleFactor

clusterloader2/testing/load/job.yaml
mborsz: Now I understand the reasoning behind this implementation. Thanks for explanation.

PR #1819, opened 2021-06-01
Add mborsz to util-images/probes/OWNERS

mborsz:
The current set of OWNERS in that directory doesn't provide a good overage (technically only wojtek-t is able to do reviews).

Add me to avoid sending all PRs there to wojtek-t.

### DIFF
--- a/util-images/probes/OWNERS
+++ b/util-images/probes/OWNERS
@@ -1,8 +1,10 @@
 approvers:
+- mborsz
 - mm4tt
 - oxddr
 - wojtek-t
 reviewers:
+- mborsz
 - mm4tt
 - oxddr
 - wojtek-t


util-images/probes/OWNERS
```diff
@@ -1,8 +1,10 @@
 approvers:
+- mborsz
```
wojtek-t, line 2: With all your contributions and detailed reviews, I actually suggest that we add you to the top-level owners in this repo.

I would be willing to approve this change, but I would like to give time for @mm4tt and @shyamjvs to objects.

But in general - please open such PR and we will let it wait for a week or so (unless everyone approve by then).

mborsz: Thanks Wojtek, I opened https://github.com/kubernetes/perf-tests/pull/1820

PR #1820, opened 2021-06-01
Add mborsz to OWNERS

mborsz:
Adds mborsz to top-level OWNERS as suggested by @wojtek-t in https://github.com/kubernetes/perf-tests/pull/1819#pullrequestreview-672742974.

/cc @gmarek @mm4tt @shyamjvs @timothysc @wojtek-t

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,5 +1,6 @@
 approvers:
 - gmarek
+- mborsz
 - mm4tt
 - shyamjvs
 - timothysc


wojtek-t:
I'm supportive, but I would like to hear from other leads explicitly.

/assign @shyamjvs 
/assign @mm4tt

wojtek-t:
/lgtm

/hold cancel

Thanks!

Issue #1575, opened 2020-11-17
Load test shouldn't delete configmaps and secrets for running pods


mborsz:
From https://github.com/kubernetes/kubernetes/issues/96038#issuecomment-728988763

It looks like we do delete deployment, configmap and secret at roughly the same time. This is incorrect in general, as there can be large delay between deployment deletion and pod deletion (e.g. if we make our tests even faster and the kube-controller-manager has a O(minutes) backlog of pods to be deleted).

During that delay, the running pod should be able to restart for whatever reason (e.g. kernel panic, machine failure).

We should modify our test to delete configmaps and secrets after we waited for all pods to be deleted.

wojtek-t:
I don't think that's really a bug per-se. Yes - technically we can delete secrets/configmaps before deleting a pod, but that's supported usecase in k8s.

PR #1822, opened 2021-06-07
Mitigate https://github.com/kubernetes/kubernetes/issues/96635 in load test

mborsz:
Mitigates https://github.com/kubernetes/kubernetes/issues/96635 in load test.

Currently if the following scenario happens:
* We delete deployment X
* We delete configmap used by deployment X
* Any node restarts (which is possible in 5k scale)
* The node will not be able to delete pod in deployment X due to https://github.com/kubernetes/kubernetes/issues/96635

With this mitigation, we delete configmaps/secrets only when we already know that all pod using that pod have been deleted.

Ref https://github.com/kubernetes/perf-tests/issues/1575
/assign @jkaniuk

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -84,9 +84,17 @@ tuningSets:
   parallelismLimitedLoad:
     parallelismLimit: {{$schedulerThroughputNamespaces}}
 # TODO(https://github.com/kubernetes/perf-tests/issues/1024): This TuningSet is used only for pod-startup-latency, get rid of it
+# Uniform5qps: for each running phase, use 5 qps.
 - name: Uniform5qps
   qpsLoad:
     qps: 5
+# Global100qps: use 100 qps globally:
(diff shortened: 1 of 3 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -130,6 +138,16 @@ steps:
       - basename: daemonset-priorityclass
         objectTemplatePath: daemonset-priorityclass.yaml
 
+- module:
```
wojtek-t, line 141: Can we add a TODO (and open an issue) to merge back those sections once the bug in upstream will be fixed?

mborsz: Added

jkaniuk: nit: "remove when k/k bug is fixed"?

mborsz: Fixed PTAL

mborsz:
/hold cancel

The comment has been addressed

PR #1840, opened 2021-06-22
Parallelize scheduler-throughput pod creation

jupblb:
**What type of PR is this?**
/kind bug

Currently, in load test we have 5k pods in huge service. Unfortunately this is too big and causes flakiness of our tests. We would like to find appropriate size of huge service so we can include it in our scalability documentation. The idea is to split huge service to several smaller deployments, and document max service size equal to the size of single deployment.

N/A

N/A

```release-note
N/A
```

```docs
N/A
```

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -20,6 +20,7 @@
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$ENABLE_API_AVAILABILITY_MEASUREMENT := DefaultParam .CL2_ENABLE_API_AVAILABILITY_MEASUREMENT false}}
 {{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
+{{$ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS := DefaultParam .CL2_ENABLE_HUGE_SERVICES_MULTIPLE_REPLICAS false}}
 # Determines number of pods per deployment. Should be a divider of .Nodes.
 {{$HUGE_SERVICES_SIZE := DefaultParam .CL2_HUGE_SERVICES_SIZE 1000}}
 {{$RANDOM_SCALE_FACTOR := 0.5}}
@@ -64,7 +65,7 @@
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
 
-{{if and $ENABLE_HUGE_SERVICES (ge .Nodes $HUGE_SERVICES_SIZE)}}
+{{if and $ENABLE_HUGE_SERVICES $ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS (ge .Nodes $HUGE_SERVICES_SIZE)}}
   {{$schedulerThroughputReplicasPerNamespace = DivideInt .Nodes $HUGE_SERVICES_SIZE}}
   {{$schedulerThroughputPodsPerDeployment = DivideInt .Nodes $schedulerThroughputReplicasPerNamespace}}
   {{$schedulerThroughputNamespaces = 1}}
@@ -195,6 +196,7 @@ steps:
     path: modules/huge-services.yaml
     params:
       action: create
+      enableMultipleDeployments: {{$ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS}}
       namespaces: {{$namespaces}}
       replicasPerNamespace: 1
       schedulerThroughputNamespaces: {{$schedulerThroughputNamespaces}}
@@ -203,6 +205,7 @@ steps:
     path: modules/scheduler-throughput.yaml
     params:
       action: create
+      enableMultipleDeployments: {{$ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS}}
       namespaces: {{$namespaces}}
       replicasPerNamespace: {{$schedulerThroughputReplicasPerNamespace}}
       schedulerThroughputNamespaces: {{$schedulerThroughputNamespaces}}
@@ -257,6 +260,7 @@ steps:
     path: modules/scheduler-throughput.yaml
     params:
       action: delete
+      enableMultipleDeployments: {{$ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS}}
       namespaces: {{$namespaces}}
       replicasPerNamespace: 0
       schedulerThroughputNamespaces: {{$schedulerThroughputNamespaces}}
@@ -267,6 +271,7 @@ steps:
     path: modules/huge-services.yaml
     params:
       action: delete
+      enableMultipleDeployments: {{$ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS}}
       namespaces: {{$namespaces}}
       replicasPerNamespace: 0
       schedulerThroughputNamespaces: {{$schedulerThroughputNamespaces}}
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -20,6 +20,8 @@
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$ENABLE_API_AVAILABILITY_MEASUREMENT := DefaultParam .CL2_ENABLE_API_AVAILABILITY_MEASUREMENT false}}
 {{$ENABLE_HUGE_SERVICES := DefaultParam .CL2_ENABLE_HUGE_SERVICES false}}
+# Determines number of pods per deployment. Should be a divider of .Nodes.
+{{$HUGE_SERVICES_SIZE := DefaultParam .CL2_HUGE_SERVICES_SIZE 1000}}
 {{$RANDOM_SCALE_FACTOR := 0.5}}
 #Variables
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
@@ -58,13 +60,20 @@
 # BEGIN scheduler-throughput section
 # TODO( https://github.com/kubernetes/perf-tests/issues/1027): Lower the number of "min-pods" once we fix the scheduler throughput measurement.
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) .Nodes}}
+{{$schedulerThroughputReplicasPerNamespace := 1}}
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
-# END scheduler-throughput section
+
+{{if and $ENABLE_HUGE_SERVICES (ge .Nodes $HUGE_SERVICES_SIZE)}}
+  {{$schedulerThroughputReplicasPerNamespace = DivideInt .Nodes $HUGE_SERVICES_SIZE}}
+  {{$schedulerThroughputPodsPerDeployment = DivideInt .Nodes $schedulerThroughputReplicasPerNamespace}}
+  {{$schedulerThroughputNamespaces = 1}}
+{{end}}
 
 # Set schedulerThroughputNamespaces to 1 on small clusters otherwise it will result
 # in an unnecessary number of namespaces.
 {{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}
+# END scheduler-throughput section
 
 # Command to be executed
 {{$EXEC_COMMAND := DefaultParam .CL2_EXEC_COMMAND nil}}
@@ -82,7 +91,7 @@ tuningSets:
 # parallel creation of deployments.
 - name: SchedulerThroughputParallel
   parallelismLimitedLoad:
-    parallelismLimit: {{$schedulerThroughputNamespaces}}
+    parallelismLimit: {{MultiplyInt $schedulerThroughputNamespaces $schedulerThroughputReplicasPerNamespace}}
 # TODO(https://github.com/kubernetes/perf-tests/issues/1024): This TuningSet is used only for pod-startup-latency, get rid of it
 # Uniform5qps: for each running phase, use 5 qps.
 - name: Uniform5qps
@@ -104,6 +113,9 @@ tuningSets:
     # as each RS changes its size from X to a uniform random value in [X/2, 3X/2].
     # To match 10 [pods/s] requirement, we need to divide saturationTime by 4.
     timeLimit: {{DivideInt $saturationTime 4}}s
+- name: DeletionTimeLimited
+  TimeLimitedLoad:
+    timeLimit: 5m
 - name: RandomizedDeletionTimeLimited
   RandomizedTimeLimitedLoad:
     timeLimit: {{$deletionTime}}s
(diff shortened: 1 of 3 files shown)


jkaniuk:
Could you make it customizable? Some env value that will set max size of the service, with the logic to split number of pods into chunks of max that size.

Probably for another PR:
In the future, we need to test larger service than number of nodes. So if env var is set higher than number of nodes, create a single service with such many pods. That is not so straightforward, as we need to reduce number of other pods to fit it.

clusterloader2/testing/load/config.yaml
```diff
@@ -57,15 +57,27 @@
...
+  {{$schedulerThroughputPodsPerDeployment = .Nodes}}
+  {{$schedulerThroughputNamespaces = 1}}
+{{end}}
+{{if and (ge .Nodes 100) (lt .Nodes 1000)}}
+  {{$totalSchedulerThroughputPods := MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS}}
+  {{$schedulerThroughputReplicasPerNamespace = 1}}
+  {{$schedulerThroughputPodsPerDeployment = .Nodes}}
+  {{$schedulerThroughputNamespaces = DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
+{{end}}
+{{if ge .Nodes 1000}}
```
jkaniuk, line 74: Instead of using constant, it should be a customizable value (1000 as default is ok) 

Even with constant, it would be better to declare it once in the header.

clusterloader2/testing/load/config.yaml
```diff
@@ -57,15 +57,27 @@
...
-{{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
+{{$schedulerThroughputReplicasPerNamespace := 0}}
+{{$schedulerThroughputPodsPerDeployment := 0}}
+{{$schedulerThroughputNamespaces := 0}}
+{{if $IS_SMALL_CLUSTER}}
+  {{$schedulerThroughputReplicasPerNamespace = 1}}
+  {{$schedulerThroughputPodsPerDeployment = .Nodes}}
+  {{$schedulerThroughputNamespaces = 1}}
+{{end}}
+{{if and (ge .Nodes 100) (lt .Nodes 1000)}}
```
jkaniuk, line 68: This duplicates `if $IS_SMALL_CLUSTER` and `if ge .Nodes 1000`.

Could you deduplicate gating logic using `else` ?

Do we even need 3 cases. I think <100 and 100>= should be enough with previous version of calculating totalPods.

clusterloader2/testing/load/config.yaml
jupblb: Added `CL2_LARGE_CLUSTER_NODES` parameter

clusterloader2/testing/load/config.yaml
jupblb: Simplified the logic down to:
```
{{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) .Nodes}}
{{$schedulerThroughputReplicasPerNamespace := 1}}

{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}

{{if $IS_LARGE_CLUSTER}}
  {{$schedulerThroughputReplicasPerNamespace = DivideInt .Nodes $LARGE_CLUSTER_NODES}}
  {{$schedulerThroughputPodsPerDeployment = DivideInt .Nodes $schedulerThroughputReplicasPerNamespace}}
  {{$schedulerThroughputNamespaces = 1}}
{{end}}
```

jupblb: We could do one more polish here with:
```
{{$schedulerThroughputNamespaces := IfThenElse (or $IS_SMALL_CLUSTER $IS_LARGE_CLUSTER) 1 $schedulerThroughputNamespaces}}
```
Please let me know what is your opinion

jupblb:
Added the `$LARGE_CLUSTER_NODES` variable with derivative `$IS_LARGE_CLUSTER`.

clusterloader2/testing/load/config.yaml
```diff
@@ -7,8 +7,10 @@
 #Constants
 # Cater for the case where the number of nodes is less than nodes per namespace. See https://github.com/kubernetes/perf-tests/issues/887
 {{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .NODES_PER_NAMESPACE 100)}}
+{{$LARGE_CLUSTER_NODES := DefaultParam .CL2_LARGE_CLUSTER_NODES 1000}}
```
jkaniuk, line 10: It is not about how big cluster is "large" but some test config marking the size of tested services that could be set arbitrarily.

CL2_HUGE_SERVICES_SIZE ?

clusterloader2/testing/load/config.yaml
```diff
@@ -7,8 +7,10 @@
...
 # See https://github.com/kubernetes/perf-tests/pull/1667#issuecomment-769642266
 {{$IS_SMALL_CLUSTER := lt .Nodes 100}}
+{{$IS_LARGE_CLUSTER := ge .Nodes $LARGE_CLUSTER_NODES}}
```
jkaniuk, line 13: I would not define this variable, I do not have a good name for it.

clusterloader2/testing/load/config.yaml
jkaniuk: Also - could you add a comment about allowed values - not larger than .Nodes and .Nodes should be divisibile by it.

We should address those restrictions later.

clusterloader2/testing/load/config.yaml
```diff
@@ -58,14 +60,21 @@
...
+{{$schedulerThroughputReplicasPerNamespace := 1}}
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
-# END scheduler-throughput section
 
 # Set schedulerThroughputNamespaces to 1 on small clusters otherwise it will result
 # in an unnecessary number of namespaces.
 {{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}
 
+{{if $IS_LARGE_CLUSTER}}
```
jkaniuk, line 71: Could you move this block above the previous with small clusters?

clusterloader2/testing/load/modules/huge-services.yaml
```diff
@@ -18,11 +15,3 @@ steps:
     objectBundle:
     - basename: huge-service
       objectTemplatePath: service.yaml
-{{if $is_deleting}}
-- name: Sleeping after deleting huge services
```
jkaniuk, line 22: As I understand, it is being replaced with DeletionTimeLimited. Could you explain how it exactly works?

marseel: I've suggested using DeletionTimeLimited. Let's say you have 5 actions and duration is 5 minutes. It will execute 1 action every 1 minute.

clusterloader2/testing/load/config.yaml
jupblb: The current logic distincts between a case where there are [100, 1000) nodes and more than a 1000 nodes. In the first case there are multiple namespaces and the number of pods is changed to 2 * `$MIN_PODS_IN_SMALL_CLUSTERS`, in the latter it's exactly one namespace and the number of pods per deployment is different.

I believe the most optimal way of setting this would be indeed to have the `$CL2_HUGE_SERVICES_SIZE` variable, applied only when `$ENABLE_HUGE_SERVICES` is `true`. Then we can change the final `if` block to:
```
{{if $ENABLE_HUGE_SERVICES}}
  {{$schedulerThroughputReplicasPerNamespace = DivideInt .Nodes $HUGE_SERVICES_SIZE}}
  {{$schedulerThroughputPodsPerDeployment = DivideInt .Nodes $schedulerThroughputReplicasPerNamespace}}
  {{$schedulerThroughputNamespaces = 1}}
{{end}}
```
Does this sound good to you?

clusterloader2/testing/load/modules/huge-services.yaml
jupblb: It is what Marcel said. As per code comment: `TimeLimitedLoad defines a load that spreads operations over given time.`

jkaniuk: What if we increase huge services size. Would this sleep be needed once again?

clusterloader2/testing/load/config.yaml
jupblb: I guess we could name it `HUGE_SERVICES_PODS_PER_DEPLOYMENT` instead of `HUGE_SERVICES_SIZE`

clusterloader2/testing/load/config.yaml
jkaniuk: It is 1 in both cases but the cause is different. I do not think this would increase readability.

clusterloader2/testing/load/modules/huge-services.yaml
jupblb: As far as I understand no, because we would have `n < 5 (iff n = 5000)` operations still spread over 5 minutes

clusterloader2/testing/load/config.yaml
jupblb: Ack

clusterloader2/testing/load/modules/huge-services.yaml
jkaniuk: Once again - 1000 is just a default value. We will increase it in the future. Target is 10,000

clusterloader2/testing/load/config.yaml
jkaniuk: Both names are fine, HUGE_SERVICES_SIZE is concise and readable for me.

> Does this sound good to you?
Yes, that was exactly my suggestion.

clusterloader2/testing/load/modules/huge-services.yaml
jupblb: We could try to make the time limit a parameter rather than a constant. But I have little idea what it could be... `(pods / 1000) * 1m`? Or somehow reuse `$deletionTime`?

jkaniuk: @marseel - I'm leaving that to you

clusterloader2/testing/load/config.yaml
```diff
@@ -64,15 +64,15 @@
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
 
-# Set schedulerThroughputNamespaces to 1 on small clusters otherwise it will result
-# in an unnecessary number of namespaces.
-{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}
-
-{{if $IS_LARGE_CLUSTER}}
-  {{$schedulerThroughputReplicasPerNamespace = DivideInt .Nodes $LARGE_CLUSTER_NODES}}
+{{if $HUGE_SERVICES_SIZE}}
```
jkaniuk, line 67: is > .Nodes

jupblb: wtf me, fixed with
```
{{if $ENABLE_HUGE_SERVICES}}
```

jupblb: Changed to:
```
{{if and $ENABLE_HUGE_SERVICES (ge .Nodes $HUGE_SERVICES_SIZE)}}
```

clusterloader2/testing/load/config.yaml
```diff
@@ -64,15 +64,15 @@
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
 
-# Set schedulerThroughputNamespaces to 1 on small clusters otherwise it will result
-# in an unnecessary number of namespaces.
-{{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}
-
-{{if $IS_LARGE_CLUSTER}}
-  {{$schedulerThroughputReplicasPerNamespace = DivideInt .Nodes $LARGE_CLUSTER_NODES}}
+{{if $ENABLE_HUGE_SERVICES}}
```
jkaniuk: nope ;)

jupblb: already changed to
```
{{if and $ENABLE_HUGE_SERVICES (ge .Nodes $HUGE_SERVICES_SIZE)}}
```
the previous version was made after agreeing on it here: https://github.com/kubernetes/perf-tests/pull/1840#discussion_r658611902

jkaniuk: thanks

jkaniuk: not sure if ENABLE_HUGE_SERVICES check is needed, but I'm fine both ways

clusterloader2/testing/load/modules/huge-services.yaml
marseel: I would leave it as it is. Once we have 5k or 10k services, I would like to get rid of these multiple deployments.

clusterloader2/testing/load/config.yaml
jkaniuk: After some consideration - `HUGE_SERVICES_PODS_PER_DEPLOYMENT` makes more sense

clusterloader2/testing/load/modules/huge-services.yaml
```diff
@@ -18,7 +16,7 @@ steps:
     objectBundle:
     - basename: huge-service
       objectTemplatePath: service.yaml
-{{if $is_deleting}}
+{{if (eq .action "delete") $enableMultipleDeployments}}
```
marseel, line 19: And is missing I think?

clusterloader2/testing/load/config.yaml
```diff
@@ -58,13 +61,20 @@
 # BEGIN scheduler-throughput section
 # TODO( https://github.com/kubernetes/perf-tests/issues/1027): Lower the number of "min-pods" once we fix the scheduler throughput measurement.
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) .Nodes}}
+{{$schedulerThroughputReplicasPerNamespace := 1}}
 {{$schedulerThroughputPodsPerDeployment := .Nodes}}
 {{$schedulerThroughputNamespaces := DivideInt $totalSchedulerThroughputPods $schedulerThroughputPodsPerDeployment}}
-# END scheduler-throughput section
+
+{{if and $ENABLE_HUGE_SERVICES $ENABLE_HUGE_SERVICES_MULTIPLE_DEPLOYMENTS  (ge .Nodes $HUGE_SERVICES_SIZE)}}
```
marseel, line 68: nit: "S  (" two spaces

clusterloader2/testing/load/modules/huge-services.yaml
jupblb: Thanks, fixed

clusterloader2/testing/load/config.yaml
jupblb: Thanks, fixed

clusterloader2/testing/load/config.yaml
```diff
@@ -104,6 +114,9 @@ tuningSets:
     # as each RS changes its size from X to a uniform random value in [X/2, 3X/2].
     # To match 10 [pods/s] requirement, we need to divide saturationTime by 4.
     timeLimit: {{DivideInt $saturationTime 4}}s
+- name: DeletionTimeLimited
+  TimeLimitedLoad:
+    timeLimit: 5m
```
marseel, line 119: I think you added it accidentally.  
Also, at the end we probably want to remove this 5m sleep. Could you put sleep within ifs?

marseel:
/lgtm
Please squash commits first.

marseel:
I've discussed this PR with current scalability oncall and we decided to merge it despite 2 commits to reduce flakiness.

Issue #1818, opened 2021-05-31
Network Performance tests: negative UDP latency values for some test runs.


VivekThrivikraman-est:
**What happened**:
UDP latency metrics collected while running network performance test cases (clusterloader2/testing/network) intermittently contains negative values.

**What you expected to happen**:
UDP latency should be always positive.

**How to reproduce it (as minimally and precisely as possible)**:
Issue is not seen when running locally in Kind or local vm cluster. Negative values are seen when the testcases are run in k8s CI/CD(GCE).

**Anything else we need to know?**:
It is  confirmed from the logs that Iperf2 tool which is used for collecting the metrics return these -ve values. Raised this in iperf2 forum :
https://sourceforge.net/p/iperf2/discussion/general/thread/a1bcdb67d8/

They suspect this to be clock synchronization issue(they mentioned NTP synchronized nodes also can have latency).

Aggregated metrics output for UDP attached:
[NetworkPerformanceMetrics_N_M_UDP_P2P_network_performance_udp-50_50_2021-05-21T08_39_04Z.json.txt](https://github.com/kubernetes/perf-tests/files/6568363/NetworkPerformanceMetrics_N_M_UDP_P2P_network_performance_udp-50_50_2021-05-21T08_39_04Z.json.txt)

**Environment**:
Testcases are run as part of prow job in k8s CI/CD, in GCE.

aojea:
We've been talking about this in slack, and since the VM synchronisation seems complex problem to fix , we though that the best solution moving forward will be to gather more data to get better insights, maybe we only need the trend and not the absolute values.

The doubt that we have is if it would be good to operate only with positive numbers, i.e. truncate negative values to 0

@wojtek-t what do you think?

wojtek-t:
We do clock-synchronization in GCE - I personally doubt it's clock difference. I bet it's a real bug...

VivekThrivikraman-est:
@wojtek-t is it NTP ?

hanamantagoudvk:
@wojtek-t @aojea : What do you think of considering negative values as zero. ? Shall we proceed with that idea ?

mm4tt:
It's NTP - https://cloud.google.com/compute/docs/instances/managing-instances#configure-ntp

VivekThrivikraman-est:
Thanks @mm4tt , as you mentioned we will first fix perfdash and monitor to make sure latency is always in microseconds and then make a decision.

aojea:
[...]

PR #1843, opened 2021-06-24
Adding additional metric for UDP Protocol

hanamantagoudvk:
a) Loss of packet in percentage added
b) Percentile calculation corrected

### DIFF
--- a/clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
+++ b/clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
@@ -358,6 +358,12 @@ func (npm *networkPerformanceMeasurement) calculateMetricDataValue(dataElem *mea
 	case manyToMany:
 		for _, metricResponse = range npm.metricVal {
 			if len(metricResponse) > 0 {
+				// Sometimes iperf gives negative values for latency. As short-term fix
+				// we are considering them as zero.
+				if metricIndex == udpLatencyAverage && metricResponse[metricIndex] < 0 {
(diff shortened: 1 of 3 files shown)


hanamantagoudvk:
@wojtek-t @mm4tt : could you please review this PR ?

clusterloader2/pkg/measurement/common/network/network_performance_measurement.go
```diff
@@ -358,6 +358,11 @@ func (npm *networkPerformanceMeasurement) calculateMetricDataValue(dataElem *mea
 	case manyToMany:
 		for _, metricResponse = range npm.metricVal {
 			if len(metricResponse) > 0 {
+				// Sometimes iperf gives negative values for latency. As short-term fix
+				// we are considering them as zero.
+				if metricIndex == udpLatencyAverage && metricResponse[metricIndex] < 0 {
+					continue
```
marseel, line 364: I see that later on you are computing percentile based on `aggregatePodPairMetrics ` array. Here you are skipping adding results for udpLatencyAverage if latency is lower than zero. If you want to assume that negative latency is zero as short term fix, don't you want to append this zero to array for percentile computation? 
Could you clarify maybe what negative latency means (I can imagine that adding zero/not adding zero to array might depend on it).

hanamantagoudvk: Hi @marseel , you can fine more context to this PR in the issue we raised https://github.com/kubernetes/perf-tests/issues/1818

Issue #1746, opened 2021-02-20
Improve CL2 config validation


lyzs90:
Continuation of #1636. There's still a quite bit more we can validate, for instance:
- Ensure that steps are not empty
- Ensure that the tuning set referenced in a phase has been declared
- Implement verifyMeasuremen
  - APIAvailability, EtcdMetrics, Exec, MetricsForE2E, NodeLocalDNSLatencyPrometheus, etc... e
  - Each measurement may have custom parameters

You can link to this issue in your PR and comment here that you're working on certain packages to help avoid duplicated work.
Please do NOT put Fixes #... in your Pull Request. That will close this PR. Instead put Part of #1746.

siddhartha97:
@lyzs90 newbie here, can I work on this? (:

lyzs90:
@siddhartha97 Sure go ahead please :) Try to keep the PRs small. Eg  for verifyMeasurement, we can start off with one measurement first instead of tackling all at once.

siddhartha97:
Will do! @lyzs90 Is there any specific directory i should be looking at in the repository to start off with _Implement verifyMeasurement_? 
/assign

lyzs90:
Sorry it was renamed to validateMeasurement, you implement the function here: https://github.com/kubernetes/perf-tests/blob/b9e08a85e93daff53f1a68370cbbfa4de77dffe1/clusterloader2/api/validation.go#L180:27

Tests for validations are written here: https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/api/validation_test.go

The whole list of measurement can be found here: https://github.com/kubernetes/perf-tests/tree/b9e08a85e93daff53f1a68370cbbfa4de77dffe1/clusterloader2/pkg/measurement/common

An example of a test config is the [load test](https://github.com/kubernetes/perf-tests/blob/b9e08a85e93daff53f1a68370cbbfa4de77dffe1/clusterloader2/testing/load/config.yaml).

[...]

PR #1845, opened 2021-06-25
added basic validation for validateMeasurement

rahulii:
**What type of PR is this?**
/kind feature

This PR implements valideMeasurement function

Part of #1746 

WIP

```release-note
NONE
```

```docs

```


rahulii:
```cmd
go test .
package k8s.io/perf-tests/clusterloader2/api
        imports k8s.io/perf-tests/clusterloader2/pkg/measurement
        imports k8s.io/perf-tests/clusterloader2/api: import cycle not allowed
```
/assign @marseel

marseel:
@rahulii So measurement package is importing api package. What we could do is to move validation.go to separate package so we don't have import cycle. For example moving "api/validation.go" to api/validation/validation.go" and changing package from "api" to "validation" should help.

rahulii:
> @rahulii So measurement package is importing api package. What we could do is to move validation.go to separate package so we don't have import cycle. For example moving "api/validation.go" to api/validation/validation.go" and changing package from "api" to "validation" should help.

Okay and then maybe change the import path of all the places where validation maybe be used?

rahulii:
Ah , @marseel there's an issue! validation.go uses types.go for all the Types and config(eg Namespace,etc) which is in api package , and many files use types.go as well!
```cmd
./validation.go:123:45: undefined: Object
./validation.go:135:54: undefined: NamespaceRange
./validation.go:146:49: undefined: TuningSet
./validation.go:184:51: undefined: Measurement
```

clusterloader2/api/validation/validation_test.go
```diff
@@ -441,3 +442,29 @@ func TestValidate(t *testing.T) {
...
+			name: "first-test",
+			input: api.Measurement{
+				Method : "APIAvailability",
+				Identifier : "APIAvailability",
+			},
+			expected:nil,
+		},
+	}{
+		t.Run(test.name,func(t *testing.T) {
+			v := NewConfigValidator("", &api.Config{})
```
rahulii, line 462: @marseel , so while writing the unit tests , how do I register the measurement, 
I know in factory there is a `register` method , but it takes second argument `createFunc createMeasurementFunc` , so not sure how do I call that function?
Otherwise every time test runs, it says `measurement not registered`.

marseel: You  need to create function that returns empty measurement and pass it to register function. This function is defined as: `type createMeasurementFunc func() Measurement`

rahulii: So , should that empty measurement also imlement the three methods, `Execute`, `String` and `Dispose` which will probably empty?

rahulii:
```cmd
func createAPIAvailabilityMeasurement() measurement.Measurement {
	return &apiAvailabilityMeasurement{}
}
type apiAvailabilityMeasurement struct {
}
func (a *apiAvailabilityMeasurement) Dispose() {}

func (a *apiAvailabilityMeasurement) String() string {
	return "APIAvailability"
}
func (a *apiAvailabilityMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
	return nil,nil
}
```

rahulii: @marseel , looks good?

rahulii: The test case runs fine , but there is another problem where in other test cases the measurement name in dummy and not registered!

marseel: Could you point to the test in which there is this measurement dummy?

clusterloader2/api/validation/validation_test.go
```diff
@@ -260,8 +261,8 @@ func TestVerifyStep(t *testing.T) {
 		},
 		{
 			name: "has measurements and no phases",
-			input: Step{
-				Measurements: []*Measurement{
+			input: api.Step{
+				Measurements: []*api.Measurement{
 					{
 						Method: "test1",
```
rahulii, line 267: @marseel here is the dummy measurement!

marseel: You can replace "test1" with measurement that actually exists and it will work.

rahulii: @marseel , every measurement needs to be regsitered,right? so lets say here if i use 
```
Measurements: []*api.Measurement{
					{
						Method: "APIAvailability",
					},
				},
```

still it will say its not registered! As we did in our new test case first register , should we also do same in this case?

rahulii: /cc @marseel

PR #1854, opened 2021-07-06
Added developing measurement tutorial

marseel:
**What type of PR is this?**
/kind documentation

Tutorial about developing measurements in CL2.

Fixes #

/assign @wojtek-t 
/assign @mm4tt 

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/docs/DEVELOPING_MEASUREMENT.md
+++ b/clusterloader2/docs/DEVELOPING_MEASUREMENT.md
@@ -1,0 +1,563 @@
+# Developing measurement
+
+## Prerequisite
+It's strongly recommended to get familiar with [Getting started] tutorial.
+
+Also, you can check out our [Design] of Clusterloader2.
+
+## Introduction
+
+All measurements are implemented [here][Measurements].
+You can find current Measurement interface [here][Measurement interface].
+
+Measurement interface consists of three methods:
+```golang
+Execute(config *Config) ([]Summary, error)
+Dispose()
+String() string
+```
+`Execute` method will be executed each time you specify your measurement in test config.
+Usually measurement is specified two times in test config - once when we start measuring something and second time when we want to gather results.
+Two distinguish these two cases, you need to specify parameter called `action` that takes values `start` and `gather`.
+For example, for PodStartupLatency you specify action and some additional parameters when starting measurement:
+
+```
+- Identifier: PodStartupLatency
+  Method: PodStartupLatency
+  Params:
+    action: start
+    labelSelector: group = test-pod
+    threshold: 20s
+```
+And then later you can just gather results with:
+```
+- Identifier: PodStartupLatency
+  Method: PodStartupLatency
+  Params:
+    action: gather
+```
+
+`Dispose` method can be used to clean up after the measurement is no longer used.
+
+`String()` method should return name of measurement.
+
+## Implementing simple measurement
+
+Let's start with implementing simple measurement that will measure maximum number of running pods.
+First, let's start with specifying package and measurement name:
+
+```golang
+package common
+
+const (
+	maxRunningPodsMeasurementName = "MaxRunningPods"
+)
+```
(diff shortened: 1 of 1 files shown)


clusterloader2/docs/DEVELOPING_MEASUREMENT.md
```diff
@@ -0,0 +1,579 @@
...
+		return y
+	}
+	return x
+}
+
+func (m *maxRunningPodsMeasurement) checkPod(oldObj, newObj interface{}) {
+	isDeleted := newObj == nil
+	isNewObject := oldObj == nil
+
+	newPod, _ := newObj.(*corev1.Pod)
```
wojtek-t, line 150: silently ignoring an error is an anti-pattern :)

wojtek-t: BTW - you're not handling tombstones - e.g.:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go#L333

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
```diff
@@ -0,0 +1,579 @@
...
+	oldPod, _ := oldObj.(*corev1.Pod)
+
+	defer func() {
+		m.maxRunningPods = MaxInt(m.maxRunningPods, m.currentRunningPods)
+		klog.V(2).Infof("Max: %d, current: %d", m.maxRunningPods, m.currentRunningPods)
+	}()
+
+	if isDeleted {
```
wojtek-t, line 158: I think we can make it much simpler by doing:
```
func isPodRunning(obj interface{}) bool {
  pod := // convert interface to pod
  return pod != nil && pod.Status.Phase == corev1.PodRunning
}

change := 0
if isPodRunning(oldObj) {
  change--
}
if isPodRunning(newObj) {
  change++
}
```

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
```diff
@@ -0,0 +1,579 @@
...
+
+```golang
+func MaxInt(x, y int) int {
+	if x < y {
+		return y
+	}
+	return x
+}
+
+func (m *maxRunningPodsMeasurement) checkPod(oldObj, newObj interface{}) {
```
wojtek-t, line 146: Discuss why we don't need locking?

[or rather why we're ignoring it in all measurements...]

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
```diff
@@ -0,0 +1,579 @@
...
+  selector:
+    matchLabels:
+      k8s-app: node-local-dns
+  namespaceSelector:
+    matchNames:
+      - kube-system
+```
+Any monitor needs to be created within monitoring namespace.
+Most important is specifying interval how often metric will be scraped (here once every 10 minutes).
+The more pods you want to scrape the smaller interval should be and/or Prometheus server should have more resources.
```
wojtek-t, line 415: s/smaller/larger/

[or smaller frequency]

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
marseel: Good idea, done.

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
marseel: Done.

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
marseel: Added lock

clusterloader2/docs/DEVELOPING_MEASUREMENT.md
marseel: Handled error, but I've decided to skip handling tombstone and treat it as not running pod.

PR #1862, opened 2021-07-28
Extended test with option to run Exec command

dlapcevic:
Using override for Exec command, the test will only measure the service creation latency for large backend size after executing the command.

**What type of PR is this?**

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/testing/l4ilb/config.yaml
+++ b/clusterloader2/testing/l4ilb/config.yaml
@@ -12,6 +12,9 @@
 {{$ilbWaitTimeout := DefaultParam .CL2_ILB_WAIT_TIMEOUT "10m"}}
 {{$namespaces := 1}}
 
+# Command to be executed
+{{$EXEC_COMMAND := DefaultParam .CL2_EXEC_COMMAND nil}}
+
 #Test
 name: ilbload
 namespace:
@@ -23,6 +26,8 @@ tuningSets:
 steps:
 # Mesure each of the ILB services separately, this will provide insight on how long programming
 # ILB takes as a function of number of backends.
+
+{{if not $EXEC_COMMAND}}
 - name: Starting measurements
   measurements:
   - Identifier: ServiceCreationLatencyLarge
@@ -43,6 +48,9 @@ steps:
       action: start
       waitTimeout: {{$ilbWaitTimeout}}
       labelSelector: size = {{$SMALL_BACKEND_LABEL}}
+{{end}}
+- name: Starting measurement for deployment creation
+  measurements:
   - Identifier: WaitForRunningDeployments
     Method: WaitForControlledPodsRunning
     Params:
@@ -99,7 +107,8 @@ steps:
       objectTemplatePath: dep.yaml
       templateFillMap:
         NumReplicas: {{$SMALL_BACKEND_SIZE}}
-- name: Waiting for objects creation to be completed
+{{if not $EXEC_COMMAND}}
+- name: Measuring service creation latency
   measurements:
   - Identifier: ServiceCreationLatencyLarge
     Method: ServiceCreationLatency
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -12,6 +12,9 @@
 {{$ilbWaitTimeout := DefaultParam .CL2_ILB_WAIT_TIMEOUT "10m"}}
 {{$namespaces := 1}}
 
+# Command to be executed
```
jkaniuk, line 15: Quite a lot of behaviour is changed when setting this single flag, with very generic name.

Wouldn't it be better to have separate flags controlling initial ILB testing and node pool upgrade? Does it make sense to run both tests serially?

wojtek-t: Can we extract stuff into modules and instead of "if-ing" in a bunch of places create two separate tests from the common modules?

dlapcevic: I think it would make sense to run them both serially. This way the existing test would remain the same if Exec command override is not provided, and if it is, then both are run.
It would work when removing all other {{if}}s.

Does that seem reasonable?

marseel: Sounds good to me.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -12,6 +12,9 @@
...
+{{$EXEC_COMMAND := DefaultParam .CL2_EXEC_COMMAND nil}}
```
prameshj, line 16: Rest of the test assumes that this EXEC_COMMAND will be a nodepool upgrade command, right? Can we just make the name more expliciit too? "NODEPOOL_UPGRADE_COMMAND"?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -117,6 +120,31 @@ steps:
...
       action: gather
+{{if $EXEC_COMMAND}}
+- name: Starting measurements for nodepool upgrade
+  measurements:
+  - Identifier: ServiceCreationLatencyNodepoolUpgrade
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      waitTimeout: {{$ilbWaitTimeout}}
+      labelSelector: size = {{$LARGE_BACKEND_LABEL}}
```
prameshj, line 131: We only verify time taken for the large backend ILBs to recover?

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -117,6 +120,31 @@ steps:
     Method: WaitForControlledPodsRunning
     Params:
       action: gather
+{{if $EXEC_COMMAND}}
+- name: Starting measurements for nodepool upgrade
```
prameshj, line 124: This test currently uses service.yaml which tests the default TrafficPolicy (Cluster). It would be great to have the nodepool upgrade scenario verified for "Local" traffic policy too. The service.yaml traffic policy field can be parameterized like in https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/l4lb/service.yaml.

This can be a follow up.

clusterloader2/testing/l4ilb/config.yaml
dlapcevic: Yes, done.

clusterloader2/testing/l4ilb/config.yaml
dlapcevic: I added the measurements for medium and small too.

clusterloader2/testing/l4ilb/config.yaml
dlapcevic: Yes, I will make the changes in the next PR.
Thank you for the comments.

clusterloader2/testing/l4ilb/config.yaml
```diff
@@ -170,3 +218,19 @@ steps:
     Method: ServiceCreationLatency
     Params:
       action: gather
+{{if $NODEPOOL_UPGRADE_COMMAND}}
```
marseel, line 221: Correct me if I'm wrong, but looking at implementation of ServiceCreationLatency I see that previously created measurements (for example "ServiceCreationLatencyLarge") will be collecting data throughout whole test. "waitForReady" doesn't seem to stop collecting data from informer so it will also take into account events during node pool upgrade?

marseel: I did a little bit of research how ServiceCreationLatency works, and latency is always computed based on timestamp of service creation. Also, once measurement is started before node pool upgrade, service is reachable, so also reachability time might be incorrect too.

dlapcevic:
Closing this due to a latency measuring issue with ServiceCreationLatency measurement when it is used multiple times for the same label selector.

I’m going to do the test separately using modules as @wojtek-t initially suggested.

PR #1871, opened 2021-08-10
Add SIGINT handler to handle Prometheus snapshot

jprzychodzen:
**What type of PR is this?**

/kind bug

Currently, [kubetest sends SIGINT](https://github.com/kubernetes/test-infra/blob/master/kubetest/process/process.go#L202) in case of test timeout. This is handled by ClusterLoader2 by exiting immediately.

After this changes, SIGINT will be cached and ClusterLoader2 will try to snapshot disk image.

Test timeouts can indicate problems with unresponsive kube-apiserver, so there is no communication attempt to kube-apiserver. Potentially partial data will be lost, but it's a better situation that missing a whole snapshot.

/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/prometheus/experimental.go
+++ b/clusterloader2/pkg/prometheus/experimental.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
+	"os/signal"
 	"time"
 
 	"github.com/spf13/pflag"
@@ -37,7 +39,7 @@ type prometheusDiskMetadata struct {
 
 const (
 	gcloudRetryInterval  = 20 * time.Second
-	snapshotRetryTimeout = 10 * time.Minute
+	snapshotRetryTimeout = 5 * time.Minute
 	deleteRetryTimeout   = 2 * time.Minute
 )
 
@@ -107,6 +109,16 @@ func (pc *Controller) tryRetrievePrometheusDiskMetadata() (bool, error) {
 	pc.diskMetadata.name = pdName
 	pc.diskMetadata.zone = zone
 	return true, nil
+}
+func (pc *Controller) snapshotPrometheusDiskIfEnabledSynchronized() error {
+	pc.snapshotLock.Lock()
+	defer pc.snapshotLock.Unlock()
+	if pc.snapshotted {
+		return pc.snapshotError
+	}
+	pc.snapshotted = true
+	pc.snapshotError = pc.snapshotPrometheusDiskIfEnabled()
+	return pc.snapshotError
 }
 
 func (pc *Controller) snapshotPrometheusDiskIfEnabled() error {
@@ -198,4 +210,18 @@ func (pc *Controller) tryDeletePrometheusDisk(pdName, zone string) error {
 	}
 	klog.V(2).Infof("Deleting disk finished with: %q", string(output))
 	return nil
+}
+
+func (pc *Controller) EnableTearDownPrometheusStackOnInterrupt() {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		// Block until a signal is received.
+		s := <-c
+		klog.V(2).Infof("Received signal: %v", s)
+		if err := pc.snapshotPrometheusDiskIfEnabledSynchronized(); err != nil {
+			klog.Warningf("Error while snapshotting prometheus disk: %v", err)
+		}
+		os.Exit(1)
+	}()
 }
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -89,6 +90,12 @@ type Controller struct {
 	templateMapping map[string]interface{}
 	// diskMetadata store name and zone of Prometheus persistent disk.
 	diskMetadata prometheusDiskMetadata
+	// snapshotLock makes sure that only single Prometheus snapshot is happening
+	snapshotLock sync.Mutex
+	// snapshotted is a check if the Prometheus snapshot is already done - protected by snapshotLock
+	snapshotted bool
+	// snapshotError contains error from snapshot attempt - protected by snapshotLock
+	snapshotError error
 	// ssh executor to run commands in cluster nodes via ssh
 	ssh util.SSHExecutor
 }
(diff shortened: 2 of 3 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -256,6 +263,9 @@ func (pc *Controller) TearDownPrometheusStack() error {
 	if err := client.WaitForDeleteNamespace(k8sClient, namespace); err != nil {
 		return err
 	}
+	if err := pc.snapshotPrometheusDiskIfEnabledSynchronized(); err != nil {
```
wojtek-t, line 266: It seems we want to somewhat conflicting with line 250.

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -199,3 +211,21 @@ func (pc *Controller) tryDeletePrometheusDisk(pdName, zone string) error {
...
 }
+
+func (pc *Controller) EnableTearDownPrometheusStackOnInterrupt() {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for {
+			// Block until a signal is received.
+			s := <-c
+			fmt.Println("Got signal:", s)
```
wojtek-t, line 222: klog

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -199,3 +211,21 @@ func (pc *Controller) tryDeletePrometheusDisk(pdName, zone string) error {
 	klog.V(2).Infof("Deleting disk finished with: %q", string(output))
 	return nil
 }
+
+func (pc *Controller) EnableTearDownPrometheusStackOnInterrupt() {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		for {
```
wojtek-t, line 219: Why do we need for loop?

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -199,3 +211,21 @@ func (pc *Controller) tryDeletePrometheusDisk(pdName, zone string) error {
...
+			// Block until a signal is received.
+			s := <-c
+			fmt.Println("Got signal:", s)
+			go func() {
```
wojtek-t, line 223: Why do you do that in goroutine?

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -199,3 +211,21 @@ func (pc *Controller) tryDeletePrometheusDisk(pdName, zone string) error {
...
+				if err := pc.snapshotPrometheusDiskIfEnabledSynchronized(); err != nil {
+					klog.Warningf("Error while snapshotting prometheus disk: %v", err)
+				}
+				os.Exit(1)
```
wojtek-t, line 227: Why do we need Exit here?

clusterloader2/pkg/prometheus/experimental.go
jprzychodzen: I guess that by design I wanted to handle receiving signal as fast as possible, then delegating real work to additional thread. This way we are not blocked on time required to handle signal. Something like Top and Bottom Halves for Linux kernel.

Agree that in this case we do not really need that - after first SIGINT we will exit CL2.

Removing loop and additional goroutine.

clusterloader2/pkg/prometheus/experimental.go
jprzychodzen: Right, thanks.

clusterloader2/pkg/prometheus/experimental.go
jprzychodzen: See answer for 

> Why do you do that in goroutine?

clusterloader2/pkg/prometheus/experimental.go
```diff
@@ -37,7 +39,7 @@ type prometheusDiskMetadata struct {
 
 const (
 	gcloudRetryInterval  = 20 * time.Second
-	snapshotRetryTimeout = 10 * time.Minute
+	snapshotRetryTimeout = 5 * time.Minute
```
wojtek-t, line 42: Why?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -89,6 +90,12 @@ type Controller struct {
 	templateMapping map[string]interface{}
 	// diskMetadata store name and zone of Prometheus persistent disk.
 	diskMetadata prometheusDiskMetadata
+	// snapshotted is a check if the Prometheus snapshot is already done
+	snapshotted bool
+	// snapshotError contains error from snapshot attempt
+	snapshotError error
+	// snapshotLock makes sure that only single Prometheus snapshot is happening
+	snapshotLock sync.RWMutex
```
wojtek-t, line 98: s/RWMutex/Mutex/

Also - move mutex above the fields it is protecting

wojtek-t (commented):
/approve

Couple small comments

clusterloader2/pkg/prometheus/experimental.go
jprzychodzen: After kubetest sends sigint, there is 15 minutes before it will terminate everything. Creating a snapshot takes ~30 seconds. I guess that if we do not succeed in 5 minutes it's ok to move on to remaining phases - like gathering logs.

clusterloader2/pkg/prometheus/prometheus.go
jprzychodzen: Done.

clusterloader2/pkg/prometheus/experimental.go
wojtek-t: Isn't CL2 killed at this point already?

This change is unrelated to this PR - so let's not mix it here.

jprzychodzen: This change is related.

When this PR will enter, the 15 minutes timeout after sigint will be shared between snapshot and logs. The more time we have for a snapshot, the less time we have for getting logs.

When kubetest will send sigint to current process, it will have 15 minutes timeout before sending sigkill and blocking additional processes. See https://github.com/kubernetes/test-infra/blob/master/kubetest/process/process.go#L200 for logic on Control instance.

The same global control object: https://github.com/kubernetes/test-infra/blob/c6e29bb385e0a93113937f08b61fbe1b9de5ee5c/kubetest/main.go#L52
Wrapper for executing test: https://github.com/kubernetes/test-infra/blob/c6e29bb385e0a93113937f08b61fbe1b9de5ee5c/kubetest/e2e.go#L224
Wrapper for getting logs: https://github.com/kubernetes/test-infra/blob/c6e29bb385e0a93113937f08b61fbe1b9de5ee5c/kubetest/gke.go#L478

mborsz:
Wojtek is OOO so I reviewed this.

LGTM, but please wait till Monday with submitting.

/lgtm
/hold

Issue #1872, opened 2021-08-16
Delete namespaces more rapidly when they contain lots of events


JohnRusk:
**Suggested addition**

When deleting namespaces at the end of the test, ClusterLoader2 should first delete all the events in the namespace.  If the code deletes many events in parallel, it will make cleanup much quicker.    I'm happy to create a PR, but have some questions first (below).

**Why**

I've noticed that clusterLoader2 is taking many minutes to clean up at the end of my tests.  E.g. at least 5 to 10 minutes to delete the namespace that was used in the test.  This is because my test generates lots of events.  It appears that a simple call to Kubernetes to delete a namespace is not very quick when the namespace contains lots of events.  However, I tested code which explicitly deletes the events first (processing a few hundred events in parallel).  This runs an order of magnitude more quickly.

**Questions**

I'm happy to submit a PR, but I have a design question first: how should the degree of parallelism be controlled?  Options that come to mind include

1. Just hard code something.  E.g. 100 concurrent deletion operations. (I.e. 100 goroutines, each deleting one event at a time).
2. Use a `tuningSet`.  This seems like the most idomatic clusterLoader2 approach... but... I can't think of a nice way to link a tuningSet to the automatic deletion of namespaces.
3. Use some other parameter, e.g. `eventCleanupRate` in the CL2 `namespaces` object.  That's simpler, but seems awfully specific to the details of this particular problem.
4. Use a command line parameter, with a default value.  This may in fact be the simplest option.  It won't clutter the CL2 test definition API, and a sensible default can be provided.

Does anyone have any thoughts on which might be the best way?  If we can settle on an approach, I'll submit a PR.

marseel:
Thanks cool idea, I like it. 

Out of these 4 options I like 4th most. Let's make it in a way that if someone does not specify it we won't explicitly delete events. 
So I would have default 0, which would mean that this feature is turned off and if someone wants to enable it they can. 
This will also allow us to safely deploy it (as we have multiple tests with different sizes of masters, so the same rate won't be good for all tests).
I've also checked that deleting namespaces takes around ~50s in our scalability tests for 5k nodes so in our case it might not be worth changing it as tests take around 5h to complete. 

Also, I am happy to review your changes.

JohnRusk:
@marseel  That's a great idea about making the default be zero.

I have created the PR.  Looks like I don't have rights to assign you as a reviewer. Can you assign yourself? Or just go ahead and review it?

marseel:
[...]

PR #1874, opened 2021-08-19
Fast delete events

JohnRusk:
**What type of PR is this?**

/kind feature

CL2 tests may cause a lot of events to be created in K8s. Then, at the end of the test, when CL2 cleans up the namespace, it can take quite a few minutes to delete the namespace due to the large number of events.

This PR adds optional fast deletion of events, by deleting events in parallel just prior to namespace deletion.  This considerably reduces the cleanup time at the end of tests (e.g. from 5mins to 30 seconds in one test).

Fixes https://github.com/kubernetes/perf-tests/issues/1872

```release-note
ClusterLoader2 can now clean up namespaces faster if you specify the command line parameter `--parallel-event-deletions`.  Try setting to it to 100 for smaller clusters and 200 for larger.
```

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -85,6 +85,7 @@ func initClusterFlags() {
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterInternalIPs, "master-internal-ip", "MASTER_INTERNAL_IP", nil /*defaultValue*/, "Cluster internal/private IP of the master vm, supports multiple values when separated by commas")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.APIServerPprofByClientEnabled, "apiserver-pprof-by-client-enabled", "APISERVER_PPROF_BY_CLIENT_ENABLED", true, "Whether apiserver pprof endpoint can be accessed by Kubernetes client.")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.ParallelEventDeletions, framework.ParallelEventDeletionsName, "PARALLEL_EVENT_DELETIONS", 0, "Number of events to delete in parallel before deleting namespace. For faster test cleanup. Zero means just let namespace deletion implicitly clean up events - which is slower for large event counts")
 
 	flags.StringEnvVar(&providerInitOptions.ProviderName, "provider", "PROVIDER", "", "Cluster provider name")
 	flags.StringSliceEnvVar(&providerInitOptions.ProviderConfigs, "provider-configs", "PROVIDER_CONFIGS", nil, "Cluster provider configurations")
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -48,6 +48,9 @@ type ClusterConfig struct {
 	DeleteStaleNamespaces bool
 	// TODO(#1696): Clean up after removing automanagedNamespaces
 	DeleteAutomanagedNamespaces bool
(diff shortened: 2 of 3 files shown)


JohnRusk:
I signed it

JohnRusk:
When using the new code today, I realised there was an issue with rate limiting, so I've just pushed a fix into this PR.

I'm not _totally_ happy with it, because it required me to re-implement the 8 lines or so of the event deletion function, but it does solve the problem and make a very big improvement to the speed.

If anyone knows of a better way to set custom rate limits in CL2, just for selected requests, please let me know.

clusterloader2/cmd/clusterloader.go
```diff
@@ -85,6 +85,7 @@ func initClusterFlags() {
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterIPs, "masterip", "MASTER_IP", nil /*defaultValue*/, "Hostname/IP of the master node, supports multiple values when separated by commas")
 	flags.StringSliceEnvVar(&clusterLoaderConfig.ClusterConfig.MasterInternalIPs, "master-internal-ip", "MASTER_INTERNAL_IP", nil /*defaultValue*/, "Cluster internal/private IP of the master vm, supports multiple values when separated by commas")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.APIServerPprofByClientEnabled, "apiserver-pprof-by-client-enabled", "APISERVER_PPROF_BY_CLIENT_ENABLED", true, "Whether apiserver pprof endpoint can be accessed by Kubernetes client.")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.ParallelEventDeletions, framework.ParallelEventDeletionsName, "PARALLEL_EVENT_DELETIONS", 0, "Number of events to delete in parallel before deleting namespace. For faster test cleanup. Zero means just let namespace deletion implicitly clean up events - which is slower for large event counts")
```
wojtek-t, line 88: I'm against this change.

If you want to speed up events deletion, you can already do that. it's a knob that is exposed from kube-apiserver. It's here:
https://github.com/kubernetes/kubernetes/blob/2c1c1b75465e93ecc49d0cd59efdfa1b0bdb3bbc/staging/src/k8s.io/apiserver/pkg/server/options/etcd.go#L127

This does exactly what you want at the lower level.

wojtek-t (commented):
/hold

clusterloader2/cmd/clusterloader.go
JohnRusk: @wojtek-t In the past few days, I've worked with two users who could not use the lower-level feature.  Here at Microsoft, that will be common.  E.g. if customers want to use ClusterLoader2 to benchmark an AKS cluster, they cannot set the kube-apiserver flag, because we don't expose it to them. The same will also be true of other cloud providers who provide management or control of API server config - they might not expose that flag to customers.

Obviously, I _could_ change the default value of the API server flag for _all_ AKS, but I'm reluctant to make and test that change, and apply it to _everyone_, just to benefit the smaller number who may want to benchmark their clusters with CL2.

So what do you think about re-evaluating this PR, on that basis?  If it helps, I can modify the prompt message in this PR, to point out that the kube-apiserver flag is the preferred solution, and users should only use the CL2 flag if they don't have access to the api-server flag.

wojtek-t: OK - I buy that.

Let's make change the help message for the flag as well as pring some Warning in the logs if the value is set. 

[I didn't look into the code itself - so the above is purely on the conceptual level.]

JohnRusk: On second thoughts, we've noticed that some auditing software doesn't particularly like deletion of events.  (The action gets classified as suspicious behaviour - like to how modification of logs would be similarly classified.).  So on that basis I'm going to abandon this PR.

To others who may read this in the future: if you want faster deletion of namespaces that have lots of stuff in them, your current options are 
(a) manually increase delete-collection-workers as wojtek-t describes above (if you have that level of access to your A
PI server)
(b) persuade your API server administrator to do the change for you, OR
(c) submit a PR for the K8s API server itself to increase the default (I'm not sure whether you'll succeed with that - presumably there is a reason why the current default is 1)

PR #1876, opened 2021-08-21
Add nodepool upgrade test

dlapcevic:
**What type of PR is this?**
/kind feature

Adds a test for nodepool upgrade latency for L4 ILB.

Fixes #

```release-note

```

```docs

```


tosi3k:
/ok-to-test
/lgtm

CCing Pavithra for having a final say on this.
/cc @prameshj 

Assigning Jacek for approval.
/assign @jkaniuk

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
```diff
@@ -0,0 +1,124 @@
...
+  measurements:
+  - Identifier: ServiceCreationLatency
+    Method: ServiceCreationLatency
+    Params:
+      action: start
+      waitTimeout: {{$ilbWaitTimeout}}
+      labelSelector: size = {{$NODEPOOL_UPGRADE_LABEL}}
+- name: Starting ILB recovery timer
+  measurements:
+  - Identifier: ILbRecoveryTimer
```
prameshj, line 72: nit: ILBRecoveryTimer?

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
```diff
@@ -0,0 +1,124 @@
...
+      max: {{$namespaces}}
+    replicasPerNamespace: {{$LB_REPLICAS_PER_NS}}
+    tuningSet: ILBConstantQPS
+    objectBundle:
+    - basename: ilb-npu-service
+      objectTemplatePath: service.yaml
+      templateFillMap:
+        DeploymentBaseName: ilb-npu-dep
+        ExternalTrafficPolicy: {{$EXTERNAL_TRAFFIC_POLICY}}
+        ILBSizeLabel: {{$NODEPOOL_UPGRADE_LABEL}}
```
prameshj, line 50: Why is this label needed? Can't we use "group=ilb-load"?

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
```diff
@@ -0,0 +1,124 @@
...
+  - Identifier: ExecNodepoolUpgrade
+    Method: Exec
+    Params:
+      command:
+      {{range $NODEPOOL_UPGRADE_COMMAND}}
+      - {{.}}
+      {{end}}
+- name: Waiting for ILBs to be reachable after nodepool upgrade
+  measurements:
+  - Identifier: ServiceCreationLatency
```
prameshj, line 88: This is good for the first iteration, but ideally we want to measure 1) latency from outside the cluster (on a different GCE VM) 2) A count of how many requests succeeded and how many failed. 

Usually we don't see complete blackholing, we see intermittent failures. So a failure count will give a better idea than the latency (which might be ~0 if the first request succeeds).

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
dlapcevic: The service.yaml template inside the l4ilb directory uses a size label. We add a value for the size label in order to reuse the template for multiple tests which are using measurements to select services by label.

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
jprzychodzen: Regarding different GCE VM - Can we measure from a CL2 container? CL2 is running in another project and this is something similar how we define SLOs. Externally visible and measured from the client's perspective.

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
dlapcevic: Done

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
```diff
@@ -0,0 +1,124 @@
...
+{{$LB_REPLICAS_PER_NS := DefaultParam .CL2_LB_REPLICAS_PER_NS 5}}
+{{$LB_BACKEND_SIZE := DefaultParam .CL2_LB_BACKEND_SIZE 300}}
+{{$EXTERNAL_TRAFFIC_POLICY := DefaultParam .CL2_EXTERNAL_TRAFFIC_POLICY "Cluster"}}
+{{$ilbWaitTimeout := DefaultParam .CL2_ILB_WAIT_TIMEOUT "10m"}}
+{{$ilbQPS := DefaultParam .CL2_ILB_TEST_QPS 20}}
+{{$NODEPOOL_UPGRADE_LABEL := "ilb-np-upgrade"}}
+{{$namespaces := 1}}
+
+# Command to be executed
+{{$NODEPOOL_UPGRADE_COMMAND := DefaultParam .CL2_EXEC_COMMAND nil}}
```
wojtek-t, line 17: I wouldn't call this parameter (and also the test itself) as nodepool-upgrade. This test is more generic and can be run as both node-pool upgrade, but also any other arbitrary operation.

dlapcevic: Good point. Thank you.
I've made the changes and I will run another test to make sure it still works.

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
prameshj: What is a CL2 container? Is it a pod running in a different cluster in a different project?

For internal LB, clients need to be part of the same VPC. If we can use shared VPC to achieve that, then it could work.

clusterloader2/testing/l4ilb/config-ilb-recovery.yaml
```diff
@@ -0,0 +1,124 @@
+# Timer measurement is used to measure the time between the start of an action executed by the
+# Exec command and when the ILB services are running and reachable.
+# ServiceCreationLatency measurement is used only for determining the moment when all of the
+# ILBs are reachable. Since it reports time based on the initial creation of the service, and
+# not from the start of the action, the measured time will not be shown.
```
prameshj, line 5: What does this mean? We will have a timeout of 10m(by default) for ILBs to be ready after starting the action, but we will not export the total time taken?

dlapcevic: The Timer measurement is used here for getting the time that is required for services to be reachable after an action.
The ServiceCreationLatency measurement always measures and reports time based on the initial creation time of the service, and not when the action is triggered. Since recreating the nodes will only reschedule the services, the ServiceCreationLatency would use the initial creation time of the ILBs as the start time, regardless of when we start the measurement.
That’s why the time measured by it is not useful for this test, but we can use it anyway only for getting the information when the services are all reachable again.

clusterloader2/testing/l4ilb/config-nodepool-upgrade.yaml
jprzychodzen: CL2 - container running ClusterLoader2. It lives in different project, and adding this peering would be problematic.

clusterloader2/testing/l4ilb/config-ilb-recovery.yaml
dlapcevic: Your statements are correct.
If we only want to measure the time from the moment when nodepool upgrade ends, we can just move the Timer to start after the EXEC command is finished.
We can also add another Timer to measure multiple periods.

What do you suggest?

prameshj: Ideally, we want the timer to measure the time for which ILB services were unreachable/flaky, right from when nodepool upgrade command got kicked off. Your change is already doing that. I think the issue is - even if there was no disruption, the test result will show the time interval as non-zero (time taken for nodepool upgrade to finish). So it won't really be useful.

A more useful measure is - count the number of requests to ILB VIP that resulted in failure. (non-200 status code). The timer value is fine to measure the nodepool upgrade duration. The failed_count will give us an idea of disruption and measure any performance improvement we make.

dlapcevic:
The additional argument for the Exec command can be used to specify the `--cluster-version` for example, so that we can execute a nodepool downgrade.

clusterloader2/testing/l4ilb/config-ilb-recovery.yaml
```diff
@@ -0,0 +1,134 @@
...
+  measurements:
+  - Identifier: ExecCommand
+    Method: Exec
+    Params:
+      timeout: {{$EXEC_TIMEOUT}}
+      command:
+      {{range $EXEC_COMMAND}}
+      - {{.}}
+      {{end}}
+      {{if $EXEC_ADD_ARGUMENT_ONE}}
```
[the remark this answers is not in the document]
jupblb: After 1:1 discussion: CL2 doesn't currently have support for this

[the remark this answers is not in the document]
dlapcevic: - EXEC_COMMAND is included from an override file that contains some fixed values and some that are read from the environment variables.
In my opinion this is the simplest and clearest way to parametrize values that are used in the EXEC command, such as cluster-version to which the node pool should upgrade/downgrade to.
It requires no additional work, but allows us to add specific arguments to some standard commands, so that we can run various tests exactly as required.

- After discussion, we came to a conclusion that making additional arguments into a list is not currently supported. We can do it in the future when it is supported.

clusterloader2/testing/l4ilb/config-ilb-recovery.yaml
dlapcevic: To measure success rate of requests to ILB during the nodepool upgrade requires a new measurement.
This will be next on the list for required measurements for load balancers. It will be added to this test once it is implemented.

clusterloader2/testing/l4ilb/config-ilb-recovery.yaml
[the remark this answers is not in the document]
dlapcevic: Thank you.
This allows us to use just one of the EXEC_ADD_ARGUMENTs to add the cluster version, instead of two.

Anyway, I think it’s good to leave both of the arguments until we can make it into a list, since it will let us be more flexible with adding custom values through CL2 parameters. Maybe not for this specific case, but potentially with the same or a different command.
What do you think?

PR #1890, opened 2021-09-15
Set www/ permissions in perfdash.

mborsz:
Without explicitly setting permissions for www/ static dir, `docker build` output depends on the chmod on the build workstation. This can lead to setting incorrect file permissions and rendering static file index instead of perfdash.

Let's explicitly request 0644 permissions.

/assign @mm4tt


mm4tt:
/lgtm
/approve

Thanks!

PR #1910, opened 2021-10-08
Handle removed nodes in image preloading

jkaniuk:
Handle removed nodes in image preloading.

Fix problems when doing preloading in a very large or automatically resized cluster.

/cc @mborsz

### DIFF
--- a/clusterloader2/pkg/imagepreload/imagepreload.go
+++ b/clusterloader2/pkg/imagepreload/imagepreload.go
@@ -94,7 +94,7 @@ func (c *controller) PreloadImages() error {
 
 	kclient := c.framework.GetClientSets().GetClient()
 
-	doneNodes := make(map[string]bool)
+	doneNodes := make(map[string]struct{})
 	stopCh := make(chan struct{})
 	defer close(stopCh)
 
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -156,19 +157,37 @@ func (c *controller) PreloadImages() error {
...
-func (c *controller) checkNode(obj interface{}, done map[string]bool) {
-	if obj == nil {
+func (c *controller) checkNode(set map[string]struct{}, old, new interface{}) {
+	if new != nil {
+		node := new.(*v1.Node)
+		preloaded := c.hasPreloadedImages(node)
+		c.markDone(set, node.Name, preloaded)
 		return
 	}
+	if old != nil {
```
mborsz, line 167: Not sure if I follow this line: I understand that this should catch case when node has been deleted (as you remove it from the map).

Why is it a case?
Don't we simply get checkNode([old version of node], [new version of node]) called for all the updates?

Please correct or add a comment why it is correct :)

mborsz: I missed "return" above. I think it's correct then, as this if is correct when new == nil && old != nil, which is a case for deletion.

PR #1912, opened 2021-10-13
Improve handling of automanaged namespaces

### LINKED ISSUE #1848: Preserving automanaged namespaces doesn't work properly for > 1 test config
When `clusterloader2` is run with more than one CL2 test config and we disable the deletion of automanaged namespaces (i.e. add `--delete-automanaged-namespaces=false` argument to CL2 invocation), all tests, apart from the first one, fail with the following error:

```
error while preparing test namespaces: automanaged namespaces creation failed: automanaged namespaces already created
```

We could either fix this and create a separate set of namespaces for different tests or make the tests reuse the namespaces created in the first test.

/cc @marseel 
/good-first-issue
/help-wanted

jprzychodzen:
Allow multi-config scenario where single automanaged namespace is reused between tests

**What type of PR is this?**

/kind bug

Currenly, leaving automanaged namespaces between configs breaks non-first config. This PR handles this use case by adding an explicit tracker of all created automanaged namespaces and by fixing verification in appropriate places.

Fixes #1848


clusterloader2/pkg/framework/framework.go
```diff
@@ -102,11 +104,9 @@ func (f *Framework) GetClusterConfig() *config.ClusterConfig {
...
 
 // CreateAutomanagedNamespaces creates automanged namespaces.
-func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
-	if f.automanagedNamespaceCount != 0 {
-		return fmt.Errorf("automanaged namespaces already created")
-	}
-
+func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int, allowExistingNamespaces bool, deleteAutomanagedNamespaces bool) error {
+	f.mux.Lock()
+	defer f.mux.Unlock()
```
marseel, line 109: Do we need this mutex?

marseel: nvm, we need it because `DeleteNamespaces ` is deleting concurrently.

PR #1914, opened 2021-10-15
Use the same image in daemonset and pod startup latency

mborsz:
Alternative approach for image preloading. Comparing to imagepreloading CL2 feature it has the following advantages:
* It uses no additional test time (in very large clusters 15 minutes isn't enough for creation, deletion blocks even more, see https://github.com/kubernetes/perf-tests/pull/1894).
* It avoid copying pod startup latency flag to CL2 flags which live in different repos:
    * with old imagepreload e.g. changing image used for pod startup latency will require changing
    * This increases our prowjob config complexity as not all tests are actually using pod startup latency, so we need to enable this feature on per-prowjob basis (instead of e.g. some preset)
 * It uses well known, proved to work code path, while we found already multiple bug in imagepreloading code (insufficient timeout, incorrect node changes handling) and we are aware of few more (failing when some tainted node already had that image, failing entire test instead of continuing and marking the step failure)

At this point, investing more in custom imagepreload doesn't seem to be reasonable.
If this merges, in the next step I will remove NODE_PRELOAD_IMAGES from all the presets and prow job definitions.

/assign @jkaniuk


clusterloader2/testing/load/config.yaml
```diff
@@ -317,7 +327,8 @@ steps:
       namespaces: {{$namespaces}}
       tuningSet: RandomizedDeletionTimeLimited
       testMaxReplicaFactor: {{$RANDOM_SCALE_FACTOR}}
-      daemonSetImage: k8s.gcr.io/pause:3.1
+      daemonSetImage: {{$latencyPodImage}}
+      daemonSetEnv: "after update"
```
wojtek-t, line 331: Why do we need to set it here?

mborsz: Personally I'm not fan of not setting it here as this requires custom logic in yaml parsing (https://github.com/kubernetes/perf-tests/blob/a81c576d431247523807fd85caa2ce273db5c7f7/clusterloader2/pkg/config/template_provider.go#L90), but to be consistent with the code near, I removed this.

PR #1925, opened 2021-10-29
Update k8s dependencies to 1.20.12

mborsz:
Three steps:
* Update go.mod to 1.20.12 (we cannot use 1.21+ as they require golang 1.16 and kubekins-e2e for 1.20 is using golang 1.15, so we will break support for testing 1.20)
* Update vendor
* Update code (mostly ssh client changes, some signature changes)


mborsz:
Wojtek is OOO.

/assign @marseel

PR #1928, opened 2021-10-29
CL2: Correct the port number for the kube-apiserver

### LINKED ISSUE #1927: CL2: kube-apiserver has wrong port in master-endpoints.yaml
**What happened**:
I tried enabling the prometheus server in clusterloader2, and it hung on startup with the following complaint.

```
I1029 15:25:40.765850  294166 util.go:101] 5/6 targets are ready, example not ready target: {map[endpoint:apiserver instance:1.2.3.4:443 job:master namespace:monitoring service:master] down}
```

It looks like that port number is coming from https://github.com/kubernetes/perf-tests/blob/d4c896470c6246838edde817dd41af04a181bd73/clusterloader2/pkg/prometheus/manifests/master-ip/master-endpoints.yaml#L19 .  On my control node, as usual, the kube-apiserver is listening on port 6443, not 443.  Port 443 works with the `kubernetes` Service's ClusterIP.

```
mspreitz@ubu20:~/go/src/k8s.io/perf-tests/clusterloader2$ netstat -lnt | grep 443
tcp6       0      0 :::6443                 :::*                    LISTEN     
```

**What you expected to happen**:
I expected the bringup of the prometheus stack to not hang.

**How to reproduce it (as minimally and precisely as possible)**:
I am describing as simple and direct a scenario as I can see.

**Anything else we need to know?**:
I created my kube cluster using k/k `hack/local-up-cluster.sh`.

**Environment**:
- Kubernetes version (use `kubectl version`): client and server on `master` branch, commit 89dd275dde0d19112f826699944f40c625c33e8e
- Cloud provider or hardware configuration: VirtualBox VM
- OS (e.g: `cat /etc/os-release`): Ubuntu 20.04.1 LTS
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others: my branch `really-skip-kubelet` (commit e512d8ba23fbd992300a33e1883f68958a254558) of perf-tests

MikeSpreitzer:
**What type of PR is this?**

/kind bug

This PR changes the port number for the apiserver in the master endpoints list.  This is because the apiserver normally listens on port 6443, not 443.  I assume the old setting was working because CL2 was being used in special circumstances where the apiserver was also available on port 443.  But CL2 should be broadly usable.

Fixes #1927

```release-note
clusterloader2 no longer assumes that the kube-apiserver is also available on port 443 on the control node's IP.
```

```docs

```

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/master-ip/master-endpoints.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/master-ip/master-endpoints.yaml
@@ -16,7 +16,7 @@ subsets:
     {{end}}
     ports:
       - name: apiserver
-        port: 443
+        port: 6443
       {{if not $PROMETHEUS_SCRAPE_APISERVER_ONLY}}
       - name: etcd-2379
         port: 2379


MikeSpreitzer:
It looks like pull-perf-tests-clusterloader2 and pull-perf-tests-clusterloader2-kubemark failed because the kube-apiserver was _not_ listening on 6443.  So I suppose the port will have to be made a parameter.

MikeSpreitzer:
See #1930 instead.

PR #1929, opened 2021-10-29
Install pdcsi driver in gke scalability test

leiyiz:
- updated pdcsi installation manifest
- install a default storage class when needed

**What type of PR is this?**

/kind bug

Fixes #

linked to kubernetes/kubernetes#104722
/assign @mborsz
/assign @mattcary

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/drivers/gcp-csi-driver-stable.yaml
+++ b/clusterloader2/drivers/gcp-csi-driver-stable.yaml
@@ -1,258 +1,677 @@
 # This config generated from the GCP PD CSI Driver
 # https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver
-# with the command: kustomize build deploy/kubernetes/overlays/stable
-# and an addition storage class from examples/kubernetes/zonal-sc-example.yaml
+# with the command: kustomize build deploy/kubernetes/overlays/stable-master
+# and an additional storage class from examples/kubernetes/zonal-sc-example.yaml
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: gce-pd-csi-driver
+  labels:
+    name: gce-pd-csi-driver
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  labels:
-    app: gcp-compute-persistent-disk-csi-driver
-  name: csi-controller-sa
-  namespace: default
+  name: csi-gce-pd-controller-sa
+  namespace: gce-pd-csi-driver
 ---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
+  name: csi-gce-pd-node-sa
+  namespace: gce-pd-csi-driver
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: csi-gce-pd-node-sa-win
+  namespace: gce-pd-csi-driver
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: csi-gce-pd-controller-psp
+spec:
+  fsGroup:
+    rule: RunAsAny
+  hostNetwork: true
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - emptyDir
+  - secret
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: csi-gce-pd-node-psp
+spec:
+  allowedHostPaths:
+  - pathPrefix: /var/lib/kubelet/plugins_registry/
+  - pathPrefix: /var/lib/kubelet
+  - pathPrefix: /var/lib/kubelet/plugins/pd.csi.storage.gke.io/
+  - pathPrefix: /dev
+  - pathPrefix: /etc/udev
+  - pathPrefix: /lib/udev
+  - pathPrefix: /run/udev
+  - pathPrefix: /sys
+  fsGroup:
+    rule: RunAsAny
+  hostNetwork: true
+  privileged: true
+  runAsUser:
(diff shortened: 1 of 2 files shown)


clusterloader2/drivers/gcp-csi-driver-stable.yaml
```diff
@@ -1,258 +1,670 @@
 # This config generated from the GCP PD CSI Driver
 # https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver
-# with the command: kustomize build deploy/kubernetes/overlays/stable
+# with the command: kustomize build deploy/kubernetes/overlays/stable-master
 # and an addition storage class from examples/kubernetes/zonal-sc-example.yaml
```
mattcary, line 4: "additional"

mattcary (commented):
Have you tested this on a migration-enabled cluster to make sure there's not another use of the default storage class hidden in there?

clusterloader2/drivers/gcp-csi-driver-stable.yaml
leiyiz: done

clusterloader2/run-e2e.sh
```diff
@@ -23,14 +23,14 @@ export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
 export KUBEMARK_ROOT_KUBECONFIG="${KUBEMARK_ROOT_KUBECONFIG:-${HOME}/.kube/config}"
 
 # Deploy the GCP PD CSI Driver if required
-if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
+if [[ "${DEPLOY_GCI_DRIVER:-true}" == "true" ]]; then
```
mborsz, line 26: I don't think we can change default here. This script is used also outside of gcp.

I think we should keep default and set this flag in our tests specifically.

mborsz:
Please do not update vendor/. It's simply a copy of some external repo. All changes should go to those repos.

In fact, I don't think we need any changes to vendor, do we?

leiyiz:
> Please do not update vendor/. It's simply a copy of some external repo. All changes should go to those repos.
> 
> In fact, I don't think we need any changes to vendor, do we?

so the runner.sh I think needs to be changed because the config in the `prometheus/manifests/0ssd-storage-class.yaml ` will not get reflected. maybe some bug in this repo can fix it but for now I am just trying to get pdcsi install correctly. Converted to draft for now.

mborsz:
> > Please do not update vendor/. It's simply a copy of some external repo. All changes should go to those repos.
> > In fact, I don't think we need any changes to vendor, do we?
> 
> so the runner.sh I think needs to be changed because the config in the `prometheus/manifests/0ssd-storage-class.yaml ` will not get reflected. maybe some bug in this repo can fix it but for now I am just trying to get pdcsi install correctly. Converted to draft for now.
> 

I don't think you need to update vendor to pickup changes in prometheus/manifests/0ssd-storage-class.yaml. It's used by prometheus-operator in this repo and AFAIK changing value in yaml should just work.

leiyiz:
> > > Please do not update vendor/. It's simply a copy of some external repo. All changes should go to those repos.
> > > In fact, I don't think we need any changes to vendor, do we?
> > 
> > 
> > so the runner.sh I think needs to be changed because the config in the `prometheus/manifests/0ssd-storage-class.yaml ` will not get reflected. maybe some bug in this repo can fix it but for now I am just trying to get pdcsi install correctly. Converted to draft for now.
> 
> I don't think you need to update vendor to pickup changes in prometheus/manifests/0ssd-storage-class.yaml. It's used by prometheus-operator in this repo and AFAIK changing value in yaml should just work.

Yes, I got vender/ confused with prometheus.go, indeed vendor/ doesn't needs to be changed. But just changing the yaml did not get picked up.

clusterloader2/run-e2e.sh
leiyiz: make sense, can you give me a pointer on where to change this flag? it's nowhere to be found in `perf-tests` repo

leiyiz:
test passed [here](https://gke-scalability-prow.corp.goog/view/gs/gke-scalability-prow/logs/ci-kubernetes-e2e-gci-gce-scalability-leiyiz/1456423465841070080) against kubernetes/kubernetes#104722

clusterloader2/run-e2e.sh
mborsz: You can set any env variable for a given test in test-infra repo.

Here I recommend changing this in preset-e2e-scalability-periodics-master preset which will automatically affect all periodic jobs using master branch:

https://github.com/kubernetes/test-infra/blob/2cb4ec605eed0c0db0c624a2990779fff43114b4/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L314

Then you should add this variable to other presets as described in https://github.com/kubernetes/test-infra/pull/19710#issue-730566534

The potential problem is that once we add this to preset-e2e-scalability-periodics (as described there), this will affect also release presubmits in file: 

https://github.com/kubernetes/test-infra/blob/c3b32daef3348869e9c4621b875c8de07120622b/config/jobs/kubernetes/sig-release/release-branch-jobs/1.22.yaml#L370

which are also using this presubmit. Setting this env in that file will try to use perf-tests/release-1.22 version of this code DEPLOY_GCI_DRIVER which isn't intended.

I recommend manually adding --env= DEPLOY_GCI_DRIVER=false in all release-branch-jobs/1.X.yaml that are using preset-e2e-scalability-periodics to override this preset value.

I know it's quite complicated, we can figure out details during test-infra PR reviews.

mborsz:
I think it's OK.

Please revert the DEPLOY_GCI_DRIVER default change and LGTM

leiyiz:
@mborsz

PR #1934, opened 2021-11-03
Remove clusterloader2/vendor

mborsz:
clusterloader is using go modules for quite some time and we should be able to live without vendoring all dependencies.

Other repos (like test-infra: https://github.com/kubernetes/test-infra/pull/10760) did this ~2 years ago.

According to https://maelvls.dev/go111module-everywhere/ it should be safe for go 1.13+ which works in go module mode anytime go.mod file is present.

The expected impact is breaking go up to 1.12 if perf-tests is built from GOPATH. This can be mitigated by setting GO111MODULE=on env variable.

In OSS and GKE we are using golang 1.15+ which is used by k8s 1.19+.

https://github.com/kubernetes/perf-tests/pull/1925 already requires golang 1.15+ so affected golang versions are already disallowed (golang 1.13+ was required for quite a long time).

/assign @wojtek-t
/cc @shyamjvs Please let me know if this may break some infra on your end.

/hold Let's submit this during WAW working hours.


wojtek-t:
/lgtm
/approve

To give time people to react - let's merge on Monday (lazy consensus timeout on Monday morning).
/hold

mborsz:
It's Monday morning
/hold cancel

Issue #1931, opened 2021-11-01
CL2: LegacyIsMasterNode does not consider the configured "master" node name


MikeSpreitzer:
**What happened**:
I tried applying clusterloader2 to a cluster where the master node is named "127.0.0.1".  I passed `--mastername 127.0.0.1` on the command line, and that solved most of the problems, but the node-exporter support still complained "Error while setting up prometheus stack: node-exporter requires master to be registered nodes".  I tracked this down to the logic in https://github.com/kubernetes/perf-tests/blob/d4c896470c6246838edde817dd41af04a181bd73/clusterloader2/pkg/util/cluster.go#L185-L210, which has a fixed expectation of what a control node's name looks like.

FYI, following is the CL2 command line I used.

```
mspreitz@ubu20:~/go/src/k8s.io/perf-tests/clusterloader2$ go run cmd/clusterloader.go --testconfig=config.yaml --provider=local --kubeconfig=$KUBECONFIG --v=3 --enable-prometheus-server --prometheus-storage-class-provisioner=standard --mastername 127.0.0.1 --master-internal-ip 1.2.3.4 --prometheus-scrape-kubelets=false --experimental-prometheus-snapshot-to-report-dir=true --tear-down-prometheus-server=false --prometheus-apiserver-scrape-port=6443 --prometheus-scrape-node-exporter=true
```

**What you expected to happen**:
I expected the node-exporter support to be able to find my cluster's control plane node based on the name I configured on the command line.

**How to reproduce it (as minimally and precisely as possible)**:
I think my remarks here constitute a pretty minimal scenario.

**Anything else we need to know?**:
I created my kube cluster with k/k's `hack/local-up-cluster.sh`.

**Environment**:
- Kubernetes version (use `kubectl version`): master
- Cloud provider or hardware configuration: VirtualBox VM
- OS (e.g: `cat /etc/os-release`): Ubuntu 20.04
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #1937, opened 2021-11-15
Fix the port number to collect the profiling data for the apiserver

chendave:
`443` is the tcp port for the cluster ip `10.96.0.1`, which is used by
service `kubernetes`.

Connect the port with the localhost and fetch the profiling data is not allowed,

> curl -k https://localhost:443/debug/pprof/profile?seconds=30 -o cpu2.pprof

>  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 443: Connection refused

Because of this, profiling file for apiserver is always empty,

> file 10.253.2.51_kube-apiserver_CPUProfile_load_2021-08-26T11:10:26Z.pprof
10.253.2.51_kube-apiserver_CPUProfile_load_2021-08-26T11:10:26Z.pprof: empty

We should either update the url to use clusterip:443 or just update the port
to use the port exposed by the pod directly.

Signed-off-by: Dave Chen <dave.chen@arm.com>

**What type of PR is this?**

/kind bug

Fixes #

Actually,  I am not pretty sure is this a bug or just something I missed, the original code has been last for one year, why no one report the issue?

 
```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -71,6 +71,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdKeyPath, "etcd-key", "ETCD_KEY", "/etc/srv/kubernetes/pki/etcd-apiserver-server.key", "Path to the etcd key on the master machine")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "DEPRECATED: Whether to delete all stale namespaces before the test execution.")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.KubeApiserverPort, "kube-apiserver-port", "KUBE_APISERVER_PORT", 443, "port of apiserver used to collect profiling data")
 	err := flags.MarkDeprecated("delete-stale-namespaces", "specify deleteStaleNamespaces in testconfig file instead.")
 	if err != nil {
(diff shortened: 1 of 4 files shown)


MikeSpreitzer:
My experience with #1928 -> #1930 might be relevant here.

chendave:
@MikeSpreitzer thanks Mike!

> My experience with #1928 -> #1930 might be relevant here.

I took a glance at that one, and I believe #1931 is relevant but we still fix this one separately. As they are belonging to different code path and for different purpose.

I was trying to collect `pprof` data for cpu and memory, and the port is hard coded here, 
https://github.com/kubernetes/perf-tests/blob/ef1ad81804b668626478375c794b9fa08031fc32/clusterloader2/pkg/provider/util.go#L25-L37

and is called by, 

https://github.com/kubernetes/perf-tests/blob/ef1ad81804b668626478375c794b9fa08031fc32/clusterloader2/pkg/measurement/common/profile.go#L228-L244


this pr is the easiest way to fix it, also as I said in the commit message, I am not sure why the code stands here for such a long time without any complaint.


@wojtek-t @mm4tt  could you pls take a look?  I not sure who are the most active reviewers at the moment, sorry if I ping the wrong person.

MikeSpreitzer:
@chendave : My point is that I could not find one port number that is always the right one to use for reaching the kube-apiserver.

MikeSpreitzer:
@chendave : what I am trying to suggest is that you make that port number a parameter.
/remove-lifecycle stale

chendave:
@MikeSpreitzer Got your point, will update ASAP, thanks!

chendave:
I think all the ports there should be better to configurable, this pr only focus on apiserver port since only the apiserver is an issue in my env. 

https://github.com/kubernetes/perf-tests/blob/30041333d995b8d3008bfaeeea594dfaae157be4/clusterloader2/pkg/provider/util.go#L25-L37

chendave:
@MikeSpreitzer WDYT?

PR #1939, opened 2021-11-15
Fix CSI driver for kubemark

mborsz:
/kind bug

See https://github.com/kubernetes/test-infra/pull/24357#issuecomment-969060521

/cc @leiyiz

/assign @tosi3k @wojtek-t

### DIFF
--- a/clusterloader2/run-e2e.sh
+++ b/clusterloader2/run-e2e.sh
@@ -20,6 +20,8 @@ set -o pipefail
 
 CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE[0]}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
+# If we are in kubemark (i.e. KUBEMARK_ROOT_KUBECONFIG is set), install CSI driver in this cluster.
+export CSI_DRIVER_KUBECONFIG=${KUBEMARK_ROOT_KUBECONFIG:-${KUBECONFIG}}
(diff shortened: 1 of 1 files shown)


clusterloader2/run-e2e.sh
```diff
@@ -20,6 +20,8 @@ set -o pipefail
 
 CLUSTERLOADER_ROOT=$(dirname "${BASH_SOURCE[0]}")
 export KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
+# If we are in kubemark (i.e. KUBEMARK_ROOT_KUBECONFIG is set), install CSI driver in this cluster.
+export CSI_DRIVER_KUBECONFIG=${KUBEMARK_ROOT_KUBECONFIG:-${KUBECONFIG}}
```
wojtek-t, line 24: This won't really work, because this has to be a real pod - and in kubemark cluster you only will run that pod on a hollow-pod (aka - it won't be running).

We need to do something similar to what we did with heapster:
https://github.com/kubernetes/kubernetes/blob/master/test/kubemark/resources/heapster_template.json#L47

I.e. running it in a base cluster - and somehow connect it to the kubemark cluster

mborsz: We solved this problem for prometheus already: it runs in the "root" cluster (CL2 reads KUBEMARK_ROOT_KUBECONFIG and runs there).

It should work IMHO if we set env variables correctly.

PR #1941, opened 2021-11-15
fix kubemark test install pdcsi driver

leiyiz:
cleanup for https://github.com/kubernetes/perf-tests/pull/1940 and installation is successful in test run of that pr.

/assign @mborsz


mborsz:
/lgtm
/approve

Based on the test result in the linked PR

PR #1948, opened 2021-11-24
Add endTime to Gatherer interface

Argh4k:
**What type of PR is this?**
/kind feature

This PR adds endTime to Gatherer interface. It is necessary for me to introduce new executor to be used in unit tests in next pull request. This new executor will use prometheus query engine with lazy loader backend, allowing users to load time series from file. Thanks to that we will be able to write unit tests for measurements which will run queries against executor that can parse and execute queries in the same manner as prometheus does. Therefore, we will be able to detect issues with our queries in measurement. Currently in unit tests, we use fakeExecutors which always return the same results for test case even after the query in measurement is changed.

Fixes #

no

```release-note
NONE
```

```docs

```

### DIFF
--- a/clusterloader2/pkg/measurement/common/metrics_server_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/metrics_server_prometheus.go
@@ -46,8 +46,8 @@ func init() {
 
 type metricsServerGatherer struct{}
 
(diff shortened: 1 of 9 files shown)


Argh4k:
cla bump

PR #1952, opened 2021-11-30
Extend log entry for in progress WaitForControlledPodsRunning

jprzychodzen:
Add information why sample Pod is not considered as updated

/kind feature

/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 
 	goerrors "github.com/go-errors/errors"
+	gocmp "github.com/google/go-cmp/cmp"
 	appsv1 "k8s.io/api/apps/v1"
 	batch "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -163,7 +164,7 @@ func getSelectorFromUnstrutured(obj *unstructured.Unstructured) (labels.Selector
 
 // GetIsPodUpdatedPredicateFromRuntimeObject returns a func(*corev1.Pod) bool predicate
 // that can be used to check if given pod represents the desired state of pod.
-func GetIsPodUpdatedPredicateFromRuntimeObject(obj runtime.Object) (func(*corev1.Pod) bool, error) {
+func GetIsPodUpdatedPredicateFromRuntimeObject(obj runtime.Object) (func(*corev1.Pod) error, error) {
 	switch typed := obj.(type) {
 	case *unstructured.Unstructured:
 		return getIsPodUpdatedPodPredicateFromUnstructured(typed)
@@ -172,7 +173,7 @@ func GetIsPodUpdatedPredicateFromRuntimeObject(obj runtime.Object) (func(*corev1
 	}
 }
 
-func getIsPodUpdatedPodPredicateFromUnstructured(obj *unstructured.Unstructured) (func(_ *corev1.Pod) bool, error) {
+func getIsPodUpdatedPodPredicateFromUnstructured(obj *unstructured.Unstructured) (func(_ *corev1.Pod) error, error) {
 	templateMap, ok, err := unstructured.NestedMap(obj.UnstructuredContent(), "spec", "template")
 	if err != nil {
 		return nil, goerrors.Errorf("failed to get pod template: %v", err)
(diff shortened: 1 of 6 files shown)


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -185,8 +186,11 @@ func getIsPodUpdatedPodPredicateFromUnstructured(obj *unstructured.Unstructured)
 		return nil, goerrors.Errorf("failed to parse spec.teemplate as v1.PodTemplateSpec")
 	}
 
-	return func(pod *corev1.Pod) bool {
-		return equality.Semantic.DeepDerivative(template.Spec, pod.Spec)
+	return func(pod *corev1.Pod) (bool, error) {
```
mborsz, line 189: Can we return only error where nil means that are equal?

Otherwise, how caller should handle cases when e.g. (true, non-nil) or (false, nil)?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
```diff
@@ -426,13 +427,111 @@ func changeEnv(in *v1.Pod) *v1.Pod {
...
+                    VolumeMounts: nil,
+                    ... // 12 identical fields
+            },
+    },
+    EphemeralContainers:           nil,
+    RestartPolicy:                 "",
+-   TerminationGracePeriodSeconds: nil,
++   TerminationGracePeriodSeconds: &30,
+-   ActiveDeadlineSeconds:         nil,
++   ActiveDeadlineSeconds:         &30,
```
mborsz, line 464: Is there a way to ignore fields not set in the first object?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
```diff
@@ -426,13 +427,111 @@ func changeEnv(in *v1.Pod) *v1.Pod {
...
+func noWhitespace(s string) string {
+	s = strings.ReplaceAll(s, " ", "")
+	s = strings.ReplaceAll(s, "\n", "")
+	s = strings.ReplaceAll(s, "\r", "")
+	s = strings.ReplaceAll(s, " ", "") //non-breaking space
+	return strings.ReplaceAll(s, "\t", "")
+}
+
+const fullCompareErrorDifferentEnv = `Not matching templates, diff:   v1.PodSpec{
+    Volumes:        nil,
```
mborsz, line 439: Is there a way to make it shorter, e.g. by skipping lines that do match?

clusterloader2/pkg/measurement/util/pods_test.go
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+package util
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestConvertPodsStartupStatusToString(t *testing.T) {
```
mborsz, line 24: While +1 to adding some tests, It's not clear if the string format is worth testing. It looks more like "change-detector" to me, rather than valuable test.

What kind of regression can we prevent with this?

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
jprzychodzen: Not an easy one. We could vendor this library and do changes in files like https://github.com/google/go-cmp/blob/d103655696d8ae43c4125ee61454dbf03d8e8324/cmp/report.go and https://github.com/google/go-cmp/blob/d103655696d8ae43c4125ee61454dbf03d8e8324/cmp/report_compare.go , however I guess that this would be a little too much work to achieve nicer log. Or we could write our own Reporter, but this is also a significant amount of work (and seeing from documentation not so well supported).

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
jprzychodzen: Not really - we are limited by public function API, and it does not allow us to hack around defaultReporter. We could write our own Reporter, but it seems that it's not well documented and reserved for cmp.Equal. From [cmp.Reporter](https://pkg.go.dev/github.com/google/go-cmp/cmp#Reporter)

```
Reporter is an Option that can be passed to Equal. When Equal traverses the value trees, it calls PushStep as it descends into each node in the tree and PopStep as it ascend out of the node. The leaves of the tree are either compared (determined to be equal or not equal) or ignored and reported as such by calling the Report method.
```

clusterloader2/pkg/measurement/util/pods_test.go
jprzychodzen: I wanted to confirm that providing `<nil>` error does not break formating.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
jprzychodzen: That's a good point. And this will simplify a code in few places. Let me implement this change.

jprzychodzen: Done.

clusterloader2/pkg/measurement/util/pods_test.go
jprzychodzen: Actually, removed this part. It was too noisy - even for successful 100 node run from presubmit hundred of additional log lines were added.

Moved log to only show single error in case of timeout.

PR #1955, opened 2021-12-06
Add implementaion of Pushgateway feature

jprzychodzen:
With '--enable-pushgateway' ClusterLoader2 will setup Pushgateway
instance and setup Prometheus to scrape metrics from that instance.

/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/pushgateway/deployment.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/pushgateway/deployment.yaml
@@ -1,0 +1,33 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: monitoring
+  name: pushgateway
+  labels:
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/prometheus/manifests/pushgateway/deployment.yaml
```diff
@@ -0,0 +1,28 @@
...
+  selector:
+    matchLabels:
+      k8s-app: pushgateway
+  template:
+    metadata:
+      name: pushgateway
+      labels:
+        k8s-app: pushgateway
+        visualize: "true"
+        run: pushgateway
```
marseel, line 21: Do we need these two labels "visualize" and "run"?

jprzychodzen: Not really, checked without those labels - it works.

PR #1956, opened 2021-12-10
Add separate huge service test config

jupblb:
**What type of PR is this?**
/kind feature

We want to decouple scheduler throughput from huge service test. In order to do so, a separate test needs to be implemented. Since we do not test for huge services everywhere, it is done in a separate file.

The default number of endpoints for the new huge service test is 1000 instead of the number of nodes.

There are two modules extracted from the existing `testing/load` configuration:
- `testing/load/modules/measurements.yaml` -> `testing/huge-service/modules/measurements.yaml`
- `testing/load/modules/huge-services.yaml` + `testing/load/modules/scheduler-throughput.yaml` -> `testing/huge-service/modules/service.yaml`

The config has been tested and passes our CI.

N/A

/assign @mborsz
/cc @marseel 

```release-note
A new specialized config has been introduced for the purpose of testing a service with high amount of endpoints. 
```

```docs
NONE
```

### DIFF
--- a/clusterloader2/testing/huge-service/config.yaml
+++ b/clusterloader2/testing/huge-service/config.yaml
@@ -1,0 +1,26 @@
+# Huge service test config
+{{$HUGE_SERVICE_HEADLESS := DefaultParam .CL2_HUGE_SERVICE_HEADLESS false}}
+{{$HUGE_SERVICE_ENDPOINTS := DefaultParam .CL2_HUGE_SERVICE_ENDPOINTS 1000}}
+
+name: huge-service
+namespace:
+  number: 1
+tuningSets:
+- name: Sequence
+  parallelismLimitedLoad:
+    parallelismLimit: 1
+steps:
+- module:
+    path: modules/measurements.yaml
+    params:
+      action: start
+- module:
+    path: modules/service.yaml
+    params:
+      endpoints: {{$HUGE_SERVICE_ENDPOINTS}}
+      isHeadless: {{$HUGE_SERVICE_HEADLESS}}
+      serviceName: huge-service
+- module:
+    path: modules/measurements.yaml
+    params:
+      action: gather
--- a/clusterloader2/testing/huge-service/modules/service.yaml
+++ b/clusterloader2/testing/huge-service/modules/service.yaml
@@ -1,0 +1,83 @@
+{{$endpoints := .endpoints}}
+{{$isHeadless := .isHeadless}}
+{{$serviceName := .serviceName}}
+
+steps:
+- name: Create {{$serviceName}}
+  phases:
+  - namespaceRange:
+      min: 1
+      max: 1
+    replicasPerNamespace: 1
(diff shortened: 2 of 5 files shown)


clusterloader2/testing/huge-service/config.yaml
```diff
@@ -0,0 +1,33 @@
+# Huge service test config
+{{$HUGE_SERVICE_HEADLESS := DefaultParam .CL2_HUGE_SERVICE_HEADLESS false}}
+{{$HUGE_SERVICE_ENDPOINTS := DefaultParam .CL2_HUGE_SERVICE_ENDPOINTS 1000}}
+
+name: huge-service
+namespace:
+  number: 1
+tuningSets:
+- name: Sequence
```
mborsz, line 9: We have three different tuningSets to remove single object :)

Please use single one everywhere (e.g. Sequence).

clusterloader2/testing/huge-service/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+- module:
+    path: modules/measurements.yaml
+    params:
+      action: start
+- module:
+    path: modules/service.yaml
+    params:
+      endpoints: {{$HUGE_SERVICE_ENDPOINTS}}
+      isHeadless: {{$HUGE_SERVICE_HEADLESS}}
+      namespaces: 1
```
mborsz, line 28: Do we need to make namespaces configurable? I'm afraid setting value significantly larger than 1 may trigger some timeouts on pod deletion.

clusterloader2/testing/huge-service/modules/service.yaml
```diff
@@ -0,0 +1,84 @@
+{{$endpoints := .endpoints}}
+{{$isHeadless := .isHeadless}}
+{{$namespaces := .namespaces}}
+{{$serviceName := .serviceName}}
+
+steps:
+- name: Create huge services
```
mborsz, line 7: Please make "huge services" a variable.

This way you will be able to reuse this module:
- name "huge servieces"
   headless: false
- name: "headless servvice"
  headless: true

mborsz (commented):
Overall, looks good. Few nits

clusterloader2/testing/huge-service/config.yaml
jupblb: Fixed.

clusterloader2/testing/huge-service/config.yaml
jupblb: I had the same question in my head when writing this and decided to retain similarity with scheduler-throughput. But I guess if you don't see the value in having it then it can be safely removed. Fixed.

clusterloader2/testing/huge-service/modules/service.yaml
jupblb: Going to reuse the `$serviceName` var. Fixed.

PR #1957, opened 2021-12-14
Add generic Prometheus query measurement

aleksandra-malinowska:
Basic version of a measurement allowing user to pass a custom Prometheus query via config, with the following limitations:
- only latency metrics are supported,
- the query must be parametrized by test duration,
- no checks/assertions available at this point.

Unit tests and documentation update pending.

### DIFF
--- a/clusterloader2/examples/generic_query_example.yaml
+++ b/clusterloader2/examples/generic_query_example.yaml
@@ -1,0 +1,34 @@
+# Simple example of the use of generic query measurement.
+#
+# Notes on parameters:
+#       metricName will be logged in place of "GenericPrometheusQuery"
+#       rawQuery must be parametrized by duration
+
+{{$duration := "240s"}}
+{{$namespaces := 1}}
+
+name: generic-query
+namespace:
+  number: {{$namespaces}}
+steps:
+- name: Start measurements
+  measurements:
+  - Identifier: gq
+    Method: GenericPrometheusQuery
+    Params:
+      action: start
+      metricName: API request latency
+      metricVersion: v1
+      rawQuery:  quantile_over_time(0.99, apiserver:apiserver_request_latency_1m:histogram_quantile{verb!="WATCH", subresource!~"log|exec|portforward|attach|proxy"}[%v])
+- name: Sleep
+  measurements:
+  - Identifier: sleep
+    Method: Sleep
+    Params:
+      duration: {{$duration}}
+- name: Gather measurements
+  measurements:
+  - Identifier: gq
+    Method: GenericPrometheusQuery
+    Params:
+      action: gather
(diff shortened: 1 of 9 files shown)


clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -0,0 +1,106 @@
...
+func (g *genericQueryGatherer) query(executor QueryExecutor, startTime, endTime time.Time) (*measurementutil.LatencyMetric, error) {
+	duration := endTime.Sub(startTime)
+	boundedQuery := fmt.Sprintf(g.rawQuery, measurementutil.ToPrometheusTime(duration))
+	klog.V(2).Infof("bounded query: %s, duration: %v", boundedQuery, duration)
+	samples, err := executor.Query(boundedQuery, endTime)
+	if err != nil {
+		return nil, err
+	}
+	// For now, only latency is supported.
+	return measurementutil.NewLatencyMetricPrometheus(samples)
```
jprzychodzen, line 94: I'm not sure if we want to focus summary on a specific query type - it's not needed right now.

Maybe instead of assuming query type, you could provide some aggregate of the results (min,max,avg,count)? That should be enough to see some basic data about the metric.

jprzychodzen (commented):
Thanks, that looks great.

Added a single comment about the summary.

mborsz (commented):
Overall LGTM.

I think we should write some unit test for this. In my opinion some test using FakeQueryExecutor (and validating if the duration is correctly calculated) + checking if post processing works well (e.g. extracting quantiles) should be sufficient.

clusterloader2/examples/generic_query_example.yaml
```diff
@@ -0,0 +1,35 @@
+# Simple example of the use of generic query measurement.
+#
+# Notes on parameters:
+#       metricName will be logged in place of "GenericPrometheusQuery"
+#       rawQuery must be parametrized by duration
+
+{{$duration := "240s"}}
+{{$latencyPodImage := DefaultParam .CL2_LATENCY_POD_IMAGE "k8s.gcr.io/pause:3.1"}}
```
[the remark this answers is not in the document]
aleksandra-malinowska: Done

aleksandra-malinowska:
Regarding unit tests - I'd rather add them in a separate PR, if that's OK.

Regarding latency metric - I believe it might require changes in perfdash (which relies on the specific data format generated by NewLatencyMetricPrometheus). Is this a blocker?

clusterloader2/pkg/measurement/common/generic_query_measurement.go
mborsz: I don't think we should any aggregations inside of CL2, rather to prepare a query that does e.g. <aggregation>_over_time and put it here: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md#definition

I agree that we should remove latency metric and allow any other metric (already discussed with Alex some potential solutions), but realistically speaking it will not happen today, so I don't want to block this PR. We can improve it in the next PR.

mborsz:
I would prefer to have tests in this PR, but given time constraints I think it's fine if this will happen as follow up PR.

PR #1966, opened 2022-01-12
Add measurement for cilium endpoint propagation delay

dlapcevic:
The Prometheus metric [EndpointPropagationDelay](https://github.com/cilium/cilium/blob/master/pkg/metrics/metrics.go#L200) is added to the Cilium 1.11 version. The CEP (Cilium Endpoint) propagation delay is useful for pods that are affected by network policies. It shows latency for pods from creation time, to the time when they are able to send and receive traffic.

The measurement gathers the `cilium_endpoint_propagation_delay` metric data by querying Prometheus, and then verifies if it is satisfying a defined SLO.

/kind feature
/assign @tosi3k


clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
```diff
@@ -0,0 +1,160 @@
...
+	percentileSLO, err := util.GetFloat64OrDefault(config.Params, "percentileSLO", defaultPercentileSLO)
+	if err != nil || percentileSLO == 0 {
+		klog.V(2).Infof("Using defaultPercentileSLO: %f, because percentileSLO param is invalid: %v", percentileSLO, err)
+		percentileSLO = defaultPercentileSLO
+	}
+
+	for _, buckets := range result {
+		totalEvents := buckets[bucketAllEntries]
+		acceptedDelayEvents := buckets[bucketSLO]
+		perc := (float64(acceptedDelayEvents) / float64(totalEvents)) * 100
```
tosi3k, line 134: What happens if `totalEvents` is equal to `0`? Wouldn't `perc` evaluate to `NaN` then?

Should we add some `if` statement above to prevent situations like this?

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
```diff
@@ -0,0 +1,160 @@
...
+		if perc < percentileSLO {
+			return errors.NewMetricViolationError(
```
tosi3k, line 136: IIUC, we fail if and only if we fail the SLO at any of the 10 minute-separated timestamps. If my understanding is correct, I have two QQs:
* Shall we provide the timestamp at which we fail the SLO in the error message as well?
* Do we want to report all time frames at which we fail the SLO?

clusterloader2/testing/load/modules/measurements.yaml
```diff
@@ -16,6 +16,9 @@
 {{$ENABLE_SYSTEM_POD_METRICS:= DefaultParam .ENABLE_SYSTEM_POD_METRICS true}}
 {{$ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS := DefaultParam .CL2_ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS false}}
 {{$ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS_SIMPLE := DefaultParam .CL2_ENABLE_VIOLATIONS_FOR_API_CALL_PROMETHEUS_SIMPLE true}}
+{{$ENABLE_CEP_PROPAGATION_DELAY_MEASUREMENT := DefaultParam .CL2_ENABLE_CEP_PROPAGATION_DELAY_MEASUREMENT false}}
+{{$CEP_PROPAGATION_DELAY_SLO_BUCKET := DefaultParam .CL2_CEP_PROPAGATION_DELAY_SLO_BUCKET ""}}
```
tosi3k, line 20: Nitpick: shall we add the default SLO bucket here as well, like we do for the SLO percentile line below?

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
```diff
@@ -0,0 +1,160 @@
...
+
+	percentileSLO, err := util.GetFloat64OrDefault(config.Params, "percentileSLO", defaultPercentileSLO)
+	if err != nil || percentileSLO == 0 {
+		klog.V(2).Infof("Using defaultPercentileSLO: %f, because percentileSLO param is invalid: %v", percentileSLO, err)
+		percentileSLO = defaultPercentileSLO
+	}
+
+	for _, buckets := range result {
+		totalEvents := buckets[bucketAllEntries]
+		acceptedDelayEvents := buckets[bucketSLO]
```
tosi3k, line 133: This is slightly brittle - if we provide a `bucketSLO` argument in the measurement's params that isn't defined by the metric, then this will always evaluate to `0` and the measurement will always fail.

OTOH, I believe there is no good way of statically ensuring that we provide a correct `bucketSLO` value. Could we at least point to the definition of the metric in a comment above the definition of `cepPropagationDelayQuery` variable?

tosi3k (changes requested):

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
```diff
@@ -0,0 +1,160 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
dlapcevic: Done.

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
```diff
@@ -0,0 +1,160 @@
...
+func (c *cepPropagationDelayGatherer) gatherCepPropagationDelay(executor QueryExecutor, endTime time.Time) (cepPropagationDelayMetricMap, error) {
+	samples, err := executor.Query(cepPropagationDelayQuery, endTime)
+	if err != nil {
+		return nil, err
+	}
+
+	// Query the data on fixed intervals to get accurate data
+	// from multiple snapshots. Stop when the query returns empty.
+	queryTime := endTime
+	for {
```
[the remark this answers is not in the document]
dlapcevic: You are right. I made the change to query between start and end time.

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
dlapcevic: I agree on both points. I added the link to the file where the metric and its buckets are defined.

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
dlapcevic: - I added the timestamp. It’s better to show the timestamp as well, but I think it’s not very important.
- I think it’s not necessary. Histogram accumulates the buckets over time, and usually if it fails at some point it will continue to fail in the subsequent snapshots as well.
The main purpose of this measurement is to notify us if the test doesn’t pass the SLO. Later we can use the test result to view all data of the snapshots.

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
tosi3k: Done.

tosi3k: Thanks. Done.

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
tosi3k: Done.

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
tosi3k: Sure, this makes perfect sense. Thanks.

clusterloader2/testing/load/modules/measurements.yaml
tosi3k: Oh, I didn't realize that... Would it be possible then to treat this parameter as a floating-point number (i.e. `{{$CEP_PROPAGATION_DELAY_SLO_BUCKET := DefaultParam .CL2_CEP_PROPAGATION_DELAY_SLO_BUCKET 600}}`) instead and parse it to string inside the measurement's logic using `FormatFloat` method like [here](https://go.dev/play/p/HBXtSUSDCy2)?

clusterloader2/pkg/measurement/common/cilium_endpoint_propagation_delay.go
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/testing/load/modules/measurements.yaml
dlapcevic: Yes, thank you for the suggestion.
Done. Also changed it in the unit test for custom SLOs to have float value for this parameter.

PR #1973, opened 2022-01-21
Add support for integers in Exec measurement

jprzychodzen:
Using an integer as a value in Exec measurement command throws a type assertion error.

Let's parse values as strings, as expected.

GetStringArray is only used in Exec measurement, so I feel confident with changing behavior of this function.


marseel:
/lgtm
/approve
/hold
Let's wait for Monday with merge.

PR #1974, opened 2022-01-24
Allow additional slow calls in huge-service test

jupblb:
**What type of PR is this?**
/kind bug
/kind regression

After enabling the new huge-service test configuration, there were multiple failures related to slow pod lists per namespace. Investigation showed that there are on average 1-2 slow requests each time the cluster is drastically scaled up or down.

In the load test this was not visible as the test for a larger number of nodes lasts long enough for the p99 to get below 5s threshold. The huge-service test lasts at most a few minutes which is not enough for the same thing to happen.

Therefore a new parameter is introduced for the huge-service test that allows to ignore those few regularly appearing slow calls.

N/A

This change is currently being tested.

```release-note
N/A
```

```docs
N/A
```


mborsz:
IIUC, this will start ignoring (verb, resource) pairs with <= 300 count, isn't?

What is an impact on set of (verb, resource) that will be checked?

Issue #799, opened 2019-09-16
Load test: Test run-to-completion workflow for Job objects


mm4tt:
As a part of https://github.com/kubernetes/perf-tests/issues/704 the load test was extended to cover Jobs.

The way it was implemented was kind of a shortcut, we used "pause" pods that never complete and used Jobs in the same way we use/test Deployments, i.e. we create N Jobs of size X then scale them up or down and then delete the jobs. While this was a good start to test overall performance of job-controller, we should actually use the jobs in a way they are designed to be used, i.e. don't use the pause pods, but pods that end after some time and test the run-to-completion workflow.

marseel:
@mm4tt @wojtek-t So I am actually wondering about one thing. Let's assume we change jobs to "sleep X". 
Currently we have two phases: creating job and waiting for pods to be running and second phase where we scale number of pods within job. 
After changing to "sleep X" first phase will complete job, so second part of scaling up/down job will not work (completions is also immutable). Do you have any idea how we want to handle that other than just creating job and then waiting for completion?

I have two ideas what we could do here:
1) Set parallelism == completions
2) Set parallelism < completions (for example parallelism = 1/4 of completions)
In both cases I guess it would make sense to make sleep a bit random, so we don't have many pods churning and/or created at the same time. 

What. do you think about it?

wojtek-t:
We can get rid of scaling phase for jobs at this point. Just run it to completion, ideally with parallelism < completions.

alculquicondor:
Let me know if this plan makes sense:
1. Add a `WaitForJobsCompleted` measurement method
2. In the load test:
    - Switch the image from `pause` into something that just finishes. Is there an image I can use? Simplest thing is a busybox with command `exit 0`, but let me know if I should add a new image.
    - Remove the `WaitForRunningJobs` identifier from `WaitForControlledPodsRunning` 
    - Add `WaitForJobsCompleted`
    - Make the scale a noop for jobs (keep the same parallelism, although at this point the job would already be finished)

cc @jprzychodzen

wojtek-t:
I'm actually reluctant to changing that in the current load test. We make some assumptions about number of pods etc.
I think we should have a new (more batch-oriented) test for it.
We may decide to merge them eventually, but I don't think we're there yet.

@mborsz - FYI

jprzychodzen:
I agree with splitting Jobs to another test. Currently we are strongly depending on number of Pods and changing that during the test can generate unstable results.

[...]

PR #1976, opened 2022-01-25
Add WaitForFinishedJobs measurement

alculquicondor:
**What type of PR is this?**

/kind feature

Adds a WaitForCompletedJobs measurement and a placeholder test.

Part of #799

kubernetes/enhancements#2214 and kubernetes/enhancements#2879

```release-note
Add measurement WaitForCompletedJobs that waits until Jobs complete or fail
```

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_jobs.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_jobs.go
@@ -1,0 +1,268 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+
+	"k8s.io/perf-tests/clusterloader2/pkg/framework"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/runtimeobjects"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/workerqueue"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	defaultWaitForFinishedJobsTimeout = 10 * time.Minute
+	waitForFinishedJobsName           = "WaitForFinishedJobs"
+	waitForFinishedJobsWorkers        = 1
+	checkFinishedJobsInterval         = 5 * time.Second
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/measurement/common/wait_for_jobs.go
```diff
@@ -0,0 +1,247 @@
...
+	for _, cond := range w.finishedJobs {
+		if cond == batchv1.JobComplete {
+			completed++
+		} else if cond == batchv1.JobFailed {
+			failed++
+		}
+	}
+	w.lock.Unlock()
+	klog.V(2).Infof("%v: %d/%d Jobs finished, %d completed, %d failed", w, completed+failed, len(jobKeys), completed, failed)
+	return nil
```
marseel, line 179: Do we want to return error if some of jobs failed? 
I am wondering how exactly we want to use this measurement later. 
@jprzychodzen @alculquicondor

alculquicondor: What are the implications of error? Does the framework continue with the next steps? I noticed that `WaitForControlledPodsRunning` doesn't trigger an error if all pods are not running (unless I misread).

We weren't planning purposely failing jobs for now, but we could consider in the future.

For the tests we are planning, a failed job should be unexpected. But if there was an error, it would be helpful to complete other measurements, such as gathering metrics.

marseel: Yes, it continues with next steps even if this measurement fails. Here is example when `WaitForControlledPodsRunning` returns error when some of deployments time out https://github.com/kubernetes/perf-tests/blob/c50dea13450a0c3a6aeee677124da8a832665c88/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go#L270

alculquicondor:
Sorry, I should have added a WIP notice. Thank you for your quick reviews.

The PR is now ready. I have renamed `completed` as `finished` to better distinguish from the `JobComplete` condition. This should allow us to distinguish `JobFailed` from `JobComplete` in the future, if we decide to do so.

clusterloader2/pkg/measurement/common/wait_for_jobs.go
alculquicondor: Ah, missed it. For now, it makes sense to error when some jobs didn't finish (regardless of their success)

clusterloader2/pkg/measurement/common/wait_for_jobs.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	completed, condition := finishedJobCondition(newJob)
+
+	w.lock.Lock()
+	if completed {
+		w.finishedJobs[key] = condition
+	} else {
+		delete(w.finishedJobs, key)
+	}
+	w.lock.Unlock()
```
marseel, line 228: Can you move it up and do `defer w.lock.Unlock()`?

clusterloader2/pkg/measurement/common/wait_for_jobs.go
```diff
@@ -0,0 +1,266 @@
...
+	for key := range jobKeys {
+		if cond, ok := w.finishedJobs[key]; !ok {
+			timedOut.Insert(key)
+		} else if cond == batchv1.JobComplete {
+			completed++
+		} else if cond == batchv1.JobFailed {
+			failed++
+		}
+	}
+	w.lock.Unlock()
```
marseel, line 184: Let's add `defer` here too and move it up.

clusterloader2/pkg/measurement/common/wait_for_jobs.go
```diff
@@ -0,0 +1,266 @@
...
+
+type waitForFinishedJobsMeasurement struct {
+	selector *measurementutil.ObjectSelector
+
+	queue            workerqueue.Interface
+	isRunning        bool
+	clusterFramework *framework.Framework
+	cancel           context.CancelFunc
+	lock             sync.Mutex
+	finishedJobs     map[string]batchv1.JobConditionType
```
marseel, line 74: Can you add comment that this lock is used for accessing finishedJobs map?

marseel (commented):
Only small nit comments, except for that lgtm.

clusterloader2/testing/batch/job.yaml
```diff
@@ -0,0 +1,19 @@
...
+  parallelism: {{.Replicas}}
+  completions: {{.Replicas}}
+  template:
+    metadata:
+      labels:
+        group: test-pod
+    spec:
+      containers:
+      - name: {{.Name}}
+        image: bash
```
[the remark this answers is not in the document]
alculquicondor: Yes, I plan to change this to a proper load test. But I still would like to have this merged as it serves as proof that the measurement works.

I will request to upload an image soon. I was told I need to verify the license. I might create a simple go code that just exits, otherwise.

I don't think we need sleep for now, although that could be an argument for the command.

alculquicondor (commented):
squashed

PR #1978, opened 2022-01-27
Add multiple queries and checks to generic query measurement

aleksandra-malinowska:
Follow-up to #1957. Changes:
- metric is no longer treated as latency,
- multiple queries can be defined for the same measurement instance,
- user can set threshold - all samples are expected to be lesser or equal to it.

@jprzychodzen - can you PTAL and see if the basic metric summary (avg, min, max) is what you meant? I'm not entirely sure what is the use case here (it obviously doesn't do anything for queries with top-level histogram_quantile()). If we're mostly supporting queries returning a single sample, perhaps we could instead return DataItem with names of queries as keys and results as values?

### DIFF
--- a/clusterloader2/pkg/measurement/common/generic_query_measurement.go
+++ b/clusterloader2/pkg/measurement/common/generic_query_measurement.go
@@ -20,7 +20,10 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/prometheus/common/model"
+
 	"k8s.io/klog"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
@@ -31,7 +34,9 @@ const (
 )
 
 func init() {
-	create := func() measurement.Measurement { return CreatePrometheusMeasurement(&genericQueryGatherer{}) }
+	create := func() measurement.Measurement {
+		return CreatePrometheusMeasurement(&genericQueryGatherer{})
+	}
 	if err := measurement.Register(name, create); err != nil {
 		klog.Fatalf("Cannot register %s: %v", name, err)
 	}
@@ -40,26 +45,62 @@ func init() {
 type genericQueryGatherer struct {
 	metricName    string
 	metricVersion string
-	rawQuery      string
+	queries       []genericQuery
+	unit          string
+}
+
+type genericQuery struct {
+	name         string
+	query        string
+	hasThreshold bool
+	threshold    float64
 }
 
 func (g *genericQueryGatherer) Configure(config *measurement.Config) error {
-	query, err := util.GetString(config.Params, "rawQuery")
-	if err != nil {
-		return err
-	}
 	metricName, err := util.GetString(config.Params, "metricName")
 	if err != nil {
 		return err
 	}
+
 	metricVersion, err := util.GetString(config.Params, "metricVersion")
 	if err != nil {
 		return err
 	}
 
-	g.rawQuery = query
+	unit, err := util.GetString(config.Params, "unit")
+	if err != nil {
+		return err
+	}
+
+	queries, err := util.GetMapArray(config.Params, "queries")
+	if err != nil {
+		return err
+	}
+	var genericQueries []genericQuery
+	for _, q := range queries {
+		query, err := util.GetString(q, "query")
+		if err != nil {
+			return err
+		}
+		name, err := util.GetString(q, "name")
+		if err != nil {
+			return err
+		}
+		hasThreshold := true
+		threshold, err := util.GetFloat64(q, "threshold")
+		if util.IsErrKeyNotFound(err) {
+			klog.V(2).Infof("No threshold set for %v: %v", metricName, name)
+			hasThreshold = false
+		} else if err != nil {
+			return err
+		}
+		genericQueries = append(genericQueries, genericQuery{name, query, hasThreshold, threshold})
+	}
+
 	g.metricName = metricName
 	g.metricVersion = metricVersion
+	g.unit = unit
+	g.queries = genericQueries
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -68,39 +104,58 @@ func (g *genericQueryGatherer) IsEnabled(config *measurement.Config) bool {
...
-	metric, err := g.query(executor, startTime, endTime)
+	var errs []error
+	data := map[string]float64{}
+	for _, q := range g.queries {
+		samples, err := g.query(q, executor, startTime, endTime)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(samples) != 1 {
```
mborsz, line 115: I'm wondering if also len(samples) == 0 should be accepted.

Let's take a case like I want to track latency of POST on pods, so I write query like:
```
        - name: Perc99
          query: histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{verb="POST", resource="pods}[%v])) by (le))
          threshold: 60
```

Let's say that there are no post on pods. I would expect this to not fail (as there is no sample violating the theshold).

For summary/perfdash you can log any value you want (e.g. not create a key for a query with 0 samples or log 0.0 or NaN).

clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -40,26 +45,57 @@ func init() {
...
+	for _, q := range queries {
+		query, err := util.GetString(q, "query")
+		if err != nil {
+			return err
+		}
+		name, err := util.GetString(q, "name")
+		if err != nil {
+			return err
+		}
+		threshold, err := util.GetFloat64(q, "threshold")
```
mborsz, line 88: Does it mean that theshold is a required field (I think util.GetFloat64 returns an error otherwise).

I think we should allow metric not to have any thresholds (like e.g. simply tracking some value, without validation)?

clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -68,39 +104,58 @@ func (g *genericQueryGatherer) IsEnabled(config *measurement.Config) bool {
 }
 
 func (g *genericQueryGatherer) Gather(executor QueryExecutor, startTime, endTime time.Time, config *measurement.Config) ([]measurement.Summary, error) {
```
mborsz, line 106: Can we add some unit tests for this method? They can simply use fakeQueryExecutor to returns some artificial samples. The goal is to validate:
* that threshold is properly validated (no threshold, threshold higher than value, threshold lower than value)
* that the conversion to the summary happens as we want
* what happens if there are len(samples) > 1
* what happens when len(samples) == 0

clusterloader2/pkg/measurement/common/generic_query_measurement.go
mborsz: not creating a key for a query in case of 0 samples seems to be the cleanest solution IMHO

clusterloader2/pkg/measurement/common/generic_query_measurement.go
aleksandra-malinowska: Done

clusterloader2/pkg/measurement/common/generic_query_measurement.go
aleksandra-malinowska: Added skipping queries that return no results

clusterloader2/pkg/measurement/common/generic_query_measurement.go
aleksandra-malinowska: Done

clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -40,26 +45,61 @@ func init() {
...
+		if err != nil {
+			return err
+		}
+		name, err := util.GetString(q, "name")
+		if err != nil {
+			return err
+		}
+		hasThreshold := false
+		threshold, err := util.GetFloat64(q, "threshold")
+		if err != nil {
```
mborsz, line 91: I think we should distinguish "not found" error (ignore) and other errors (fail). There is a function in util package to check if error type is not found.

mborsz:
Looks good to me. Please fix one comment, fix linter and squash commits.

mborsz:
LGMT from my side. I would like @jprzychodzen to also take a look.

clusterloader2/pkg/util/util.go
```diff
@@ -71,7 +71,12 @@ func GetMap(dict map[string]interface{}, key string) (map[string]interface{}, er
 	return getMap(dict, key)
 }
 
-// GetStringArray tries to return value from map of type map with casting to a string using fmt.Sprintf. If value doesn't exist, error is returned.
+// GetMapArray tries to return value from map of type []map. If value doesn't exist, error is returned.
+func GetMapArray(dict map[string]interface{}, key string) ([]map[string]interface{}, error) {
+	return getMapArray(dict, key)
+}
+
+// GetStringArray tries to return value from map cast to a string using fmt.Sprintf. If value doesn't exist, error is returned.
```
jprzychodzen, line 79: Are you sure you want to change this comment?

clusterloader2/pkg/util/util.go
```diff
@@ -129,11 +134,33 @@ func getMap(dict map[string]interface{}, key string) (map[string]interface{}, er
 
 	mapValue, ok := value.(map[string]interface{})
 	if !ok {
-		return nil, fmt.Errorf("type assertion error: %v is not a string", value)
```
jprzychodzen, line 132: Why this change?

clusterloader2/pkg/util/util.go
aleksandra-malinowska: I think I did, but not to this :) I meant cast to a string slice (as it doesn't return a map, nor accept a map of maps). Does that make sense?

clusterloader2/pkg/util/util.go
jprzychodzen: Ach, sorry, you are right. I've confused this part of a code with different one that was recently touched.

You are correct, we are casting to map[string]interface{}.

clusterloader2/pkg/util/util.go
jprzychodzen: Right, it should include a string slice information. Thanks.

jprzychodzen:
Thanks for this change, single nit if you are improving the comment.

/lgtm
/hold

PR #1980, opened 2022-01-29
Revert "Install bind-tools in dnsperfgo image."

prameshj:
Reverts kubernetes/perf-tests#1968

Installing bind-tools somehow prevents nslookup from doing in-parallel A and AAAA lookups. We do want to test this scenario since it can trigger iptables race conditions - https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts.


prameshj:
/close

Closing in favor of https://github.com/kubernetes/perf-tests/pull/1982

PR #1984, opened 2022-01-31
Add image for sleep command

alculquicondor:
**What type of PR is this?**

/kind feature

Adds an image with a program that sleeps for a given duration. This will be useful to test Jobs, which run to completion, as opposed to run forever.

Part of #799

```release-note
NONE
```

### DIFF
--- a/util-images/sleep/Makefile
+++ b/util-images/sleep/Makefile
@@ -1,0 +1,17 @@
+PROJECT = k8s-staging-perf-tests
+IMG = gcr.io/$(PROJECT)/sleep
+TAG = v0.0.1
+
+all: push
+
+.PHONY: build
+build:
+	docker build --pull -t $(IMG):$(TAG) .
+	docker tag $(IMG):$(TAG) $(IMG):latest
+	@echo Built $(IMG):$(TAG) and tagged with latest
+
+.PHONY: push
+push: build
+	docker push $(IMG):$(TAG)
+	docker push $(IMG):latest
+	@echo Pushed $(IMG) with :latest and :$(TAG) tags
--- a/util-images/sleep/README.md
+++ b/util-images/sleep/README.md
@@ -1,0 +1,10 @@
+# Sleep
+
+Simple program that sleeps for a provided duration.
+
+## Building and Releasing
+
+1. Increment the `TAG` in the Makefile.
+2. `make build`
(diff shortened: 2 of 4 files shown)


util-images/sleep/main.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+func main() {
+	flag.Parse()
+	input := flag.Arg(0)
+	var duration time.Duration
+	if input != "" {
+		var err error
+		duration, err = time.ParseDuration(input)
```
marseel, line 41: I believe that for testing terminating endpointslices we would like to have program that sleeps forever and when it catches SIGTERM then it sleeps for specified time and exits, right?

wojtek-t: This one is for a different usecase (for bath).

But I agree that instead of building too many different images for the similar purpose, we should extend this one.
But that can probably be done in a subsequent PR by @andrewsykim

util-images/sleep/Makefile
```diff
@@ -0,0 +1,17 @@
+PROJECT = k8s-testimages
```
marseel, line 1: Please change to `PROJECT = k8s-staging-perf-tests`

util-images/sleep/README.md
```diff
@@ -0,0 +1,10 @@
+# Sleep
+
+Simple program that sleeps for a provided duration.
+
+## Building and Releasing
+
+1. Increment the `TAG` in the Makefile.
+2. `make build`
+3. Test changes with `docker run gcr.io/k8s-testimages/sleep:latest -- ...`
```
marseel, line 9: s/k8s-testimages/k8s-staging-perf-tests/

marseel:
Except for project name, lgtm

util-images/sleep/Makefile
alculquicondor: Done

util-images/sleep/README.md
alculquicondor: Done

alculquicondor:
Can you help with uploading the image to gcr?

marseel:
I've uploaded image to gcr

/lgtm
/approve

util-images/sleep/main.go
andrewsykim: ack -- noted for when we add this test in v1.24

PR #1992, opened 2022-02-15
Added parsing of generic Pantheon queries results to perfdash

pawbana:
**What type of PR is this?**
/kind feature

Right now results generated by generic queries are not parsed or displayed perfdash. This pr adds this functionality to perfdash.

This is a draft. I don't like this solution as it encodes test label in file name but it changes least amount of code.

Alternative approach:
- Add label with metric name to the results json.
- Make it possible for result file parsers to change test label
  - For example change `getBuildData` so it takes parser as argument and parser functions would either return custom test label or populate results map themself. This would require to change signatures of parser functions.

```release-note
Perfdash now properly displays generic queries results.
```

```docs

```


mborsz:
> Alternative approach:
> 
> * Add label with metric name to the results json.
> * Make it possible for result file parsers to change test label
>   
>   * For example change `getBuildData` so it takes parser as argument and parser functions would either return custom test label or populate results map themself. This would require to change signatures of parser functions.

I'm wondering if we can't simply drop the changes metrics-downloader.go entirely?

IIUC what will happen is that we will parse all files with name like GenericPrometheusQuery_{metric_name}.json and then produce output like:
image

It's not beautiful, but dead simple and should be sufficient for our purposes.

pawbana:
Yes removing modifications from `metrics-downloader.go` would make perfdash behave like that.

Maybe I'm too pedantic but I think it be better to have plots nicely categorized even with the cost of some more changes to the code.

clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -160,18 +160,22 @@ func (g *genericQueryGatherer) query(q genericQuery, executor QueryExecutor, sta
...
-func (g *genericQueryGatherer) createSummary(name string, data map[string]float64) (measurement.Summary, error) {
+func (g *genericQueryGatherer) createSummary(metricName string, data map[string]float64) (measurement.Summary, error) {
 	content, err := util.PrettyPrintJSON(&measurementutil.PerfData{
 		Version: g.metricVersion,
 		DataItems: []measurementutil.DataItem{
 			{
 				Data: data,
 				Unit: g.unit,
+				Labels: map[string]string{
+					"TestName": metricName,
```
mborsz, line 171: "MetricName" instead of TestName?

perfdash/metrics-downloader.go
```diff
@@ -158,7 +158,7 @@ func (g *Downloader) artifactName(jobAttrs Tests, file string) string {
 }
 
 func getResultCategory(metricsFileName string, filePrefix string, category string, artifacts []string) string {
-	if len(artifacts) <= 1 {
+	if len(artifacts) <= 1 || strings.HasPrefix(metricsFileName, GenericPrometheusQueryMeasurementName) {
```
mborsz, line 161: Please add a bool "enforceConstantCategory" or something like that to TestDescription, set it for GenericMeasurements and check it here.

This way, we will pattern match files only in a single place (and e.g. if we will change to some other model, e.g. metadata-based instead of file-name based) it will be easier to reason about changes to be made.

mborsz:
Two nits. LGTM Otherwise

clusterloader2/pkg/measurement/common/generic_query_measurement.go
pawbana: Done

pawbana:
I think adding labels will strike a good balance between complicating code and usability. After adding labels perfdash will behave like this:
image

perfdash/config.go
```diff
@@ -34,9 +34,10 @@ import (
 
 // TestDescription contains test name, output file prefix and parser function.
 type TestDescription struct {
-	Name             string
-	OutputFilePrefix string
-	Parser           func(data []byte, buildNumber int, testResult *BuildData)
+	Name                  string
+	OutputFilePrefix      string
+	Parser                func(data []byte, buildNumber int, testResult *BuildData)
+	forceConstantCategory bool
```
mborsz, line 40: Please start with "F" (large latter) to be consistent with other fields here.

clusterloader2/pkg/measurement/common/generic_query_measurement_test.go
```diff
@@ -147,6 +148,10 @@ func TestGather(t *testing.T) {
 			for _, s := range tc.notWantedData {
 				assert.False(t, strings.Contains(content, s), "summary contains extra data: got: %v, want: no %v", content, s)
 			}
+			testNameLabelPattern := fmt.Sprintf("\"labels\":\\s*\\{\\s*\"MetricName\":\\s*\"%v\"\\s*\\}", tc.name)
+			if match, _ := regexp.MatchString(testNameLabelPattern, content); !match {
```
mborsz, line 152: Do not ignore error.

Proposed fix:
* move regexp to some global var (or top variable in this test) and use testNameLabelPattern :=  regexp.MustParse()
* here use testNameLabelPattern.MatchString which returns no error: https://pkg.go.dev/regexp#Regexp.MatchString

clusterloader2/pkg/measurement/common/generic_query_measurement_test.go
```diff
@@ -147,6 +148,10 @@ func TestGather(t *testing.T) {
 			for _, s := range tc.notWantedData {
 				assert.False(t, strings.Contains(content, s), "summary contains extra data: got: %v, want: no %v", content, s)
 			}
+			testNameLabelPattern := fmt.Sprintf("\"labels\":\\s*\\{\\s*\"MetricName\":\\s*\"%v\"\\s*\\}", tc.name)
```
mborsz, line 151: metricNameLabelPattern?

perfdash/config.go
pawbana: Thx, done

clusterloader2/pkg/measurement/common/generic_query_measurement_test.go
pawbana: Thx, done

clusterloader2/pkg/measurement/common/generic_query_measurement_test.go
pawbana: I don't think I can move regexp to global variable as test case name is part of the pattern.
Added error check.

PR #1995, opened 2022-02-24
Add PROMETHEUS_SCRAPE_MASTER_KUBELETS

mborsz:
Add PROMETHEUS_SCRAPE_MASTER_KUBELETS override that can be used to scrape only master kubelets, without scraping node kubelets.

Cases:
* PROMETHEUS_SCRAPE_MASTER_KUBELETS is not set, we will fallback to the current behaviour (which is to scrape only if we scrape kubelets on nodes)
* PROMETHEUS_SCRAPE_MASTER_KUBELETS is set, we will ignore PROMETHEUS_SCRAPE_KUBELETS

/assign @tosi3k

### DIFF
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -73,6 +73,7 @@ type PrometheusConfig struct {
 	ScrapeEtcd                 bool
 	ScrapeNodeExporter         bool
 	ScrapeKubelets             bool
+	ScrapeMasterKubelets       bool
 	ScrapeKubeProxy            bool
 	ScrapeKubeStateMetrics     bool
 	ScrapeMetricsServerMetrics bool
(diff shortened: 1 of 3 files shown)


mborsz:
/hold

Wait, I will change it slightly

mborsz:
/hold cancel

Let's continue with the current approach. I was thinking about how to actually set this mapping in ours tests. We can use CL2_PROMETHEUS_SCRAPE_MASTER_KUBELETS preset to do that. I see that this is slightly inconsistent with other mappings in prometheus package, but seems to be overall a good path in future. This approach also allows us to distinguish case when PROMETHEUS_SCRAPE_MASTER_KUBELETS is not set (and fallback to the PROMETHEUS_SCRAPE_KUBELETS for backward compatibility) which is not possible/more ugly using flags.

mborsz:
As discussed offline, I changed this to use flags (to be consistent with code in this area) and allowed both PROMETHEUS_SCRAPE_MASTER_KUBELETS and PROMETHEUS_SCRAPE_KUBELETS to control master scraping.

This way, if PROMETHEUS_SCRAPE_MASTER_KUBELETS is not set, we will be backward compatible. If PROMETHEUS_SCRAPE_MASTER_KUBELETS is set, we will scrape master.

PR #1998, opened 2022-03-01
Add job load test: create multiple jobs based on the number of nodes

### LINKED ISSUE #799: Load test: Test run-to-completion workflow for Job objects

alculquicondor:
**What type of PR is this?**

/kind feature

Fixes #799

It was not trivial to gather metrics from kcm, so I'm leaving that as a follow up.

Because of this, I increased the resolution of WaitForJobsToFinish gather interval.

```release-note
Add a load test for the Job controller
```

### DIFF
--- a/clusterloader2/testing/batch/config.yaml
+++ b/clusterloader2/testing/batch/config.yaml
@@ -1,22 +1,31 @@
-{{$mode := (DefaultParam .MODE "Indexed")}}
-{{$pods_per_node_per_size := 20}}
-{{$total_pods_per_size := MultiplyInt .Nodes $pods_per_node_per_size}}
-{{$small_job_size := 5}}
-{{$small_jobs_count := DivideInt $total_pods_per_size $small_job_size}}
-{{$medium_job_size := 20}}
-{{$medium_jobs_count := DivideInt $total_pods_per_size $medium_job_size}}
-{{$large_job_size := 400}}
-{{$large_jobs_count := DivideInt $total_pods_per_size $large_job_size}}
+{{$MODE := DefaultParam .MODE "Indexed"}}
+{{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .NODES_PER_NAMESPACE 100)}}
+{{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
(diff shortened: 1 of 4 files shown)


alculquicondor:
This is now ready for review @jprzychodzen

clusterloader2/testing/batch/overrides.yaml
```diff
@@ -0,0 +1 @@
+MODE: Indexed
```
ArangoGutierrez, line 1: nit, newline

clusterloader2/testing/batch/config.yaml
```diff
@@ -1,12 +1,22 @@
...
 
 namespace:
   number: 1
 
 tuningSets:
-- name: Uniform1qps
+- name: Uniform5qps
   qpsLoad:
-    qps: 1
+    qps: 5
```
jprzychodzen, line 19: Are you sure you want to use so low constant value?

For CL2 test we are using `--env=CL2_LOAD_TEST_THROUGHPUT=50` to determine QPS in saturation tunning test.

clusterloader2/testing/batch/overrides.yaml
alculquicondor: I decided to remove the file instead.

clusterloader2/testing/batch/config.yaml
```diff
@@ -1,12 +1,22 @@
...
+{{$small_jobs_count := DivideInt $total_pods_per_size $small_job_size}}
+{{$medium_job_size := 20}}
+{{$medium_jobs_count := DivideInt $total_pods_per_size $medium_job_size}}
+{{$large_job_size := 400}}
+{{$large_jobs_count := DivideInt $total_pods_per_size $large_job_size}}
+
 name: batch
 
 namespace:
   number: 1
```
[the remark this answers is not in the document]
alculquicondor: Updated the template to have parameters similar to `load`.

PR #1983, opened 2022-01-31
Use the new dnsclient in scale tests.

prameshj:
Use the new DNS client that pushes up metrics and lookups names using GO net.LookupHost.

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs

```

PR #1999, opened 2022-03-02
Revert "Use the new dnsperfgo client in dns load tests."

dlapcevic:
Reverts kubernetes/perf-tests#1962

This will make the tests use the old dns client again.

The plan is to make some changes to the new client and then reactivate the new version for the scale tests.

The change on the new client is in: https://github.com/kubernetes/perf-tests/pull/1982


dlapcevic:
Agreed with @marseel that we will not revert to the old client but rather immediately update to 1.2.0 client, after https://github.com/kubernetes/perf-tests/pull/1983 is merged.

PR #2005, opened 2022-03-07
Extend GenericPrometheusQuery to support multiple streams for single query

mborsz:
Use case:
* I want to get container usage, breakdown by container name

Usage:

```
  - Identifier: gq
    Method: GenericPrometheusQuery
    Params:
      action: start
      metricName: API request latency
      metricVersion: v1
      unit: ms
      dimensions:
      - verb
      - resource
      - subresource
      queries:
        - name: Perc99
          query: histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket[%v])) by (verb, resource, subresource, le))
          threshold: 60
        - name: Perc90
          query: histogram_quantile(0.9, sum(rate(apiserver_request_duration_seconds_bucket[%v])) by (verb, resource, subresource, le))
        - name: Perc50
          query: histogram_quantile(0.5, sum(rate(apiserver_request_duration_seconds_bucket[%v])) by (verb, resource, subresource, le))
          threshold: 5
```

See tests for more examples.

Semantics:
* Still no more than a single value for any dimension combination
* All samples must be below threshold (if set)

Depends on https://github.com/kubernetes/perf-tests/pull/2003

/assign @jprzychodzen
/cc @aleksandra-malinowska

### DIFF
--- a/clusterloader2/pkg/measurement/common/generic_query_measurement.go
+++ b/clusterloader2/pkg/measurement/common/generic_query_measurement.go
@@ -54,6 +54,7 @@ type StartParams struct {
 	MetricVersion string
 	Queries       []GenericQuery
 	Unit          string
+	Dimensions    []string
 }
 
 // TODO(mborsz): github.com/go-playground/validator or similar project?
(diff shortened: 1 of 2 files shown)


mborsz:
Rebased, PTAL

clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -104,38 +105,64 @@ func (g *genericQueryGatherer) IsEnabled(config *measurement.Config) bool {
...
+					Labels: labels,
+				}
+				dataItem.Labels["MetricName"] = g.MetricName
+				dataItems[k] = dataItem
+			}
+
+			val := float64(sample.Value)
+			_, ok = dataItem.Data[q.Name]
+			if ok {
+				errs = append(errs, errors.NewMetricViolationError(q.Name, fmt.Sprintf("too many samples for %s: query returned %v streams, expected 1", k, len(samples))))
```
jprzychodzen, line 150: I'm not sure that len(samples) is the number of streams with the same key. Could you check?

clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -104,38 +105,64 @@ func (g *genericQueryGatherer) IsEnabled(config *measurement.Config) bool {
...
+			if !ok {
+				dataItem = &measurementutil.DataItem{
+					Data:   make(map[string]float64),
+					Unit:   g.Unit,
+					Labels: labels,
+				}
+				dataItem.Labels["MetricName"] = g.MetricName
+				dataItems[k] = dataItem
+			}
+
```
jprzychodzen, line 146: I think that 'else' here would mean that we already parsed this specific key?

jprzychodzen: Ok, I see, this is ok, because we can have multiple queries that attach to the same DataItem

mborsz: We enter the if only if the key is seen for the first time (so we don't this means that the specific key is present already).

clusterloader2/pkg/measurement/common/generic_query_measurement.go
mborsz: Done

clusterloader2/pkg/measurement/common/generic_query_measurement.go
mborsz: Ohh, I see what you meant.

So the entry here is created for key, but values are stored for (key, q.Name).

jprzychodzen: Could you move lines 136 - 145 to separate function 'ensureDataItemForKey' or something like this?

jprzychodzen:
/lgtm

Thanks

mborsz:
/hold

Let me resolve all comments (I see some comments on chat)

clusterloader2/pkg/measurement/common/generic_query_measurement_test.go
```diff
@@ -150,6 +169,158 @@ func TestGather(t *testing.T) {
...
+					{
+						Metric: model.Metric{
+							model.LabelName("d1"): model.LabelValue("d1-val1"),
+							model.LabelName("d2"): model.LabelValue("d2-val1"),
+						},
+						Value: model.SampleValue(2),
+					},
+				},
+			},
+			wantErr: "too many samples",
```
jprzychodzen, line 309: Why we do not need full error message here?

mborsz: Bug in test code. I actually left the substring behaviour to avoid change detector tests.

clusterloader2/pkg/measurement/common/generic_query_measurement.go
mborsz: Done

mborsz:
/hold cancel

All comments addressed

PR #2012, opened 2022-03-08
Cleanup validation and unify execute Measurement/Phase

marseel:
**What type of PR is this?**
/kind cleanup

Removes unnecessary checks, adds validation for Measurements and unifies Measurement execute with Phase execute.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/api/validation_test.go
+++ b/clusterloader2/api/validation_test.go
@@ -241,6 +241,64 @@ func TestVerifyPhase(t *testing.T) {
 	}
 }
 
+func TestVerifyMeasurement(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		input    Measurement
+		expected bool
+	}{
+		{
+			name: "Identifier specified only",
+			input: Measurement{
+				Identifier: "Measurement",
+			},
+			expected: true,
+		},
+		{
+			name: "Instances specified only",
+			input: Measurement{
+				Instances: []*MeasurementInstanceConfig{
+					{
+						Identifier: "Measurement1",
+					},
+					{
+						Identifier: "Measurement2",
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "Both identifier and instances specified",
+			input: Measurement{
+				Identifier: "Measurement",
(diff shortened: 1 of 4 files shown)


clusterloader2/api/validation_test.go
```diff
@@ -264,6 +323,7 @@ func TestVerifyStep(t *testing.T) {
 				Measurements: []*Measurement{
 					{
 						Method: "test1",
+						Identifier: "measurement1",
```
Argh4k, line 326: Nit: we could also add it to "has phases and measurements", to ensure that this test case fails because it has both phases and measurements, not because it has invalid measurement.

clusterloader2/pkg/measurement/measurement_executor.go
```diff
@@ -24,18 +24,26 @@ import (
...
+func Execute(mm Manager, m *api.Measurement) errors.ErrorList {
 	if m.Identifier != "" {
-		return mm.Execute(m.Method, m.Identifier, m.Params)
+		return executeSingleMeasurement(mm, m)
 	}
 	return executeWrapperMeasurement(mm, m)
 }
 
-func executeWrapperMeasurement(mm Manager, m *api.Measurement) error {
+func formatError(method, identifier string, err error) error {
```
Argh4k, line 34: Are we going to use it anywhere else? Right now it is only used once as far as I can see.

Argh4k (approved):

clusterloader2/api/validation_test.go
marseel: Yea, true. The way these tests work right now is quite error-prone - we are only checking that specific configuration is correct/not correct and we don't check why.

clusterloader2/pkg/measurement/measurement_executor.go
marseel: Done, but I think we should refactor this code and test it better.

clusterloader2/api/validation_test.go
marseel: Done.

Argh4k: Agree

PR #2015, opened 2022-03-09
Shows substeps in junit

marseel:
**What type of PR is this?**
Adds measurements/phases as separate steps for reporting - Each measurement/phase will have separate row in junit.xml, which will result in separate row in testgrid.

Depends on PR https://github.com/kubernetes/perf-tests/pull/2012

/kind feature

Fixes #

```release-note

```

```docs

```


mborsz:
Two questions:
* Do we have a guarantee about the order of substeps? I remember @tosi3k was adding "[step: X]" to get some order guarantees.
* Is it a good idea to split phases into separate substeps? My concern is that without a human readable name it will be hard to figure out what it does:
image

PR #2016, opened 2022-03-10
pod startup logging

pohly:
**What type of PR is this?**

/kind cleanup

Running the pod startup test at -v=2 produces too much (periodic status dumps in WaitForPods) and confusing log output (WaitForPods called while cleaning up).

See commit messages  for details. These problems were discussed on Slack: https://kubernetes.slack.com/archives/C09QZTRH7/p1646660039727209

```release-note
NONE
```

/cc @marseel

### DIFF
--- a/clusterloader2/pkg/measurement/util/wait_for_pods.go
+++ b/clusterloader2/pkg/measurement/util/wait_for_pods.go
@@ -21,8 +21,11 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	clientset "k8s.io/client-go/kubernetes"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog"
 )
 
@@ -41,6 +44,7 @@ type WaitForPodOptions struct {
 	CallerName          string
 	WaitForPodsInterval time.Duration
 	SilentProgress      bool
+	Lister              corev1listers.PodLister
 
 	// IsPodUpdated can be used to detect which pods have been already updated.
 	// nil value means all pods are updated.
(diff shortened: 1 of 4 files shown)


marseel:
@pohly 
looks like tests are failing right now.

pohly:
Looking at the job history, it seems to be specific to this PR.

> measurement call WaitForControlledPodsRunning - WaitForSchedulerThroughputDeployments error: metric WaitForControlledPodsRunning has not been started

That may be a result of the new `Dispose` call.

Did you have time to look into whether closing the measurement during "gather" is okay?

pohly:
Yes, "gather" gets used repeatedly for the same measurement:
https://github.com/kubernetes/perf-tests/blob/30041333d995b8d3008bfaeeea594dfaae157be4/clusterloader2/testing/density/config.yaml#L186-L209

That means we need an explicit "stop" action.

clusterloader2/pkg/measurement/util/wait_for_pods.go
```diff
@@ -49,31 +54,69 @@ type WaitForPodOptions struct {
...
+		} else {
+			klog.V(2).Infof("%s: %s: completed successfully", options.CallerName, options.Selector.String())
+		}
+	}()
+
+	// The periodic status dumps below only make sense when there is a
+	// single WaitForPods running, for example in the WaitForRunningPods
+	// measurements. When WaitForControlledPodsRunning invokes it for every
+	// single deployment in parallel, then the log output is too verbose.
+	klogVProgress := klog.V(2)
```
mborsz, line 105: While I agree that dumping all the data every 5s is way to verbose, I have used those logs in the past for debugging of some test cases.

Can we maybe simply change them to log only if they have changed  since the previous dump or something like that? Wouldn't that be sufficient?

pohly: At -v2 it now only logs on changes. At -v5 you still get periodic dumps.

mborsz: Usually we are debugging already finished runs so there is no way to change verbosity flag for them.

clusterloader2/pkg/measurement/util/wait_for_pods.go
```diff
@@ -49,31 +54,69 @@ type WaitForPodOptions struct {
 // WaitForPods waits till desired number of pods is running.
 // Pods are be specified by namespace, field and/or label selectors.
 // If stopCh is closed before all pods are running, the error will be returned.
-func WaitForPods(clientSet clientset.Interface, stopCh <-chan struct{}, options *WaitForPodOptions) error {
-	ps, err := NewPodStore(clientSet, options.Selector)
-	if err != nil {
-		return fmt.Errorf("pod store creation error: %v", err)
+func WaitForPods(clientSet clientset.Interface, stopCh <-chan struct{}, options *WaitForPodOptions) (finalErr error) {
+	var lister func() []*corev1.Pod
+	if options.Selector.FieldSelector == "" && options.Lister != nil {
```
mborsz, line 59: I like this change, but I'm concerned about cpu resource usage growth of clusterloader.

We have done a test with this PR cherry picked and it looks like the CPU usage grows a lot:

* Without this PR:
image

* With this PR:
image

This make sense as now, for each deployment being updated we will be scanning all pods (in a given namespace) and filter pods manually, every 5s. Before this PR, this load was done by kube-apiserver (which is also bad).

Can we optimize this somehow?

mborsz: We can try to use indexers like in https://github.com/kubernetes/kubernetes/blob/4e4a868e138ca1ecff7d126859f01822db1b5ee1/pkg/controller/daemon/daemon_controller.go#L190-L192

but this would require changing WaitForControlledPodsRunning API to ask user for label to index by. If not provided, fallback to old watch per deployment logic.

clusterloader2/pkg/measurement/util/wait_for_pods.go
pohly: `SilentProgress: true` can also become `SilentProgress: false`, i.e. dump progress by default. Then when I run it, I can disable the excessive output. It just seemed to me that the other way around would be a more sensible default.

clusterloader2/pkg/measurement/util/wait_for_pods.go
pohly: I'm not sure if or when I'll have time for such an improvement. I could take this part out of the PR and submit it separately, then perhaps the rest can be merged.

wojtek-t: > We can try to use indexers like in https://github.com/kubernetes/kubernetes/blob/4e4a868e138ca1ecff7d126859f01822db1b5ee1/pkg/controller/daemon/daemon_controller.go#L190-L192

> but this would require changing WaitForControlledPodsRunning API to ask user for label to index by. If not provided, fallback to old watch per deployment logic.


In practise, I don't think we need to over-optimize it here. I believe that having a static index by namespace should be enough.
here is my proposal on how to do that: https://github.com/kubernetes/perf-tests/pull/2056

PR #2018, opened 2022-03-15
Implement prometheus-based container CPU/Memory measurement

mborsz:
/kind feature

This PR adds a new prometheus-based measurement that collects CPU/Memory usage of containers.

/assign @marseel 
Let me verify in presubmit that it actually works (queries already verified) and change default.
/hold


mborsz:
Actually we may need https://github.com/kubernetes/test-infra/pull/25627 to properly validate this in presubmit here (otherwise we won't have any metric here).

PR #2024, opened 2022-03-18
 k8s-infra: Switch to registry.k8s.io

ameukam:
Related:
  - https://github.com/kubernetes/k8s.io/issues/3411

/kind feature

Switch the scalability jobs to the new proxy for Kubernetes container images.


ameukam:
cc @marseel @mborsz @wojtek-t

PR #2027, opened 2022-03-21
Add chaos monkey measurement

aleksandra-malinowska:
Allows enabling chaos monkey only for part of the test.

### DIFF
--- a/clusterloader2/pkg/measurement/common/chaos_monkey_measurement.go
+++ b/clusterloader2/pkg/measurement/common/chaos_monkey_measurement.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"sync"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog"
 	"k8s.io/perf-tests/clusterloader2/api"
 	"k8s.io/perf-tests/clusterloader2/pkg/chaos"
@@ -33,7 +34,9 @@ const (
 
 func init() {
 	create := func() measurement.Measurement {
-		return &chaosMonkeyMeasurement{}
+		return &chaosMonkeyMeasurement{
+			killedNodes: sets.NewString(),
+		}
 	}
 	if err := measurement.Register(chaosMonkeyMeasurementName, create); err != nil {
 		klog.Fatalf("Cannot register %s: %v", chaosMonkeyMeasurementName, err)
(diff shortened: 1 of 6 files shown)


clusterloader2/pkg/measurement/common/chaos_monkey_measurement.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		c.stopChannel = make(chan struct{})
+		c.chaosMonkey = chaos.NewMonkey(config.ClusterFramework.GetClientSets().GetClient(), config.CloudProvider)
+		c.chaosMonkeyWaitGroup, err = c.chaosMonkey.Init(api.ChaosMonkeyConfig{&c.NodeFailureConfig}, c.stopChannel)
+		if err != nil {
+			close(c.stopChannel)
+			return nil, fmt.Errorf("error while creating chaos monkey: %v", err)
+		}
+		return nil, nil
+	case "stop":
```
jprzychodzen, line 69: Is it possible to add case "pause", "resume" (even as a follow up)?

NodeKiller ensures that single VM will be restarted once. Running multiple measurements disables this feature.

aleksandra-malinowska: Pretty sure it's possible. Not sure how complicated it would be, given that current chaos monkey implementation doesn't expect such usage.

Is there any reason to enforce single VM restart throughout the test? It seems like it weakens the test to introduce this assumption, but I don't have much context here.

jprzychodzen: It's related to restart count metric. With enabled chaos monkey we are ok with single restart of each component on each node. Current configuration overwrites this (see https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/chaosmonkey/ignore_node_killer_container_restarts_100.yaml).

We need to match ignored restart to numbers in this configuration, otherwise we will alert on restarting components.

Alternatively we could incorporate connection between Chaos Monkey and restart count measurement to discover and ignore restart generated by Chaos Monkey, but it seems like a much more work to do.

jprzychodzen:
/lgtm

Please follow up with 'pause' and 'resume'

jprzychodzen:
/lgtm

@mborsz Can you approve?

Issue #1311, opened 2020-06-05
Eliminate latency phase from density test


wojtek-t:
PodStartup from SaturationPhase signiifcantly dropped recently:

![Screenshot from 2020-06-05 13-58-57](https://user-images.githubusercontent.com/10743879/83873843-d01e3500-a734-11ea-9d6a-1c62b8977f5b.png)

This seem to be the diff between those two runs:
https://github.com/kubernetes/kubernetes/compare/bded41a81...1700acb03

We should understand why that happened, as if this is expected (not a bug somewhere), this would potentially allow us to achieve our long-standing goal to get rid of latency phase from density test completely.

@mm4tt @mborsz - FYI

wojtek-t:
Nothing interesting happened neither in test-infra nor in perf-tests repo at that time...

mm4tt:
Interesting. Looks like we **had** a similar drop in load test pod startup, but it got back to normal after a few runs and it completely doesn't coincide in time:
http://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=E2E&metricname=LoadPodStartup&Metric=pod_startup 
![5L9Y6Eh6jaZ](https://user-images.githubusercontent.com/2604887/83875775-00b39e00-a738-11ea-865e-d01f4312e8eb.png)

Strange...

mborsz:
I checked https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1335992827380764672/ where we see 57s latency for only stateless pods.

Sample slow pod: 

```
{test-xrxpvb-40/small-deployment-63-677d88d774-xgj2d 1m5.155890502s}
```

Timeline:
* Created at 18:05:11.015687
```
I1207 18:05:11.015687      11 event.go:291] "Event occurred" object="test-xrxpvb-40/small-deployment-63-677d88d774" kind="ReplicaSet" apiVersion="apps/v1" type="Normal" reason="SuccessfulCreate" message="Created pod: small-deployment-63-677d88d774-xgj2d"
```

* Scheduled at 18:06:10.760230
```
I1207 18:06:10.760230      11 scheduler.go:592] "Successfully bound pod to node" pod="test-xrxpvb-40/small-deployment-63-677d88d774-xgj2d" node="gce-scale-cluster-minion-group-2-v779" evaluatedNodes=500 feasibleNodes=500
```
* Running at 18:06:15.139943

```
[...]

PR #2030, opened 2022-03-24
Simplify resource requirements for load test

wojtek-t:
This should help a bit with variance of pod-startup times - a bit more context can be found in:
https://github.com/kubernetes/perf-tests/issues/1311#issuecomment-1077405626

It also makes our configs a bit simpler.

/assign @mborsz


clusterloader2/testing/load/deployment.yaml
```diff
@@ -1,6 +1,8 @@
-{{$HostNetworkMode := DefaultParam .CL2_USE_HOST_NETWORK_PODS false}}
-{{$CpuRequest := DefaultParam .CpuRequest "10m"}}
-{{$MemoryRequest := DefaultParam .MemoryRequest "10M"}}
+{{$HostNetworkMode := DefaultParam .CL2_USE_HOST_NETWORK_PODS false}}i
```
mborsz, line 1: "i" suffix?

wojtek-t: good catch - fixed

wojtek-t:
@mborsz  - PTAL (I fixed the test by adding default values to simple-deployment.yaml)

PR #2033, opened 2022-03-25
Implement perfdash support for ContainerRestarts measurement.

mborsz:
/kind feature

/assign @jprzychodzen


mborsz:
/hold

It's quite to read. Let me change it slightly

mborsz:
/hold cancel

I changed the code to present all containers on a single graph. The reasoning is that it should be 0 for most of the containers and having one view makes it easier to find restarting components.

perfdash/parser.go
```diff
@@ -52,6 +52,30 @@ func createRequestCountData(data *perftype.DataItem) error {
...
+
+func parseContainerRestarts(data []byte, buildNumber int, testResult *BuildData) {
+	build := fmt.Sprintf("%d", buildNumber)
+	restarts := make([]ContainerRestartsInfo, 0)
+	if err := json.Unmarshal(data, &restarts); err != nil {
+		klog.Errorf("error parsing JSON in build %d: %v %s", buildNumber, err, string(data))
+		return
+	}
+
+	restartsbyContainer := make(map[string]float64)
```
jprzychodzen, line 70: nit: s/restartsbyContainer/restartsByContainer

perfdash/parser.go
```diff
@@ -52,6 +52,30 @@ func createRequestCountData(data *perftype.DataItem) error {
...
+	for _, restart := range restarts {
+		restartsbyContainer[restart.Container] = restartsbyContainer[restart.Container] + float64(restart.RestartCount)
```
jprzychodzen, line 72: I'm not sure that 'Container Name' is the best key here. Actually 'ResourceUsage' has a nice pattern of 'simplified(Pod)/Container'. Maybe this is something that we can follow?

What do you think about 
`key = Namespace + "/" + RemoveDisambiguationInfixes(Pod) + "/" + Container`?

perfdash/parser.go
mborsz: Looks like both parts (restartsbyContainer and restartsbyContainer) are the same?

image

perfdash/parser.go
mborsz: I was thinking about including pod name in the key, but the problem is that for static pods (which are a very important use case for this) it includes hostname, e.g. fluentbit-gke-something-something-something and the something are changing from test run to test run.

RemoveDisambiguationInfixes will not strip hostnames which is necessary to make it useful for pods running as static pods , do you want to change it to strip hostname part?

How to do that cleanly? Strip everything after "-gke"? It sounds error-prone.

Personally I feel like container name is good enough and still quite clean. If we better key, we can change it later.

perfdash/parser.go
jprzychodzen: There is capital **B**

perfdash/parser.go
jprzychodzen: FTR: We could configure this 'cloud provider specific' regex as flag for Perfdash. But it seems like an overblown approach.

LGTM here, let's handle it the way it is. If we will need better naming, we can figure out this as needed.

perfdash/parser.go
mborsz: :D thanks!

Done

PR #2034, opened 2022-03-29
Add Windows measurements for soak tests

oprinmarius:
**What type of PR is this?**
/kind feature

Add Windows measurements for
- node storage
- open file handles
- Windows processes
- Windows network traffic

Part of https://github.com/kubernetes/kubernetes/issues/108621

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
+++ b/clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
@@ -49,7 +49,11 @@ const (
 	// bytes issued to I/O operations
 	processQueryTop10 = `topk(10, sum by (process) (windows_process_io_bytes_total{process!~"Idle|Total|System"}))`
 	// process metrics file name prefix
-	processMetricsName                        = "WindowsProcesses"
+	processMetricsName = "WindowsProcesses"
+	// total bytes received and transmitted by interface
+	networkQueryTop10 = `topk(10, sum by (nic) (windows_net_bytes_total))`
+	// network metrics file name prefix
+	networkMetricsName                        = "WindowsNetwork"
 	currentWindowsResourceUsageMetricsVersion = "v1"
 )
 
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -31,13 +31,33 @@ import (
 const (
 	windowsResourceUsagePrometheusMeasurementName = "WindowsResourceUsagePrometheus"
 	// get top 10 non-system processes with highest cpu usage within 1min query window size
-	cpuUsageQueryTop10 = `topk(10, sum by (process) (irate(wmi_process_cpu_time_total{process!~"Idle|Total|System"}[5m]) / on(job) group_left wmi_cs_logical_processors) * 100)`
+	cpuUsageQueryTop10 = `topk(10, sum by (process) (irate(windows_cpu_time_total{process!~"Idle|Total|System"}[5m]) / on(job) group_left windows_cs_logical_processors) * 100)`
```
jsturtevant, line 34: This was 10 ten processes not just cpu, I think knowing top process would be valuable.

oprinmarius: Thank you for noticing this. Initially when I changed the name of this metric I didn't have the process collector enabled and I could only find windows_cpu_time_total metric
Going to submit a change soon with the correct metric name.

jsturtevant:
/ok-to-test

This looks like it would work with the newest versions of `windows_exporter`.  Did the gce windows_exporter node installer get upgraded? Wondering if it would still require the older `wmi_process_cpu_time_total` metric names.

jsturtevant:
> This looks like it would work with the newest versions of `windows_exporter`. Did the gce windows_exporter node installer get upgraded? Wondering if it would still require the older `wmi_process_cpu_time_total` metric names.

I think this is fine as long as the GCE installer and jobs are updated to using the latest `windows_exporter`.  

@ibabou are you still running the perf tests for windows using the old wmi exporter?

jsturtevant:
@ibabou The latest exporter should be compatible, the project was just renamed when it moved under the Prometheus organization.  If this change went it you would want https://github.com/prometheus-community/windows_exporter/releases/download/v0.18.1/windows_exporter-0.18.1-amd64.exe as the url.  

Any chance you can test this change with that url?  The alternative here is use a if statement to apply the current metric names but upgrading seems more appropriate long term.  You'll likely want some of the updates that went into the latest `windows_exporter`, i believe it has some per improvements and things like support for containerd.

You should also be able to use a lot of the work we are doing with the hostrpocess windows_exporter once you are ready to switch. Very little support was specific to Azure and the installation is cleaner with the Hostprocess support that @oprinmarius has been working on.

ibabou:
@jsturtevant Thanks for the clarification! Let's go ahead then with the change, and I'll follow-up with the owners of those scalability jobs to get it updated! and will figure if there is any issues with the new setup.

Re the hostprocess based exporter, agree, I had it disabled just because it's not supported yet on GKE. but this will be the way going forward.

clusterloader2/pkg/measurement/common/slos/windows_node_resource_usage.go
```diff
@@ -99,6 +119,90 @@ func convertToMemoryPerfData(samples []*model.Sample) *measurementutil.PerfData
...
+}
+
+func convertToOpenFilesPerfData(samples []*model.Sample) *measurementutil.PerfData {
+	perfData := &measurementutil.PerfData{Version: currentWindowsResourceUsageMetricsVersion}
+	for _, sample := range samples {
+		item := measurementutil.DataItem{
+			Data: map[string]float64{
+				"Open File Handles": math.Round(float64(sample.Value)),
+			},
+			Unit: "Files",
```
jsturtevant, line 146: This handle count is Unit of `Files` and below the hands are `Handles`.  Should this be consistent with Unit of `Handles`

Issue #2035, opened 2022-03-29
Prometheus support for Azure


oprinmarius:
**What would you like to be added**:
- Windows Exporter as a HostProcess container

- Node Exporter as a daemonset

- Add Azure-Disk drivers

- Update ssd storageClass to support azure-disk configuration

**Why is this needed**:

Currently GCloud SSH is being used to install WMI exporter on Windows nodes and to provision the node-exporter manifest on the master nodes.

part of https://github.com/kubernetes/kubernetes/issues/108621

PR #2041, opened 2022-04-04
Add Windows Exporter as a HostProcess container

oprinmarius:
Signed-off-by: Marius Oprin <moprin@cloudbasesolutions.com>

**What type of PR is this?**

/kind feature

* Add Windows Exporter as a HostProcess container

```release-note

```

```docs

```


clusterloader2/pkg/prometheus/manifests/exporters/windows_node_exporter/windows-exporter-servicemonitor.yaml
```diff
@@ -0,0 +1,19 @@
...
+  endpoints:
+  - interval: 30s
+    port: http
+  jobLabel: k8s-app
+  namespaceSelector:
+    matchNames:
+    - kube-system
+  selector:
+    matchLabels:
+      k8s-app: windows-exporter
```
jsturtevant, line 19: Service monitor and PodMonitor aren't both needed.   This label won't match the Windows-exporter Deamon set above either. it would need to be `app: windows-exporter`  The namespace selector is off too.

I am not sure this projects perfered method (servicemonitor vs podmonitor)

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -154,6 +156,12 @@ func NewController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *Control
 		// Backward compatibility.
 		clusterLoaderConfig.PrometheusConfig.ScrapeNodeExporter = mapping["PROMETHEUS_SCRAPE_NODE_EXPORTER"].(bool)
 	}
+	if _, exists := mapping["PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER"]; !exists {
+		mapping["PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER"] = clusterLoaderConfig.PrometheusConfig.ScrapeWindowsNodeExporter
+	} else {
+		// Backward compatibility.
+		clusterLoaderConfig.PrometheusConfig.ScrapeWindowsNodeExporter = mapping["PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER"].(bool)
```
jsturtevant, line 163: If you are keeping this for backwards compat I think you would also need to keep `additionalScrapeConfigs` that set the windows node ip in the file [clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml](https://github.com/kubernetes/perf-tests/pull/2041/files#diff-50b6a49c70a72aec86e5668edf14744f9ea763b5fd8fdfa2eeb62d131529ceee)?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -199,12 +207,6 @@ func (pc *Controller) SetUpPrometheusStack() error {
 	if err := client.CreateNamespace(k8sClient, namespace); err != nil {
 		return err
 	}
-	// If enabled scraping windows node, need to setup windows node and template mapping
-	if isWindowsNodeScrapingEnabled(pc.templateMapping, pc.clusterLoaderConfig) {
```
jsturtevant, line 203: Same here for back compat?

clusterloader2/pkg/prometheus/manifests/exporters/windows_node_exporter/windows-exporter-daemonset.yaml
```diff
@@ -0,0 +1,46 @@
...
+        app: windows-exporter
+    spec:
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\system"
+      hostNetwork: true
+      initContainers:
+        - name: configure-firewall
+          image: mcr.microsoft.com/windows/nanoserver:1809
```
jsturtevant, line 24: this will work on WS 2019/2022 since this is a hostprocess container but in the future we should make this a multi arch image to save some space (the image gets pulled but is not actually used so on WS2022 we would have to pull the full image where as on 2019 the base layer is probably already pulled)

jsturtevant (commented):
Nice to see HostProcess containers being used with the windows_exporter!

clusterloader2/pkg/prometheus/prometheus.go
oprinmarius: I added them back

clusterloader2/pkg/prometheus/prometheus.go
mborsz: Please revert

PR #2042, opened 2022-04-04
Add node-exporter as a daemonset

oprinmarius:
Signed-off-by: Marius Oprin <moprin@cloudbasesolutions.com>

**What type of PR is this?**

/kind feature
/kind design

* Add Node Exporter as a deamonset

```release-note

```

```docs

```


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -213,9 +211,9 @@ func (pc *Controller) SetUpPrometheusStack() error {
 		return err
 	}
 	if pc.clusterLoaderConfig.PrometheusConfig.ScrapeNodeExporter {
-		if err := pc.runNodeExporter(); err != nil {
```
mborsz, line 216: I think the reason why I implemented this in such weird way (ssh to nodes) is as the daemonsets approach doesn't work in kubemark. https://github.com/kubernetes/perf-tests/pull/596#issuecomment-510500749

clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/nodeExporter-serviceMonitor.yaml
```diff
@@ -0,0 +1,31 @@
...
+    app.kubernetes.io/component: exporter
+    app.kubernetes.io/name: node-exporter
+    app.kubernetes.io/part-of: kube-prometheus
+    app.kubernetes.io/version: 1.3.1
+  name: node-exporter
+  namespace: monitoring
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 15s
```
mborsz, line 14: What was maximum node scale you tested with interval: 15s?

I suspect that prometheus may have a problem with scraping e.g. 15k nodes with 15s interval with a single instance.

mborsz: I guess we may need some tricky logic like in 

https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubeProxy.yaml#L18

if we expect increasing interval.

clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/nodeExporter-daemonset.yaml
```diff
@@ -0,0 +1,117 @@
...
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
+          runAsGroup: 65532
+          runAsNonRoot: true
+          runAsUser: 65532
+      hostNetwork: true
+      hostPID: true
+      nodeSelector:
+        kubernetes.io/os: linux
```
mborsz, line 99: Before this PR, we scraped node-exporter only from master nodes.

Now we scrape from all nodes which is a huge difference from performance perspective (1 master vs e.g. few thousands targets to scrape).

I'm fine with this change, but this will require ensuring that scrape interval is low enough to make it work, or sharding promethus, ensuring that prometheus has enough resources (cpu/memory) to handle them and running some larger scale tests to validate that this works.

jsturtevant: would it make sense to keep this label selector to just the control plane nodes?  Seems like it could be a big change to include it on all the Linux nodes. Are there any use cases for having this run on all the nodes?

jsturtevant: would it make sense to keep this label selector to just the control plane nodes?  Seems like it could be a big change to include it on all the Linux nodes. Are there any use cases for having this run on all the nodes?

clusterloader2/pkg/prometheus/prometheus.go
jsturtevant: I am not familiar with kubemark.  Why do the control planes in kubemark have the `NetworkUnavaiable` condition?

clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/nodeExporter-daemonset.yaml
mborsz: Yes, this makes sense. You can also add a flag to include in on all linux nodes if needed, but by default please scrape only masters.

clusterloader2/pkg/prometheus/prometheus.go
mborsz: Kubemark has incorrect network setup and we never had a time to cleanup this. While that time, the kubemark was quite important to us, now the focus has changed and we rarely use it.

I'm fine with any implementation that won't be scraping on masters on kubemark clusters.

clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/nodeExporter-daemonset.yaml
jsturtevant: in 1.24, many nodes in clusters will be marked with `node-role.kubernetes.io/control-plane` see https://github.com/kubernetes/kubernetes/pull/107533

PR #2043, opened 2022-04-04
Add support for Azure Disk storageClass

oprinmarius:
**What type of PR is this?**

/kind feature

* Adds support for Azure-Disk storageClass

```release-note

```

```docs

```

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
@@ -6,8 +6,17 @@ apiVersion: storage.k8s.io/v1
 metadata:
   name: ssd
 provisioner: {{$PROMETHEUS_STORAGE_CLASS_PROVISIONER}}
+{{if eq $PROMETHEUS_STORAGE_CLASS_PROVISIONER "kubernetes.io/azure-disk"}}
+parameters:
+  kind: Managed
+  storageaccounttype: {{$PROMETHEUS_STORAGE_CLASS_VOLUME_TYPE}}
+  cachingmode: ReadOnly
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
```diff
@@ -7,7 +7,18 @@ metadata:
   name: ssd
 provisioner: {{$PROMETHEUS_STORAGE_CLASS_PROVISIONER}}
 parameters:
+  {{if eq $PROMETHEUS_STORAGE_CLASS_PROVISIONER "kubernetes.io/gce-pd"}}
```
mborsz, line 10: This stops setting type for PROMETHEUS_STORAGE_CLASS_PROVISIONER != kubernetes.io/gce-pd which is a breaking changes and will break some gke tests.

Can you remove this if and set PROMETHEUS_STORAGE_CLASS_VOLUME_TYPE to empty value when needed?
IIUC empty value shouldn't get overridden by DefaultParam above: https://github.com/kubernetes/perf-tests/blob/d8d8c4b6016376f3871432dd221e44649bb73837/clusterloader2/pkg/config/template_functions.go#L202-L202.

oprinmarius: `type` is not a valid parameter for azure-disk storageClass and must be replaced with other parameters. I will update this

marosset:
Can the azuredisk drivers be deployed with either https://github.com/kubernetes-sigs/azuredisk-csi-driver/tree/master/charts/v1.15.0 or https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/docs/install-csi-driver-v1.15.0.md ?

PR #2044, opened 2022-04-05
Revert " k8s-infra: Switch to registry.k8s.io"

mborsz:
Reverts kubernetes/perf-tests#2024

This breaks internal GKE tests. Let's rollback and figure out why.

/assign @marseel

### DIFF
--- a/clusterloader2/testing/density/deployment.yaml
+++ b/clusterloader2/testing/density/deployment.yaml
@@ -16,7 +16,7 @@ spec:
         group: {{.Group}}
     spec:
       containers:
-      - image: registry.k8s.io/pause:3.1
+      - image: k8s.gcr.io/pause:3.1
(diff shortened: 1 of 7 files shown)


mborsz:
We know that the problem is that GKE tests have no access to images stored outside of gcr.io.

We are checking possible options.

BenTheElder:
IMHO: GKE should probably set unreachable registry domains to be reached by a mirror (e.g. container mirror config) and / OR pin the version of this tool in use.

Do we have any idea when we can re-revert this?

ameukam:
> We know that the problem is that GKE tests have no access to images stored outside of gcr.io.
> 
> We are checking possible options.

FYI the k/k repo will flip to `registry.k8s.io`soon https://github.com/kubernetes/kubernetes/pull/109938.

PR #2049, opened 2022-04-13
Improve logging in EtcdMetrics

mborsz:
Right now it can fail due to sshResult.Code !=0 and we won't even know why it failed (or what sshResult.Code value was).

/assign @marseel


clusterloader2/pkg/measurement/common/etcd_metrics.go
```diff
@@ -196,6 +196,8 @@ func (e *etcdMetricsMeasurement) getEtcdMetrics(host string, provider provider.P
 	cmd := fmt.Sprintf("curl http://localhost:%d/metrics", port)
 	if samples, err := e.sshEtcdMetrics(cmd, host, provider); err == nil {
 		return samples, nil
+	} else {
```
wojtek-t, line 199: nit - above you error.

So instead of adding else, just log it unconditionally here (it's semantically equivalent)

PR #2053, opened 2022-04-25
Log node allocatable on cluster loader startup

wojtek-t:
Ref https://github.com/kubernetes/perf-tests/issues/1311#issuecomment-1108525343

/assign @mborsz


wojtek-t:
/hold

This is a bit too verbose now - will take a look at it later.

PR #2056, opened 2022-05-02
Avoid instantiating separate pod store for every object managed by clusterloader

wojtek-t:
/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -27,10 +27,12 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -497,6 +528,32 @@ func (w *waitForControlledPodsRunningMeasurement) getObjectKeysAndMaxVersion() (
...
 }
 
+type wrapperPodStore struct {
+	indexer       cache.Indexer
+	namespace     string
+	labelSelector labels.Selector
+}
+
+func (w *wrapperPodStore) List() []*v1.Pod {
+	objects, err := w.indexer.ByIndex("metadata.namespace", w.namespace)
```
mborsz, line 538: I'm wondering if this is sufficient.

Before this PR, we used do filtering server side which saved tons of CPU here.
Now we will be iterating all pods in a namespace every 5s from multiple WaitForPods calls.

Do we have some data about CPU usage of pod running the test after this change? How it changed wrt to old pattern?

I was thinking about using some user-provided indexing key, but your approach is better, if it solves the problem properly.

wojtek-t: User-provided stuff is always more error prone, so I wanted to avoid that.
And in practice our namespaces are relatively small.

I was relying a bit on the graph that you pasted on the other PR showing that without any indexing it was using ~2 cores. With indexing, it should on average be using ~50x less in our load test...

We should run it against large test, but I'm wondering if we already have any way to automatically grab the information about the test resource usage itself?

mborsz: There is no automatic way. You need to go to the stackdriver of cluster running a prow pod and check cpu usage there (you should have an access). This is how the graph on the other PR was created.

pohly: You could build a test binary, then run that binary under `/usr/bin/time -v`. That should give you a total of the resource consumption of the test driver itself, with output directly in the log file.

mborsz: The CPU graph for presubmit of this PR:
image

You can compare this with graphs in https://github.com/kubernetes/perf-tests/pull/2016#discussion_r843925817

Overall, it looks good to me.

mborsz:
I managed to take a pprof for a period of increased memory usage

image

I'm not sure how much it helps.

wojtek-t:
The alternative merged - closing.

PR #2057, opened 2022-05-05
stop using PSP API which is removed in 1.25

deads2k:
found while investigating the 100 test on https://github.com/kubernetes/kubernetes/pull/108797 

Not sure if it's a root cause of the failures or not.  I am interested in where I'd find logs in the job that would show this failure.  https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/108797/pull-kubernetes-e2e-gce-100-performance/1522188489787969536 being an example job.


deads2k:
cc @wojtek-t

PR #2066, opened 2022-05-13
clusterloader2: Fix error of probes about metrics-server

yangjunmyfm192085:
**What type of PR is this?**
/kind bug

When using `InClusterAPIServerRequestLatency` to measure the latency of metrics-server, need to add `serviceMonitor` and other information, and update the probe image

Fixes #

```release-note
"NONE"
```

```docs
"NONE"
```


yangjunmyfm192085:
Plan to use perf-test in `kubernetes-sigs/metrics-server`, which requires the probe resources, fix the bugs
/cc @mborzx @mm4tt

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-deployment.yaml
```diff
@@ -4,27 +4,27 @@ metadata:
...
   template:
     metadata:
       labels:
-        probe: metrics-server-probe
+        probe: kube-client
     spec:
       containers:
         - name: prober
-          image: gcr.io/k8s-testimages/probes:v0.0.4
+          image: gcr.io/k8s-testimages/probes:v0.0.5
```
yangjunmyfm192085, line 20: @mm4tt, I build the image using `https://github.com/kubernetes/perf-tests/blob/master/util-images/probes/Dockerfile ` but don't have permission to push `gcr.io/k8s-testimages/probes:v0.0.5`.
Need help to push an image `gcr.io/k8s-testimages/probes:v0.0.5`

yangjunmyfm192085: /cc @wojtek-t @mborsz @mm4tt

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-deployment.yaml
```diff
@@ -4,27 +4,27 @@ metadata:
   namespace: probes
   name: metrics-server-prober
   labels:
-    probe: metrics-server-probe
+    probe: kube-client
```
wojtek-t, line 7: Hmm - I actually likes the metrics-server-probe (or sth like that).

kube-client seems too generic too me.

Why are you changing this?

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-role.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
wojtek-t, line 2: What's the point of splitting into multiple files? Keeping whole BRA-related stuff in one file as it was actually is pretty convinient to use...

yangjunmyfm192085: Because I found that some resources were not created successfully  in one file, so splitting into multiple files.
And I found that dnsLookup and network are also separated

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-deployment.yaml
yangjunmyfm192085: yeah, thanks for review. I have updated all suggestions

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-role.yaml
yangjunmyfm192085: ok, I have I have rollbacked to use one file.

yangjunmyfm192085 (commented):
/cc @wojtek-t I have updated all suggestions.

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-rbac.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: v1
+kind: List
+items:
+- apiVersion: rbac.authorization.k8s.io/v1
```
wojtek-t, line 4: I don't understand this...

k8s doesn't support list containing multiple types of objects...

I still don't understand why you're touching it - what didn't work with the original file?

yangjunmyfm192085: the issue is here
https://github.com/kubernetes/perf-tests/blob/e6c9306927161ab03460f2d7f0d284e78a732c2e/clusterloader2/pkg/framework/framework.go#L275
According to the original file, only the first object can be parsed from the file, and the following cannot be parsed.

yangjunmyfm192085 (commented):
/test pull-perf-tests-clusterloader2-e2e-gce-scale-performance-manual

clusterloader2/pkg/measurement/common/probes/manifests/metricsServer/metrics-server-prober-rbac.yaml
yangjunmyfm192085: and I have observed that other manifest files have only one resource.

wojtek-t: OK - that makes sense.

wojtek-t:
/assign @marseel 

@marseel - can you please help with pushing the new image?

wojtek-t:
/approve

But we need to push the image before lgtm-ing.

marseel:
I've pushed 0.0.5, but please change also util-images/probes/Makefile version from 0.0.4 to 0.0.5

yangjunmyfm192085 (commented):
/cc @[marseel](https://github.com/marseel)

yangjunmyfm192085:
> I've pushed 0.0.5, but please change also util-images/probes/Makefile version from 0.0.4 to 0.0.5

Hi, @marseel I have applied all modifications, any other updates?

PR #2073, opened 2022-05-19
Use performance metrics from the latest dnsperfgo client

dlapcevic:
Use the Prometheus metrics generated by the dnsperfgo (https://github.com/kubernetes/perf-tests/tree/master/dns/dnsperfgo) pods.
1. Make changes to deployments that run these pods to expose the port for the metrics.
2. Add DNS test for in-cluster K8s hostnames to the load test.

/kind feature
/assign @marseel


clusterloader2/testing/load/modules/dns-k8s-hostnames.yaml
```diff
@@ -0,0 +1,27 @@
+{{$ENABLE_DNSTESTS := DefaultParam .CL2_ENABLE_DNSTESTS false}}
```
marseel, line 1: Can we have separate flag different from CL2_ENABLE_DNSTESTS? If we use the same flag we are risking that currently passing tests break.

dlapcevic: Added a new flag `USE_ADVANCED_DNSTEST` to guard the new functionality in `deployment.yaml`, `measurement.yaml` and `dns-k8s-hostnames.yaml`.

We should remove this flag after it's confirmed that it works as intended on a subset of tests, so that this runs for all tests that have `ENABLE_DNSTESTS`.

clusterloader2/testing/load/deployment.yaml
```diff
@@ -22,12 +25,14 @@ spec:
...
+{{if and .EnableDNSTests $USE_ADVANCED_DNSTEST}}
+        dns-test: dnsperfgo
+{{end}}
     spec:
       hostNetwork: {{$HostNetworkMode}}
       containers:
 {{if .EnableDNSTests}}
-      # TODO need to push this image.
-      - image: gcr.io/k8s-staging-perf-tests/dnsperfgo:v1.2.0
+      - image: gcr.io/k8s-staging-perf-tests/dnsperfgo:v1.3.0
```
marseel, line 35: Can you guard new image too, please?

PR #2075, opened 2022-05-26
Add control-plane toleration for Prometheus for Windows tests

oprinmarius:
**What type of PR is this?**
/kind feature

Adds control-plane tolerations for Prometheus when running Windows Tests in Azure

part of #2035

Windows Soak tests running in Azure do not spawn Linux worker nodes.
Prometheus service has to be scheduled on the control-plane nodes.

```release-note

```

```docs

```


clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -43,13 +44,21 @@ spec:
   - key: "monitoring"
     operator: "Exists"
     effect: "NoSchedule"
+  # Windows test jobs spawn clusters without Linux nodes. Add a toleration for Prometheus to be scheduled on the control-plane nodes
```
marosset, line 47:
```suggestion
  # Windows jobs spawn clusters without Linux worker nodes. Add a toleration for Prometheus to be scheduled on the control-plane nodes
```

clusterloader2/pkg/prometheus/manifests/0prometheus-operator-deployment.yaml
```diff
@@ -49,6 +50,13 @@ spec:
           readOnlyRootFilesystem: true
       nodeSelector:
         kubernetes.io/os: linux
+      # Windows test jobs spawn clusters without Linux nodes. Add a toleration for Prometheus to be scheduled on the control-plane nodes
```
marosset, line 53:
```suggestion
      # Windows jobs spawn clusters without Linux worker nodes. Add a toleration for Prometheus to be scheduled on the control-plane nodes
```

marosset:
/lgtm
The behavior of running windows containers on the worker nodes and linux containers on the control-plane nodes is consistent with kubernetes/kubernetes e2e tests.
This is documented at https://github.com/kubernetes-sigs/windows-testing#method-1-using-an-existing-cluster

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```diff
@@ -1,5 +1,6 @@
 {{$PROMETHEUS_SCRAPE_KUBELETS := DefaultParam .PROMETHEUS_SCRAPE_KUBELETS false}}
 {{$PROMETHEUS_SCRAPE_WINDOWS_NODES := DefaultParam .PROMETHEUS_SCRAPE_WINDOWS_NODES false}}
+{{$PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER := DefaultParam .PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER false}}
```
jsturtevant, line 3: can we resuse the `PROMETHEUS_SCRAPE_WINDOWS_NODES`? feels awkward to have to different windows scrape settings

jackfrancis:
FYI using this PR as a guinea pig to test these changes:

https://github.com/kubernetes/test-infra/pull/26397/files#r883159740

clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
oprinmarius: PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER  is the variable we introduced when we switched to use Windows Exporter as a HostProcess and we need to use it for Azure jobs

PROMETHEUS_SCRAPE_WINDOWS_NODES  variable is used by the GCE jobs to install WMI exporter on the nodes via gcloud ssh command

I used the new PROMETHEUS_SCRAPE_WINDOWS_NODE_EXPORTER variable to add taints to the control-plane nodes and to remove storage configuration for Prometheus so I don't alter the other jobs that might be running Windows tests in GCE

jsturtevant: thanks for the clarification.  If we can eventually remove the GCE specific code and use just the hostprocess install method, we can eventually drop PROMETHEUS_SCRAPE_WINDOWS_NODES.  My understanding is that GCE doesn't support hpc currently but will in the future.

PR #2076, opened 2022-05-26
Add 'do not evict' annotation to Prometheus

aleksandra-malinowska:
The annotation prevents cluster autoscaler from evicting the pod, reducing potential for disruption to measurements.


marseel:
FTR in 0.50.0 prometheus-operator, there was bug introduced for podMetadata, which was fixed in 0.53.0. Currently we have 0.48.0 prometheus-operator, but if we decide to upgrade version, we should keep it in mind. 
https://github.com/prometheus-operator/prometheus-operator/pull/4422

/lgtm
/approve

PR #2079, opened 2022-06-05
Make exec_deployment.yaml path configurable

jupblb:
**What type of PR is this?**
/kind feature

Allows setting custom path to `exec_deployment.yaml`. Necessary if clusterloader is ran from another directory than `perf-tests/clusterloader2`.

N/A

This does not introduce a breaking change.

```release-note
NONE
```

```docs

```


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus_test.go
```diff
@@ -49,7 +52,6 @@ var (
 
 func turnOffLoggingToStderrInKlog(t *testing.T) {
 	if klogLogToStderr {
-		klog.InitFlags(nil)
```
jupblb, line 52: For some reason having `InitFlags` in `exec_service.go` changed code initialisation in a way that caused `flags.go#init()` (which already has `klog.InitFlags(nil)`) to fire before this line, causing a flag init collision (nasty bug). Solved by adding an empty import to `flags` package.

PR #2080, opened 2022-06-06
Update CL2 APIResponsivenessPrometheus threasholds

### LINKED ISSUE #1908: CL2 doesn't distinguish between single pod and collection of pods deletion in api-latency measurement
For example, a test can fail with:

```
[measurement call APIResponsivenessPrometheus - APIResponsivenessPrometheus error: top latency metric: there should be no high-latency requests, but: [got: &{Resource:events Subresource: Verb:DELETE Scope:namespace Latency:perc50: 33.374999999s, perc90: 38.74s, perc99: 39.873999999s Count:11 SlowCount:10}; expected perc99 <= 1s]]
```

The scope is namespaces, so it's actually a DELETECOLLETION call which shouldn't be a subject for 1s time limit.

In addition, the grafana dashboards should be changed to reflect this distinction. We should have separate graphs for resource level mutations (1s limit) and separate for collection mutations.

azylinski:
**What type of PR is this?**

/kind cleanup

1. Update thresholds to reflect updated api_call_latency SLO definition: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md#definition
2. Fix/Simplify the getSLOThreshold() logic - before, it would not work for DELETECOLLETION (`verb:DELETE` `scope:namespace`).

Fixes #1908

Change tests classification criteria.

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -39,9 +39,8 @@ const (
 
 	// Thresholds for API call latency as defined in the official K8s SLO
 	// https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md
-	resourceThreshold  time.Duration = 1 * time.Second
-	namespaceThreshold time.Duration = 5 * time.Second
-	clusterThreshold   time.Duration = 30 * time.Second
+	singleResourceThreshold    time.Duration = 1 * time.Second
+	multipleResourcesThreshold time.Duration = 30 * time.Second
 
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
```
wojtek-t, line 2: Please don't change this boilerplate for existing files.

azylinski: Done.

PR #2084, opened 2022-06-08
Add custom nodeSelector for Prometheus

jprzychodzen:
Allow custom nodeSelector labels to harden Prometheus Pods

I need to do manual tests to confirm that everything behaves as expected
/hold


jprzychodzen:
/unhold

/assign @mborsz 

Tested manually that parameter work.

clusterloader2/pkg/prometheus/manifests/0prometheus-operator-deployment.yaml
```diff
@@ -49,6 +50,9 @@ spec:
           readOnlyRootFilesystem: true
       nodeSelector:
         kubernetes.io/os: linux
+        {{if ne $PROMETHEUS_NODE_SELECTOR ""}}
```
mborsz, line 53: Nit: Do we need to if for != "".

It looks like simply writing
```
nodeSelector:
  {{$PROMETHEUS_NODE_SELECTOR}}
```

will work the same way.

mborsz:
/lgtm
/approve
/hold

To decide if you want to address nit.

jprzychodzen:
/unhold

@mborsz PTAL

PR #2094, opened 2022-06-29
Add GCP Compute API quotas usage measurement

tosi3k:
This change adds a GCE-related API quotas usage measurement to the `load` test.

Underneath, we extend the CL2's capabilities to use the [Google Cloud Managed Service for Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus) as a backend instead of the in-cluster Prometheus instance. This backend be enabled through `prometheusClient: managed` parameter in the Prometheus-based measurements' params in the test scenario config and works only for a limited number of providers.

/sig scalability

### DIFF
--- a/clusterloader2/pkg/measurement/common/prometheus_measurement.go
+++ b/clusterloader2/pkg/measurement/common/prometheus_measurement.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	prom "k8s.io/perf-tests/clusterloader2/pkg/prometheus/clients"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
 
@@ -56,7 +57,14 @@ type prometheusMeasurement struct {
 }
 
 func (m *prometheusMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
-	if config.PrometheusFramework == nil {
+	prometheusClient, err := util.GetStringOrDefault(config.Params, "prometheusClient", "inCluster")
+	if err != nil {
+		return nil, err
+	}
+	if prometheusClient != "inCluster" && prometheusClient != "managed" {
+		return nil, fmt.Errorf("unknown Prometheus client")
+	}
+	if prometheusClient == "inCluster" && config.PrometheusFramework == nil {
 		klog.Warningf("%s: Prometheus is disabled, skipping the measurement!", config.Identifier)
 		return nil, nil
 	}
@@ -86,8 +94,17 @@ func (m *prometheusMeasurement) Execute(config *measurement.Config) ([]measureme
 			return nil, err
 		}
 
-		c := config.PrometheusFramework.GetClientSets().GetClient()
-		executor := measurementutil.NewQueryExecutor(c)
+		var pc prom.Client
+		switch prometheusClient {
+		case "inCluster":
+			pc = prom.NewInClusterPrometheusClient(config.PrometheusFramework.GetClientSets().GetClient())
+		case "managed":
+			pc, err = config.CloudProvider.GetManagedPrometheusClient()
+			if err != nil {
+				return nil, fmt.Errorf("error while creating managed Prometheus client: %w", err)
+			}
+		}
+		executor := measurementutil.NewQueryExecutor(pc)
 
 		summary, err := m.gatherer.Gather(executor, m.startTime, time.Now(), config)
 		if err != nil {
--- a/clusterloader2/pkg/measurement/util/prometheus.go
+++ b/clusterloader2/pkg/measurement/util/prometheus.go
@@ -17,7 +17,6 @@ limitations under the License.
 package util
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -28,8 +27,8 @@ import (
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/common/model"
 	"k8s.io/apimachinery/pkg/util/wait"
-	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
+	prom "k8s.io/perf-tests/clusterloader2/pkg/prometheus/clients"
 )
 
 const (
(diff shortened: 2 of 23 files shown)


clusterloader2/testing/load/modules/measurements.yaml
```diff
@@ -151,6 +151,21 @@ steps:
...
+      metricVersion: v1
+      prometheusClient: managed
+      unit: Queries
+      queries:
+      - name: default
+        query: sum by (quota_metric) (delta(serviceruntime_googleapis_com:quota_rate_net_usage{monitored_resource="consumer_quota", quota_metric="compute.googleapis.com/default"}[%v]))
+      - name: list_requests
+        query: sum by (quota_metric) (delta(serviceruntime_googleapis_com:quota_rate_net_usage{monitored_resource="consumer_quota", quota_metric="compute.googleapis.com/list_requests"}[%v]))
+      - name: read_requests
+        query: sum by (quota_metric) (delta(serviceruntime_googleapis_com:quota_rate_net_usage{monitored_resource="consumer_quota", quota_metric="compute.googleapis.com/read_requests"}[%v]))
```
marseel, line 168: For testing, this metric is fine, but I would prefer not to relay on it as values will depend on length of test etc.

tosi3k: Done.

clusterloader2/pkg/measurement/util/prometheus.go
```diff
@@ -85,15 +89,71 @@ type promResponseData struct {
...
+
+func NewManagedPrometheusClient(client *http.Client) PrometheusClient {
+	return &managedPrometheusClient{
+		client: client,
+		uri:    fmt.Sprintf("https://monitoring.googleapis.com/v1/projects/%s/location/global/prometheus/api/v1/query", os.Getenv("PROJECT")),
+	}
+}
+
 // Query executes given prometheus query at given point in time.
 func (e *PrometheusQueryExecutor) Query(query string, queryTime time.Time) ([]*model.Sample, error) {
```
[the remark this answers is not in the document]
tosi3k: Done.

clusterloader2/pkg/measurement/common/prometheus_measurement.go
```diff
@@ -56,7 +59,14 @@ type prometheusMeasurement struct {
...
 func (m *prometheusMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
-	if config.PrometheusFramework == nil {
+	prometheusClient, err := util.GetStringOrDefault(config.Params, "prometheusClient", "inCluster")
+	if err != nil {
+		return nil, err
+	}
+	if prometheusClient != "inCluster" && prometheusClient != "managed" {
+		return nil, fmt.Errorf("unknown Prometheus client")
+	}
+	if (config.PrometheusFramework == nil && prometheusClient == "inCluster") || (prometheusClient == "managed" && (config.CloudProvider.Name() != provider.GKEName || config.CloudProvider.Name() != provider.GCEName)) {
```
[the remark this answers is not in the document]
tosi3k: Thanks for the remark, done.

clusterloader2/pkg/measurement/common/prometheus_measurement.go
```diff
@@ -56,10 +59,21 @@ type prometheusMeasurement struct {
...
+		return nil, err
+	}
+	if prometheusClient != "inCluster" && prometheusClient != "managed" {
+		return nil, fmt.Errorf("unknown Prometheus client")
+	}
+	if config.PrometheusFramework == nil && prometheusClient == "inCluster" {
 		klog.Warningf("%s: Prometheus is disabled, skipping the measurement!", config.Identifier)
 		return nil, nil
 	}
+	if cp := config.CloudProvider.Name(); cp != provider.GKEName && cp != provider.GCEName && prometheusClient == "managed" {
```
marseel, line 73: Can we add flag to provider that specifies whether managed prometheus is supported and set it for gke and gce to true?

tosi3k: Good suggestion, done.

clusterloader2/testing/load/modules/measurements.yaml
```diff
@@ -151,6 +151,27 @@ steps:
...
+    Method: GenericPrometheusQuery
+    Params:
+      action: {{$action}}
+      metricName: Quota usage
+      metricVersion: v1
+      prometheusClient: managed
+      unit: QPMs
+      queries:
+      - name: default perc99
+        query: quantile_over_time(0.99, sum by (quota_metric) (irate(serviceruntime_googleapis_com:quota_rate_net_usage{monitored_resource="consumer_quota", quota_metric="compute.googleapis.com/default"}[1m]))[%v:]) * 60
```
mborsz, line 164: Maybe we can use "dimensions" feature or "GenericPrometheusQuery" to avoid 3 copies of query?

tosi3k: SG, done.

tosi3k:
Some transient `gcloud`-related issues in `pull-perf-tests-clusterloader2`.

/retest

clusterloader2/testing/load/modules/measurements.yaml
```diff
@@ -151,6 +151,21 @@ steps:
       - name: Perc50
         query: quantile_over_time(0.50, sum by (container) (container_memory_working_set_bytes / 1024 / 1024)[%v:])
 {{end}}
+  - Identifier: Quotas total usage
```
mborsz, line 154: Is it really OK that we enable it everywhere, even for other cloud providers?

I understand that we ignore this later, but this seems to be quite misleading. I think it would be better to flag guard this and enable it only where needed.

clusterloader2/pkg/measurement/common/prometheus_measurement.go
```diff
@@ -56,10 +58,21 @@ type prometheusMeasurement struct {
...
+	}
+	if prometheusClient != "inCluster" && prometheusClient != "managed" {
+		return nil, fmt.Errorf("unknown Prometheus client")
+	}
+	if prometheusClient == "inCluster" && config.PrometheusFramework == nil {
 		klog.Warningf("%s: Prometheus is disabled, skipping the measurement!", config.Identifier)
 		return nil, nil
 	}
+	if prometheusClient == "managed" && !config.CloudProvider.Features().SupportManagedPrometheus {
+		klog.Warningf("%s: Google Cloud Managed Service for Prometheus is disabled for %q cloud provider, skipping the measurement!", config.CloudProvider, config.Identifier)
```
mborsz, line 73: If I configured the measurement in a way that is not supported, I think it should be an error, not warning.

Do we really need this check?
I understand that we want to avoid breaking tests for other cloud providers, but IMHO this should happen on the config level, not here.

clusterloader2/testing/load/modules/measurements.yaml
tosi3k: Hid the measurement under a flag guard.

clusterloader2/pkg/measurement/common/prometheus_measurement.go
tosi3k: Removed the warning && added managed Prometheus getter to the cloud provider interface.

clusterloader2/pkg/provider/util.go
```diff
@@ -22,6 +22,10 @@ import (
 	sshutil "k8s.io/kubernetes/test/e2e/framework/ssh"
 )
 
+const (
+	noManagedPrometheus = "no managed Prometheus service for this cloud provider"
```
mborsz, line 26: Nit: Please use ErrNoManagedPrometheus = errors.New("no managed Prometheus service for this cloud provider")

This way it will be possible to check error kind without string matching.

Can be done as follow up PR

tosi3k: Done.

PR #2095, opened 2022-06-30
Implement profiling in clusterloader

mborsz:
Pprof for clusterloader.

It can be used e.g. to understand memory usage in https://github.com/kubernetes/perf-tests/pull/2056.

/assign @jupblb

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -19,6 +19,8 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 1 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -255,6 +259,13 @@ func main() {
 		klog.Exitf("Flag parse failed: %v", err)
 	}
 
+	if profiling {
```
jupblb, line 262: If `profiling` is `false` does that mean that pprof won't run?

mborsz: Yes, the pprof runs logic only when /debug/pprof is accessed

PR #2103, opened 2022-07-14
Avoid instantiating separate pod store for every object managed by clusterloader attempt #3

mborsz:
Based on https://github.com/kubernetes/perf-tests/pull/2056 and https://github.com/kubernetes/perf-tests/pull/2016

The PR tries to avoid creating a separate pod store/watch for each object, instead it uses single informer and client-side filtering.

This increases cpu usage as the filtering is happening client-side to values like (for 5k node performance test):

image

Baseline:
image

As per pprofs ~70% at the cpu spike are being spent on the labels matching in pod selector (kind of expected in this approach), I try to mitigate this effect by reducing interval of checking if pods are running to 20s (from 5s).

The memory usage grows by ~10% in this approach (compared to separate watches).

(TODO(mborsz): Paste results when done)

/hold (waiting for test results with 20s interval)
/kind feature
/assign @wojtek-t


clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -497,6 +551,32 @@ func (w *waitForControlledPodsRunningMeasurement) getObjectKeysAndMaxVersion() (
...
+	objects, err := w.indexer.ByIndex("metadata.namespace", w.namespace)
+	if err != nil {
+		// This should only happen if index isn't registered, which effectively means programmers error.
+		klog.Errorf("podIndexer.ByIndex failed: %v", err)
+		return nil
+	}
+	pods := []*v1.Pod{}
+	for _, o := range objects {
+		pod := o.(*v1.Pod)
+		if w.labelSelector.Matches(labels.Set(pod.Labels)) {
```
wojtek-t, line 570: If this is our main performance bottleneck in this approach, maybe we can do something smarter here?

Obviously can hack this around by hardcoding the additional index for "name" (which would address load test), but we shouldn't do that. But maybe we can do something better, given we have "wrapperPodStore" per deployment?

If it would be keeping some cache (say from RV-> [yet/no] saying if a given object matches our selector or not), would that give us something?
It would mean that "Matches" would be called exactly once for each object version, but would acceessing the this cache (hash-map) actually be faster? (and won't cause significant memory increase)?

Or taking a step further:
- let's stay it would be storing: "objects that were matching selector in the previous call and RV from that point"
- then to get what is currently matching selector we can basically:
  (a) the the items that were previously matching selector and check if they are still in the list
  (b) check if objects with RV > "that previous RV" are matching selector

That makes the cache here proportional to the size of result (which means that memory impact should definitely be fine) and the comparisons should also be faster...

mborsz: As discussed offline, I decided to use ownerReference as index key. The presubmit has passed (it works in 100 node scale). I'm running 5k node test now to see performance gain.

It think this PR is ready for review.

mborsz:
/hold cancel
/retest

I updated the PR description with 5k node test results.
The PR is ready for review.

clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
```diff
@@ -0,0 +1,135 @@
...
+		podsIndexer: podsInformer.Informer().GetIndexer(),
+		podsSynced:  podsInformer.Informer().HasSynced,
+		rsIndexer:   rsInformer.Informer().GetIndexer(),
+		rsSynced:    rsInformer.Informer().HasSynced,
+	}, nil
+}
+
+// WaitForCacheSync waits for all required informers to be initialized.
+func (p *ControlledPodsIndexer) WaitForCacheSync(ctx context.Context) bool {
+	return cache.WaitForNamedCacheSync("PodsIndexer", ctx.Done(), p.podsSynced, p.podsSynced)
```
wojtek-t, line 78: should be rsSynced as the last argument

clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
```diff
@@ -0,0 +1,135 @@
...
+func (p *ControlledPodsIndexer) PodsControlledBy(obj interface{}) ([]*corev1.Pod, error) {
+	metaAccessor, err := meta.Accessor(obj)
+	if err != nil {
+		return nil, fmt.Errorf("object has no meta: %w", err)
+	}
+	typeAccessor, err := meta.TypeAccessor(obj)
+	if err != nil {
+		return nil, fmt.Errorf("object has unknown type: %w", err)
+	}
+	res, err := p.appendPodsControlledBy(nil, metaAccessor.GetUID())
```
wojtek-t, line 91: In case of deployments this step is wasted.
Can you instead change it to sth like:
```
... 

var podOwners []string
switch typeAccessor.GetKind() {
case "Deployment":
  replicaSets, err := p.rsIndexer.ByIndex(...)
  if err != nil { ...}
  for _, rs := range replicaSets {
    podOwners = append(podOwners, rs.GetUID())
  }
default:
  podOwners = append(podOwners, metaAccessor.GetUID())
}

for _, owner := range podOwners {
  res, err := appendPodsControllerBy(res, owner)
  if err != nil {...}
}
```

clusterloader2/pkg/measurement/util/object_store.go
```diff
@@ -98,13 +99,53 @@ func NewPodStore(c clientset.Interface, selector *ObjectSelector) (*PodStore, er
...
+func (s *OwnerReferenceBasedPodStore) List() ([]*v1.Pod, error) {
+	pods, err := s.controlledPodsIndexer.PodsControlledBy(s.owner)
+	if err != nil {
+		return nil, err
+	}
+
+	// Using ownerReference as proxy for pods controlled by a given object should be pretty accurate,
+	// but technically the API is defined on labelSelectors.
+	// Let's check them to work properly in uncommon cases (like e.g. change of label selector in deployment).
+	// TODO(mborsz): Is it really needed?
```
wojtek-t, line 141: I don't think it's really needed - it can only speed up things a bit.

Technically what every controller is doing is when the pod changes (and it was previously controller by it) it will orphan it (by removing its ownerRef). So checking for it is only an optimization.

If you would like to be 100% accurate you would need to go over all other pods and see if others shouldn't be picked up (which controller didn't yet adopt). But again - it's an optimization as controller will eventually adopt them.

wojtek-t: I suggest adding a comment along the lines what I wrote above instead and getting rid of this code.

wojtek-t (commented):
I like this approach - just added two more technical comments.

clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
mborsz: Done

mborsz:
PTAL

wojtek-t:
/lgtm
/approve

/hold
@mborsz - cancel hold when you will be ready (not sure if you still wanted to run some larger scale tests)

mborsz:
/hold cancel

The test results in 5k are already in first comment.
/cc @jupblb

PR #2104, opened 2022-07-20
Cleanup of WaitForPods

mborsz:
Few changes:
* Rename WaitForPodsWithLister to WaitForPods
* Remove Selector from WaitForPodOptions (as in the current state it doesn't need to match podStore, which may be misleading), as it's used only for providing human readable identifier of pods (this is being moved to PodLister interface).
* Remove GetSelectorFromRuntimeObject which is no longer used anywhere

/assign @wojtek-t


wojtek-t:
/lgtm
/approve

Thanks!

PR #2107, opened 2022-07-25
Add sleep after deleting LBs in l4lb test

panslava:
**What type of PR is this?**
/kind bug

Add sleep to wait for LB resources to be cleaned up, before deleting 

In the current state, cluster gets deleted, and 10 minutes (default wait to clean up resources) is not enough to remove all LBs and Deployments, that is why tests are failing


clusterloader2/testing/l4lb/config.yaml
```diff
@@ -87,6 +90,12 @@ steps:
       objectTemplatePath: service.yaml
     - basename: lb-dep
       objectTemplatePath: dep.yaml
+- name: Sleep
```
marseel, line 93: As discussed offline, right now we can have sleep, but in the future we would like to have something like "WaitForServices" so we don't need to have this sleep.

PR #2121, opened 2022-08-22
Use rsIndexer instead of indexing manually

mborsz:
The goal is to simplify the code by reusing existing blocks: there already exists a rsIndexer that allows us to store objects along with some indexes for them (this is what is used by informers).

We can use here to avoid manual map operations.

/assign @tosi3k

### DIFF
--- a/clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
+++ b/clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"sync"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
```diff
@@ -253,8 +257,16 @@ func (p *ControlledPodsIndexer) PodsControlledBy(obj interface{}) ([]*corev1.Pod
...
+		}
+		for _, replicaSet := range replicaSets {
+			replicaSet, ok := replicaSet.(*appsv1.ReplicaSet)
+			if !ok {
+				return nil, fmt.Errorf("expected *appsv1.ReplicaSet; got: %T", replicaSet)
+			}
+			podOwners = append(podOwners, replicaSet.GetUID())
 		}
 		p.UIDLock.Unlock()
 	default:
```
tosi3k, line 272: QQ: does the `default` path would work for bare `ReplicaSet` objects controlled by CL2?

If not, could we open an issue for that and link it here in a comment for documentation's sake?

tosi3k (changes requested):

clusterloader2/pkg/measurement/util/controlled_pods_indexer.go
mborsz: I think it should -- the pods set the owner directly so we should be able to get them the same way we get for deployment (deployments resolves to resplicasets, replicasets to pods)

PR #2128, opened 2022-08-24
Remove artificial pod creation/update/delete rate limiting

mborsz:
Right now, the test tries to spread load to generate expected load on the control plane. It has been introduced when master was able to handle max 10 pods per second and was overloading with higher throughput.

Now we are in a state when it is able to handle 100 pods per second throughput (we have at least few manual runs with this PR, the metrics look comparable to regular runs).

Let's remove the artificial throttling to:
* make the test harder (it also validates property of k8s master that prevents overload)
* to make the test faster in happy case

The feature is controlled by CL2_RATE_LIMIT_POD_CREATION (default true) to keep default behavior.

We will gradually swap all tests to false and then remove the env var.

/assign @marseel

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -12,6 +12,7 @@
 {{$PODS_PER_NODE := DefaultParam .PODS_PER_NODE 30}}
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$DELETE_TEST_THROUGHPUT := DefaultParam .CL2_DELETE_TEST_THROUGHPUT $LOAD_TEST_THROUGHPUT}}
+{{$RATE_LIMIT_POD_CREATION := DefaultParam .CL2_RATE_LIMIT_POD_CREATION true}}
 {{$BIG_GROUP_SIZE := DefaultParam .BIG_GROUP_SIZE 250}}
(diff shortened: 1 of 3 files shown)


clusterloader2/testing/load/modules/reconcile-objects.yaml
```diff
@@ -75,7 +76,8 @@ steps:
       action: start
       checkIfPodsAreUpdated: {{$CHECK_IF_PODS_ARE_UPDATED}}
       labelSelector: group = load
-      operationTimeout: 15m
+      operationTimeout: {{$operationTimeout}}
+      exactOperationTimeout: true
```
marseel, line 80: I think we could move `exactOperationTimeout` to be CL2 flag/env that will change how WaitForPods works for whole test instead of particular measurements.

mborsz:
It looks like the change guarding doesn't work. The generated data leaks qps: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/2128/pull-perf-tests-clusterloader2/1562701334413578240/artifacts/generatedConfig_load.yaml

clusterloader2/testing/load/config.yaml
```diff
@@ -168,7 +169,13 @@ steps:
     params:
       actionName: "create"
       namespaces: {{$namespaces}}
+      {{if .RATE_LIMIT_POD_CREATION}}
```
mborsz, line 172: This should be {{if $RATE_LIMIT_POD_CREATION}}

PR #2129, opened 2022-08-24
Simplify scheduler throughput module

mborsz:
Few steps:
* Removes SCHEDULER_THROUGHPUT_SIZE as it's not set anywhere
* Assume we always create a single replica per namespace
* In clusters <= 1000 nodes use single deployment with 1000 pods
* In clusters > 1000 nodes, use single deployment with .Nodes pods
* We still allow to customize SCHEDULER_THROUGHPUT_PODS_PER_DEPLOYMENT as we use that in some tests
* Changes tuningsets to remove custom logic 

/assign @jupblb

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -19,9 +19,6 @@
 {{$MEDIUM_STATEFUL_SETS_PER_NAMESPACE := DefaultParam .MEDIUM_STATEFUL_SETS_PER_NAMESPACE 1}}
 {{$ENABLE_CHAOSMONKEY := DefaultParam .ENABLE_CHAOSMONKEY false}}
 {{$ENABLE_API_AVAILABILITY_MEASUREMENT := DefaultParam .CL2_ENABLE_API_AVAILABILITY_MEASUREMENT false}}
-{{$SCHEDULER_THROUGHPUT_SIZE := DefaultParam .CL2_SCHEDULER_THROUGHPUT_SIZE .Nodes}}
-# Determines number of pods per deployment. Should be a divider of $SCHEDULER_THROUGHPUT_SIZE.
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -61,14 +60,9 @@
 # TODO( https://github.com/kubernetes/perf-tests/issues/1027): Lower the number of "min-pods" once we fix the scheduler throughput measurement.
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) .Nodes}}
 {{$schedulerThroughputReplicasPerNamespace := 1}}
-{{$schedulerThroughputPodsPerDeployment := .Nodes}}
+{{$schedulerThroughputPodsPerDeployment := MinInt $totalSchedulerThroughputPods $SCHEDULER_THROUGHPUT_PODS_PER_DEPLOYMENT}}
```
jupblb, line 63: If `.Nodes < 1000` then `min(max(1000, .Nodes), .Nodes)` is `.Nodes`. If I understood PR description well, we should have 1,000 pods for clusters with less than 1,000 nodes. So we should probably have `MaxInt` here?

jupblb (changes requested):

clusterloader2/testing/load/config.yaml
```diff
@@ -61,14 +60,9 @@
 # TODO( https://github.com/kubernetes/perf-tests/issues/1027): Lower the number of "min-pods" once we fix the scheduler throughput measurement.
 {{$totalSchedulerThroughputPods := MaxInt (MultiplyInt 2 $MIN_PODS_IN_SMALL_CLUSTERS) .Nodes}}
 {{$schedulerThroughputReplicasPerNamespace := 1}}
```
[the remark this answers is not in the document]
mborsz: Removed

clusterloader2/testing/load/config.yaml
mborsz: Fixed.

mborsz:
PTAL

PR #2130, opened 2022-08-25
Simplify ControlledPodsIndexer

mborsz:
We don't have to count pods manually, we can try to use podIndexer for this purpose.

/assign @tosi3k

### DIFF
--- a/clusterloader2/pkg/measurement/util/controlled_pods_indexer_test.go
+++ b/clusterloader2/pkg/measurement/util/controlled_pods_indexer_test.go
@@ -311,3 +311,55 @@ func TestControlledPodsIndexer_PodsControlledBy_PodUpdate(t *testing.T) {
 
 	assert.ElementsMatch(t, got, want)
 }
+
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/util/controlled_pods_indexer_test.go
```diff
@@ -292,3 +292,57 @@ func TestControlledPodsIndexer_PodsControlledBy_PodUpdate(t *testing.T) {
...
+	time.Sleep(1 * time.Second)
+
+	want := []*corev1.Pod{}
+	got, err := p.PodsControlledBy(deployment)
+	if err != nil {
+		t.Errorf("PodsIndexer.PodsControlledBy() error = %v, wantErr %v", err, nil)
+		return
+	}
+
+	if !equality.Semantic.DeepEqual(got, want) {
```
tosi3k, line 336: Please change to `assert.ElementsMatch(got, want)` as we removed importing of `equality` package in #2136.

PR #2134, opened 2022-08-25
Remove artificial pod creation/update/delete rate limiting #2

mborsz:
Roll forward https://github.com/kubernetes/perf-tests/pull/2128, this time with fixed change guarding.

/assign @marseel 
/hold for me to check generatedConfig this time.

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -73,6 +73,7 @@
 {{$SLEEP_AFTER_EXEC_DURATION := DefaultParam .CL2_SLEEP_AFTER_EXEC_DURATION "0s"}}
 
 {{$latencyPodImage := DefaultParam .CL2_LATENCY_POD_IMAGE "k8s.gcr.io/pause:3.1"}}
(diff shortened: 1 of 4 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -315,7 +333,7 @@ steps:
     path: /modules/configmaps-secrets.yaml
     params:
       actionName: delete
-      tuningSet: Global100qps
+      tuningSet: default
```
marseel, line 336: Doesn't it simplify test for 100 nodes?

mborsz: As discussed offline, it's true but the point of this part of the test isn't to stress any specific dimension or kube-apiserver, rather to mitigate https://github.com/kubernetes/kubernetes/issues/96635

mborsz:
Diff for generatedConfig_load.yaml: https://gist.github.com/mborsz/d6539bdc32dad389a63a143d8bba668b

It's mainly tuningset rename + changing qps value, which is expected and considered to be safe.

/hold cancel

PR #2138, opened 2022-09-01
Fall back to using ADC when setting up GCE PD CSI driver in CL2

tosi3k:
So far, it wasn't possible to set up GCE PD CSI driver in CL2 without providing service account credentials to the driver's controller.

This PR adds an option to reuse Application Default Credentials if `GOOGLE_APPLICATION_CREDENTIALS` env var is not set during CL2 execution.

/assign @jkaniuk

### DIFF
--- a/clusterloader2/run-e2e.sh
+++ b/clusterloader2/run-e2e.sh
@@ -29,14 +29,36 @@ export AZUREDISK_CSI_DRIVER_VERSION="v1.15.0"
 export AZUREDISK_CSI_DRIVER_INSTALL_URL="https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/${AZUREDISK_CSI_DRIVER_VERSION}/deploy/install-driver.sh"
 export WINDOWS_USE_HOST_PROCESS_CONTAINERS=true
 
-# Deploy the GCP PD CSI Driver if required
+# Deploy the GCE PD CSI Driver if required
 if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
-   if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
-      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
-      exit 1
+   if [[ -n "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
+      kubectl --kubeconfig "${CSI_DRIVER_KUBECONFIG}" apply -f "${CLUSTERLOADER_ROOT}"/drivers/gcp-csi-driver-stable.yaml
+      kubectl --kubeconfig "${CSI_DRIVER_KUBECONFIG}" create secret generic cloud-sa --from-file=cloud-sa.json="${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" -n gce-pd-csi-driver
+   else
+      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS is unset."
+      echo "Falling back to using Application Default Credentials for GCE PD CSI driver deployment."
+      if [[ ! -x "$(command -v yq)" ]]; then
+         echo "yq must be installed to set up GCE PD CSI Driver with Application Default Credentials."
+         echo "Please install this tool from https://github.com/mikefarah/yq."
+         exit 1
+      fi
+      # Running yq to patch GCE PD CSI Driver manifests so that it runs using
+      # Application Default Credentials instead of creating service account
+      # keys, hence avoiding security risks. See
+      # https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/610
+      # for more details.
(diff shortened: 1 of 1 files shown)


tosi3k:
/hold

I want to test it end-to-end before it merges.

clusterloader2/run-e2e.sh
```diff
@@ -29,14 +29,27 @@ export AZUREDISK_CSI_DRIVER_VERSION="v1.15.0"
...
 if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
    if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
-      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
-      exit 1
+      # Running yq to patch GCE PD CSI Driver manifests so that it runs using
+      # Application Default Credentials instead of creating service account
+      # keys, hence avoiding security risks. See
+      # https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/610
+      # for more details.
+      yq eval -i '
```
jkaniuk, line 40: Do we have any guarantees that there is yq? What if there is not?

clusterloader2/run-e2e.sh
```diff
@@ -29,14 +29,27 @@ export AZUREDISK_CSI_DRIVER_VERSION="v1.15.0"
...
+         with(select(.metadata.name == "csi-gce-pd-controller" and .kind == "Deployment").spec.template.spec;
+            del(.volumes[] | select(.name == "cloud-sa-volume")) |
+            with(.containers[] | select(.name == "gce-pd-driver");
+               del(.env[] | select(.name == "GOOGLE_APPLICATION_CREDENTIALS")) |
+               del(.volumeMounts[] | select(.name == "cloud-sa-volume"))
+            )
+         )' "${CLUSTERLOADER_ROOT}"/drivers/gcp-csi-driver-stable.yaml
```
jkaniuk, line 47: It should create a copy, not modify in place. In subsequent run, CL2 would behave differently with in-place modification.

clusterloader2/run-e2e.sh
tosi3k: There is, if one runs CL2 from `kubekins-e2e`: https://github.com/kubernetes/test-infra/blob/ed307a9f6b5fe93c4004e4a2d803fa791c9318c0/images/kubekins-e2e/Dockerfile#L73-L76

Added a precautionary `if` statement with instruction on how to mitigate this.

clusterloader2/run-e2e.sh
tosi3k: Good point; done.

clusterloader2/run-e2e.sh
```diff
@@ -29,14 +29,34 @@ export AZUREDISK_CSI_DRIVER_VERSION="v1.15.0"
 export AZUREDISK_CSI_DRIVER_INSTALL_URL="https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/${AZUREDISK_CSI_DRIVER_VERSION}/deploy/install-driver.sh"
 export WINDOWS_USE_HOST_PROCESS_CONTAINERS=true
 
-# Deploy the GCP PD CSI Driver if required
+# Deploy the GCE PD CSI Driver if required
 if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
    if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
-      echo "Env var E2E_GOOGLE_APPLICATION_CREDENTIALS must be set to deploy driver"
```
jkaniuk, line 35: We've lost this old message about application credentials. Is that not relevant anymore?

jkaniuk: Also - is there any way to distinguish between the case when user forgot credentials and when forgot yq? Current message could be quite confusing.

clusterloader2/run-e2e.sh
```diff
@@ -29,14 +29,34 @@ export AZUREDISK_CSI_DRIVER_VERSION="v1.15.0"
 export AZUREDISK_CSI_DRIVER_INSTALL_URL="https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/${AZUREDISK_CSI_DRIVER_VERSION}/deploy/install-driver.sh"
 export WINDOWS_USE_HOST_PROCESS_CONTAINERS=true
 
-# Deploy the GCP PD CSI Driver if required
+# Deploy the GCE PD CSI Driver if required
 if [[ "${DEPLOY_GCI_DRIVER:-false}" == "true" ]]; then
    if [[ -z "${E2E_GOOGLE_APPLICATION_CREDENTIALS:-}" ]]; then
```
jkaniuk, line 34: I would reverse E2E_GOOGLE_APPLICATION_CREDENTIALS check - first case with credentials, then longer case of yq

clusterloader2/run-e2e.sh
tosi3k: Added informative message about the lack of credentials.

PR #2140, opened 2022-09-08
Add DNS lookup error percentage threshold instead of error count

dlapcevic:
Changes:
1. Modify generic_query_measurement to support PromQL queries that have multiple duration placeholders `[%v]`.
2. Add DNS lookup error percentage query to the advanced DNS test measurement in load test config, and move the threshold from DNS lookup error count to the DNS lookup error percentage query, with 0.1% threshold, meaning that there must be at least 99.9% of successful DNS lookups.

/kind feature
/assign @marseel

### DIFF
--- a/clusterloader2/pkg/measurement/common/generic_query_measurement.go
+++ b/clusterloader2/pkg/measurement/common/generic_query_measurement.go
@@ -19,6 +19,7 @@ package common
 import (
 	goerrors "errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/generic_query_measurement.go
```diff
@@ -182,7 +183,8 @@ func (g *genericQueryGatherer) String() string {
 
 func (g *genericQueryGatherer) query(q GenericQuery, executor QueryExecutor, startTime, endTime time.Time) ([]*model.Sample, error) {
 	duration := endTime.Sub(startTime)
-	boundedQuery := fmt.Sprintf(q.Query, measurementutil.ToPrometheusTime(duration))
+	// Replace all provided duration placeholders (%v) with the test duration.
+	boundedQuery := strings.ReplaceAll(q.Query, "%v", fmt.Sprintf("%v", measurementutil.ToPrometheusTime(duration)))
```
marseel, line 187: I think it could be  `measurementutil.ToPrometheusTime(duration)` instead of `fmt.Sprintf("%v", measurementutil.ToPrometheusTime(duration))`?

dlapcevic: You are right. The `ToPrometheusTime` returns a string. Thanks for point it out.

I made the change.

PR #2147, opened 2022-10-03
Use DNS performance test and metrics in huge-services

dlapcevic:
Changes:

1. In `load` test, create a module `dns-performance-metrics.yaml` out of `DNSPerformanceMetrics` measurement.
2. Use `load` test modules `dns-k8s-hostnames.yaml` and `dns-performance-metrics.yaml` in the `huge-service` test.

/kind feature
/assign @jupblb


jupblb:
Note for the future (not actionable): we should probably have some another directory for common test modules.

/lgtm
/assign @marseel

PR #2157, opened 2022-10-12
Fix presubmits

jupblb:
Fix presubmits
Example run: https://pastebin.com/vmP3vRCd

/assign @marseel


marseel:
As you are deleting compare directory and also making some other changes, can you please split it into two separate commits or PRs?

jupblb:
Split into two separate commits

PR #2161, opened 2022-10-12
Updated the go.mod with version 0.23.3

Ganiredi:
**What type of PR is this?**

Changed the go.mod client go version with v0.23.3 and able to build the binary.

Right now in the k8s upstream using only v0.20.12 and we are trying to test against the version k8s 1.22 and 1.23 which is getting failed ,but this build will be passed with the versions that were updated

Fixes #

```release-note

```

```docs

```


mborsz:
I don't think we are ready for 0.23 as it require golang 1.17.

We still have environments where we use golang 1.16. We can bump client to 0.22 which works with golang 1.16 if you want.

clusterloader2/go.mod
```diff
@@ -3,61 +3,80 @@ module k8s.io/perf-tests/clusterloader2
...
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.20.12
-	k8s.io/metrics => k8s.io/metrics v0.20.12
-	k8s.io/mount-utils => k8s.io/mount-utils v0.20.12
-	k8s.io/node-api => k8s.io/node-api v0.20.12
-	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.20.12
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.20.12
-	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.20.12
-	k8s.io/sample-controller => k8s.io/sample-controller v0.20.12
-)
+replace k8s.io/api => k8s.io/api v0.23.3
```
mborsz, line 6: Please keep existing structure (single replace)

Ganiredi: @mborsz I have added a new commit

hakuna-matatah:
> > I don't think we are ready for 0.23 as it require golang 1.17.
> > We still have environments where we use golang 1.16. We can bump client to 0.22 which works with golang 1.16 if you want.
> 
> @mborsz can you please share where golang 1.16 is being used (Kubernetes 1.24 has moved to 1.18 Go) and if we can upgrade to 1.17 atleast?

 1.22  k8s is on 1.16 @prateekgogia.

mengqiy:
>I don't think we are ready for 0.23 as it require golang 1.17.

>We still have environments where we use golang 1.16. We can bump client to 0.22 which works with golang 1.16 if you want.

@mborsz I don't quite understand why there is a hard requirement for CL2. As long as the clusterloader CLI tool is using a compatible client-go version to talk the clusters, the golang runtime should not really be a concern here. Is version skew your concern? What is your concern?

mborsz:
> > I don't think we are ready for 0.23 as it require golang 1.17.
> 
> > We still have environments where we use golang 1.16. We can bump client to 0.22 which works with golang 1.16 if you want.
> 
> Maciej Borsz I don't quite understand why there is a golang runtime hard requirement for CL2. As long as the clusterloader CLI tool is using a compatible client-go version to talk the clusters under test, the golang version that compiles clusterloader CLI should not really be a concern here. Is version skew your concern? What is your concern?

The problem is that we have some google-internal tests that are still using kubekins for 1.22 (golang 1.16) and perf-tests master branch to build CL2. If we update client-go in CL2 master to version of client-go that requires 1.17, there is a chance that the build will break (it was happening in the past).

I'm not the fan of this approach either and we are slowly migrating out of it, but frankly speaking we are not there yet.

mborsz:
/lgtm
/approve

Thanks!

mengqiy:
@Ganiredi The PR title and description don't reflect what has been done in this PR (e.g. version number).
Also this PR has been reverted in https://github.com/kubernetes/perf-tests/pull/2162, since it caused SIGSEGV.
There is a chance that this problem has been fixed in latest patch version of client-go `v0.22.15`. Please ensure you use the latest patch version next time.

mborsz:
Let's try v0.22.15. It seems to be likely that this will contain some fix.

@Ganiredi can you prepare a PR for v0.22.15?

Ganiredi:
@mborsz  Will prepare a new PR using latest patch version.

PR #2163, opened 2022-10-18
Add request-benchmark CL2 scenario

jupblb:
**What type of PR is this?**
/kind feature

Followup to https://github.com/kubernetes/perf-tests/pull/2156

N/A

In a single run the benchmark may create a number of config maps and perform GET/LIST operations on them.

/assign @mborsz 
/assign @marseel

### DIFF
--- a/clusterloader2/testing/request-benchmark/config.yaml
+++ b/clusterloader2/testing/request-benchmark/config.yaml
@@ -1,0 +1,72 @@
+# Request benchmark
+{{$configMapBytes := DefaultParam .CL2_BENCHMARK_CONFIG_MAP_BYTES 1000}}
+{{$configMapGroup := DefaultParam .CL2_BENCHMARK_CONFIG_MAP_GROUP "benchmark-config-map"}}
+{{$configMapNumber := DefaultParam .CL2_BENCHMARK_CONFIG_MAP_NUMBER 1}}
+{{$benchmarkReplicas := DefaultParam .CL2_BENCHMARK_PODS 1}}
+
+{{$inflight := DefaultParam .CL2_BENCHMARK_INFLIGHT 10}}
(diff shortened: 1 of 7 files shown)


clusterloader2/testing/request-benchmark/config.yaml
```diff
@@ -0,0 +1,72 @@
...
+    objectBundle:
+    - basename: {{$configMapGroup}}
+      objectTemplatePath: configmap.yaml
+      params:
+        bytes: {{$configMapBytes}}
+        group: {{$configMapGroup}}
+- module:
+    path: modules/benchmark-deployment.yaml
+    params:
+      inflight: {{$inflight}}
```
mborsz, line 52: Nit: reorder fields? "pods" is a different category (how many pods) than "uri" and "inflight" (what each pod should be doing)?. Also maybe some other name to make it more clear? "replicas"?

mborsz: +1 to "replicas" as it's the name used later in the spec.

mborsz (commented):
Just one nit

clusterloader2/testing/request-benchmark/modules/measurements.yaml
```diff
@@ -0,0 +1,25 @@
...
+    Params:
+      action: {{$action}}
+      metricName: Container CPU
+      metricVersion: v1
+      unit: cores
+      dimensions:
+      - container
+      queries:
+      - name: Perc99
+        query: quantile_over_time(0.99, sum by (container) (rate(container_cpu_usage_seconds_total[1m]))[%v:])
```
mborsz, line 21: The open question (we can address it here or in the next iteration [e.g. when the test is already running]):

- What should be a desired unit of result? Now you return CPU/s, Personally I would vote for "CPU/s per inflight" as this is the number we are looking for.

PR #2166, opened 2022-10-19
support building multi-arch image for perfdash

mengqiy:
Perfdash currently only release amd64 image.
This PR adds support for building multi-arch images for perfdash using docker buildx.
It follows https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/


mengqiy:
It's not clear to me what the release process for perdash container image. I don't find any GitHub action for it.

PR #2168, opened 2022-10-20
Extend benchmark duration

jupblb:
#### What type of PR is this?
/kind bug

We noticed that variance of different indicators is higher at the beginning of the test so this PR adds 1m before the CPU measurement starts so that we are sure this does not influence results.

N/A

/assign @mborsz 
/assign @marseel

### DIFF
--- a/clusterloader2/testing/request-benchmark/config.yaml
+++ b/clusterloader2/testing/request-benchmark/config.yaml
@@ -47,6 +47,10 @@ steps:
         bytes: {{$configMapBytes}}
         group: {{$configMapGroup}}
 - module:
+    path: modules/measurements.yaml
+    params:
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/request-benchmark/config.yaml
```diff
@@ -46,6 +46,10 @@ steps:
       params:
         bytes: {{$configMapBytes}}
         group: {{$configMapGroup}}
+- module:
+    path: modules/measurements.yaml
+    params:
+      name: noise
```
marseel, line 52: baseline?

jupblb: Fixed.

PR #2172, opened 2022-10-26
Add stringer for pods status

Argh4k:
#### What type of PR is this?

/kind feature

Add stringer interface for pods status to improve logging output readability

Fixes #


clusterloader2/pkg/measurement/util/pods.go
```diff
@@ -23,6 +23,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+//go:generate stringer -type=status
```
mborsz, line 26: How does it work, does it require some tooling to regenerate?

Argh4k: https://pkg.go.dev/golang.org/x/tools/cmd/stringer 
This can be regenerated via:
`go generate` or `stringer -type=status` in utils directory when you have stringer tool installed.

mborsz: I’m afraid that lack of boilerplate will block this approach. Each file in the repo needs the header with some legal stuff. can we add it here?

PR #2174, opened 2022-11-03
migrate to klog v2

aojea:
/kind cleanup

get rid of klog v1 since was deprecated and polluting the go mod , it also can cause that some logging doesn't work

It seems my vscode linter/formatted is doing some change too, but is on comments , that should be fine

### DIFF
--- a/clusterloader2/go.mod
+++ b/clusterloader2/go.mod
@@ -55,8 +55,7 @@ require (
 	k8s.io/client-go v0.22.15
 	k8s.io/component-base v0.22.15
 	k8s.io/component-helpers v0.22.15
-	k8s.io/klog v1.0.0
-	k8s.io/klog/v2 v2.30.0 // indirect
+	k8s.io/klog/v2 v2.30.0
 	k8s.io/kubelet v0.22.15
 	k8s.io/kubernetes v1.22.15
 )
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
@@ -29,7 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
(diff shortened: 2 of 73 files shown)


clusterloader2/go.mod
```diff
@@ -55,8 +55,7 @@ require (
 	k8s.io/client-go v0.22.15
 	k8s.io/component-base v0.22.15
 	k8s.io/component-helpers v0.22.15
-	k8s.io/klog v1.0.0
-	k8s.io/klog/v2 v2.30.0 // indirect
```
aojea, line 59: this is the main thing

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
```diff
@@ -40,9 +40,11 @@ const (
 // ReplicasWatcher is a struct that allows to check a number of replicas at a given time.
 // Usage:
 // var rw ReplicasWatcher = (...)
-// if err := rw.Start(stopCh); err != nil {
-//   panic(err);
-// }
+//
+//	if err := rw.Start(stopCh); err != nil {
```
mborsz, line 44: I'm not sure if I understand this change :)

this will result in code snipped with weird formatting (some part indentend, some not).

mborsz:
Thanks for the PR!

One question for the comment style. Other than LGTM

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
marseel: Probably some automatic formatting in vs code or sth similar :)

wojtek-t: I was thinking about that too, but in theory var can be defined somewhere else (e.g. this snippet can be inside some nested statement)
So I think it's fine :)

[Although I don't have strong opinion here :) ]

mborsz:
/hold cancel

I'm fine with @wojtek-t reasoning :)

clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
aojea: vscode magic, somehow some code has tabulations

Issue #2176, opened 2022-11-03
Pod startup latency calculation phases are not efficient and may result in error for larger clusters in ClusterLoader2


hakuna-matatah:
**What happened**:

This line [here](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go#L285) results in throwing server side errors specially for larger load tests when events can't be listed from APIServer in one minute, reason being we have a default server side TTL for a request to be 60seconds:

```

-request-timeout duration     Default: 1m0s
--
  | An optional field indicating the duration a handler must keep a request open before timing it out. This is the default request timeout for requests but may be overridden by flags such as --min-request-timeout for specific types of requests.

```

**What you expected to happen**:
Have a client-side mechanism that should be able to wait more than a minute to sync up all the events to not run into server side ttl limit.

**How to reproduce it (as minimally and precisely as possible)**:
On a heavy clusters like 5k and above with many events, this can be reproduced when listing all events at once.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.23
- Cloud provider or hardware configuration: aws
- OS (e.g: `cat /etc/os-release`): amazon-linux
- Kernel (e.g. `uname -a`):  N/A
- Install tools: N/A
- Network plugin and version (if this is a network-related bug): 
- Others: N/A

Issue #2006, opened 2022-03-07
Pod startup time phases are inaccurate in longer tests.


wojtek-t:
gatherScheduleTimes:
https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go#L268
is currently listing events at the end of the test.

Given that events by default have 1h TTL, for measurements across longer periods it just returns incomplete results. Given that we don't 100% accuracy, we should switch to a mechanism that is similar to the one that slo-monitor is using (added in #1477 ).

harshanarayana:
@wojtek-t  Is someone working on this still or can I take a stab at this ?

PR #2178, opened 2022-11-03
use informers for pod events instead of Listing

hakuna-matatah:
#### What type of PR is this?
/kind bug

/kind failing-test

It's a quick fix to help calculate pod_startup_latencies effectively for large clusters and not worry about running into Apiserver side ttl issues nor worry about events being expired after 1h for larger clusters.

Fixes #

It fixes these issues 
- https://github.com/kubernetes/perf-tests/issues/2176
- https://github.com/kubernetes/perf-tests/issues/2006

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -61,6 +61,7 @@ func createPodStartupLatencyMeasurement() measurement.Measurement {
 		podStartupEntries: measurementutil.NewObjectTransitionTimes(podStartupLatencyMeasurementName),
 		podMetadata:       measurementutil.NewPodsMetadata(podStartupLatencyMeasurementName),
 		eventQueue:        workqueue.New(),
+		schedEventQueue:   workqueue.New(),
 	}
 }
 
@@ -70,12 +71,14 @@ type eventData struct {
 }
 
 type podStartupLatencyMeasurement struct {
-	selector  *util.ObjectSelector
-	isRunning bool
-	stopCh    chan struct{}
+	selector    *util.ObjectSelector
+	isRunning   bool
+	stopCh      chan struct{}
+	stopSchedCh chan struct{}
 	// This queue can potentially grow indefinitely, beacause we put all changes here.
 	// Usually it's not recommended pattern, but we need it for measuring PodStartupLatency.
 	eventQueue        *workqueue.Type
+	schedEventQueue   *workqueue.Type
 	podStartupEntries *measurementutil.ObjectTransitionTimes
 	podMetadata       *measurementutil.PodsMetadata
 	threshold         time.Duration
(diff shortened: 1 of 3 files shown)


mengqiy:
@hakuna-matatah Your commit is not associated with your account properly. I guess you need to amend it and push again.

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -146,6 +148,29 @@ func (p *podStartupLatencyMeasurement) start(c clientset.Interface) error {
...
 	go p.processEvents()
+
+	selector := fields.Set{
+		"involvedObject.kind": "Pod",
+		"source":              schedulerName,
+	}.AsSelector().String()
+
+	p.stopSchedCh = make(chan struct{})
+
+	e := informer.NewInformer(
```
tosi3k, line 159: Could we use [`Controller`](https://github.com/kubernetes/client-go/blob/cbe28cf468c43969d2c170e079f556becfc7adb7/tools/cache/controller.go#L116-L123), a slightly lower-lever primitive than `Informer` and pass a large (e.g. 10k) `WatchListPageSize` in the [Config](https://github.com/kubernetes/client-go/blob/cbe28cf468c43969d2c170e079f556becfc7adb7/tools/cache/controller.go#L39-L79)?

The reason I'm asking this is that in large clusters there's a tendency to have O(hundreds of thousands) events and listing them using default page size ([500](https://github.com/kubernetes/client-go/blob/ec0f33729df006747d8685920630bacb364bf3bc/tools/pager/pager.go#L31)) may result in informer's initial list getting timed out.

tosi3k (changes requested):

clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
hakuna-matatah: >listing them using default page size ([500](https://github.com/kubernetes/client-go/blob/ec0f33729df006747d8685920630bacb364bf3bc/tools/pager/pager.go#L31)) may result in informer's initial list getting timed out.

Oh! how will it timeout  IIUC ? I have ensured we are not relying on client side timeout defined [here](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/util/informer/informer.go#L84-L91) for this use-case,  instead I'm calling directly  `Run` method here [here](https://github.com/kubernetes/perf-tests/blob/b7234715ccb6626f7e9b92cba5e89204e083728b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go#L173) . Am i misinterpreting what you are trying to imply here ?

tosi3k: Ah, I see... we deliberately set the timeout in our wrappers around informers to make sure that the initial list (the one responsible for the cache's sync) completes in a reasonable time. In a clusters with O(xxx k) events this will take ages (O(a few minutes)) if using the default page size and because of that possibly run into the "too old resource version" during the initial list. I'd strongly suggest using the larger page size and hence the `Controller` primitive for the List+Watch pattern for events like we do in https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/loadbalancer_nodesync_latency.go.

CC @mborsz for his thoughts as I'll be OOO for the rest of the week.

PR #2179, opened 2022-11-03
Disable resource usage measurement for Autopilot provider

tosi3k:
Currently, to get the resource usage of containers, the resource usage measurement queries `nodes/proxy` subresource underneath to access the `/stats/summary` endpoint of kubelets. Autopilot provider doesn't permit the user to access that subresource at all.

/assign @marseel

### DIFF
--- a/clusterloader2/pkg/measurement/common/resource_usage.go
+++ b/clusterloader2/pkg/measurement/common/resource_usage.go
@@ -58,6 +58,12 @@ type resourceUsageMetricMeasurement struct {
 // - start - Starts resource metrics collecting.
 // - gather - Gathers and prints current resource usage metrics.
 func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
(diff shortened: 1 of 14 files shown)


clusterloader2/pkg/measurement/common/resource_usage.go
```diff
@@ -58,6 +59,12 @@ type resourceUsageMetricMeasurement struct {
 // - start - Starts resource metrics collecting.
 // - gather - Gathers and prints current resource usage metrics.
 func (e *resourceUsageMetricMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
+	cloudProvider := config.ClusterFramework.GetClusterConfig().Provider
+	if cloudProvider.Name() == provider.AutopilotName {
```
marseel, line 63: Could you do it as cloud provider feature https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/provider/provider.go#L36:6 and not filter it by provider name, please?

tosi3k: Done.

PR #2181, opened 2022-11-04
bundle: Add mutex profile for components other than etcd

MadhavJivrajani:
#### What type of PR is this?

/kind cleanup

This PR enables mutex profiles for components other than etcd.

Fixes #


marseel:
Thanks for contribution!

/lgtm
/approve

PR #2182, opened 2022-11-07
kube-proxy iptables partial restore failures

aojea:
kube-proxy implemented a new feature to do only partial restore of iptables rules to improve its performance.

The feature adds a new metrics that signals the operator if something goes wrong during the partial resttore. If this metrics is different than zero it most probably a bug in the code.

/kind feature

Ref: https://github.com/kubernetes/enhancements/pull/3454/files#r979617744

### DIFF
--- a/clusterloader2/testing/load/modules/measurements.yaml
+++ b/clusterloader2/testing/load/modules/measurements.yaml
@@ -92,6 +92,18 @@ steps:
     Method: NetworkProgrammingLatency
     Params:
       action: {{$action}}
+  - Identifier: Kube-proxy partial iptables restore failures
+    Method: GenericPrometheusQuery
+    Params:
+      action: {{$action}}
+      metricName: KubeProxyIptablesRestoreFailures
+      metricVersion: v1alpha1
+      unit: failures
+      queries:
+      - name: Total
+        query: sum(kubeproxy_sync_proxy_rules_iptables_partial_restore_failures_total)
+        requireSamples: false # It is a feature gate and may not be enabled
+        threshold: 0
 {{end}}
 {{if $PROMETHEUS_SCRAPE_KUBE_STATE_METRICS}}
   - Identifier: KubeStateMetricsLatency


aojea:
/hold

I need help with the unit test, I don't know if the query is wrong or what happen but the executor in the test doesn't get samples.
However,  I can do `sum(kubeproxy_sync_proxy_rules_iptables_partial_restore_failures_total)`in a local cluster to obtain the total of that metrics. It is also documented in the prometheus documentation

/assign @wojtek-t @marseel 
/cc @danwinship

clusterloader2/pkg/measurement/common/kube_proxy_iptables_restore_failures.go
```diff
@@ -0,0 +1,100 @@
...
+
+const (
+	kubeProxyIptablesFailuresName = "KubeProxyIptablesRestoreFailures"
+	// The metric definition
+	// https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/metrics/metrics.go#L129-L138
+	// IptablesPartialRestoreFailuresTotal is the number of iptables *partial* restore
+	// failures (resulting in a fall back to a full restore) that the proxy has seen.
+	// This should not happen unless there is a kube-proxy error on the code
+	// Ref: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3453-minimize-iptables-restore
+	kubeProxyIptablesFailuresQuery = `sum(kubeproxy_sync_proxy_rules_iptables_partial_restore_failures_total)`
```
aojea, line 37: my prometheus knowledge is very minimal, maybe this is the wrong query?

clusterloader2/pkg/measurement/common/kube_proxy_iptables_restore_failures.go
```diff
@@ -0,0 +1,100 @@
...
+	return summaries, nil
+}
+
+func (c *kubeProxyIptablesFailuresGatherer) gatherKproxyIptablesFailures(executor QueryExecutor, startTime, endTime time.Time) (*iptablesRestoreFailures, error) {
+	measurementEnd := time.Now()
+	samples, err := executor.Query(kubeProxyIptablesFailuresQuery, measurementEnd)
+	if err != nil {
+		return nil, err
+	}
+	if len(samples) != 1 {
```
aojea, line 79: if I do a sum how many samples should I expect?

marseel:
Could you take a look, please?
/assign @Argh4k

clusterloader2/pkg/measurement/common/kube_proxy_iptables_restore_failures.go
```diff
@@ -0,0 +1,100 @@
...
+
+	summaries := []measurement.Summary{measurement.CreateSummary(kubeProxyIptablesFailuresName, "json", content)}
+	if iptablesFailures.Total > 0 {
+		return summaries, errors.NewMetricViolationError("kubeproxy_sync_proxy_rules_iptables_partial_restore_failures_total", fmt.Sprintf("iptables restore failures greater than 0, got: %d", iptablesFailures.Total))
+	}
+	return summaries, nil
+}
+
+func (c *kubeProxyIptablesFailuresGatherer) gatherKproxyIptablesFailures(executor QueryExecutor, startTime, endTime time.Time) (*iptablesRestoreFailures, error) {
+	measurementEnd := time.Now()
```
Argh4k, line 74: It seems that you are using time.Now() as endTime, instead of endTime passed via argument. This probably breaks tests as data loaded by lazyloader starts at time.Unix(0, 0). I can see that in test you have properly passed start and end time, but you are not respecting it here

aojea: copy paste thing

aojea: same error :/ after removing this line

danwinship:
I don't know prometheus well, but I'll just point out that you also need to set the `MinimizeIPTablesRestore` feature gate when starting kube-proxy or the metric is guaranteed to be 0 anyway. (The metric itself is not feature-gated, so it will exist, it's just that it would be never be non-0.)

aojea:
> I don't know prometheus well, but I'll just point out that you also need to set the `MinimizeIPTablesRestore` feature gate when starting kube-proxy or the metric is guaranteed to be 0 anyway. (The metric itself is not feature-gated, so it will exist, it's just that it would be never be non-0.)

/hold

ah, good point, how can I do so this test only runs when this featuregate is set?

clusterloader2/pkg/measurement/common/kube_proxy_iptables_restore_failures.go
```diff
@@ -0,0 +1,99 @@
+/*
```
mborsz, line 1: Do we need to create dedicated measurement for this metric? If you want to simply gather value, show in perf dash, fail if exceeds threshold, then probably Generic Prometheus measurement is a way to go.

mborsz: https://github.com/kubernetes/perf-tests/blob/34b4a7a772ea2293a82ae13a1dc29fe19bd439cd/clusterloader2/testing/load/modules/measurements.yaml#L162

aojea: I was just following the docs, no special need AFAIK ... I really think you'll know better ... the goal is to fail if that metrics is > 0 ...

aojea: done

aojea:
> Well, I was more concerned in the other direction; you need to set the feature gate in the test run before this test is even useful.

I wasn't thinking that further, I just wanted to have a test that only fail when it should ... now that we have that we have to figure out how to set the feature gate on the jobs 😄

mborsz:
> > Well, I was more concerned in the other direction; you need to set the feature gate in the test run before this test is even useful.
> 
> I wasn't thinking that further, I just wanted to have a test that only fail when it should ... now that we have that we have to figure out how to set the feature gate on the jobs 😄

If you want to run one-off, then probably approach like in https://github.com/kubernetes/kubernetes/pull/113695 is prefered -- you open PR with artificial Feature Gate change and run presubmit there

If you want to run periodics, you can set KUBE_FEATURE_GATES var for CI jobs: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L133

aojea:
> If you want to run periodics, you can set KUBE_FEATURE_GATES var for CI jobs: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml#L133

I think that in the KEP we kind of agree to run periodics to we can have enough signal for graduation

aojea:
is the error related? I can't see how

```
F1108 15:01:44.087597   29265 clusterloader.go:303] Cluster verification error: no schedulable nodes in the cluster
2022/11/08 15:01:44 process.go:155: Step '/home/prow/go/src/k8s.io/perf-tests/run-e2e.sh cluster-loader2 --nodes=100 --prometheus-scrape-node-exporter --provider=kubemark --report-dir=/logs/artifacts --testconfig=testing/load/config.yaml --testconfig=testing/huge-service/config.yaml --testconfig=testing/access-tokens/config.yaml --testoverrides=./testing/experiments/enable_restart_count_check.yaml --testoverrides=./testing/experiments/use_simple_latency_query.yaml --testoverrides=./testing/overrides/kubemark_load_throughput.yaml' finished in 3m33.769282639s
2022/11/08 15:01:44 e2e.go:776: Dumping logs for kubemark master to GCS directly at path: gs://sig-scalability-logs/pull-perf-tests-clusterloader2-kubemark/1589991412294750208
2022/11/08 15:01:44 process.go:153: Running: /workspace/log-dump.sh /logs/artifacts gs://sig-scalability-logs/pull-perf-tests-clusterloader2-kubemark/1589991412294750208
Checking for custom logdump instances, if any
Using gce provider, skipping check for LOG_DUMP_SSH_KEY and LOG_DUMP_SSH_USER
Project: k8s-presubmit-scale-6
Network Project: k8s-presubmit-scale-6
Zone: us-east1-b
Dumping logs temporarily to '/tmp/tmp.4xBe9oJ531/logs'. Will upload to 'gs://sig-scalability-logs/pull-perf-tests-clusterloader2-kubemark/1589991412294750208' later.
Dumping logs from master locally to '/tmp/tmp.4xBe9oJ531/logs'
Changing logfiles to be world-readable for download
Copying 'kube-apiserver.log kube-apiserver-audit.log kube-scheduler.log kube-controller-manager.log etcd.log etcd-events.log glbc.log cluster-autoscaler.log kube-addon-manager.log konnectivity-server.log fluentd.log kubelet.cov startupscript.log' from e2e-2182-0bf8a-kubemark-master
Specify --start=74404 in the next get-serial-port-output invocation to get only the new output starting from here.
scp: /var/log/glbc.log*: No such file or directory
scp: /var/log/cluster-autoscaler.log*: No such file or directory
scp: /var/log/konnectivity-server.log*: No such file or directory
scp: /var/log/fluentd.log*: No such file or directory
scp: /var/log/kubelet.cov*: No such file or directory
scp: /var/log/startupscript.log*: No such file or directory
ERROR: (gcloud.compute.scp) [/usr/bin/scp] exited with return code [1].
```

aojea:
@mborsz can you review again?

mborsz:
/lgtm
/approve

Thanks, looks good!

PR #2183, opened 2022-11-08
Turn storage scalability tests to green

wojtek-t:
Ref https://github.com/kubernetes/kubernetes/issues/109915

Before the failure are understood, let's simply disable the non-critical metrics in this test to turn it back to green.

```release-note
NONE
```

/kind bug
/sig scalability

/assign @mborsz @marseel


wojtek-t:
@marseel @mborsz  - friendly ping

PR #2188, opened 2022-11-09
Add checking all IPs in Service reachibility verification

panslava:
This is adding support for Dual Stack Kubernetes services

#### What type of PR is this?
/kind feature

When checking if service is reachable we were checking only first Service IP
This PR adds checking all IPs, this adds support for DualStack services, which can have 2 IPs

### DIFF
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -19,6 +19,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"net"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -336,16 +337,28 @@ func (p *pingChecker) run() {
...
+				for _, ingress := range p.svc.Status.LoadBalancer.Ingress {
+					ips = append(ips, ingress.IP)
+				}
+				port = p.svc.Spec.Ports[0].Port
+			}
+			for _, ip := range ips {
+				address := net.JoinHostPort(ip, fmt.Sprint(port))
+				command := fmt.Sprintf("curl %s", address)
+				_, err = execservice.RunCommand(pod, command)
+				if err != nil {
```
tosi3k, line 359: Consider using [ErrorList](https://github.com/kubernetes/perf-tests/blob/4a4b12023fe092bd4bfd69762c356bcc2f776667/clusterloader2/pkg/errors/error_list.go#L24-L28) for collecting errors from attempted connections to all the IP addresses rather than immediately breaking from the loop on receiving the first non-nil error here.

tosi3k (changes requested):

clusterloader2/pkg/measurement/common/service_creation_latency.go
panslava: I agree that would be useful, if we wanted to return errors, or log them

But here we just want to check if both IPs are reachable, what is the benefit of checking second one, if we know, that first one already failed?

tosi3k: Ah, I didn't read the rest of this measurement's logic. I thought we're plugging this error through somewhere higher up the call stack and wanted to know if, for instance, IPv6 network programming failed but IPv4 did not  🤷.

In that case I'm okay with leaving it as it is.

PR #2190, opened 2022-11-09
feat: allow overriding the storage class used for prometheus pvc

eddycharly:
#### What type of PR is this?

/kind feature

This PR allows overriding the storage class used for prometheus pvc.
When working with kind, we need to use the already installed storage class (`standard`).


realshuting:
Would love to see this gets accepted otherwise the test stuck waiting for the Prometheus stack to become available🙏

eddycharly:
@marseel do i need to do something for kubemark ?

marseel:
nope, this is just regression caused by https://github.com/kubernetes/perf-tests/issues/2192
I will merge it manually

PR #2200, opened 2022-11-18
CL2: Use embed FS instead of relying on $GOPATH.

mborsz:
#### What type of PR is this?
/kind cleanup

Clusterloader now reads a lot from $GOPATH which is problematic (e.g. when is executed from a different path or when binary has been built with different version than is now available in $GOPATH).

This is a first PR that aims to break this dependency in favour of embed.FS: we will embed necessary files (such as manifests which are considered a part of measurements). The test config files (such as load/config.yaml) will be still stored externally as they are not part of the testing framework.

This PR implements necessary interface and migrates the first that reads data in runtime: exec measurement as POC.

See https://github.com/kubernetes/perf-tests/pull/2199 for ~goal where are we going.

/assign @jupblb 

Fixes #

### DIFF
--- a/clusterloader2/pkg/execservice/exec_service.go
+++ b/clusterloader2/pkg/execservice/exec_service.go
@@ -19,6 +19,7 @@ package execservice
 import (
 	"bytes"
 	"context"
+	"embed"
 	"fmt"
 	"math/rand"
 	"os/exec"
@@ -47,11 +48,15 @@ const (
 	execPodCheckTimeout  = 2 * time.Minute
 
 	execServiceName = "Exec service"
+	deploymentYaml  = "manifest/exec_deployment.yaml"
 )
 
 var (
 	lock     sync.Mutex
 	podStore *measurementutil.PodStore
+
+	//go:embed manifest
+	manifestFS embed.FS
 )
 
 func InitFlags(c *config.ExecServiceConfig) {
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/execservice/exec_service.go
```diff
@@ -63,12 +70,16 @@ func InitFlags(c *config.ExecServiceConfig) {
...
 	)
 	flags.StringEnvVar(
-		&c.DeploymentYaml,
+		&unusedDeploymentYamlFlagValue,
 		"exec-deployment-yaml",
 		"EXEC_DEPLOYMENT_YAML",
 		"pkg/execservice/manifest/exec_deployment.yaml",
 		"Path to execservice deployment yaml.",
 	)
+	err := flags.MarkDeprecated("exec-deployment-yaml", "exec_deployment.yaml is embed during a build now.")
```
jupblb, line 79: I'm not sure anyone uses this flag besides myself (I added it so I could build cl2 using nix). I don't think it'd be a huge problem if we removed it.

mborsz:
/hold cancel

Thanks!

mborsz:
Sorry, I misunderstood the comment.

/hold

clusterloader2/pkg/execservice/exec_service.go
mborsz: You mean remove completely?

I can do that! :)

PR #2205, opened 2022-11-23
clusterloader2: use new apiserver latency SLI

dgrisonnet:
#### What type of PR is this?
/kind cleanup

In Kubernetes 1.23.0 a new request latency metric more suitable for SLOs since it doesn't account for webhooks latency was introduced. Since this new metric is better suited to clusterloader2 than the general-purpose one, it would make sense to update the existing Prometheus rules.

Note that this implies moving from a STABLE metric to an ALPHA one, but considering that the new metric has been around for already 3 releases and that it has a real-life scenario attached to it, the metric is very unlikely to be removed in the future.

This metric has been renamed in Kubernetes 1.26 (https://github.com/kubernetes/kubernetes/pull/112679) which means that the queries will only work on clusters from 1.26 onward. Would that be problematic for the support policy of clusterloader2?

This PR will require the release of Kubernetes 1.26 as well as a new version of metrics-server based on the apiserver code post 1.26. For these reasons, I will put it on hold for now.
/hold

/cc @marseel

### DIFF
--- a/clusterloader2/pkg/measurement/common/testdata/api_responsiveness_prometheus/rules.yml
+++ b/clusterloader2/pkg/measurement/common/testdata/api_responsiveness_prometheus/rules.yml
@@ -2,17 +2,17 @@ groups:
 - name: apiserver.rules
   rules:
   - expr: |
-      histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
+      histogram_quantile(0.99, sum(rate(apiserver_request_sli_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency:histogram_quantile
     labels:
       quantile: "0.99"
   - expr: |
-      histogram_quantile(0.9, sum(rate(apiserver_request_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
+      histogram_quantile(0.9, sum(rate(apiserver_request_sli_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency:histogram_quantile
     labels:
       quantile: "0.90"
   - expr: |
-      histogram_quantile(0.5, sum(rate(apiserver_request_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
+      histogram_quantile(0.5, sum(rate(apiserver_request_sli_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency:histogram_quantile
     labels:
       quantile: "0.50"
@@ -86,17 +86,17 @@ groups:
 - name: apiserver.1m.rules
   rules:
   - expr: |
-      histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket[1m])) by (resource,  subresource, verb, scope, le))
+      histogram_quantile(0.99, sum(rate(apiserver_request_sli_duration_seconds_bucket[1m])) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency_1m:histogram_quantile
     labels:
       quantile: "0.99"
   - expr: |
-      histogram_quantile(0.9, sum(rate(apiserver_request_duration_seconds_bucket[1m])) by (resource,  subresource, verb, scope, le))
+      histogram_quantile(0.9, sum(rate(apiserver_request_sli_duration_seconds_bucket[1m])) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency_1m:histogram_quantile
     labels:
       quantile: "0.90"
   - expr: |
-      histogram_quantile(0.5, sum(rate(apiserver_request_duration_seconds_bucket[1m])) by (resource,  subresource, verb, scope, le))
+      histogram_quantile(0.5, sum(rate(apiserver_request_sli_duration_seconds_bucket[1m])) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency_1m:histogram_quantile
     labels:
       quantile: "0.50"
--- a/clusterloader2/pkg/measurement/common/testdata/api_responsiveness_prometheus/rules.yml
+++ b/clusterloader2/pkg/measurement/common/testdata/api_responsiveness_prometheus/rules.yml
@@ -2,17 +2,24 @@ groups:
 - name: apiserver.rules
   rules:
   - expr: |
-      histogram_quantile(0.99, sum(rate(apiserver_request_sli_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
+      rate(apiserver_request_duration_seconds_bucket[5m]) and on (job) kubernetes_build_info{major="[0-1]",minor=~"^([0-9]|1[0-9]|2[0-2])$"}
+      or
+      rate(apiserver_request_slo_duration_seconds_bucket[5m]) and on (job) kubernetes_build_info{major="1",minor=~"^2[3-5]$"}
+      or
+      rate(apiserver_request_sli_duration_seconds_bucket[5m]) and on (job) kubernetes_build_info{major="1",minor=~"^(2[6-9]|[3-9][0-9])$"}
+    record: apiserver:request_latency:rate5m
+  - expr: |
+      histogram_quantile(0.99, sum(apiserver:request_latency:rate5m) by (resource,  subresource, verb, scope, le))
     record: apiserver:apiserver_request_latency:histogram_quantile
     labels:
       quantile: "0.99"
   - expr: |
-      histogram_quantile(0.9, sum(rate(apiserver_request_sli_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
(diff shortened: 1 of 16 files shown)


tosi3k:
Thanks for the pull request.

>This metric has been renamed in Kubernetes 1.26 (kubernetes/kubernetes#112679) which means that the queries will only work on clusters older than 1.26.

Judging by the linked PR's contents I think you rather meant this will work for Kubernetes 1.26+ (since you use the new name of the metric rather than the old one), right?

tosi3k:
Regarding compatibility - WDYT about moving this change under some flag that would either use the old or the new metric (`apiserver_request_duration_seconds_bucket` vs. `apiserver_request_sli_duration_seconds`, respectively) in `api_responsiveness_prometheus.go`?

By default, we'd still use the old metric but user could override that behavior by setting the flag in order to switch to the webhook-less latency metric.

For the configuration of dashboards I believe we could just include two separate charts for both metrics for now rather than fully replace the old metric in the graphs.

dgrisonnet:
That's a good idea to make it compatible, but I don't think the users should have this level of awareness. It would be best in my opinion if this was seamless and clusterloader2 would just use the better-suited metrics.

I can see two options to make it so that users don't have to know about this technical detail:

1. Create a recording rule to use the latest metrics available. It would look something like this:
```
apiserver_request_sli_duration_seconds
or
apiserver_request_slo_duration_seconds
or
apiserver_request_duration_seconds
```
The potential problem with that approach is that if we don't aggregate any labels like in the example above, the evaluation will take quite some time considering the sheer amount of metrics to return. Also, since recording rules are evaluated on a periodical basis (configurable), the metrics won't be as fresh as they used to be before.

2. Add Kubernetes version awareness to clusterloader2 to be able to select particular metrics depending on the version of the cluster it is testing against

tosi3k:
>Add Kubernetes version awareness to clusterloader2 to be able to select particular metrics depending on the version of the cluster it is testing against

I like this idea very much!

I think that we might just focus on the K8s minor (instead of having some sophisticated Semantic Versioning version parsing) against which we run ClusterLoader2.

We could fetch this piece of information simply from the `/version` endpoint at the beginning of CL2 execution, store this information somewhere [here](https://github.com/kubernetes/perf-tests/blob/c02e422cc303f0d59b4303cedd7a8c58ec0a4549/clusterloader2/pkg/framework/framework.go#L42-L53) and use this knob later on in the internals of API responsiveness and Metrics Server measurements from the `measurement.Config` var we input to the `Gather` methods to decide which Prometheus query (i.e. metric) to run.

dgrisonnet:
Thank you for the pointers, I will look adding this functionality

wojtek-t:
@dgrisonnet - will you have time to push this forward?
We will need this also in the context of APF graduation, so I'm really interesting in it.

dgrisonnet:
I'll try to spend some time finishing this in the upcoming weeks.

/remove-lifecycle stale

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
```diff
@@ -10,17 +10,17 @@ spec:
   - name: apiserver.rules
     rules:
     - expr: |
-        histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
+        histogram_quantile(0.99, sum(rate(apiserver_request_sli_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
```
dgrisonnet, line 13: @wojtek-t do you perhaps have any suggestions on how we should handle the versioning of the manifests?

The less intrusive option that I can see would be to add a recording rule to select the metric and always use the newest available. e.g:
```
expr: |
  (
    apiserver_request_sli_duration_seconds_bucket
    or
    apiserver_request_slo_duration_seconds_bucket
    or
    apiserver_request_duration_seconds_bucket
  )
```
The problems if we were to do that are:
1. since we don't do any aggregation, the evaluation of this new recording rule could take quite some time
2. freshness will now depend on both the scrape interval and the evaluation interval

wojtek-t: I'm not super familiar with this, so let me ask some questions:
1) What is this manifest used for?
  [Do I understand correctly that it's effectively an input to prometheus server and then our measurements (SLOs) effectively call these queries (which prometheus server has now somehow precomputed?]

2) Would it be an option that we effectively will have:
- all possible queries listed here in manifest (would have a query on a metric that doesn't exist break something)?
- and just the measurements (SLOs) checking cluster version and based on that chosing the right query?

dgrisonnet: 1. Yes, that is exactly how you understood it. This particular manifest seems to be used in a couple a places such as the dashboards and the SLO measurements: https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go#L52

2. If the metric doesn't exist, the query will not yield any result, so as long as we don't use it, nothing will break. We could create recording rules (precomputed queries) for all the metrics and then select in code which one to use depending on the cluster version and that would solve the problem for the measurements.

It might be a bit more challenging for the dashboards though, I will try to see if I can find a query that would make this work.

dgrisonnet:
The code changes are done, but I still need to figure out how I can test all the areas that I've touched.

If anyone wants to have a look at the PR in the meantime, all the logic for selecting metrics depending on the cluster version has been added in the last two commits.

wojtek-t:
I guess this test is explaining more:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/perf-tests/2205/pull-perf-tests-verify-test/1651882214188650496

dgrisonnet:
Thanks for the pointers, I'll try to reproduce the failures locally.

clusterloader2/pkg/measurement/util/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+package util
+
+import "k8s.io/perf-tests/clusterloader2/pkg/measurement"
+
+func GetApiserverSLI(config *measurement.Config) string {
+	if config.ClusterVersion.Major == "1" && config.ClusterVersion.Minor < "23" {
+		return "apiserver_request_duration_seconds"
+	} else if config.ClusterVersion.Major == "1" && config.ClusterVersion.Minor < "26" {
```
wojtek-t, line 24: nit: drop "else"

clusterloader2/pkg/measurement/util/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+		return "apiserver_request_slo_duration_seconds"
+	}
+	return "apiserver_request_sli_duration_seconds"
+}
+
+func GetApiserverLatency(config *measurement.Config) string {
+	if config.ClusterVersion.Major == "1" && config.ClusterVersion.Minor < "23" {
+		return "apiserver:apiserver_request_latency_1m:histogram_quantile"
+	} else if config.ClusterVersion.Major == "1" && config.ClusterVersion.Minor < "26" {
```
wojtek-t, line 33: nit: drop "else"

wojtek-t (commented):
I looked through the code and it all looks fine. I didn't try to parse all the queries and I'm fairly sure it's some of the queries (also from looking into test results).

dgrisonnet:
Yes there was an extra parentheses in the queries. I also took the opportunity to fix a segfault that I introduced in the unit tests, so normally they should be green now. But I still haven't tested the feature end-to-end.

wojtek-t:
> But I still haven't tested the feature end-to-end.

The test at head, we will get for free from the presubmit.
So we only need to test it against e.g. 1.26 and 1.22.
[Technically, 1.22 is out of support already, so probably testing 1.26 is enough...]

wojtek-t:
I didn't validate the metrics, for presubmits but they look reasonable (so hopefully the right metric is used).

Ideally, we should try to run it against 1.26 cluster too before merging.

clusterloader2/pkg/measurement/common/testdata/api_responsiveness_prometheus/rules.yml
```diff
@@ -2,17 +2,24 @@ groups:
 - name: apiserver.rules
   rules:
   - expr: |
-      histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket[5m])) by (resource,  subresource, verb, scope, le))
+      rate(apiserver_request_duration_seconds_bucket[5m]) and on (job) kubernetes_build_info{major="[0-1]",minor=~"([0-9]|1[0-9]|2[0-2])"}
+      or
+      rate(apiserver_request_slo_duration_seconds_bucket[5m]) and on (job) kubernetes_build_info{major="1",minor=~"2[3-5]"}
+      or
+      rate(apiserver_request_sli_duration_seconds_bucket[5m]) and on (job) kubernetes_build_info{major="1",minor=~"(2[6-9]|[3-9][0-9])"}
```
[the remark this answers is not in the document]
dgrisonnet: Good catch :+1:

dgrisonnet:
I tested on both 1.25.8 and 1.27.0 locally and the metrics are correctly selected.

Is there a way I could easily check that I didn't break the dashboards?

wojtek-t:
> I tested on both 1.25.8 and 1.27.0 locally and the metrics are correctly selected.

Great - thanks!

> Is there a way I could easily check that I didn't break the dashboards?

This looks fine to me - we can always fix them later if we missed something.

/lgtm
/approve

/hold cancel

PR #2206, opened 2022-11-23
Allow to optionally deploy kube-state-metrics scrapper on GKE clusters

oginskis:
I am running performance tests on GKE clusters and would like to measure several metrics provided by the `kube-state-metrics` scrapper (such as `kube_pod_info`).

I tested the change locally and verified that the scraper was successfully deployed if the `prometheus-scrape-kube-state-metrics` flag was enabled and I also could query the metrics in Prometheus.

Is there any particular reason why it's an option only for GCE for now?


oginskis:
@marseel please have a look
cc @alculquicondor

PR #2207, opened 2022-11-24
wait for services to be deleted

### LINKED ISSUE #2111: Create WaitForServices step
On service creation, deletion, that would be nice to have a step, which waits for services to be created/deleted

**Why is this needed**:
Services of type 'LoadBalancer' provisioned not instantly, so it takes time to create/delete them
It became a problem for l4lb test, cause I needed to add sleep to wait for 100 ilbs to be deleted https://github.com/kubernetes/perf-tests/pull/2107 
This could be better solved by having special function which waits for the services

HuShaoRu:
#### What type of PR is this?

/kind feature

It adds a step `waitForDeletion` in ServiceCreationLatency to wait for services to be deleted, and it can also get the time duration of deletion.

Fixes #2111


gaopenghigh:
Can this PR be merged?

PR #2215, opened 2022-12-14
cl2: Scrape block profiles instead of mutex ones

MadhavJivrajani:
#### What type of PR is this?

/kind bug cleanup

In k/k, contention-profiling can be enabled using the --contention-profiling flag. However, this enables block profiling and not mutex profiling. CL2 as of now scrapes mutex profiles and not block profiles, because of which inspite of enabling contention profiling on control plane components, the profiles are still empty.

As seen here for the API Server, `--contention-profiling` enables block profiling:
https://github.com/kubernetes/kubernetes/blob/68f808e6db388b249dae8c6c9dfb078bc0947b49/staging/src/k8s.io/apiserver/pkg/server/config.go#L919-L920

and similarly for other components.

### DIFF
--- a/clusterloader2/pkg/measurement/common/bundle/test_metrics.go
+++ b/clusterloader2/pkg/measurement/common/bundle/test_metrics.go
@@ -57,8 +57,8 @@ func createTestMetricsMeasurement() measurement.Measurement {
 	if metrics.etcdMemoryProfile, err = measurement.CreateMeasurement("MemoryProfile"); err != nil {
 		klog.Errorf("%v: etcdMemoryProfile creation error: %v", metrics, err)
 	}
-	if metrics.etcdMutexProfile, err = measurement.CreateMeasurement("MutexProfile"); err != nil {
-		klog.Errorf("%v: etcdMutexProfile creation error: %v", metrics, err)
+	if metrics.etcdBlockProfile, err = measurement.CreateMeasurement("BlockProfile"); err != nil {
+		klog.Errorf("%v: etcdBlockProfile creation error: %v", metrics, err)
 	}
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/bundle/test_metrics.go
```diff
@@ -218,26 +218,26 @@ func (t *testMetrics) Execute(config *measurement.Config) ([]measurement.Summary
 		appendResults(&summaries, errList, summary, executeError(t.etcdCPUProfile.String(), action, err))
 		summary, err = execute(t.etcdMemoryProfile, etcdGatherConfig)
 		appendResults(&summaries, errList, summary, executeError(t.etcdMemoryProfile.String(), action, err))
-		summary, err = execute(t.etcdMutexProfile, etcdGatherConfig)
```
marseel, line 221: For Etcd I can see that mutex profiles are actually working, check https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-performance/1602710237461090304/artifacts/

MadhavJivrajani: Yeah etcd actually enables both block and mutex: 
https://github.com/etcd-io/etcd/blob/8e3fc380dcaf44c37bb5dc19f1832981225061cc/pkg/debugutil/pprof.go#L28

I think we may have to introduce a new type called `BlockProfile`, enable that for all control plane components and then use `MutexProfile` only for etcd.

marseel: Sure, sounds good.

PR #2218, opened 2022-12-15
Set priorityclass in prometheus

mborsz:
The goal is to prevent eviction of prometheus when some pods with higher priority are created as part of the test.

/hold (for me to ensure in manual run that the goal is achieved)

/assign @tosi3k


mborsz:
Also keeping grafana is useful to track progress during tests with preemption.

/hold cancel

tosi3k:
> /lgtm

I think some transient bug prevented the reintroduction of LGTM :/.

/lgtm

PR #2221, opened 2022-12-21
Adds monitoring toleration to prometheus operator.

ashishranjan738:
#### What type of PR is this?
/kind bug

This commit adds monitoring toleration to prometheus operator, reason being currently the CL2_PROMETHEUS_NODE_SELECTOR are getting applied to both the prometheus operator and prometheus custom resource, however incase of situations when we have tainted nodes with monitoring and CL2_PROMETHEUS_NODE_SELECTOR is applied for the same node, the prometheus operator will fail to get schedule because it doesn't tolerate the monitoring taint.

Fixes #

Nothing ! just hoping for a quick review.

Signed-off-by: Ashish Ranjan <rnshis@amazon.com>


tosi3k:
/ok-to-test

Could you add some more context? I have trouble understanding your use case based on the PR's description, including what node selectors have to do with taints and tolerations.

PR #2222, opened 2022-12-28
Add timeout to exec command used for polling single master health

Argh4k:
#### What type of PR is this?
/kind feature

With the current approach the whole operation of polling single master health via exec command can get stuck for a latency of respective POST call, which in case of overloaded cluster can take O(minutes) .

Fixes #

### DIFF
--- a/clusterloader2/pkg/execservice/exec_service.go
+++ b/clusterloader2/pkg/execservice/exec_service.go
@@ -134,10 +134,10 @@ func TearDownExecService(f *framework.Framework) error {
 	return nil
 }
 
-// RunCommand executes given command on a pod in cluster.
-func RunCommand(pod *corev1.Pod, cmd string) (string, error) {
+// RunCommand executes given command on a pod in cluster. Context is passed to the exec command.
+func RunCommand(ctx context.Context, pod *corev1.Pod, cmd string) (string, error) {
 	var stdout, stderr bytes.Buffer
-	c := exec.Command("kubectl", "exec", fmt.Sprintf("--namespace=%v", pod.Namespace), pod.Name, "--", "/bin/sh", "-x", "-c", cmd)
+	c := exec.CommandContext(ctx, "kubectl", "exec", fmt.Sprintf("--namespace=%v", pod.Namespace), pod.Name, "--", "/bin/sh", "-x", "-c", cmd)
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/execservice/exec_service.go
```diff
@@ -145,6 +145,18 @@ func RunCommand(pod *corev1.Pod, cmd string) (string, error) {
 	return stdout.String(), nil
 }
 
+func RunCommandWithTimeout(pod *corev1.Pod, cmd string, timeout time.Duration) (string, error) {
+	var stdout, stderr bytes.Buffer
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
mborsz, line 150: IMHO more idiomatic way to do that is "RunCommandWithContext" approach where you take context as a parameter and e.g. caller passes context with timeout or cancel function or overall something more precise that context.Background

Argh4k: Hmm, I think we could do even better and just create one function RunCommand() where ctx is the first argument. I

mborsz: +1

clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -210,6 +213,11 @@ func (a *apiAvailabilityMeasurement) initFields(config *measurement.Config) erro
 			return err
 		}
 		a.hostPollTimeoutSeconds = hostPollTimeoutSeconds
+		hostPollExecTimeoutSeconds, err := util.GetIntOrDefault(config.Params, "hostPollExecTimeoutSeconds", 10)
```
mborsz, line 216: How about using util.GetDurationOrDefault?

This will allow us to drop "seconds" suffix, as it will be clear from the context that the unit is what you specified in the config (e.g. "1s", "10m") and allow to drop seconds to time.Duration convertion in line 99.

Argh4k: I just wanted to be consistent with hostPollTimeoutSeconds. But I think we can change both of them tbh

mborsz:
/lgtm
/approval

Thanks!

PR #2226, opened 2023-01-17
Network policy latency test clients v.0.0.1

dlapcevic:
A golang client application that runs in pods inside a K8s cluster, and measures network policy enforcement latency for changes to pods and network policies. 

Details are in the README.md file.

Main purpose: Test network policy performance and scalability by using it with [ClusterLoader2](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2) framework. 

/kind feature


network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,545 @@
...
+
+	if c.config.testPodCreation {
+		if err := c.measurePodCreation(); err != nil {
+			klog.Errorf("Pod creation test failed, error: %v", err)
+		}
+	} else {
+		if err := c.measureNetPolicyCreation(); err != nil {
+			klog.Errorf("Pod creation test failed, error: %v", err)
+		}
+	}
```
alan-kut, line 160: testPodCreation is vague.
Maybe it would be fine if this is additional test you can run, e.g.

if testPodCreation {
   c.measurePodCreation()
}

but in the current form it is not really a boolean whether to test pod creation but rather a flag for picking a mode what to test.

Why can't we just test both of them?
What is the usecase for this flag?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,545 @@
...
+	klog.Infof("--workerCount=%d", c.config.workerCount)
+
+	return nil
+}
+
+// measureNetPolicyCreation runs the network policy enforcement latency test for
+// network policy creation.
+func (c *TestClient) measureNetPolicyCreation() error {
+	klog.Infof("Starting to measure network policy enforcement latency on network policy creation")
+	podList, err := c.targetPodList()
```
alan-kut, line 241: What about getTargetPodList() or listTargetPods()?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,545 @@
...
+
+// targetPodList returns a list of pods based on the test client config, for
+// target namespace and target label selector, with a limit to the listed pods.
+func (c *TestClient) targetPodList() ([]corev1.Pod, error) {
+	podList, err := c.k8sClient.CoreV1().Pods(c.config.targetNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: c.config.targetLabelSelector, Limit: int64(c.config.expectedTargets)})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list pods for selector %q in namespace %q, error: %v", c.config.targetLabelSelector, c.config.targetNamespace, err)
+	}
+
+	if len(podList.Items) == 0 {
```
alan-kut, line 277: Shouldn't it be equal to expectedTargets?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,548 @@
...
+// network policy creation.
+func (c *TestClient) measureNetPolicyCreation() error {
+	klog.Infof("Starting to measure network policy enforcement latency on network policy creation")
+	podList, err := c.targetPodList()
+	if err != nil {
+		return fmt.Errorf("failed to get the pod list, error: %v", err)
+	}
+	klog.Infof("%d pods listed", len(podList))
+
+	c.policyCreatedTime = &timeWithLock{lock: sync.RWMutex{}}
```
alan-kut, line 247: Why is it intialized here, not in the constructor?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,548 @@
...
+
+	if c.policyCreatedTime.time == nil {
+		networkPolicy, err := c.k8sClient.NetworkingV1().NetworkPolicies(c.config.testClientNamespace).Get(context.TODO(), c.config.allowPolicyName, metav1.GetOptions{})
+		if err != nil {
+			klog.Warningf("Failed to get network policies for pod %q in namespace %q with IP %q: %v", target.name, c.config.testClientNamespace, target.ip, err)
+			failed = true
+		}
+
+		policyCreateTime = networkPolicy.GetCreationTimestamp().Time
+		c.policyCreatedTime.time = &policyCreateTime
```
alan-kut, line 497: This is still executed, even if the get failed.

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,548 @@
...
+// metric for pod creation.
+func (c *TestClient) reportReachedTimeForPodCreation(target *targetSpec, reachedTime time.Time) {
+	latency := reachedTime.Sub(target.startTime)
+	metrics.PolicyEnforceLatencyPodCreation.Observe(latency.Seconds())
+	klog.Infof("Pod %q in namespace %q reached %v after pod IP was assigned", target.name, target.namespace, latency)
+}
+
+// reportReachedTimeForPolicyCreation records network policy enforcement latency
+// metric for network policy creation.
+func (c *TestClient) reportReachedTimeForPolicyCreation(target *targetSpec, reachedTime time.Time) {
```
alan-kut, line 479: Why not to return errors?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,548 @@
...
+		networkPolicy, err := c.k8sClient.NetworkingV1().NetworkPolicies(c.config.testClientNamespace).Get(context.TODO(), c.config.allowPolicyName, metav1.GetOptions{})
+		if err != nil {
+			klog.Warningf("Failed to get network policies for pod %q in namespace %q with IP %q: %v", target.name, c.config.testClientNamespace, target.ip, err)
+			failed = true
+		}
+
+		policyCreateTime = networkPolicy.GetCreationTimestamp().Time
+		c.policyCreatedTime.time = &policyCreateTime
+	} else {
+		policyCreateTime = *c.policyCreatedTime.time
```
alan-kut, line 499: Do you expect this method to be called multiple times? It seems it does nothing after first call.

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,548 @@
...
+
+	for {
+		select {
+		case <-c.MainStopChan:
+			return
+		default:
+		}
+
+		_, err := utils.RunCommand(command)
+		if err == nil {
```
alan-kut, line 458: if err != nil {
  continue + some comment why it's expected
}
the code if err is not nil?

alan-kut (changes requested):
I haven't reviewed it fully yet.
Here is first batch of comments.

PTAL

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: I changed the flag name to `podCreationMode`.

I think it’s better to have one test client to run just one of the tests (modes), because of the use case and to keep it simpler.
You can run both at the same time by deploying them in separate pods for each mode.

The flag determines which mode to use. They work differently.
Default is network policy enforcement latency for network policy creation. It lists target pods and tries to reach them (no pod watcher).
When the flag is `True`, it tests network policy enforcement latency for pod creation (creates pod watcher).

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Changed name to `listTargetPods()`.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Changed the flag name to `maxTargets`, because that's what it actually does.
There will be an error if there are no target pods -- the test cannot carry on.
It will not fail if there are more than `maxTargets`, it will just limit the kube API list call to that value.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: The test client runs in one of the modes, and it instantiates objects for the specific mode once that mode is starting to run.
I think it's better to instantiate it closer to use, rather than doing it early for all variables, even the ones not to be used.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Fixed.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: I don't see the benefit of returning an error here. Logging a warning in case it fails is enough.
It's used in `recordFirstSuccessfulRequest()`, which is called many times as a separate go routine and won't impact the main execution.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Yes. It will be called every time a target is reached. To avoid getting the network policy from apiserver every time, I cache it the first time it’s retrieved.

After the first call, it will reuse the cached `c.policyCreateTime.time`.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Done.
Added `err != nil {continue}` and a comment for it.
“CURL command returns errors for requests denied by a policy, or when the IP is unreachable. Everything else is considered a success”

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,552 @@
...
+
+	if c.config.podCreationMode {
+		if err := c.measurePodCreation(); err != nil {
+			klog.Errorf("Pod creation test failed, error: %v", err)
+		}
+	} else {
+		if err := c.measureNetPolicyCreation(); err != nil {
+			klog.Errorf("Pod creation test failed, error: %v", err)
+		}
+	}
```
alan-kut, line 160: I don't like it. If I'm not mistake majority of the code is for either pod or policy.
Some common code could be extracted to some utils or library.

network/tools/network-policy-enforcement-latency/pkg/client.go
alan-kut: what if there is just one pod returned?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,552 @@
...
+		c.targetPodsMap.lock.Unlock()
+
+		// Process only up to the expected number of pods.
+		if haveMaxTargets {
+			c.stopInformerChan()
+			return
+		}
+
+		// Do the measurements only if it hasn't already been done for this pod.
+		if !targetUsed {
```
alan-kut, line 419: if targetUsed {
  return
}

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,552 @@
...
+		return
+	}
+
+	pod, ok := objList[0].(*corev1.Pod)
+	if !ok {
+		klog.Warningf("processNextPodCreationEvent() failed to convert obj (%T) to *corev1.Pod", objList[0])
+		return
+	}
+
+	if len(pod.Status.PodIP) > 0 {
```
alan-kut, line 399: if len(...) == 0 {
  return
}

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,552 @@
...
+
+		// Do the measurements only if it hasn't already been done for this pod.
+		if !targetUsed {
+			target := &targetSpec{
+				ip:        pod.Status.PodIP,
+				name:      podName,
+				namespace: pod.GetNamespace(),
+				startTime: startTime,
+			}
+			go c.recordFirstSuccessfulRequest(target, true)
```
alan-kut, line 426: You don't want to wait for it?

network/tools/network-policy-enforcement-latency/pkg/client.go
```diff
@@ -0,0 +1,552 @@
...
+		} else {
+			policyCreateTime = networkPolicy.GetCreationTimestamp().Time
+			c.policyCreatedTime.time = &policyCreateTime
+		}
+	} else {
+		policyCreateTime = *c.policyCreatedTime.time
+	}
+
+	c.policyCreatedTime.lock.Unlock()
+	if failed {
```
alan-kut, line 507: What happens in this case?
Don't we lose some test data because of this?
Doesn't it mean that request was sent before the policy was deployed?
Can't we wait for the policy before testing reachability?

alan-kut (changes requested):

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Policy creation test starts sending traffic before allow policy is created.
Once allow policy is created, the requests to pods will succeed, and the policy will exist to measure latency for its creation.
If the policy cannot be retrieved here, it means the setup is incorrect. The warning log will make it clear that the specified allow policy is not found. No metrics should be generated from it.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: No. It contains the functionality of recording the first successful request, by generating metrics.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: You are right. Most of the functionality besides initialization and sending a request is not common to the modes.
I will extract the common code to utils package and separate it into 2 applications. We can run both in the scale tests.
Sounds good?

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: It will run the test for one pod. Send requests to it until it's reached, and generate one latency metric data point for it.

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Done

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Done

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Fixed

network/tools/network-policy-enforcement-latency/pkg/client.go
dlapcevic: Split into two applications.

dlapcevic:
I refactored the code. Extracted all common code into separate package and split the modes into two applications `pod-creation-reachability-latency` and `policy-creation-enforcement-latency`.
@alan-kut Please take another look.

alan-kut (approved):

dlapcevic:
Hey @marseel, can you give LGTM for this one, so I can merge it?
@alan-kut reviewed it.

PR #2144, opened 2022-09-25
Add qperf support to netperf for network benchmarking

PratikDeoghare:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Currently netperf only supports `iperf`. 
`iperf` measures throughput only. 
`qperf` allows measuring both throughput and latency. It will also allow comparing numbers across the tools. 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.**:

```docs
NONE
```

PR #2227, opened 2023-01-23
Fix netperf docker image build failure

bitmeal:
##### What
- Update netperf container base from `debian:jessie` to `debian:bullseye`
- Add `CFLAGS=-fcommon` to build of netperf binary

##### Why
- The default image `sirot/netperf-latest` used by the netperf test crashes when given the `-testFrom` and `-testTo` parameters. These parameters cannot be turned off, but are implicitly set when omitted by user; introduced by kubernetes/perf-tests#2144.
- `debian:jessie` container on docker hub is ~2 years old: apt repo keys are expired, thus preventing dependency installation and container build
- After upgrade to `debian:bullseye`, with a newer compiler, netperf needs `-fcommon` flag for building (as workaround); see HewlettPackard/netperf#46

##### Relevance/Value
Updating the container is necessary to run the tests at all, fixing a regression introduced by the additionally passed parameters from kubernetes/perf-tests#2144

```
NONE
```

```
NONE
```

#### What type of PR is this?
/kind bug
/kind regression

### DIFF
--- a/network/benchmarks/netperf/Dockerfile
+++ b/network/benchmarks/netperf/Dockerfile
@@ -31,7 +31,7 @@ COPY go.mod go.mod
 
 RUN go build -o nptests
 
-FROM debian:jessie
+FROM debian:bullseye
 ENV LD_LIBRARY_PATH=/usr/local/lib
 
 MAINTAINER Girish Kalele <gkalele@google.com>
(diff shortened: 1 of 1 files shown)


marseel:
Could you take a look at it Dorde?
/assign @dlapcevic

dlapcevic:
Thanks @bitmeal.
/approve

PR #2232, opened 2023-02-14
Switch to registry.k8s.io

ameukam:
Reference issue:
  - https://github.com/kubernetes/perf-tests/issues/2231

Part of:
  - https://github.com/kubernetes/k8s.io/issues/4738

Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

### DIFF
--- a/clusterloader2/docs/GETTING_STARTED.md
+++ b/clusterloader2/docs/GETTING_STARTED.md
@@ -228,7 +228,7 @@ spec:
         group: test-pod
     spec:
       containers:
-      - image: k8s.gcr.io/pause:3.1
+      - image: registry.k8s.io/pause:3.1
         name: {{.Name}}
 ```
 ## Execute test
--- a/clusterloader2/drivers/gcp-csi-driver-stable.yaml
+++ b/clusterloader2/drivers/gcp-csi-driver-stable.yaml
@@ -478,7 +478,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: k8s.gcr.io/sig-storage/csi-provisioner:v2.2.1
+        image: registry.k8s.io/sig-storage/csi-provisioner:v2.2.1
         livenessProbe:
           failureThreshold: 1
           httpGet:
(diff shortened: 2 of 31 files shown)


mborsz:
/hold

We have had some problems with using registry.k8s.io in testing environments without public internet access.
We need to test this change in such environment before merging or adapt the tests running there to have access to registry.k8s.io.

ameukam:
> testing environments without public internet access.

which environments are talking about ? related to the Kubernetes project ?

mborsz:
The last time this change has broken GKE internal tests (https://github.com/kubernetes/perf-tests/pull/2044) which are using private nodes without internet access. I'm not sure what is a state now and if we expect to registry.k8s.io to work in such cases. I asked the current oncall to investigate this issue.

aojea:
> The last time this change has broken GKE internal tests (#2044) which are using private nodes without internet access. I'm not sure what is a state now and if we expect to registry.k8s.io to work in such cases. I asked the current oncall to investigate this issue.

cc: @BenTheElder

BenTheElder:
So from an OSS perspective, GKE internal test environment issues should really not be blocking upstream changes unless they're surfacing a bug ...

Kubernetes has a merged KEP to cease allowing any new images in GCR for 1.27+, and SIG K8s Infra could choose to shut down GCR even sooner as all images are available at registry.k8s.io and Kubernetes is massively over-budget (think $4M spend rate on $3M GCP budget), primarily due to content download costs.

Kubernetes has been on registry.k8s.io since 1.25, released August '22.

I see the internal bug and will follow-up on GKE's options there.

Argh4k:
Hi, we updated gke internal tests and we are ready to go with this change.

ameukam:
@mborsz @marseel @wojtek-t PTAL.

marseel:
/approve
but waiting for someone from GKE team to unhold when they are ready to merge.

aojea:
as a followup we should update the pause image version, right now kubernetes uses the `3.9` version and this repo seems to be useing `3.1` , I don't know if that can be an issue though

vaibhav2107:
Fixes: #2231

PR #2234, opened 2023-02-15
Add network policy enforcement latency measurement

dlapcevic:
**Description**

A new measurement that utilizes network policy enforcement latency test clients.
https://github.com/kubernetes/perf-tests/tree/master/network/tools/network-policy-enforcement-latency 

This measurement is intended to be used in the ClusterLoader2 load test.
https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load

Network policy enforcement latency is the required time for any change to pods or network policies to be reflected in the system. It's measured by simulating a real scenario with workloads that are connecting to each other, and measuring the time it takes for the network policies to be honored for new pods or network policy changes. The relevant scalability and performance dimensions are pod and policy churn rate (create, update, delete).

**Usage**

The measurement tests network policy enforcement latency for two cases:

1. Created network policies
Deploy the test clients (setup and run) with "testType" flag set to "policy-creation" after creating the target pods.

2. Created pods that are affected by network policies
Deploy the test clients (setup and run) with "testType" flag set to "pod-creation”, before creating the target pods. Target pods are all pods that have the specified label: { targetLabelKey: targetLabelValue }.

The test is set up by creating the required resources, including the network policy enforcement latency test client pods that measure the latencies and generate metrics for them.

/kind feature

### DIFF
--- a/clusterloader2/pkg/measurement/common/network-policy/manifests/clusterrole.yaml
+++ b/clusterloader2/pkg/measurement/common/network-policy/manifests/clusterrole.yaml
@@ -1,0 +1,11 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{.Name}}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["get"]
--- a/clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-pod-creation.yaml
+++ b/clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-pod-creation.yaml
@@ -1,0 +1,52 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    test: {{.TestClientLabel}}
+    type: {{.TypeLabelValue}}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        name: {{.Name}}
+        test: {{.TestClientLabel}}
+        type: {{.TypeLabelValue}}
+    spec:
+      # Use separate nodes to avoid consuming CPU/Memory resources on default
+      # nodes where all deployments of the performance test run.
+      nodeSelector:
+        {{.TestClientNodeSelectorKey}}: {{.TestClientNodeSelectorValue}}
+      tolerations:
+      - key: {{.TestClientNodeSelectorKey}}
+        operator: Equal
+        value: {{.TestClientNodeSelectorValue}}
+        effect: NoSchedule
+      containers:
+      - name: net-policy-latency-client
+        ports:
+        - containerPort: {{.MetricsPort}}
+          name: npdelaymetrics
+          protocol: TCP
+        imagePullPolicy: Always
+        image: gcr.io/k8s-staging-perf-tests/network-policy-enforcement-latency/pod-creation-reachability-latency:v0.0.1
+        command:
+        - sh
+        - -c
+        - ./pod-creation-reachability-latency
+          -HostNamespace="{{.Namespace}}"
+          -TargetLabelSelector="{{.TargetLabelSelector}}"
+          -TargetNamespace="{{.TargetNamespace}}"
+          -TargetPort={{.TargetPort}}
+          -MaxTargets={{.MaxTargets}}
+          -MetricsPort={{.MetricsPort}}
+        resources:
+          requests:
+            cpu: 200m
+            memory: 100Mi
+      serviceAccountName: {{.ServiceAccountName}}
--- a/clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
+++ b/clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
@@ -1,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    test: {{.TestClientLabel}}
+    type: {{.TypeLabelValue}}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: {{.Name}}
+  template:
+    metadata:
+      labels:
+        name: {{.Name}}
+        test: {{.TestClientLabel}}
+        type: {{.TypeLabelValue}}
+    spec:
+      # Use separate nodes to avoid consuming CPU/Memory resources on default
+      # nodes where all deployments of the performance test run.
+      nodeSelector:
+        {{.TestClientNodeSelectorKey}}: {{.TestClientNodeSelectorValue}}
+      tolerations:
+      - key: {{.TestClientNodeSelectorKey}}
+        operator: Equal
+        value: {{.TestClientNodeSelectorValue}}
+        effect: NoSchedule
+      containers:
+      - name: net-policy-latency-client
+        ports:
+        - containerPort: {{.MetricsPort}}
+          name: npdelaymetrics
+          protocol: TCP
+        imagePullPolicy: Always
+        image: gcr.io/k8s-staging-perf-tests/network-policy-enforcement-latency/policy-creation-enforcement-latency:v0.0.1
+        command:
+        - sh
+        - -c
+        - ./policy-creation-enforcement-latency
+          -HostNamespace="{{.Namespace}}"
+          -TargetLabelSelector="{{.TargetLabelSelector}}"
+          -TargetNamespace="{{.TargetNamespace}}"
+          -TargetPort={{.TargetPort}}
+          -MaxTargets={{.MaxTargets}}
+          -MetricsPort={{.MetricsPort}}
+          -AllowPolicyName={{.AllowPolicyName}}
+        resources:
+          requests:
+            cpu: 200m
+            memory: 100Mi
+      serviceAccountName: {{.ServiceAccountName}}
--- a/clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-apiserver.yaml
+++ b/clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-apiserver.yaml
@@ -1,0 +1,22 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    test: {{.TestClientLabel}}
+spec:
+  podSelector:
+    matchLabels:
+      test: {{.TestClientLabel}}
+  policyTypes:
+  - Egress
+  egress:
+  - ports:
+    - port: 443
+      protocol: TCP
+    - port: 80
+      protocol: TCP
+    to:
+    - ipBlock:
+        cidr: {{.kubeAPIServerIP}}/32
--- a/clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-target-pods.yaml
+++ b/clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-target-pods.yaml
@@ -1,0 +1,25 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    type: {{.TypeLabelValue}}
+spec:
+  podSelector:
+    matchLabels:
+      type: {{.TypeLabelValue}}
+  policyTypes:
+  - Egress
+  egress:
+  - to:
+    - podSelector:
+        matchLabels:
+          {{.TargetLabelKey}}: {{.TargetLabelValue}}
+{{if .OnlyTargetNamespace}}
+      namespaceSelector:
+        matchLabels:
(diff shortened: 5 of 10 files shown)


clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
```diff
@@ -0,0 +1,53 @@
...
+        value: {{.TestClientNodeSelectorValue}}
+        effect: NoSchedule
+      containers:
+      - name: net-policy-latency-client
+        ports:
+        - containerPort: {{.MetricsPort}}
+          name: npdelaymetrics
+          protocol: TCP
+        imagePullPolicy: Always
+        image: gcr.io/k8s-staging-perf-tests/network-policy-enforcement-latency/policy-creation-enforcement-latency:v0.0.1
```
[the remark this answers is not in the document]
dlapcevic: I'll push another one for v0.0.1, because nothing is yet using it, and the existing one has a bug, so it shouldn't even be used. So, we can just keep it v0.0.1 here.

clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-pod-creation.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  namespace: {{.Namespace}}
+  labels:
+    test: {{.TestClientLabel}}
+    type: {{.TypeLabelValue}}
+spec:
+  replicas: 1
```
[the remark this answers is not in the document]
dlapcevic: The way it's implemented now, it wouldn't make sense to have multiple replicas of the same pod, because it will target the same pods.
I want to deploy exactly one pod per target namespace, and I use a deployment with 1 replica to do so.

The pods are intended to run on a separate set of nodes that is specific for hosting test client pods, to avoid interfering with the load test.

There is a node selector specified, and a comment for it:
```
      # Use separate nodes to avoid consuming CPU/Memory resources on default
      # nodes where all deployments of the performance test run.
```

clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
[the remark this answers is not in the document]
dlapcevic: I pushed `v0.0.1` again with the fix https://github.com/kubernetes/perf-tests/pull/2233 to https://gcr.io/k8s-staging-perf-tests/network-policy-enforcement-latency

clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
```
marseel, line 1: Maybe it would make sense to combine this file and `dep-test-client-pod-creation.yaml` as diff is only 3 lines and seems like all arguments are almost the same?

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
```diff
@@ -0,0 +1,568 @@
...
+
+	nps.framework = config.ClusterFramework
+	nps.k8sClient = config.ClusterFramework.GetClientSets().GetClient()
+
+	namespaceList, err := nps.k8sClient.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Target namespaces are only those that have a "test-" prefix.
```
marseel, line 219: Shouldn't this be `framework.GetAutomanagedNamespacePrefix()`?

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
```diff
@@ -0,0 +1,568 @@
...
+	if err := nps.createTestClientDeployments(depTemplateMap, policyCreationTest, depPolicyCreationFilePath); err != nil {
+		return err
+	}
+
+	maxRetries := 5
+	retryDuration := 10 * time.Second
+	allReady := false
+
+	klog.V(2).Infof("Waiting for policy creation test client pods to be running")
+	for retries := 0; retries < maxRetries; retries++ {
```
marseel, line 323: I think you could use `WaitForPods` here.

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
```diff
@@ -0,0 +1,568 @@
...
+	}
+
+	policyLoadQPS, err := util.GetIntOrDefault(config.Params, "policyLoadQPS", defaultPolicyLoadQPS)
+	if err != nil {
+		klog.Errorf("Failed getting parameter policyLoadQPS value, error: %v", err)
+		return
+	}
+
+	policiesPerNs := policyLoadCount / len(nps.targetNamespaces)
+	qpsSleepDuration := (1 * time.Second) / time.Duration(policyLoadQPS)
```
marseel, line 463: You could use `rate.NewLimiter` instead

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
```diff
@@ -0,0 +1,568 @@
...
+					klog.Errorf("Error while creating load network policy for label selector 'name=%s': %v", podSelectorLabelValue, err)
+				}
+			}()
+
+			time.Sleep(qpsSleepDuration)
+		}
+	}
+
+	expectedFinishTime := time.Now().Add(time.Duration(policyLoadCount/policyLoadQPS) * time.Second)
+	// Add a short time buffer before completing, to not affect next test steps.
```
marseel, line 492: Hm, maybe let's switch to rate.NewLimiter as mentioned above and make api calls inline instead of separate goroutine and let's get rid of this part.

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
dlapcevic: It should. Fixed.

clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
dlapcevic: Yes. I parameterized AppName and AppVersion, and merged the files into one.

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
dlapcevic: Yes, it's much better. Done.

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
dlapcevic: Thanks. Done.

clusterloader2/pkg/measurement/common/network-policy/manifests/dep-test-client-policy-creation.yaml
dlapcevic: I reverted it to two files, because I actually couldn't make it work.
`command:` arguments couldn't be parsed correctly when some arguments are inside of `{{if}}`s.

Anyway I think it's good to have two files as these are two different applications, and it'll be easier to extend and maintain it this way.

clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-apiserver.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - Egress
+  egress:
+  - ports:
+    - port: 443
+      protocol: TCP
+    - port: 80
+      protocol: TCP
+    to:
+    - ipBlock:
+        cidr: {{.kubeAPIServerIP}}/32
```
marseel, line 22: Probably you could use something like:
```
  egress:
    - toEntities:
        - kube-apiserver
 ```
instead (https://docs.cilium.io/en/v1.12/policy/language/#entities-based). But it's minor, non-blocking.

**Comment by marseel** [comment]
/lgtm

As it's a non-trivial measurement I would like someone else to take a look too.

File: clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-apiserver.yaml (hunk 8 above)
Comment (dlapcevic) [reply clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-apiserver.yaml:22]: `toEntities` and `fromEntities` would work only for Cilium implementation I think.
I don’t see these fields as part of K8s network policy API.
I think we should keep it as is.

Comment (marseel) [reply clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-apiserver.yaml:22]: ah yea, I totally forgot it's NetworkPolicy and not CiliumNetworkPolicy.

File: clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-target-pods.yaml [hunk 9]
```diff
@@ -0,0 +1,25 @@
...
+  - to:
+    - podSelector:
+        matchLabels:
+          {{.TargetLabelKey}}: {{.TargetLabelValue}}
+{{if .OnlyTargetNamespace}}
+      namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: {{.TargetNamespace}}
+{{else}}
+      namespaceSelector: {}
```
tosi3k, line 24: Nit: is this actually needed?

clusterloader2/pkg/measurement/common/network-policy/manifests/clusterrole.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{.Name}}
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["get", "list", "watch"]
```
tosi3k, line 11: Nitpick: `"watch"` for networkpolicies isn't required, is it?

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
```diff
@@ -0,0 +1,567 @@
...
+
+	if err := nps.framework.ApplyTemplatedManifests(manifestsFS, clusterRoleBindingFilePath, templateMap); err != nil {
+		return fmt.Errorf("error while creating clusterrolebinding: %v", err)
+	}
+
+	return nil
+}
+
+func (nps *networkPolicyEnforcementMeasurement) run(config *measurement.Config) error {
+	// Should this be mandatory? Fail if not provided.
```
tosi3k, line 259: Should this get a TODO or is it an unfinished leftover?

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+
+	if err := nps.framework.ApplyTemplatedManifests(manifestsFS, clusterRoleBindingFilePath, templateMap); err != nil {
+		return fmt.Errorf("error while creating clusterrolebinding: %v", err)
+	}
+
+	return nil
+}
+
+func (nps *networkPolicyEnforcementMeasurement) run(config *measurement.Config) error {
```
tosi3k, line 258: Could we add some check to `run` that `setup` was run beforehand?

tosi3k (changes requested):
Nitpicks, apart from that LGTM.

clusterloader2/pkg/measurement/common/network-policy/manifests/policy-egress-allow-target-pods.yaml
dlapcevic: Yes, because we are allowing ingress traffic from all test namespaces and not just from the one that the pods are in and policies applied.
More info: https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/05-allow-traffic-from-all-namespaces.md
```

clusterloader2/pkg/measurement/common/network-policy/manifests/clusterrole.yaml
dlapcevic: Apparently neither `"list"` is being used. I'm removing both `"list"` and `"watch"` for network policies.
They are being used by the test client app https://github.com/kubernetes/perf-tests/tree/master/network/tools/network-policy-enforcement-latency

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
dlapcevic: Removed.

clusterloader2/pkg/measurement/common/network-policy/network-policy-enforcement-latency.go
dlapcevic: Done. Used `nps.framework == nil` to check if setup was run before.

dlapcevic:
@tosi3k @marseel can we please merge now?

dlapcevic:
@marseel can we please merge now?

tosi3k:
I think you just need to cancel the hold :).

/hold cancel

dlapcevic:
Thanks @tosi3k. I didn't know that I had a permission to do it.

PR #2240, opened 2023-03-01
Add a way to customize namespace deletion timeout

mborsz:
#### What type of PR is this?

/kind feature

This PR makes it possible to customize namespace deletion timeout, as it's not possible that all tests everywhere can fit in 10m.

This PR also uses the possibility to customize it in huge service which in some cases takes more than 10m to cleanup (it's short and creates tons of events).

Fixes #

/assign @Argh4k

### DIFF
--- a/clusterloader2/api/default.go
+++ b/clusterloader2/api/default.go
@@ -19,6 +19,7 @@ package api
 import (
 	"fmt"
 
+	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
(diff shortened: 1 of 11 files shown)


wojtek-t:
This is failing compilation:
```
F0301 17:01:47.592668   16871 clusterloader.go:378] Test compilation failed: [config reading error: decoding failed: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field NamespaceConfig.namespace.deleteNamespaceTimeout of type time.Duration]
```

mborsz:
Let's try one more time

PR #2242, opened 2023-03-03
fix(clusterloader2): fix daemonset node counter

wrype:
daemonset in load test count node error when exist taints

/kind bug

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
@@ -154,14 +154,12 @@ func (n *NodeCounter) shouldRun(obj interface{}) (bool, error) {
 		return false, fmt.Errorf("unexpected type of obj: %v. got %T, want *corev1.Node", obj, obj)
 	}
 	matched, err := podMatchesNodeAffinity(n.affinity, node)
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/util/runtimeobjects/replicaswatcher.go
```diff
@@ -150,7 +154,14 @@ func (n *NodeCounter) shouldRun(obj interface{}) (bool, error) {
 		return false, fmt.Errorf("unexpected type of obj: %v. got %T, want *corev1.Node", obj, obj)
 	}
 	matched, err := podMatchesNodeAffinity(n.affinity, node)
-	return !node.Spec.Unschedulable && matched, err
+	tolerate := true
```
mborsz, line 157: Can we add an unit test that covers this change?

wrype: ok, have any suggestions about testcase?

mborsz: Maybe something like: a pod that wants to run on a given node, without toleration it wouldn't be possible (test case 1), but it has toleration it's OK (test case 2)

PR #2249, opened 2023-03-29
Update `pause` container image version. 

kishen-v:
This PR contains the changes to use the more recent `pause:3.9` container image for deployments.


mborsz:
/lgtm
/approve

Out of curiosity, why are we doing this?

PR #2254, opened 2023-04-24
use new debian image in build_kube_cross for golang job

Argh4k:
#### What type of PR is this?
/kind bug

ci-build-and-push-k8s-at-golang-tip job is currently broken because it is using debian stretch and it was moved to archive.debian.org (https://lists.debian.org/debian-devel-announce/2023/03/msg00006.html).

Fixes #


tosi3k:
/lgtm

Thanks!

PR #2255, opened 2023-04-25
WIP: Add a perf test for the API Streaming feature

p0lyn0mial:
#### What type of PR is this?
/kind feature

The test start with generating 400 secrets in a single namespace. Each secret holding 1MB of payload.
Next, the test starts measuring memory consumption of the API server. Then the test starts 4 informers to retrieve data.

Fixes #

https://github.com/kubernetes/enhancements/issues/3157


clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,84 @@
+name: huge-secrets
+tuningSets:
+- name: Uniform10qps
+  qpsLoad:
+    qps: 10
+steps:
+- name: Create well known namespaces
```
p0lyn0mial, line 7: I decided to manually create ns because the app (which will be defined in a separate PR) needs to know the target ns

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,84 @@
...
+          objectTemplatePath: "deployment.yaml"
+          templateFillMap:
+            Replicas: 1
+- name: Wait for the secret informer pod to be running
+  measurements:
+    - Identifier: WaitForControlledPodsRunning
+      Method: WaitForControlledPodsRunning
+      Params:
+        action: gather
+- name: Measure the API server memory usage
```
p0lyn0mial, line 67: this is untested because `kind` (my local env) is not supported.

clusterloader2/testing/watch-list/deployment.yaml
```diff
@@ -0,0 +1,22 @@
...
+    matchLabels:
+      group: watch-list
+  template:
+    metadata:
+      labels:
+        group: watch-list
+    spec:
+      containers:
+      # TODO: change the image ref
+      - image: quay.io/polynomial/simple-watch:latest
```
p0lyn0mial, line 19: the actual test is defined here https://github.com/kubernetes/perf-tests/pull/2256

clusterloader2/testing/watch-list/deployment.yaml
```diff
@@ -0,0 +1,22 @@
...
+        name: {{.Name}}
+        command: ["/usr/bin/app"]
```
p0lyn0mial, line 21: change it to `watch-list`

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,84 @@
...
+  measurements:
+    - Identifier: MemoryProfile
+      Method: MemoryProfile
+      Params:
+        action: gather
+- name: Delete well known namespaces
```
p0lyn0mial, line 73: How do I know that my test has finished?

clusterloader2/testing/watch-list/config.yaml
mborsz: Alternatively you may want to use "namespaces" with specified "prefix: huge-services": https://github.com/kubernetes/perf-tests/blob/4da7097e35b13225ed05bf3dfcb842864c11fae4/clusterloader2/api/types.go#L140-L140

Then you namespace should be named "huge-services-0".

This will avoid using "namespaceRange: null" which is slightly hacky :)

clusterloader2/testing/watch-list/config.yaml
mborsz: I'm not sure if now CL2 supports waiting for pod termination.

I think it may be easier to achieve desired effect in the current clusterloader model with:
* pod image are probing until they are kiled
* CL2 starts deployment, waits for it to be started
* Sleeps 5m
* CL2 deletes deployment, waits for all pods to be deleted

mborsz: Alternatively you can migrate to Jobs and use "WaitForFinishedJobs" https://github.com/kubernetes/perf-tests/blob/4da7097e35b13225ed05bf3dfcb842864c11fae4/clusterloader2/pkg/measurement/common/wait_for_jobs.go#L17-L17

clusterloader2/testing/watch-list/config.yaml
mborsz: Yes, this will collect memory profiles, but I don't think it will gather number of bytes (maybe you can infer this somehow from pprofs).

You may want to use TestMetrics (https://github.com/kubernetes/perf-tests/blob/4da7097e35b13225ed05bf3dfcb842864c11fae4/clusterloader2/testing/access-tokens/config.yaml#L60-L67)

which gathers quite a lot of metrics (including cpu/memory absolute values)/pprofs and other stuff that may be useful.

p0lyn0mial: mhm, ok, is there a way to turn profiles into a dashboard so that it easier to compare/visualise?

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: oh, didn't know i can to that, that would be awesome, let me try:)

p0lyn0mial: looks like it works, thanks for the tip!

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: decided to take the first approach, added a sleep, seems to be working.

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,74 @@
...
+- name: Uniform10qps
+  qpsLoad:
+    qps: 10
+steps:
+- name: Create secrets
+  phases:
+  - namespaceRange:
+      min: 1
+      max: 1
+      basename: huge-secrets
```
mborsz, line 19: I'm not sure if the namespaceRange is necessary when you set namespace.prefix on top.

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,74 @@
+# $testDuration controls
+# for how long the test should be executed
+# expressed in minutes
+{{$testDuration := 5}}
```
mborsz, line 4: Nit: personally I would suggest testDuration to be expressed in time.Duration-parsable format like "5m", "10s". Then it will be self explanatory and comments like "expressed in minutes" won't be necessary

mborsz:
I think it looks good. I suggest changing MemoryProfile to TestMetrics as this will get more data like MemoryProfile and resource usage.

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: Also when I remove it I get the following error:
```
F0426 13:50:19.769252   46728 clusterloader.go:378] Test compilation failed: [config reading error: decoding failed: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal object into Go struct field Step.steps.phases of type []*api.Phase]
exit status 1
```

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: oh, it looks great!

p0lyn0mial: do you know how memory-related metrics are gathered from the api server?

p0lyn0mial: okay, it looks like by default (no params defined) it will bring memory stats :)

clusterloader2/testing/watch-list/config.yaml
mborsz: Sorry, my comment makes no sense.

I was thinking about namespaceRange.basename which seems to be redundant with namespace.prefix.

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: np, done.

p0lyn0mial:
@mborsz many thanks for the review. I applied you comments. I'll keep it as `WIP` until we merge https://github.com/kubernetes/perf-tests/pull/2256

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,71 @@
...
+      Method: WaitForControlledPodsRunning
+      Params:
+        action: gather
+- name: Waiting {{$testDuration}} before cancelling the test
+  measurements:
+    - Identifier: Wait
+      Method: Sleep
+      Params:
+        duration: {{$testDuration}}
+- name: Collect the API server memory usage
```
mborsz, line 66: Nit: s/API server memory usage/test metrics or measurements?

mborsz:
> 

Yes, I think we are good here. Let's merge https://github.com/kubernetes/perf-tests/pull/2256 and then this PR.

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: done

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,71 @@
...
+      objectBundle:
+        - basename: huge-secret
+          objectTemplatePath: role.yaml
+        - basename: huge-secret
+          objectTemplatePath: roleBinding.yaml
+        - basename: huge-secret
+          objectTemplatePath: "deployment.yaml"
+          templateFillMap:
+            Replicas: 1
+            Duration: {{$testDuration}}
```
wojtek-t, line 53: What does it give us? Even if the binary will finish, the pod itself will get restarted (the default restart policy is Always), no?

Also - by definition deployment is for long-running stuff. Should we instead of `Job` here?

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,71 @@
...
+- name: Wait for the secret informer pod to be running
+  measurements:
+    - Identifier: WaitForControlledPodsRunning
+      Method: WaitForControlledPodsRunning
+      Params:
+        action: gather
+- name: Waiting {{$testDuration}} before cancelling the test
```
wojtek-t, line 60: Given my comment above - if we would use Job - instead of sleeping, I would just wait for all pods to finish here (i.e. effectively the job completion).

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: > What does it give us? Even if the binary will finish, the pod itself will get restarted (the default restart policy is Always), no?

yes, that is correct.

> Also - by definition deployment is for long-running stuff. Should we instead of Job here?

I don't have a preference. We could switch to a Job but we need to define when our test ends.

I think we would need `testDuration` anyway even for a job, do you agree?

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: np, we can switch to a Job.

clusterloader2/testing/watch-list/config.yaml
wojtek-t: We would need testDuration - but that would only be an argument to the binary itself.

And then test finishes when job finishes.

p0lyn0mial: yes, okay let me rework the manifest, thanks.

p0lyn0mial: okay, switched to a Job, PTAL.

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,62 @@
+{{$testDuration := "5m"}}
+name: huge-secrets
```
wojtek-t, line 2: I personally wouldn't call this test "huge-secrets", because it's not what it is really testing.
We're really testing streaming lists.

So maybe "streaming-list"? Or "watch-list" ?

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,62 @@
...
+namespace:
+  number: 1
+  prefix: "huge-secrets"
```
wojtek-t, line 5: Same here

wojtek-t: And everywhere else :)

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,62 @@
...
+    tuningSet: Uniform10qps
+    objectBundle:
+    - basename: huge-secret
+      objectTemplatePath: "secret.yaml"
+- name: Start measurements
+  measurements:
+    - Identifier: TestMetrics
+      Method: TestMetrics
+      Params:
+        action: start
```
wojtek-t, line 27: I would also add verification of api call latency

https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/load/modules/measurements.yaml#L52-L60

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: okay, will change to `watch-list`

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: done :)

clusterloader2/testing/watch-list/config.yaml
```diff
@@ -0,0 +1,73 @@
...
+      Params:
+        action: gather
+        timeout: 10m
+- name: Collecting test measurements
+  measurements:
+    - Identifier: TestMetrics
+      Method: TestMetrics
+      Params:
+        action: gather
+    - Identifier: APIResponsivenessPrometheusSimple
```
p0lyn0mial, line 67: > I would also add verification of api call latency

done, please also check if the param i'm passing are okay/correct.

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: BTW will the verification fail if the api calls will be slow/above some threshold?

wojtek-t:
/lgtm
/approve

[the kubemark presubmit is currently failing - we're trying to debug it]

PR #2256, opened 2023-04-25
Add utility image used in testing the API streaming feature

p0lyn0mial:
#### What type of PR is this?
/kind feature

It adds utility image used in testing the API streaming feature. (https://github.com/kubernetes/perf-tests/pull/2255)

The binary starts X number of secret informers to the target namespaces and waits until they are fully synchronised. Then the test is repeated until specified timeout has elapsed.

Fixes #

xref: https://github.com/kubernetes/enhancements/issues/3157


util-images/watch-list/main.go
```diff
@@ -0,0 +1,77 @@
...
+	var kubeconfig string
+	var targetNamespace string
+	var informerCount int
+	var testTimeout time.Duration
+
+	klog.InitFlags(flag.CommandLine)
+	flag.StringVar(&kubeconfig, "kubeconfig", "", "path to kubeconfig.")
+	flag.StringVar(&targetNamespace, "namespace", "huge-secrets-0", "namespace that host secrets to list. If empty a default (huge-secrets-0) value will be used.")
+	flag.IntVar(&informerCount, "count", 4, "the number of informers per targetNamespace to run. If empty a default (4) value will be used.")
+	flag.DurationVar(&testTimeout, "timeout", time.Minute, "timeout duration for the test")
```
p0lyn0mial, line 27: I need to figure out the timeout for the test. I'd start with 5 min or so.

util-images/watch-list/Makefile
```diff
@@ -0,0 +1,17 @@
+PROJECT = k8s-staging-perf-tests
+IMG = gcr.io/$(PROJECT)/watch-list
```
p0lyn0mial, line 2: The makefile is untested because i don’t have permissions to the registry

util-images/watch-list/go.mod
```diff
@@ -0,0 +1,49 @@
+module k8s.io/perf-tests/watch-list
+
+go 1.20
+
+require (
+	k8s.io/apimachinery v0.27.0
```
p0lyn0mial, line 6: Are we okay with these levels ?

mborsz:
looks good to me. The only potential change is whether the binary should finish after some time or the CL2 should kill it (see comments in https://github.com/kubernetes/perf-tests/pull/2255).

Please fix linter.

Also I can push the images (I should have permissions) once we submit the code (or are close to submission).

util-images/watch-list/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+all: push
+
+.PHONY: build
+build:
+	docker build --pull -t $(IMG):$(TAG) .
+	docker tag $(IMG):$(TAG) $(IMG):latest
+	@echo Built $(IMG):$(TAG) and tagged with latest
+
+.PHONY: push
```
p0lyn0mial, line 13: BTW: How can I publish this image ? Is it a manual process or done by some job in the CI?

mborsz: It's manual process. I think I will have to do that due permissions, just let me know when you are ready to push.

util-images/watch-list/main.go
```diff
@@ -0,0 +1,77 @@
...
+	flag.Parse()
+
+	// TODO: turn it into a flag so that we can compare test results(watch-list on/off)
```
p0lyn0mial, line 30: how can we do this? are we going to have two separate runs ?

p0lyn0mial:
> looks good to me. The only potential change is whether the binary should finish after some time or the CL2 should kill it (see comments in https://github.com/kubernetes/perf-tests/pull/2255).

Okay, thanks, I'll have a look. I'm fine with doing whatever is easier. 

I just want to make sure that the test had a chance to send some requests over some period of time and we managed to collect memory stats.

util-images/watch-list/Makefile
p0lyn0mial: Ok, I will. Thanks!

mborsz:
> 



> > looks good to me. The only potential change is whether the binary should finish after some time or the CL2 should kill it (see comments in #2255).
> 
> Okay, thanks, I'll have a look. I'm fine with doing whatever is easier.
> 
> I just want to make sure that the test had a chance to send some requests over some period of time and we managed to collect memory stats.

I think other tests work in a way that pods are running indefinitely and we are killing them from CL2 after some period of time, e.g. https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/request-benchmark so this is a default implementation I would suggest.

mborsz:
Actually https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/request-benchmark looks very similar to what you are trying to achieve:
* it already creates some (potentially large configmaps)
* runs some pod doing some probing to generate load on kube-apiserver
* gathers apiserver's metrics

You can see it running here: https://k8s-testgrid.appspot.com/sig-scalability-benchmarks#gce-benchmark-requests-1

Maybe you can just adapt the test instead of creating a new one (e.g. make the image used in that test controllable and put your watch image)?

util-images/watch-list/go.mod
mborsz: LGTM

util-images/watch-list/main.go
mborsz: I think this needs to be infinity given that in CL2 config you kill the pod after some time.

p0lyn0mial:
> EDIT: Feel free to ignore the suggestion as both tests are rather small so there is no that much duplication between them

yeah, as always, it is a tradeoff between clarity and code duplication.

I would keep https://github.com/kubernetes/perf-tests/pull/2255 for clarity but `request-benchmark` binary looks interesting.

I would have to slightly change it so that: 
 - it knows how to deal with the `WATCH` requests. (i'd like to bring all data)
 - it knows when to stop a watch request and start a new one
 - ^ would be different for normal watch requests and for the ones used specifically by the streaming (watch-list) feature.

In addition to that we wouldn't test functional changes that were made to the client-go library anymore. However this should be okay since this is a performance test.

util-images/watch-list/main.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	v1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+func main() {
+	var kubeconfig string
```
wojtek-t, line 34: nit:
I would make these vars a file-level ones and then move flag definition to a separate function (addFlags? registerFlags?)

It's the pattern we generally use in k/k, e.g.:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/options/audit.go#L263

wojtek-t (commented):
Just one small comment - overall LGTM

util-images/watch-list/main.go
p0lyn0mial: done

p0lyn0mial:
I applied the last batch of comments, PTAL.

Let's wait until all CI jobs are green before tagging. Thanks!

util-images/watch-list/main.go
```diff
@@ -0,0 +1,103 @@
...
+	klog.Info("Exiting the test app")
+}
+
+func registerFlags() {
+	klog.InitFlags(flag.CommandLine)
+
+	flag.StringVar(&kubeconfig, "kubeconfig", "", "path to kubeconfig.")
+	flag.StringVar(&targetNamespace, "namespace", "huge-secrets-1", "namespace that host secrets to list. If empty a default (huge-secrets-1) value will be used.")
+	flag.IntVar(&informerCount, "count", 4, "the number of informers per targetNamespace to run. If empty a default (4) value will be used.")
+	flag.DurationVar(&testTimeout, "timeout", time.Minute, "timeout duration for the test")
```
mborsz, line 91: I think that setting binary-level timeout is not compatible with the current approach in https://github.com/kubernetes/perf-tests/pull/2255 where you delete pods after sleep period.

What will be happening is that the binary will be restarting every 1m during the $testDuration (5m) period.

p0lyn0mial: oh, this is just a default timeout, I was planing to set a new one directly from the test.

p0lyn0mial: here: https://github.com/kubernetes/perf-tests/pull/2255/files#diff-58ef704e64da515ecf846e260d5440f01f331c7b67741b51368887d99d7b8d16R22

p0lyn0mial:
> I'm afraid we are not able to upgrade golang more to than 1.19 in that presubmit, as the master branch of perf-tests is still being used in some places against k8s 1.23 which is using 1.19.

> Probably we need to pick other versions of dependencies, is it doable?

the feature i'm trying to test was introduced to 1.27 but let me just change the go.mod version to go1.19 and try to see if this helps

p0lyn0mial:
>/home/prow/go/pkg/mod/k8s.io/client-go@v0.27.0/tools/cache/synctrack/lazy.go:29:15: undefined: atomic.Pointer
note: module requires Go 1.20

OK, it doesn't work, it looks like the above error message applies to the `client-go` module which requires golang in version 1.20 (https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/go.mod#L5)

mborsz:
> would it be okay to temporarily disable the linter for the util-images/watch-list ?

Yes, I think we can do that.

You need to change `verify/test.sh` and `verify/verify-lint.sh` files (probably you can do that as part of the PR).

verify/test.sh
```diff
@@ -21,11 +21,20 @@ set -o pipefail
 KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
 cd "$KUBE_ROOT"
 
+# WARNING: we exclude the following paths from linting
+# because the CI lacks the required golang version (1.20)
+EXCLUDE_PATHS=("util-images/watch-list")
```
p0lyn0mial, line 26: should I create an issue in this repo so that we don't forget to fix it once we switch to go1.20?

p0lyn0mial:
@mborsz it looks like the CI is happy. Before we merge this PR could you please build and push the image to the registry?

verify/test.sh
mborsz: Yes, please do :)

mborsz:
> Maciej Borsz it looks like the CI is happy. Before we merge this PR could you please build and push the image to the registry?

I pushed this as `gcr.io/k8s-staging-perf-tests/watch-list:v0.0.1`

Full log https://gist.github.com/mborsz/4b17a5e3664a17315bd627c5525f6fae

/lgtm
/approve

PR #2258, opened 2023-05-05
Don't use default kubeconfig for clusterloader

wojtek-t:
Fix https://github.com/kubernetes/kubernetes/issues/104809

This effectively moved clusterloader calls to `workload-low` PriorityLevel (matched by service-accounts FlowSchema)


mborsz:
/hold cancel

Wojtek pinged me offline that kubemark works

PR #2260, opened 2023-05-09
API Streaming test: allow for overwrite enableWatchListFeature flag

p0lyn0mial:
#### What type of PR is this?
/kind feature

The flag controls whether to set` ENABLE_CLIENT_GO_WATCH_LIST_ALPHA` env var. 
It will allow to have two separate CI jobs (https://github.com/kubernetes/test-infra/pull/29464) for comparing results with the WatchList feature on/off.

xref: https://github.com/kubernetes/enhancements/issues/3157

Fixes #

### DIFF
--- a/clusterloader2/testing/watch-list/config.yaml
+++ b/clusterloader2/testing/watch-list/config.yaml
@@ -1,3 +1,4 @@
+{{$enableWatchListFeature := DefaultParam .CL2_ENABLE_WATCH_LIST_FEATURE false}}
 {{$testDuration := "5m"}}
 name: watch-list
 namespace:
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/watch-list/config.yaml
```diff
@@ -1,3 +1,4 @@
+{{$enableWatchListFeature := DefaultParam .ENABLE_WATCH_LIST_FEATURE false}}
```
marseel, line 1: you could change name to something with `CL2_` prefix, like this: `CL2_ENABLE_WATCH_LIST_FEATURE`.
Then you don't need `overrides/watch_list.yaml` file and can just specify `- --env=CL2_ENABLE_WATCH_LIST_FEATURE=true` in job spec :)

wojtek-t: +1

wojtek-t (commented):
Other than the single comment, LGTM

clusterloader2/testing/watch-list/config.yaml
p0lyn0mial: pushed, PTAL.

PR #2261, opened 2023-05-09
watch-list: define resources request and limit (mem, cpu) for the test pod

p0lyn0mial:
#### What type of PR is this?
/kind feature

16GB = 400MB (secrets) * 16 (informers) * 2 (store: old, new) + some headroom to accommodate future growth

Fixes #

### DIFF
--- a/clusterloader2/testing/watch-list/job.yaml
+++ b/clusterloader2/testing/watch-list/job.yaml
@@ -13,6 +13,13 @@ spec:
       containers:
         - name: {{.Name}}
           image: gcr.io/k8s-staging-perf-tests/watch-list:v0.0.1
+          resources:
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/watch-list/job.yaml
```diff
@@ -13,6 +13,13 @@ spec:
       containers:
         - name: {{.Name}}
           image: gcr.io/k8s-staging-perf-tests/watch-list:v0.0.1
+          resources:
+            requests:
+              memory: "16Gi"
+              cpu: "6"
```
p0lyn0mial, line 19: i'm not sure if we need 6 cores but since we will have the whole node let's be greedy.

PR #2265, opened 2023-05-12
Scrape k8s apiserver only for aws provider

marseel:
#### What type of PR is this?
/kind bug

Scrape k8s apiserver only metrics for AWS provider (added `ShouldPrometheusScrapeApiserverOnly : true` + sorted other options)
CL2 was unable to scrape kcm metrics:
```
4/8 targets are ready, example not ready target: {map[endpoint:kube-controller-manager instance:192.168.149.39:10257 job:master namespace:monitoring service:master] down}
```


shyamjvs:
Thanks Marseel!

PR #2266, opened 2023-05-22
perf-dash: add "benchmark" as name also for BenchmarkPerfScheduling

pohly:
#### What type of PR is this?

/kind bug
/kind failing-test

This is required because an empty name is no longer supported: the perf-dashboard is run with --allow-parsers-matching-all-tests=false which causes perfdash to skip current configuration for BenchmarkPerfResults as it does not have name set (https://github.com/kubernetes/perf-tests/blob/4674704f45e5d9785741b0399426cbfced37a648/perfdash/metrics-downloader.go#L165-L167).

The corresponding change for writing that file with "benchmark" is in https://github.com/kubernetes/kubernetes/pull/118164.

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -514,11 +514,7 @@ var (
 				Parser:           parsePerfData,
 			}},
 			"BenchmarkPerfResults": []TestDescription{{
-				// Expected file prefix string is constructed as
-				// OutputFilePrefix+"_"+Name. Given we currently generate
-				// file prefixed with "BenchmarkPerfScheduling_" followed by a date,
-				// set the Name to an empty string.
-				Name:             "",
+				Name:             "bechmark",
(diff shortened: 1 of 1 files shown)


perfdash/config.go
```diff
@@ -514,11 +514,7 @@ var (
 				Parser:           parsePerfData,
 			}},
 			"BenchmarkPerfResults": []TestDescription{{
-				// Expected file prefix string is constructed as
-				// OutputFilePrefix+"_"+Name. Given we currently generate
-				// file prefixed with "BenchmarkPerfScheduling_" followed by a date,
-				// set the Name to an empty string.
-				Name:             "",
+				Name:             "bechmark",
```
kerthcet, line 517: It should be `benchmark` cc @pohly

PR #2268, opened 2023-05-22
watch-list: allow for specifying custom api latency thresholds

p0lyn0mial:
#### What type of PR is this?
for watch-list-off scenario
a higher latency is expected
due to the size of the total payload

one this PR merges i'm going to open 
a new PR with an overwrite

Fixes #


clusterloader2/testing/watch-list/config.yaml
```diff
@@ -73,3 +74,4 @@ steps:
         enableViolations: true
         useSimpleLatencyQuery: true
         summaryName: APIResponsivenessPrometheus_simple
+        customThresholds: {{YamlQuote $customApiCallThresholds 4}}
```
p0lyn0mial, line 77: I specified custom threshold only for  `gathering` phase.

PR #2272, opened 2023-05-22
clusterloader2: Use immediate poll for prom health checks

learnitall:
#### What type of PR is this?

/kind bug

This commit changes `poll.Wait` to `poll.WaitImmediate` in the logic that is used to wait for Prometheus to be ready. Before this commit, cl2 would always wait 30 seconds before checking if Prometheus is healthy, as this is the default interval used to wait between unsuccessful checks. By using `WaitImmediate`, this initial 30s is skipped, which speeds up consecutive runs where Prometheus is already deployed and ready.


marseel:
Good idea, please sing CLA :) 
/ok-to-test

PR #2274, opened 2023-05-24
[WIP] Allow for creating CL2 kubeconfig objects in non-default namespace

wojtek-t:


clusterloader2/run-e2e.sh
```diff
@@ -80,23 +81,25 @@ if [[ "${DEPLOY_AZURE_CSI_DRIVER:-false}" == "true" ]]; then
 fi
 
 # Create a dedicated service account for cluster-loader.
-kubectl --kubeconfig "${KUBECONFIG}" create serviceaccount cluster-loader
-kubectl --kubeconfig "${KUBECONFIG}" create clusterrolebinding cluster-loader --clusterrole=cluster-admin --serviceaccount=default:cluster-loader
+# Create the namespace if it doesn't exist.
+kubectl --kubeconfig "${KUBECONFIG}" create namespace "${KUBECONFIG_NAMESPACE}" --dry-run=client -o yaml | kubectl --kubeconfig "${KUBECONFIG}" apply -f -
+kubectl --kubeconfig "${KUBECONFIG}" create serviceaccount --namespace="${KUBECONFIG_NAMESPACE}" cluster-loader
+kubectl --kubeconfig "${KUBECONFIG}" create clusterrolebinding --namespace="${KUBECONFIG_NAMESPACE}" cluster-loader --clusterrole=cluster-admin --serviceaccount=default:cluster-loader
```
jprzychodzen, line 87: AFAIK clusterrole and clusterrolebinding is global, not namespace scope, we don't need `--namespace="${KUBECONFIG_NAMESPACE}"` here. Also, we should target something like `--serviceaccount=${KUBECONFIG_NAMESPACE}:cluster-loader`

PR #2276, opened 2023-05-25
watch-list: define a job type for the perf-dashboard

p0lyn0mial:
#### What type of PR is this?
a separate job type will allow to display metrics
collected during watchlist per tests like
https://perf-dash.k8s.io/#/?jobname=gce-100Nodes-master&metriccategoryname=E2E&metricname=LoadResources&PodName=kube-apiserver-e2e-big-master%2Fkube-apiserver&Resource=memory

Fixes #

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -623,6 +623,23 @@ var (
 		},
 	}
 
+	watchListDescriptions = TestDescriptions{
+		"E2E": {
+			"LoadResources": []TestDescription{{
+				Name:             "watch-list",
+				OutputFilePrefix: "ResourceUsageSummary",
+				Parser:           parseResourceUsageData,
+			}},
+		},
+		"APIServer": {
+			"LoadResponsiveness_PrometheusSimple": []TestDescription{{
+				Name:             "watch-list",
+				OutputFilePrefix: "APIResponsivenessPrometheus_simple",
+				Parser:           parsePerfData,
+			}},
+		},
+	}
+
 	jobTypeToDescriptions = map[string]TestDescriptions{
 		"performance":  performanceDescriptions,
 		"benchmark":    benchmarkDescriptions,
@@ -630,6 +647,7 @@ var (
 		"storage":      storageDescriptions,
 		"throughput":   throughputDescriptions,
 		"windows":      windowsDescriptions,
+		"watchlist":    watchListDescriptions,
 	}
 )
 


perfdash/config.go
```diff
@@ -623,13 +623,24 @@ var (
 		},
 	}
 
+	watchListDescriptions = TestDescriptions{
+		"E2E": {
+			"LoadResources": []TestDescription{{
+				Name:             "load",
```
marseel, line 629: This should be `watch-list` I think? Based on file in artifacts ` ResourceUsageSummary_watch-list_2023-05-24T08:35:20Z.json`

perfdash/config.go
```diff
@@ -623,13 +623,24 @@ var (
 		},
 	}
 
+	watchListDescriptions = TestDescriptions{
```
wojtek-t, line 626: Why do we even need to introduce this? Why can't we use the existing performance?

p0lyn0mial: honesty i don't understand why it doesn't work.

p0lyn0mial: the tests use the `performance` type - https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-experimental-periodic-jobs.yaml#L198 atm

perfdash/config.go
p0lyn0mial: I would also add something that would allow me to show latencies

perfdash/config.go
p0lyn0mial: I started a slack thread if you want to chime in - https://kubernetes.slack.com/archives/C09QZTRH7/p1684911329191689

marseel: It doesn't work due to the fact that perfdash has predefined file names here: https://github.com/kubernetes/perf-tests/pull/2276/files#diff-17340b391bfccaca6d6cf3409b032676038adfa8dead9fb15f65557a99bd098eR82
and `watch-list/config.yaml` has test name `watch-list` and not `load`

perfdash/config.go
p0lyn0mial: perhaps
```
"LoadResponsiveness": []TestDescription{{
				Name:             "watch-list",
				OutputFilePrefix: "APIResponsiveness",
			}},
```

?

p0lyn0mial: or

```
"LoadResponsiveness_PrometheusSimple": []TestDescription{{
				Name:             "watch-list",
				OutputFilePrefix: "APIResponsivenessPrometheus_simple",
			}},
```
?

perfdash/config.go
wojtek-t: This is... weird.

Can we then instead parametrized the map and reuse it? (i.e. make load/watch-list a parameter of function returning this map)?

p0lyn0mial: If we want to keep the new type then I don't see value in making the map reusable. IMO it would make the code less readable. 

What I would like to see is to make the whole mechanism more generic. Ideally if add a new job would not require defining new types.

wojtek-t: > I would appreciate if the whole mechanism could be more generic. Ideally if add a new job would not require defining new types.

I would like that too, but I don't want to overinvest in perfdash...

perfdash/config.go
Argh4k: If you want to you can run perfdash locally to play with and find all the metrics that you need. To do that modify perfdash/Makefile by replacing githubConfigDir with  `https://api.github.com/repos/kubernetes/test-infra/contents/config/jobs/kubernetes/sig-scalability` and replace builds with some reasonable number like 2 so you don't have to wait minutes for perdash to scrape all the data. After that just run `make run` from perfdash directory and you should have perfdash hosted at localhost:8080

wojtek-t:
OK, let's merge it to unblock stuff. And we can improve in a follow up.

/lgtm
/approve

p0lyn0mial:
thanks, i have also opened https://github.com/kubernetes/test-infra/pull/29604

PR #2281, opened 2023-06-05
watch-list/job: run two replicas per namespace

p0lyn0mial:
#### What type of PR is this?
/kind feature

Running more than one instance of the watch-list perf test we will put more load on the api-server.
An increased load on the server will allows us for better comparison/understanding scaling characteristic of the streaming API feature.

A single instance (`job`) of the performance test requires `6 CPUs, 16 RAM`. The test runs on an `e2-standard-16` machine with `16 CPU, 64 GB`. 

Fixes #

### DIFF
--- a/clusterloader2/testing/watch-list/config.yaml
+++ b/clusterloader2/testing/watch-list/config.yaml
@@ -36,13 +36,13 @@ steps:
       Method: APIResponsivenessPrometheus
       Params:
         action: start
-- name: Start the secret informer
+- name: Start the secret informers
   phases:
     - namespaceRange:
         min: 1
         max: 1
         basename: watch-list
-      replicasPerNamespace: 1
+      replicasPerNamespace: 2
       tuningSet: Uniform10qps
       objectBundle:
         - basename: watch-list-secret


clusterloader2/testing/watch-list/config.yaml
```diff
@@ -36,13 +36,13 @@ steps:
...
         action: start
-- name: Start the secret informer
+- name: Start the secret informers
   phases:
     - namespaceRange:
         min: 1
         max: 1
         basename: watch-list
-      replicasPerNamespace: 1
+      replicasPerNamespace: 2
```
wojtek-t, line 45: This looks fine, but I think we probably need to first increase the cluster-size to have 2 nodes (I think 2 instances won't fit on a current single node, right?)

p0lyn0mial: I think we should be fine.

A single instance (`job`) of the performance test requires `6 CPUs`, `16 RAM`. The test runs on an `e2-standard-16` machine with `16 CPU, 64 GB`. It seems we should be able to run the second `job`.

wojtek-t: ok - let's see

PR #2284, opened 2023-06-05
Dns switch to endpointslices

aojea:
/kind cleanup


aojea:
/assign @wojtek-t 
/cc @dpasiukevich 

It took me more than expected but we finally get it working

verify/verify-lint.sh
```diff
@@ -37,6 +37,8 @@ MODULE_BASED=$(find . -type d -name vendor -prune \
 set +e
 status=0
 
+golangci-lint version
```
dpasiukevich, line 40: Is it needed?

dpasiukevich (commented):
Thanks for the PR! I'm curious - what is the root cause of the lint failure?

verify/verify-lint.sh
aojea: is it useful for debugging

aojea:
> Thanks for the PR! I'm curious - what is the root cause of the lint failure?

the golangci-lint version was old

https://github.com/kubernetes/test-infra/pull/29690

dlapcevic (approved):
lgtm

dlapcevic:
We need approval from OWNERS (https://github.com/kubernetes/perf-tests/blob/master/OWNERS).
@wojtek-t could you please approve?

aojea:
/assign @marseel @mborsz 

they may help us here too

mborsz:
/lgtm
/approve

BTW: Maybe we can remove vendor completely? I don't think it's needed nowadays. Sample PR for other components removing vendor: https://github.com/kubernetes/perf-tests/pull/1934

aojea:
> BTW: Maybe we can remove vendor completely? I don't think it's needed nowadays. Sample PR for other components removing vendor: #1934

I was carrying on with current state, I was tempted to do it honestly, does not seem necessary

dlapcevic:
I just built and released the new image `1.4.0` of dnsperfgo and it's 56 MB in size. It grew 10 MB from the previous version `1.3.0`.
That's quite big for such a small app. Is it due to vendor libraries?

dpasiukevich:
> Is it due to vendor libraries?

I expect it to be the case. The prev kubernetes versions were kinda old
```
k8s.io/apimachinery v0.20.2
k8s.io/client-go v0.20.2
```

And now it's 0.27.2.

aojea:
```
cd dns/dnsperfgo
go build
ls -altrh 
-rwxr-x---  1 aojea primarygroup  45M Jun  7 09:57 dnsperfgo
go install github.com/nikolaydubina/go-binsize-treemap@latest
go tool nm -size dnsperfgo | go-binsize-treemap > binsize.svg

```
binsize

aojea:
I still don't understand well the diff between the reported size and the actual size

PR #2285, opened 2023-06-07
Bump dnsperfgo to 1.4.0

dlapcevic:
Release dnsperfgo `1.4.0` to include the change to use EndpointSlices instead of Endpoints https://github.com/kubernetes/perf-tests/pull/2284


dlapcevic:
cc @dpasiukevich @aojea

PR #2286, opened 2023-06-07
CL2 load test bump dnsperfgo image to 1.4.0

dlapcevic:
New dnsperfgo image is released `1.4.0` to use EndpointSlices instead of Endpoints, effectively supporting over 1k endpoints behind a single service https://github.com/kubernetes/perf-tests/pull/2285


dlapcevic:
cc @dpasiukevich

PR #2290, opened 2023-06-14
chore: pkg imported more than once

testwill:
#### What type of PR is this?

/kind cleanup

package "k8s.io/api/core/v1" is being imported more than once (ST1019)
Fixes #


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects_test.go
```diff
@@ -52,21 +51,21 @@ var (
 	image         = "gcr.io/some-project/some-image"
 )
 
-var node1 = corev1.Node{
```
prateekgogia, line 55: v1 by itself doesn't make much sense unless someone check whats v1 referring. 
I would argue, if we are making this effort we should go in another direction and `'s/v1/corev1/g'`

PR #2295, opened 2023-07-10
Take full duration into account in Prometheus-based measurements

tosi3k:
This makes sure that we take a look at the full interval which we are supposed to measure some Prometheus-based SLI rather than truncating it to full minutes.

/kind bug
/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/measurement/util/prometheus.go
+++ b/clusterloader2/pkg/measurement/util/prometheus.go
@@ -174,8 +174,5 @@ func (qr *promResponseData) UnmarshalJSON(b []byte) error {
 
 // ToPrometheusTime returns prometheus string representation of given time.
 func ToPrometheusTime(t time.Duration) string {
-	if t < time.Minute {
-		return fmt.Sprintf("%ds", int64(t)/int64(time.Second))
-	}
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/util/prometheus.go
```diff
@@ -174,8 +174,13 @@ func (qr *promResponseData) UnmarshalJSON(b []byte) error {
...
+	minutes := int64(t) / int64(time.Minute)
+	seconds := (int64(t) % int64(time.Minute)) / int64(time.Second)
+	if minutes == 0 {
+		return fmt.Sprintf("%ds", seconds)
 	}
-	return fmt.Sprintf("%dm", int64(t)/int64(time.Minute))
+	if seconds == 0 {
+		return fmt.Sprintf("%dm", minutes)
+	}
+	return fmt.Sprintf("%dm%ds", minutes, seconds)
```
mborsz, line 185: Can we just produce output in seconds instead of doing some calculations here?

Like 1012312312s?

tosi3k: Done

Issue #2275, opened 2023-05-25
Add module to exercise list calls in CL2


mengqiy:
**What would you like to be added**:

Clusterloader2 seems to lack tests to exercise list calls.

**Why is this needed**:

We have an SLO for list calls in https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md. But CL2 doesn't explicitly exercise it.

anson627:
> **What would you like to be added**:
> 
> Clusterloader2 seems to lack tests to exercise list calls.
> 
> **Why is this needed**:
> 
> We have an SLO for list calls in https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md. But CL2 doesn't explicitly exercise it.

agree, list calls are the most expensive, and  most likely to overload API server

mengqiy:
[watch-list](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/watch-list) sololy exercises the list calls. We can leverage it be embed it in the load test to create a mixture of all workload types.

PR #2302, opened 2023-07-20
support list objects in load tests

mengqiy:
#### What type of PR is this?

/kind feature
/kind api-change

Provide a load test coverage to exercise list calls.

Fixes https://github.com/kubernetes/perf-tests/issues/2275

This PR requires https://github.com/kubernetes/perf-tests/pull/2258.

### DIFF
--- a/clusterloader2/testing/list/README.md
+++ b/clusterloader2/testing/list/README.md
@@ -1,0 +1,16 @@
+## List load test
+
+List load test perform the following steps:
+- Create configmaps and secrets in a set of namespaces.
+  - The namespaces used here are managed by CL2.
+  - The number of namespaces can be specified using `CL2_LIST_NAMESPACE_NUMBER`.
+  - The size of a single configmap and secret can be speficed using `CL2_LIST_CONFIG_MAP_BYTES` and `CL2_LIST_SECRET_BYTES` respectively.
+  - The number of configmaps and secrets can be specified using `CL2_LIST_CONFIG_MAP_NUMBER` and `CL2_LIST_SECRET_NUMBER` respectively.
+- Create RBAC rules to allow lister pods to access these configmaps and secrets
+- Create lister pods using deployment in a separate namespace to list configmaps and secrets.
+  - The namespace is created using `namespace.yaml` as a template, but its lifecycle is managed by CL2.
+  - The number of replicas for the lister pods can be specified using `CL2_LIST_BENCHMARK_PODS`.
+  - Each lister pod will maintain `CL2_LIST_CONCURRENCY` number of inflight list requests.
+- Measurement uses [`APIResponsivenessPrometheusSimple`](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/README.md) to meansure API latency for list configmaps and secrets calls.
+
+The lister pods leverage https://github.com/kubernetes/perf-tests/tree/master/util-images/request-benchmark to create in-cluster list load.
--- a/clusterloader2/testing/list/README.md
+++ b/clusterloader2/testing/list/README.md
@@ -3,7 +3,7 @@
 List load test perform the following steps:
 - Create configmaps and secrets in a set of namespaces.
   - The namespaces used here are managed by CL2.
-  - The number of namespaces can be specified using `CL2_LIST_NAMESPACE_NUMBER`.
+  - The number of namespaces is derived from `NODES_PER_NAMESPACE`. Namespaces will be 2, if there are no nodes.
   - The size of a single configmap and secret can be speficed using `CL2_LIST_CONFIG_MAP_BYTES` and `CL2_LIST_SECRET_BYTES` respectively.
   - The number of configmaps and secrets can be specified using `CL2_LIST_CONFIG_MAP_NUMBER` and `CL2_LIST_SECRET_NUMBER` respectively.
 - Create RBAC rules to allow lister pods to access these configmaps and secrets
--- a/clusterloader2/testing/list/clusterrole.yaml
+++ b/clusterloader2/testing/list/clusterrole.yaml
@@ -6,7 +6,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - configmaps
-  - secrets
+  - "*"
   verbs:
   - list
--- a/clusterloader2/testing/list/clusterrole.yaml
+++ b/clusterloader2/testing/list/clusterrole.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: list-clusterrole
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - secrets
+  verbs:
+  - list
--- a/clusterloader2/testing/list/clusterrolebinding.yaml
+++ b/clusterloader2/testing/list/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{.Name}}
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: list-benchmark-0
+roleRef:
+  kind: ClusterRole
+  name: list-clusterrole-0
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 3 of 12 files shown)


mborsz:
Few questions:
* What is a context here? Is it a part of some larger effort/some KEP? Why just LISTs and not other request types as well? What signal you want to get? latency of the calls, control plane resource consumption, some APF metrics?
* Does it really need to be a part of load/config.yaml? The test is already quite complex and it looks like the lists don't really need to be part of that test. Probably a separate test would work better.
* Why the decision on implementing this natively in cluster loader? Overall, if possible the approach where we implement some in-cluster running prober exporting prometheus metric is preferred as allows to decouple the complexity from the framework
* We already have e.g. https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/request-benchmark test that is able to excercise any request type you want. It's running periodically e.g. https://testgrid.k8s.io/sig-scalability-experiments#consistent-list-from-cache-on to excercise a performance of list calls in some scenarios. Is that test in some way not sufficient? Maybe you can extend the test?

mengqiy:
> That what we discussed during the meeting - it should be an in-cluster component sending those lists.

I just realized that I missed this part after revisiting the meeting notes. I will revise it.

mengqiy:
Updated the test to use the request benchmark image as the in-cluster requests sender.
The list test in its own test instead of part of `load` test.
@mborsz @wojtek-t PTAL

clusterloader2/testing/list/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: {{.Name}}
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: list-benchmark-0
+roleRef:
+  kind: ClusterRole
+  name: list-clusterrole-0
```
shyamjvs, line 11: seems like this mismatches with the role name used above `list-clusterrole`?

shyamjvs:
Meta comment - did we discuss the option of extending the existing load tests with the list call functionality (https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/load) instead of defining a separate test scenario dedicated to it? The former will automatically include this as part of our periodic CI jobs.

clusterloader2/testing/list/clusterrolebinding.yaml
shyamjvs: or maybe the suffix is added in a parametrized way?

shyamjvs:
ah ignore the above, I just caught up with this comment from wojtek:

> I wanted it to be a separate module and separate phase. But a separate test is probably even better.

clusterloader2/testing/list/clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: list-clusterrole
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - secrets
```
shyamjvs, line 10: maybe allow * instead of updating the types each time?

clusterloader2/testing/list/config.yaml
```diff
@@ -0,0 +1,189 @@
...
+  parallelismLimitedLoad:
+    parallelismLimit: 10
+steps:
+- name: Setup namespace for list benchmark pods
+  phases:
+  - replicasPerNamespace: 1
+    tuningSet: Sequence
+    objectBundle:
+    - basename: list-benchmark
+      objectTemplatePath: namespace.yaml
```
shyamjvs, line 33: Trying to understand this - so we're creating the objects itself in $namespaceNumber many namespaces and creating a separate namespace to run the pods that list those objects?

clusterloader2/testing/list/config.yaml
```diff
@@ -0,0 +1,189 @@
...
+      min: 1
+      max: {{$namespaceNumber}}
+    tuningSet: Sequence
+    replicasPerNamespace: {{$secretNumber}}
+    objectBundle:
+    - basename: {{$secretGroup}}
+      objectTemplatePath: secret.yaml
+      templateFillMap:
+        bytes: {{$secretBytes}}
+        group: {{$secretGroup}}
```
shyamjvs, line 80: might wanna combine these two phases into a single step - they can happen in parallel

clusterloader2/testing/list/deployment.yaml
```diff
@@ -0,0 +1,37 @@
+{{$image := DefaultParam .CL2_BENCHMARK_IMAGE "gcr.io/k8s-testimages/perf-tests-util/request-benchmark:latest"}}
```
shyamjvs, line 1: Where is the source code for this image? Would help to link that in a comment here so folks can cross-reference.

clusterloader2/testing/list/config.yaml
```diff
@@ -0,0 +1,189 @@
...
+{{$secretNumber := DefaultParam .CL2_LIST_SECRET_NUMBER 3000}}
+
+{{$listReplicas := DefaultParam .CL2_LIST_BENCHMARK_PODS 1}}
+{{$inflight := DefaultParam .CL2_LIST_CONCURRENCY 2}}
+{{$qps := DefaultParam .CL2_LIST_QPS -1}}
+
+{{$namespacedPodCpu := DefaultParam .CL2_LIST_BENCHMARK_POD_CPU "400m"}}
+{{$namespacedPodMemory := DefaultParam .CL2_LIST_BENCHMARK_POD_MEMORY "800Mi"}}
+{{$clusterScopedPodCpu := DefaultParam .CL2_LIST_BENCHMARK_POD_CPU "2000m"}}
+{{$clusterScopedPodMemory := DefaultParam .CL2_LIST_BENCHMARK_POD_MEMORY "4000Mi"}}
```
shyamjvs, line 17: Any opportunity to get rid of some of these variables either by using some sane defaults or auto-calculating based on the load parameters?

More options usually mean more confusion and harder to adopt by other devs :)

shyamjvs:
Another meta comment: it would be useful (for other devs to reuse this utility) if you can add a small readme under `testing/list/` explaining the setup (how we create namespaces, objects in those namespaces, a "lister" pods deployment and necessary roles/bindings, measurements).

clusterloader2/testing/list/config.yaml
mengqiy: Correct.
Suppose there are N namespace for creating objects (e.g. configmaps), we will only be able to have N, 2N, 3N ... list benchmarking pods depending on how many replicas of list benchmarking pods per namespace.
But if we separate these 2 kinds of namespaces, we can have any number of list benchmarking pods. IMO this approach is more flexible.

clusterloader2/testing/list/deployment.yaml
mengqiy: Will add.
It's here: https://github.com/kubernetes/perf-tests/tree/master/util-images/request-benchmark

clusterloader2/testing/list/clusterrole.yaml
mengqiy: Done

clusterloader2/testing/list/config.yaml
mengqiy: Done

clusterloader2/testing/list/config.yaml
mengqiy: Done

mengqiy:
@shyamjvs PTAL

clusterloader2/testing/list/config.yaml
```diff
@@ -0,0 +1,192 @@
+# List benchmark
+{{$namespaceNumber := DefaultParam .CL2_LIST_NAMESPACE_NUMBER 2}}
```
hakuna-matatah, line 2: can we just use `NODES_PER_NAMESPACE` variable that we use for other test configs ? Wondering if we really need this ? Could it be a factor of nodes and determine from that ?

mengqiy: Updated

clusterloader2/testing/list/README.md
```diff
@@ -0,0 +1,16 @@
+## List load test
+
+List load test perform the following steps:
+- Create configmaps and secrets in a set of namespaces.
+  - The namespaces used here are managed by CL2.
+  - The number of namespaces can be specified using `CL2_LIST_NAMESPACE_NUMBER`.
```
hakuna-matatah, line 6: This has to be changed to reflect [this](https://github.com/kubernetes/perf-tests/pull/2302#discussion_r1370579581)

mengqiy: Thanks for catching this.
Updated

mengqiy:
@hakuna-matatah PTAL

PR #2306, opened 2023-07-28
issue:2287: find maximum number of informers that doesn't impact latency.

p0lyn0mial:
#### What type of PR is this?
After https://github.com/kubernetes/perf-tests/pull/2305 the latency increased to `30s`. 
With https://github.com/kubernetes/perf-tests/pull/2303 it was around `1s`.

Between `1s` and `30s` the CPU usage didn't significantly increase.
Everything indicates that the egress is being throttled.

According to my calculations we should be below the egress limit (2 GB/s) with `6` or `5` informers.
This PR set the value of the informers to `6`.

Once we find our sweet spot I'm going to add a metric that will measure the latency for streaming.
Then we update the value of informers to initial value of `16` and compare the latencies one more time.

xref: https://github.com/kubernetes/perf-tests/issues/2287

Fixes #

### DIFF
--- a/clusterloader2/testing/watch-list/job.yaml
+++ b/clusterloader2/testing/watch-list/job.yaml
@@ -22,5 +22,5 @@ spec:
               cpu: "6"
           command: [ "watch-list" ]
           # TODO(p0lyn0mial): bring back 16 informers
-          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=8", "--namespace=watch-list-1", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
+          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=6", "--namespace=watch-list-1", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
       restartPolicy: Never


wojtek-t:
> With https://github.com/kubernetes/perf-tests/pull/2303 it was around 1s.

It was more like 6-7 seconds looking into perf-dash.

/lgtm
/approve

clusterloader2/testing/watch-list/job.yaml
```diff
@@ -22,5 +22,5 @@ spec:
               cpu: "6"
           command: [ "watch-list" ]
           # TODO(p0lyn0mial): bring back 16 informers
-          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=8", "--namespace=watch-list-1", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
+          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=6", "--namespace=watch-list-1", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
```
aojea, line 25: nit, we should expand the comment to explain this magic number, is going to be hard to trace back from the git logs

p0lyn0mial: This is just temporary to gather some data. I am going to bring back the previous value once we finish with the experiment.

PR #2312, opened 2023-09-04
[WIP] Add Hubble test suite

siwiutki:
#### What type of PR is this?

/kind feature

This PR adds a test suite for Hubble metrics and relay.

This is a draft version of this PR, but it's rather close to final intended version. There are still several issues, which are listed below, but the layout of this test suite and steps are how it's intended to look like in the final version.

Prometheus server in this test suite uses up to around 62 GB of memory, assuming current small load of ~5 flow/sec per node. Further tests with higher flow load are yet to be done to assess Prometheus memory usage in that scenario. 

This test suite assumes that Hubble and Cilium is deployed the same way as on GKE cluster with GKE DPv2 observability and metrics enabled. That includes:
* cilium agent daemonset has port named "flowmetrics" with Hubble metrics exposed
* certgen and secrets are generated under the same names, in kube-system namespace
* hubble-relay is deployed as deployment/hubble-relay in kube-system namespace, with open ports 4245 for server and port with name "metrics" for hubble-relay metrics
* hubble-relay deployment has k8s-app: hubble-relay label
* svc/hubble-relay is set up for hubble-relay in kube-system, with port forwarding 4245 to 443

Any comments to make it more generic are appreciated.

Issues:
1. ~~Test doesn't fail when GenericPrometheusQuery measures a value that's above threshold~~ - apparently there is `enableViolations` param that can be added to this measurement that achieves exactly that. By default it's false, though I'd expect it to be true. Well, added `enableViolations: true`.
1. grpc_server_handled_total metric scraped from hubble-relay component has only zero value where non-zero is expected
1. ~~As of now, grpc_server_handled_total metric scraped from hubble-relay component can only output "OK" grpc_code due to implementation of this metric in OSS. This is planned to be addressed in OSS, but in the case it won't change, GenericPrometheusQuery step for that metric should be removed~~ - this is actually not true, and this comment came from wrong understanding of this metric. Metric correctly tracks gRPC call status (and there can be codes other than "OK"), but it's not enough to track hubble-relay availability. https://github.com/cilium/cilium/pull/28217 is planned to address that.
1. ~~High Prometheus server usage, due to nature of this test, which might be mitigated by assigning a highmem instance for Prometheus~~ - limited the scraped metrics, now only pod_flow_handler metric is scraped from hubble-relay


clusterloader2/go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	k8s.io/client-go v0.22.15
 	k8s.io/component-base v0.22.15
 	k8s.io/component-helpers v0.22.15
+	k8s.io/klog v1.0.0 // indirect
```
aojea, line 58: let's try to find out who is pulling this and fix it there, we should only have `k8s.io/klog/v2`

clusterloader2/pkg/measurement/common/hubble/manifests/hubble-cli.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: hubble-cli-cl2
+  namespace: kube-system
+spec:
+  containers:
+    - name: hubble-cli
+      image: quay.io/cilium/hubble:latest
```
aojea, line 9: is this the image we want to test?

aojea: I'm not familiar enough with this framework, or is this image replaced later on the `setup` step?

clusterloader2/pkg/measurement/common/hubble/manifests/hubble-cli.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: v1
+kind: Pod
```
dlapcevic, line 2: Maybe it should be a deployment.

dlapcevic (approved):

PR #2314, opened 2023-09-08
service_creation_latency: limit number of calls, exit on success

cezarygerard:
/kind cleanup

### DIFF
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -48,7 +48,7 @@ const (
 	defaultServiceCreationLatencyTimeout = 10 * time.Minute
 	defaultCheckInterval                 = 10 * time.Second
 	pingBackoff                          = 1 * time.Second
-	pingChecks                           = 10
+	pingChecks                           = 3
 
 	creatingPhase     = "creating"
 	ipAssigningPhase  = "ipAssigning"
@@ -393,6 +393,7 @@ func (p *pingChecker) run() {
 			success++
 			if success == pingChecks {
 				p.creationTimes.Set(key, phaseName(reachabilityPhase, p.svc.Spec.Type), time.Now())
+				return
 			}
 		}
 	}
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -351,9 +351,6 @@ func (p *pingChecker) run() {
 		case <-p.stopCh:
 			return
 		default:
-			if _, exists := p.creationTimes.Get(key, phaseName(reachabilityPhase, p.svc.Spec.Type)); exists {
-				return
-			}
 			// TODO(#685): Make ping checks less communication heavy.
 			pod, err := execservice.GetPod()
 			if err != nil {


clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -393,6 +393,7 @@ func (p *pingChecker) run() {
 			success++
 			if success == pingChecks {
 				p.creationTimes.Set(key, phaseName(reachabilityPhase, p.svc.Spec.Type), time.Now())
+				return
```
aojea, line 396: I don't think  that is necessary wrong to assert 10 times, maybe is indeed excessive , but the main problem here is that this doing an exec per command.

Another flaw in this code is that if no `ips`are obtained , this will consider it as a success.

We need to do at least two changes

aojea:
please see suggestion

```diff
diff --git a/clusterloader2/pkg/measurement/common/service_creation_latency.go b/clusterloader2/pkg/measurement/common/service_creation_latency.go
index 6b69dedfb..8176851e0 100644
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -345,7 +345,6 @@ func (p *pingChecker) run() {
                klog.Errorf("%s: meta key created error: %v", p.callerName, err)
                return
        }
-       success := 0
        for {
                select {
                case <-p.stopCh:
@@ -358,7 +357,6 @@ func (p *pingChecker) run() {
                        pod, err := execservice.GetPod()
                        if err != nil {
                                klog.Warningf("call to execservice.GetPod() ended with error: %v", err)
-                               success = 0
                                time.Sleep(pingBackoff)
                                continue
                        }
@@ -377,23 +375,23 @@ func (p *pingChecker) run() {
                                }
                                port = p.svc.Spec.Ports[0].Port
                        }
+                       if len(ips) == 0 {
+                               klog.Warningf("no ips found for service: %+v", p.svc)
+                               time.Sleep(pingBackoff)
+                               continue
+                       }
                        for _, ip := range ips {
-                               address := net.JoinHostPort(ip, fmt.Sprint(port))
-                               command := fmt.Sprintf("curl %s", address)
+                               command := fmt.Sprintf(`date; for i in $(seq 1 %d); do echo "$(date) Try: ${i}"; curl -g -q -s --max-time 15 --connect-timeout 1 %s; echo; done`, pingChecks, net.JoinHostPort(ip, fmt.Sprint(port)))
                                _, err = execservice.RunCommand(context.TODO(), pod, command)
                                if err != nil {
                                        break
                                }
                        }
                        if err != nil {
-                               success = 0
                                time.Sleep(pingBackoff)
                                continue
                        }
-                       success++
-                       if success == pingChecks {
-                               p.creationTimes.Set(key, phaseName(reachabilityPhase, p.svc.Spec.Type), time.Now())
-                       }
+                       p.creationTimes.Set(key, phaseName(reachabilityPhase, p.svc.Spec.Type), time.Now())
                }
        }
 }
```

clusterloader2/pkg/measurement/common/service_creation_latency.go
aojea: oh, so this never returned before 🤦

cezarygerard: we should get rid for this lines

for the simple return here:

```
			if success == pingChecks {
				p.creationTimes.Set(key, phaseName(reachabilityPhase, p.svc.Spec.Type), time.Now())
				return
                         }
```

aojea: agree, is confusing to do the check at the end if we are explicitily setting it here ... unless there are multiple of these running in parlallel? do you know if that is even possible? the same service gets called several times?

aojea:
/assign @wojtek-t 

doing 3 checks instead of 10 does not sounds bad, specially if each check is a new `kubectl exec` call, that is very expensive, an alternative is to do the multiple checks in one call to the pod as in https://github.com/kubernetes/perf-tests/pull/2314#issuecomment-1723288007

The logic to return from the forever loop checks for the condition to exit at the beginning of the loop, I do not know if this plugins may execute the same service in parallel and this is needed to avoid overlapping https://github.com/kubernetes/perf-tests/pull/2314#discussion_r1329203519

PR #2318, opened 2023-09-26
Add CL2 prefix to parameter names in network policy metrics

dlapcevic:
Make parameter names in network policy metrics manifest consistent by adding the CL2 prefix.

/kind cleanup


marseel:
/lgtm
/approve
Thanks!

PR #2393, opened 2023-10-24
feat: Add support for specifying a list of existing namespaces

### LINKED ISSUE #2392: Add support for specifying list of namespaces for deployments

It would be good to have the ability to specify where certain resources are deployed in cases where a specific pre-existing namespace must be used. This would be an alternative to using generated namespaces or the "use existing namespaces" PR from ~2020. The mentioned PR still enforces namespace suffixes which is not always ideal. 

**Why is this needed**:

Certain k8s resources are only meant to be deployed in certain namespaces. This would allow the user to specify such processes.

kmichelsenG:
#### What type of PR is this?

/kind feature

This PR allows users to specify a list of pre-existing namespaces rather than using the generated ones. It is important in scenarios where certain resources are restricted to only be deployed in specific namespaces. 

Fixes #2392

### DIFF
--- a/clusterloader2/api/types.go
+++ b/clusterloader2/api/types.go
@@ -94,10 +94,17 @@ type Module struct {
 // In a given namespace range (or cluster scope if no range is specified)
 // it defines the number and the configuration of managed objects.
 type Phase struct {
-	// NamespaceRange defines the set of namespaces in which objects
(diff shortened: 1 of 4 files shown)


mborsz:
/ok-to-test

Looks good to me. Please sign CLA.

PR #2409, opened 2023-11-10
Add a template helper function for string slices

x13n:
#### What type of PR is this?

/kind feature

This is necessary for generating templates that iterate over values from a dynamically provided variable.

Fixes #


wojtek-t:
Retest won't help - test are broken now. We need to fix them...

PR #2416, opened 2023-11-15
Enable KubeState metrics for EKS and AWS providers

hakuna-matatah:
#### What type of PR is this?
Enable KubeState metrics for EKS and AWS providers. 
This is required to gather prom metrics dataset about state changes of k8s objects during the tests.
Enabling with just `PROMETHEUS_SCRAPE_KUBE_STATE_METRICS` flag doesn't work for EKS and AWS providers due to this explicit check with `AND` operation [here](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/prometheus.go#L266)

Fixes #


hakuna-matatah:
@wojtek-t PTAL when you get a chance

Issue #2417, opened 2023-11-15
Validate In cluster network Latencies using probes when enabled


hakuna-matatah:
**What would you like to be added**:
 Ability to measure/validate In cluster network latencies SLOs for cloud-providers during CL2 load test when enabled with thresholds. 
It should be backwards compatible, so in order for that, set the default value to "0s" to not validate by default today based on condition [here](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/measurement/common/probes/probes.go#L231) 

**Why is this needed**:

I have discussed this in Sig-Scalability meeting couple weeks back with @wojtek-t  about needing this support to measure these In Cluster network latency SLOs that are already defined [here](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/network_latency.md) 

It gives us the opportunity to catch regressions if any over time.

PR #2408, opened 2023-11-10
Feature gate to enable N/w Programming Latencies SLO measurement for …

hakuna-matatah:
…Load tests

#### What this PR does / why we need it:

This PR gives the ability to measure/validate Network Programming Latencies SLOs for cloudproviders during CL2 load test when enabled. 
It's backwards compatible, it defaults to feature being disabled.

#### Which issue(s) this PR fixes:

Fixes #
https://github.com/kubernetes/perf-tests/issues/2407

PR #2418, opened 2023-11-15
Feature gate to enable In cluster n/w Latencies SLO measurement for cloud providers

hakuna-matatah:
…Load tests

This PR gives the ability to measure/validate In cluster network latencies SLOs for cloudproviders during CL2 load test when enabled. 
It's backwards compatible, it defaults to not validate this SLO by defaulting the threshold value to 0.

Fixes #
https://github.com/kubernetes/perf-tests/issues/2417


hakuna-matatah:
@wojtek-t This is for ability to measure In cluster networking SLOs when enabled similar to other [PR](https://github.com/kubernetes/perf-tests/pull/2408) 
Please let me know if you have any questions.

PR #2420, opened 2023-11-17
[ClusterLoader2] run-e2e.sh improvements

jprzychodzen:
* Allows re-runs by checking if K8s objects exist before creating them
* We only mutate temporary config, which prevents breakages for a local workflow

#### What type of PR is this?

/kind bug

After #2258 `run-e2e` failed on re-runs on the same cluster, and was messing with local KUBECONFIG. This change ensures that we only create service account and appropriate K8s configuration if it's missing. Additionally, we are creating a temporary configuration for CL2 usage. 

/assign @wojtek-t


wojtek-t:
/lgtm
/approve

Thanks!

PR #2445, opened 2023-11-29
Add option to dns performance test to install additional YAMLs

rahulkjoshi:
#### What type of PR is this?
/kind feature

Provides the capability to apply additional manifests before starting a DNS performance test. This can be used to install other features that might need to be scaled tested (e.g. egress gateways, network policies, etc.)

### DIFF
--- a/clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
+++ b/clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"embed"
 	"fmt"
+	"os"
+	"path/filepath"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,6 +30,7 @@ import (
 	"k8s.io/perf-tests/clusterloader2/pkg/framework"
 	"k8s.io/perf-tests/clusterloader2/pkg/framework/client"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
 )
 
@@ -44,6 +47,7 @@ scraped and results verified to not cross the specified thresholds.
 const (
 	dnsPerfK8sHostnamesMeasureName = "DNSPerformanceK8sHostnames"
 	dnsPerfTestNamespace           = "dns-perf-test"
+	dnsPerfTestDeploymentName      = "dns-perf-client"
 	dnsPerfTestPermissionsName     = "dns-test-client"
 	serviceAccountFilePath         = "manifests/serviceaccount.yaml"
 	clusterRoleFilePath            = "manifests/clusterrole.yaml"
@@ -78,6 +82,9 @@ type dnsPerfK8sHostnamesMeasurement struct {
 	qpsPerClient int
 	// testDurationMinutes is the duration in minutes for DNS client pods to run.
 	testDurationMinutes int
+	// additionalManifestFilePath is the optional location of an extra manifest
+	// to install when setting up the test.
+	additionalManifestFilePath string
 }
 
 func (m *dnsPerfK8sHostnamesMeasurement) Execute(config *measurement.Config) ([]measurement.Summary, error) {
@@ -88,6 +95,10 @@ func (m *dnsPerfK8sHostnamesMeasurement) Execute(config *measurement.Config) ([]
 
 	if err = client.CreateNamespace(m.k8sClient, m.testClientNamespace); err != nil {
 		return nil, fmt.Errorf("error while creating namespace: %v", err)
+	}
+
+	if err = m.createAdditionalManifests(); err != nil {
+		return nil, fmt.Errorf("failed to apply additional manifests: %v", err)
 	}
 
 	if err = m.createDNSClientPermissions(); err != nil {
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
```diff
@@ -182,6 +221,35 @@ func (m *dnsPerfK8sHostnamesMeasurement) deleteDNSClientPermissions() error {
...
+
+	ctx, cancel := context.WithTimeout(context.TODO(), 2*time.Minute)
+	defer cancel()
+	if _, err := measurementutil.WaitForPods(ctx, podStore, options); err != nil {
+		return fmt.Errorf("wating for DNS client pods to delete: %v", err)
+	}
+	return nil
+}
+
 func (m *dnsPerfK8sHostnamesMeasurement) cleanUp() error {
```
aojea, line 253: don't you have to cleanup all the additional manifests that you have installed?

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
```diff
@@ -78,6 +82,9 @@ type dnsPerfK8sHostnamesMeasurement struct {
 	qpsPerClient int
 	// testDurationMinutes is the duration in minutes for DNS client pods to run.
 	testDurationMinutes int
+	// additionalManifestFilePath is the optional location of an extra manifest
```
aojea, line 85: I'm not familiar with this framework but is risky to allow passing unstructured configuration to tests

rahulkjoshi: Yeah I understand the concern. My rationale here is a want to install GKE specific CRDs that probably don't need to be hard coded into the perf test framework.

I feel like the unstructured configuration should be ok since this is used for running perf tests and shouldn't be running in anyones production environment anyhow.

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
rahulkjoshi: I wasn't able to figure a good way to cleanup unstructured manifests. I can play around with invoking a `kubectl delete`

dlapcevic: There must be a way to parse a manifest file and issue a delete object request.
Try to see if you can use this https://github.com/rahulkjoshi/perf-tests/blob/master/clusterloader2/pkg/test/simple_test_executor.go#L331

Otherwise, `kubectl delete` would be ok.

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
dlapcevic: Design-wise I'd say it's acceptable. Although very flexible, nobody is forced to use it. Users will take the responsibility.

@aojea, do you mean there is a security risk?

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
```diff
@@ -182,6 +221,35 @@ func (m *dnsPerfK8sHostnamesMeasurement) deleteDNSClientPermissions() error {
 	return m.k8sClient.RbacV1().ClusterRoleBindings().Delete(context.TODO(), dnsPerfTestPermissionsName, metav1.DeleteOptions{})
 }
 
+func (m *dnsPerfK8sHostnamesMeasurement) deleteDNSClient() error {
```
dlapcevic, line 224: The clients are deleted when the namespace is deleted in the cleanup phase.

I assume it's best practice to separately clean up the pods to identify any issues with the clean up.

dlapcevic (approved):
LGTM. Thank you Rahul.

dlapcevic:
@rahulkjoshi please assign to @marseel for approval and merging, once you're ready for it.

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
aojea: ah, no security, just in the design of tests I always encourage people to be opinionated, tests are meant to be super stable and solid, so you run them over a black box, and you have clear inputs and you know exactly what output you expect, so the black box is correct because you assert on the outputs. Flexibility also creates more complexity to maintain it ...
If the inputs are flexible the outputs start to be flexible too ... is up to you, just sharing, maybe in this case is different ... just take it into account

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
rahulkjoshi: Added cleanup code to remove the additional manifests

clusterloader2/pkg/measurement/common/dns/dns_performance-k8s-hostnames.go
rahulkjoshi: Added cleanup code to delete the additional manifests.

marseel:
@rahulkjoshi @dlapcevic 
Can I get a bit more context why are we doing this? 
It seems to me that we are creating any arbitrary manifest within DNS measurement (I don't see how these two things are related)
I don't really see what's the point if iwe could do the same thing in yaml test config.
Something like this with NetworkPolicies: https://github.com/kubernetes/perf-tests/blob/82625fa63d6669af6b378bb6f680c36772141097/clusterloader2/testing/load/modules/reconcile-objects.yaml#L122
One difference I see is that above one creates objects in test-X namespaces rather then the one used by DNS measurement `dns-perf-test`.
In that case we can either make DNS Measurement to run within regular "test" namespaces or we can add option to Phase to specify namespace names rather than range of namespaces that are auto-created: https://github.com/marseel/perf-tests/blob/83f46c7db453955aa142c1c1bf42311846fa270a/clusterloader2/api/types.go#L100

dlapcevic:
The purpose is in the description:
```
Provides the capability to apply additional manifests before starting a DNS performance test. This can be used to install other features that might need to be scaled tested (e.g. egress gateways, network policies, etc.)
```

We want to be able to apply custom manifests that are related to DNS. I don’t see a simple way to do it without having to add these manifests here in the perf-tests repo. It’s much simpler to allow us to pass them in.
However, Marcel might be right that it doesn’t belong within the DNS measurement, and we could move it out to a separate measurement, and run this measurement just before the DNS measurement.

PR #2189, opened 2022-11-09
fix: master kubelets service monitor in CL2

eddycharly:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

This PR fixes master kubelets service monitor in CL2.

The service monitor lacks `bearerTokenFile` leading to 401 when scraping master kubelets.

PR #2463, opened 2023-12-13
Add bearer token to apiserver prometheus target

marseel:
#### What type of PR is this?

/kind bug

This is followup from https://github.com/kubernetes/perf-tests/pull/2189 to add bearerTokenFile for apiserver.
We've observed the same issue for clusters created with kops on GCE.


marseel:
Seems like adding token breaks kubemarks:
```
I1213 11:22:43.209089      11 httplog.go:132] "HTTP" verb="GET" URI="/metrics" latency="827.166µs" userAgent="Prometheus/2.25.0" audit-ID="d5a688d0-c6e7-43ba-ba51-4d2794867fa8" srcIP="10.64.1.33:40554" resp=401
```
Converting to draft till I fix it.

PR #2486, opened 2023-12-26
Update x/crypto to v0.17.0

marseel:
#### What type of PR is this?

/kind bug

Update x/crypto to a newer version to try to resolve ssh issues.


upodroid (approved):
LGTM

Thanks

PR #2483, opened 2023-12-20
Add an option to inject additional measurements to load config

Argh4k:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This PR allows to inject additional measurements via CL2_ADDITIONAL_MEASURMENT_MODULES to load/config.yaml.

#### Which issue(s) this PR fixes:
Fixes #

PR #2494, opened 2024-01-08
Fix YAML injections in the network policy enforcement test

dlapcevic:
Remove label key templating from the YAML files to pass the YAML injection checks introduced in https://github.com/kubernetes/perf-tests/pull/2421

Example error message from test failure:
```
reading template (deployment.yaml) error: replacing placeholders error: YAML Injection Detected
```

Similar for the Network Policy Enforcement measurement:
```
NetworkPolicyEnforcement - NetworkPolicyEnforcement error: error while creating test client deployment: TemplateToObject error: replacing placeholders error: YAML Injection Detected
```

/kind bug


dlapcevic:
Hi @marseel, could we merge this since the failures are due to N2 machine quotas, and not related to the change.
Another PR has just been merged despite these failures https://github.com/kubernetes/perf-tests/pull/2483

I tested the change. Our tests that were previously broken are now passing with this.

PR #2529, opened 2024-01-27
add simple HPA scale test

mengqiy:
#### What type of PR is this?

/kind feature

This PR adds a simple scale test to test HPA controller to check if it can keep up with the load of reconcile all HPAs object within the reconciliation interval (default: 15s).

If a cluster has thousands of HPAs objects, it may not be able to reconcile them all every 15s. HPA controller's workqueue will consistently have backlog. This test can help catch it.

It was previously discussed in SIG scalability call: https://docs.google.com/document/d/1hEpf25qifVWztaeZPFmjNiJvPo-5JX1z0LSvvVY5G2g/edit#heading=h.mh1dh917dnmi

### DIFF
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -76,6 +76,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdCertificatePath, "etcd-certificate", "ETCD_CERTIFICATE", "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt", "Path to the etcd certificate on the master machine")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdKeyPath, "etcd-key", "ETCD_KEY", "/etc/srv/kubernetes/pki/etcd-apiserver-server.key", "Path to the etcd key on the master machine")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.KCMPort, "KCM-port", "KCM_PORT", ports.KubeControllerManagerPort, "Port of the kube-controller-manager")
 	flags.BoolEnvVar(&clusterLoaderConfig.ClusterConfig.DeleteStaleNamespaces, "delete-stale-namespaces", "DELETE_STALE_NAMESPACES", false, "DEPRECATED: Whether to delete all stale namespaces before the test execution.")
 	err := flags.MarkDeprecated("delete-stale-namespaces", "specify deleteStaleNamespaces in testconfig file instead.")
 	if err != nil {
--- a/clusterloader2/pkg/measurement/common/kcm_metrics.go
+++ b/clusterloader2/pkg/measurement/common/kcm_metrics.go
@@ -1,0 +1,223 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/prometheus/common/model"
+	"k8s.io/klog/v2"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/provider"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const (
+	kcmWorkQueueDepthMetricName    = "KCMWorkQueueDepthMetrics"
+	kcmWorkQueueDepthMetricVersion = "v1"
+	hpaResource                    = "horizontalpodautoscaler"
+)
+
+func init() {
+	if err := measurement.Register(kcmWorkQueueDepthMetricName, createKCMWorkQueueDepthMeasurement); err != nil {
+		klog.Fatalf("Cannot register %s: %v", kcmWorkQueueDepthMetricName, err)
+	}
+}
+
+func createKCMWorkQueueDepthMeasurement() measurement.Measurement {
+	return &kcmWorkQueueDepthMeasurement{
+		depthMetrics: map[string]float64{},
+		stopCh:       make(chan struct{}),
+	}
+}
+
+type kcmWorkQueueDepthMeasurement struct {
+	// depthMetrics is the minimum workqueue depth for each resouce in the leader.
+	depthMetrics      map[string]float64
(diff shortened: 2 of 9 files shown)


clusterloader2/cmd/clusterloader.go
```diff
@@ -76,6 +76,7 @@ func initClusterFlags() {
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdCertificatePath, "etcd-certificate", "ETCD_CERTIFICATE", "/etc/srv/kubernetes/pki/etcd-apiserver-server.crt", "Path to the etcd certificate on the master machine")
 	flags.StringEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdKeyPath, "etcd-key", "ETCD_KEY", "/etc/srv/kubernetes/pki/etcd-apiserver-server.key", "Path to the etcd key on the master machine")
 	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.EtcdInsecurePort, "etcd-insecure-port", "ETCD_INSECURE_PORT", 2382, "Inscure http port")
+	flags.IntEnvVar(&clusterLoaderConfig.ClusterConfig.KCMPort, "KCM-port", "KCM_PORT", ports.KubeControllerManagerPort, "Port of the kube-controller-manager")
```
hakuna-matatah, line 79: Nit: to make it consistent with other params, could we use "kcm-port" in lower case for the first part like how we have it for `etcd-insecure-port` ?

clusterloader2/pkg/measurement/common/kcm_metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unexpected error (code: %d) in ssh connection to master: %#v", sshResult.Code, err)
+	} else if sshResult.Code != 0 {
+		return nil, fmt.Errorf("failed running command: %s on the host: %s, result: %+v", cmd, host, sshResult)
+	}
+	data := sshResult.Stdout
+	return measurementutil.ExtractMetricSamples(data)
+}
+
+func getMaxOrMin(existing, new map[string]float64, max bool) map[string]float64 {
```
hakuna-matatah, line 215: You don't seem to use `max` param at all, this function always seems to return the `min`, do we want to strip the `max param` and change the function name to `getMin` ?

clusterloader2/pkg/measurement/common/kcm_metrics.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+const (
+	kcmWorkQueueDepthMetricName    = "KCMWorkQueueDepthMetrics"
+	kcmWorkQueueDepthMetricVersion = "v1"
+	hpaResource                    = "horizontalpodautoscaler"
+)
+
+func init() {
+	if err := measurement.Register(kcmWorkQueueDepthMetricName, createKCMWorkQueueDepthMeasurement); err != nil {
```
hakuna-matatah, line 40: Given this file is named as `kcm_metrics.go`, can we make it generic and flexible to measure other KCM metrics in future if we want ? Instead of pinning it down to `workqueuedepth` ? 

`
if err := measurement.Register(kcmMetricsName, createKCMMetricsMeasurement); err != nil { 
klog.Fatalf("Cannot register %s: %v", kcmMetricsName, err)
	}
}
`
wdyt ?

clusterloader2/pkg/measurement/common/kcm_metrics.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func createKCMWorkQueueDepthMeasurement() measurement.Measurement {
+	return &kcmWorkQueueDepthMeasurement{
+		depthMetrics: map[string]float64{},
+		stopCh:       make(chan struct{}),
+	}
+}
+
+type kcmWorkQueueDepthMeasurement struct {
```
hakuna-matatah, line 52: `kcmMetricsMeasurement` to give room for other metrics measurement in kcm   ?

clusterloader2/pkg/measurement/common/kcm_metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	// depthMetrics is the minimum workqueue depth for each resouce in the leader.
+	depthMetrics      map[string]float64
```
hakuna-matatah, line 54: could we change this to `metric  *kcmMetrics`  ?

And define kcmMetrics as something like 

```
type kcmMetrics struct { 
depthMetrics map[string]float64
....
...
}
```
This way it gives us room to declare other kcm metrics in future that we want to measure ?

clusterloader2/pkg/measurement/common/kcm_metrics.go
```diff
@@ -0,0 +1,223 @@
...
+					metricsFromHosts = append(metricsFromHosts, workQueueDepthMetrics)
+				}
+				if len(metricsFromHosts) == 0 {
+					continue
+				}
+				// For one single data point in each KCM instance, we take the max,
+				// since the leader will have larger workqueue backlog.
+				result := metricsFromHosts[0]
+				for i := 1; i < len(metricsFromHosts); i++ {
+					result = getMaxOrMin(result, metricsFromHosts[i], true)
```
hakuna-matatah, line 141: Do you think you are getting the `max` here ? Because the `getMaxOrMin` function doesn't seem to take the boolean into consideration. So i think you might be getting always the `min` here IIUC ?

clusterloader2/cmd/clusterloader.go
shyamjvs: +1

clusterloader2/pkg/measurement/common/kcm_metrics.go
```diff
@@ -0,0 +1,223 @@
+/*
```
shyamjvs, line 1: I see component metrics related files are all placed under `https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/pkg/measurement/common/metrics`.  If there are two parts to this file - collecting metrics and measurement logic, can you move the former to that location and rename this file to explain what the measurement we are making is? Just like the others - https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/pkg/measurement/common

clusterloader2/testing/hpa/README.md
```diff
@@ -0,0 +1,18 @@
+## HPA controller load test
+
+Assumptions:
+- Cluster master noded are accessible via SSH.
```
shyamjvs, line 4: nodes

clusterloader2/testing/hpa/README.md
```diff
@@ -0,0 +1,18 @@
...
+- The `/metrics` endpoint in the kube-controller-manager (KCM) can be accessed by anonymous users.
+  `--authorization-always-allow-paths` can be useful to achieve this.
+
+HPA load test perform the following steps:
+- Create replicasets, services, HPAs in a set of namespaces.
+  - The namespaces used here are managed by CL2.
+  - The number of HPAs in each namespace can be specified using `CL2_HPA_PER_NAMESPACE`.
+- Periodically measure the HPA controller workqueue depth in each KCM instance
+  - In each scrape, we use the max value among them since the leader will have more items in the workqueue.
+  - We keep tracking the smallest value over time.
```
shyamjvs, line 14: Can you clarify this - what's the significance of tracking the smallest value over time?

shyamjvs:
Looks like there are some legit linter errors causing the verify-lint job to fail. Can you fix them?

clusterloader2/testing/hpa/README.md
```diff
@@ -0,0 +1,18 @@
...
+Assumptions:
+- Cluster master noded are accessible via SSH.
+- The `/metrics` endpoint in the kube-controller-manager (KCM) can be accessed by anonymous users.
+  `--authorization-always-allow-paths` can be useful to achieve this.
+
+HPA load test perform the following steps:
+- Create replicasets, services, HPAs in a set of namespaces.
+  - The namespaces used here are managed by CL2.
+  - The number of HPAs in each namespace can be specified using `CL2_HPA_PER_NAMESPACE`.
+- Periodically measure the HPA controller workqueue depth in each KCM instance
```
hakuna-matatah, line 12: `Periodically` -  could you add details on what's the period/interval here  ? every second ?

PR #2533, opened 2024-01-30
[ClusterLoader2] Enable measuring API availability for public and/or private IPs

jprzychodzen:
Enable API Availability measurement for public IPs, also add a flag to enable/disable private/public IPs

#### What type of PR is this?

/kind feature

### DIFF
--- a/clusterloader2/pkg/measurement/common/api_availability_measurement.go
+++ b/clusterloader2/pkg/measurement/common/api_availability_measurement.go
@@ -58,6 +58,10 @@ type apiAvailabilityMeasurement struct {
 	summaries           []measurement.Summary
 	clusterLevelMetrics *apiAvailabilityMetrics
 	threshold           float64
+	// Should we check public IPs of the host VMs
+	useHostPublicIPs bool
+	// Should we check internal IPs of the host VMs
+	useHostInternalIPs bool
 	// Metrics per host internal IP.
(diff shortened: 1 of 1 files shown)


mborsz:
/hold

Is the change backward compatible? i.e. if someone was using the api availability on internal IPs what will happen?

Can we adjust defaults to match the previous behavior (i.e. true for internal, false for public)?

jprzychodzen:
> Is the change backward compatible?

No, it's not in a current shape. Let's go with internal IPs default to true if `config.ClusterLoaderConfig.ExecServiceConfig.Enable` and `len(config.ClusterFramework.GetClusterConfig().MasterInternalIPs) != 0`. With public IPs default to `false` it will be completly backward compatible.

jprzychodzen:
kubemark test is falling with
```
 - Setting minimum CPU platform is not supported for the selected machine type e2-standard-8.
Failed to create master instance due to non-retryable error
2024/01/30 13:18:34 process.go:155: Step 'test/kubemark/start-kubemark.sh' finished in 1m20.74023594s
```

mborsz:
> kubemark test is falling with
> 
> ```
>  - Setting minimum CPU platform is not supported for the selected machine type e2-standard-8.
> Failed to create master instance due to non-retryable error
> 2024/01/30 13:18:34 process.go:155: Step 'test/kubemark/start-kubemark.sh' finished in 1m20.74023594s
> ```

 I will take care of it

PR #2575, opened 2024-02-22
Add an option to inject additional objects to load config

alan-kut:
#### What type of PR is this?
/kind feature

This PR allows to inject additional objects via CL2_ADDITIONAL_OBJECTS_MODULES to load/config.yaml.

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -78,6 +78,7 @@
 {{$defaultQps := DefaultParam .CL2_DEFAULT_QPS (IfThenElse (le .Nodes 500) 10 100)}}
 
 {{$ADDITIONAL_MEASUREMENT_MODULES := DefaultParam .CL2_ADDITIONAL_MEASURMENT_MODULES nil}}
+{{$ADDITIONAL_PHASES_MODULES := DefaultParam .CL2_ADDITIONAL_PHASES_MODULES nil}}
 
 name: load
 namespace:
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -78,6 +78,7 @@
 {{$defaultQps := DefaultParam .CL2_DEFAULT_QPS (IfThenElse (le .Nodes 500) 10 100)}}
 
 {{$ADDITIONAL_MEASUREMENT_MODULES := DefaultParam .CL2_ADDITIONAL_MEASURMENT_MODULES nil}}
+{{$ADDITIONAL_OBJECTS_MODULES := DefaultParam .CL2_ADDITIONAL_OBJECTS_MODULES nil}}
```
tosi3k, line 81: Could we rename this to `ADDITIONAL_PHASES_MODULES` and `CL2_ADDITIONAL_PHASES_MODULES`?

This sounds more in line with the CL2's API where a single step can be either a list of measurements or phases.

alan-kut: Done

clusterloader2/testing/load/config.yaml
```diff
@@ -207,6 +208,15 @@ steps:
       smallJobSize: {{$SMALL_GROUP_SIZE}}
       smallJobsPerNamespace: 1
 
+{{if $ADDITIONAL_PHASES_MODULES}}
```
mborsz, line 211: Nit: is this if necessary? Range over nil should be legal operation.

alan-kut: I followed what was done in the https://github.com/kubernetes/perf-tests/pull/2483

PR #2599, opened 2024-03-15
Add NEG Latency test

alexkats:
Add NEG Latency test

/kind feature

### DIFF
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -55,6 +56,8 @@ const (
 	reachabilityPhase = "reachability"
 	deletingPhase     = "deleting"
 	deletedPhase      = "deleted"
+
+	ingressType = "ingress"
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -361,21 +493,33 @@ func (p *pingChecker) run() {
...
-				ips = p.svc.Spec.ClusterIPs
-				port = p.svc.Spec.Ports[0].Port
-			case corev1.ServiceTypeNodePort:
-				ips = []string{pod.Status.HostIP}
-				port = p.svc.Spec.Ports[0].NodePort
-			case corev1.ServiceTypeLoadBalancer:
-				for _, ingress := range p.svc.Status.LoadBalancer.Ingress {
-					ips = append(ips, ingress.IP)
+			isIngress := p.ingress != nil
+			if isIngress {
```
dlapcevic, line 497: This is getting very nested and many ifs for ingress.
It would be more readable and easier to follow if they are decoupled.
Maybe one if in the beginning and then the rest can be generalized.

dlapcevic (commented):
Because NEG is a GCP specific concept I think it's better to move the measurement into a separate directory, like `.../common/gcp/neg`.

Same for the test. Move to `.../testing/gcp/neg`.

clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -168,12 +171,33 @@ func (s *serviceCreationLatencyMeasurement) start() error {
 			s.queue.Add(&f)
 		},
 	)
-	return informer.StartAndSync(i, s.stopCh, informerSyncTimeout)
+	if err := informer.StartAndSync(svcInformer, s.stopCh, informerSyncTimeout); err != nil {
+		return err
+	}
+	ingressInformer := informer.NewInformer(
```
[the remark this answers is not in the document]
alexkats: Makes sense, thanks, changed

clusterloader2/pkg/measurement/common/service_creation_latency.go
alexkats: Agree, improved readability there

dlapcevic (approved):

alexkats:
> Because NEG is a GCP specific concept I think it's better to move the measurement into a separate directory, like `.../common/gcp/neg`.
> 
> Same for the test. Move to `.../testing/gcp/neg`.

That's true that NEG is a GCP specific thing, but at the same time we already have L4 tests that test ingress controller, and it already heavily relies on NEGs and other GCP only concepts. And this test in that sense is not much different.

PR #2620, opened 2024-03-21
increase the duration to 15m

mengqiy:
This is a small change to the list load test.
It changes the duration from 5m to 15m to give it a bit more time to exercise and allow more time to see potential memory usage increase.


hakuna-matatah:
/lgtm

Thanks for following up on this David. 
May be in future we could parameterize `wait time` of this test and leave it to enduser to choose it how much time they would want to run it.

PR #2643, opened 2024-04-12
implement dns-propagation-prober

besher-massri:
Implementation of a DNS propagation prober, measuring the propagation time of statefulSet workload pods from the point they are running to the point they are discoverable.

/kind feature


aojea:
/ok-to-test

You need to sign the CLA :)

besher-massri:
resigned cla

util-images/probes/cmd/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2024 The Kubernetes Authors.
```
aojea, line 2: no need to update this

util-images/probes/pkg/dnsPropagation/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+		Namespace: common.ProbeNamespace,
+		Name:      "dns_propagation_seconds",
+		Help:      "Gauge of the time (in seconds) it took for the pods in a statefulSet to be discoverable after they start running.",
+	}, []string{"namespace", "service", "podName"})
+	// DNSPropagationCount calculate the number of DNS propagation checks performed.
+	DNSPropagationCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: common.ProbeNamespace,
+		Name:      "dns_propagation_count",
+		Help:      "Counter of the number of DNS propagation checks performed.",
+	}, []string{"namespace", "service", "podName"})
```
aojea, line 36: should not be better to use a histogram https://prometheus.io/docs/concepts/metric_types/#histogram

util-images/probes/pkg/dnsPropagation/probe.go
```diff
@@ -0,0 +1,242 @@
...
+	for idx := range indices {
+		indices[idx] = idx
+	}
+	rand.Shuffle(len(indices), func(i, j int) { indices[i], indices[j] = indices[j], indices[i] })
+	indices = indices[:sampleCount]
+	return indices
+}
+func runSinglePod(client kubernetes.Interface, url string, podName string, namespace string, interval time.Duration) time.Duration {
+	klog.V(4).Infof("Starting dns propagation calculation for pod %s ...", url)
+	tick := time.NewTicker(interval)
```
aojea: you need to stop the ticker or it will leak resourcecs

https://pkg.go.dev/time#NewTicker
```suggestion
	tick := time.NewTicker(interval)
	defer tick.Stop()
```

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,242 @@
...
+		trials++
+		time.Sleep(1 * time.Second)
+	}
+	return getPodRunningTimeFromClient(client, podName, namespace, false)
+}
+
+func getNamespacesFromClient(client kubernetes.Interface, namespace string) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+	namespaceList, err := client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
```
aojea, line 209: namespaces get a label with the name when they are created so you can use to query directly for the namespace name with a label selector
```
  labels:
      kubernetes.io/metadata.name: <name of the namespace>
```

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,242 @@
...
+	for _, namespaceItem := range namespaceList.Items {
+		if strings.HasPrefix(namespaceItem.Name, namespace) {
+			return namespaceItem.Name, nil
+		}
+	}
+	return "", errors.New("No namespace were found that matches the prefix: " + namespace)
+}
+
+func getPodRunningTimeFromClient(client kubernetes.Interface, podName string, namespace string, resourceVersionEnabled bool) (time.Time, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
aojea, line 222: this is too short, also the client has retries and timeouts by default, any particular reason you want to cancel the context?

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	for _, namespaceItem := range namespaceList.Items {
+		if strings.HasPrefix(namespaceItem.Name, namespace) {
+			return namespaceItem.Name, nil
+		}
+	}
+	return "", errors.New("No namespace were found that matches the prefix: " + namespace)
+}
+
+func getPodRunningTimeFromClient(client kubernetes.Interface, podName string, namespace string, resourceVersionEnabled bool) (time.Time, error) {
```
tosi3k, line 221: IMHO we do not have to rely on quorum reads from etcd - watch cache is expected to be behind up to O(x)-O(xx) ms wrt. the real state in etcd. If it's sth above 0.5s then sth is definitely wrong going on in the cluster anyway.

Thus, I think we can simply always query with RV=0.

util-images/probes/pkg/dnspropagation/probe.go
tosi3k: I'm not so sure - single object GETs (especially from the watch cache) should be pretty quick.

Our SLO threshold for API call latency for single object API calls is 1s in 99th percentile, so I think 3s is a pretty sane assumption here as some additional buffer.

If we go above that value, then the cluster is definitely overloaded and not really usable anyway :P.

tosi3k: Also, why not use a watch here instead of periodic polling?

tosi3k (changes requested):

util-images/probes/pkg/dnspropagation/probe.go
besher-massri: @aojea the problem is that the namespace name is different each time, but they share a prefix, hence the need to list then search.

util-images/probes/pkg/dnsPropagation/metrics.go
besher-massri: the reason here is that we don't really have bins per se, since we have exactly one measurement per label. Also timestamp of the measurement isn't important, so I wanted to choose a metric that wouldn't get it confused with histogram to store a single value per label.

util-images/probes/pkg/dnspropagation/probe.go
besher-massri: fixed

util-images/probes/pkg/dnsPropagation/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+
+package dnsPropagation
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/perf-tests/util-images/probes/pkg/common"
+)
+
+var (
+	// DNSPropagationSeconds calculate the propagation time of a given statefulSet the average difference between the pod running and pod discoverable timestamps.
```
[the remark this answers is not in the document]
besher-massri: fixed

util-images/probes/pkg/dnsPropagation/probe.go
besher-massri: thanks, fixed

util-images/probes/cmd/main.go
besher-massri: fixed

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,234 @@
...
+		klog.V(2).Infof("dns propagation probe complete, waiting until the test finishes...")
+		time.Sleep(10 * time.Second)
+	}
+}
+func runProbe(client kubernetes.Interface) {
+	klog.V(1).Infof("DNS propagation probe started")
+	var namespace string
+	for {
+		klog.V(2).Infof("Get namespace from client based on prefix %s", *namespacePrefix)
+		fullNamespace, err := getNamespacesFromClient(client, *namespacePrefix)
```
tosi3k, line 95: This might work randomly if you have multiple namespaces matching the prefix.

If you intend to execute this container image from clusterloader2/pkg/measurement/common/probes/probes.go, then, there, you will have access to `*measurement.Config`: https://github.com/kubernetes/perf-tests/blob/32357a4c9514cca8cd61b4c327547adc686b68a3/clusterloader2/pkg/measurement/interface.go#L29-L49 which has the `ClusterFramework` field: https://github.com/kubernetes/perf-tests/blob/32357a4c9514cca8cd61b4c327547adc686b68a3/clusterloader2/pkg/framework/framework.go#L44-L56 which has two fields that could be of interest to you:

- `automanagedNamespacePrefix`
- `automanagedNamespaces`

IMHO we can get simpler (overall) approach working here:
- Pass the index of the namespace of our interest to the measurement through `Params`,
- Reconstruct the namespace name using that index and aforementioned `automanagedNamespacePrefix` field,
- Use the reconstructed namespace name when filling out the template for the prober's deployment.

Thoughts?

tosi3k (changes requested):

util-images/probes/pkg/dnspropagation/probe.go
besher-massri: Done

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,234 @@
...
+
+	if err != nil {
+		klog.Fatalf("Failed to build kubeClient, error:%v", err)
+	}
+	//TODO deprecated as of Go 1.20. To remove when go version gets bumped
+	rand.Seed(time.Now().UnixNano())
+	runProbe(clientset)
+	for {
+		klog.V(2).Infof("dns propagation probe complete, waiting until the test finishes...")
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
besher-massri: Done

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,221 @@
...
+		time.Sleep(1 * time.Second)
+	}
+	return getPodRunningTimeFromClient(client, podName, namespace)
+}
+
+// getPodRunningTimeFromClient fetches the running time of the given pod.
+// It returns the running time of the pod, or an error if the pod is not found or if the running time cannot be fetched.
+func getPodRunningTimeFromClient(client kubernetes.Interface, podName string, namespace string) (time.Time, error) {
+	options := metav1.GetOptions{ResourceVersion: "0"}
+	pod, err := client.CoreV1().Pods(namespace).Get(context.Background(), podName, options)
```
tosi3k, line 209: Nitpick: use `context.TODO()`: https://pkg.go.dev/context#TODO.

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,221 @@
...
+		klog.V(2).Infof("dns propagation probe complete, waiting until the test finishes...")
+		time.Sleep(60 * time.Second)
+	}
+}
+
+// runProbe runs the DNS propagation probe.
+func runProbe(client kubernetes.Interface) {
+	klog.V(1).Infof("DNS propagation probe started")
+	var wg sync.WaitGroup
+	ch := make(chan DNSPodPropagationResult)
```
tosi3k, line 94: Nitpick: we can increase the channel's capacity to match the number of goroutines so we do not block on sending data through it.

util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -0,0 +1,221 @@
...
+			"podName":   propagationResult.podName,
+		}
+		DNSPropagationSeconds.With(labels).Set(propagationResult.duration.Seconds())
+		DNSPropagationCount.With(labels).Inc()
+		durations = append(durations, propagationResult.duration.Seconds())
+	}
+	klog.V(2).Infof("Finished calculating DNS propagation for all sample pods")
+
+	if len(durations) == 0 {
+		klog.Warningf("DNS propagation probe has zero obseravtions")
```
tosi3k, line 129: Supernit: typo.

tosi3k (changes requested):
Please also add a short entry about the probe in the README file (in util-images/probes).

util-images/probes/pkg/dnspropagation/probe.go
besher-massri: DONE

util-images/probes/pkg/dnspropagation/probe.go
besher-massri: DONE

util-images/probes/pkg/dnspropagation/probe.go
besher-massri: DONE

PR #2649, opened 2024-04-18
DNS propagation measurement in CL2

besher-massri:
Implementation of a DNS propagation measurement in CL2, using the DNS propagation prober.

Follow up on #2643

/kind feature

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
@@ -63,6 +63,9 @@ spec:
       record: probes:dns_lookup_latency:histogram_quantile
       labels:
         quantile: "0.50"
+    - expr: |
+        sum(max_over_time(probes_dns_propagation_seconds[1h]))/sum(max_over_time(probes_dns_propagation_count[1h]))
+      record: probes:dns_propagation_average
   - name: kube-proxy.rules
     rules:
     - expr: |
--- a/clusterloader2/go.mod
+++ b/clusterloader2/go.mod
@@ -40,6 +40,7 @@ require (
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/safetext v0.0.0-20230106111101-7156a760e523
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/montanaflynn/stats v0.7.1 // indirect
 	github.com/onsi/ginkgo v1.16.5
 	github.com/prometheus/client_model v0.6.0
 	github.com/prometheus/common v0.26.0
--- a/clusterloader2/go.sum
+++ b/clusterloader2/go.sum
@@ -708,6 +708,8 @@ github.com/modern-go/reflect2 v1.0.2/go.mod h1:yWuevngMOJpCy52FWWMvUC8ws7m/LJsjY
 github.com/mohae/deepcopy v0.0.0-20170603005431-491d3605edfb/go.mod h1:TaXosZuwdSHYgviHp1DAtfrULt5eUgsSMsZf+YrPgl8=
 github.com/monochromegane/go-gitignore v0.0.0-20200626010858-205db1a8cc00/go.mod h1:Pm3mSP3c5uWn86xMLZ5Sa7JB9GsEZySvHYXCTK4E9q4=
 github.com/montanaflynn/stats v0.0.0-20171201202039-1bf9dbcd8cbe/go.mod h1:wL8QJuTMNUDYhXwkmfOly8iTdp5TEcJFWZD2D7SIkUc=
+github.com/montanaflynn/stats v0.7.1 h1:etflOAAHORrCC44V+aR6Ftzort912ZU+YLiSTuV8eaE=
+github.com/montanaflynn/stats v0.7.1/go.mod h1:etXPPgVO6n31NxCd9KQUMvCM+ve0ruNzt6R8Bnaayow=
 github.com/morikuni/aec v1.0.0/go.mod h1:BbKIizmSmc5MMPqRYbxO4ZU0S0+P200+tUnFx7PXmsc=
 github.com/mrunalp/fileutils v0.5.0/go.mod h1:M1WthSahJixYnrXQl/DFQuteStB1weuxD2QJNHXfbSQ=
 github.com/mschoch/smat v0.0.0-20160514031455-90eadee771ae/go.mod h1:qAyveg+e4CE+eKJXWVjKXM4ck2QobLqTDytGJbLLhJg=
--- a/clusterloader2/pkg/measurement/common/probes/manifests/dnsPropagation/dns-propagation-prober-deployment.yaml
+++ b/clusterloader2/pkg/measurement/common/probes/manifests/dnsPropagation/dns-propagation-prober-deployment.yaml
@@ -1,0 +1,48 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: probes
+  name: dns-propagation-prober
+  labels:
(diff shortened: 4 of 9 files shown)


clusterloader2/pkg/measurement/util/propagation_metric.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	DNSPropagationProbeSampleCount, err := util.GetInt(config.Params, "DNSPropagationProbeSampleCount")
+	if err != nil {
+		DNSPropagationProbeSampleCount = defaultDNSPropagationProbeSampleCount
+	}
+	//calculating namespace
+	DNSPropagationProbeNamespaceIndex, err := util.GetInt(config.Params, "DNSPropagationProbeNamespaceIndex")
+	namespacePrefix := config.ClusterFramework.GetAutomanagedNamespacePrefix()
+	klog.Infof("GetAutomanagedNamespacePrefix %s", namespacePrefix)
```
tosi3k, line 76: Please use verbosity levels when using `Infof`.

clusterloader2/pkg/measurement/util/propagation_metric.go
```diff
@@ -0,0 +1,95 @@
...
+	DNSPropagationProbeStatefulSet, err := util.GetString(config.Params, "DNSPropagationProbeStatefulSet")
+	if err != nil {
+		DNSPropagationProbeStatefulSet = ""
+	}
+	DNSPropagationProbePodCount, err := util.GetInt(config.Params, "DNSPropagationProbePodCount")
+	if err != nil {
+		DNSPropagationProbePodCount = 0
+	}
+
+	DNSPropagationProbeSampleCount, err := util.GetInt(config.Params, "DNSPropagationProbeSampleCount")
```
tosi3k, line 69: Nit: `GetIntOrDefault`?

Dittos for the above getters.

clusterloader2/pkg/measurement/util/propagation_metric.go
```diff
@@ -0,0 +1,95 @@
...
+	namespacePrefix := config.ClusterFramework.GetAutomanagedNamespacePrefix()
+	klog.Infof("GetAutomanagedNamespacePrefix %s", namespacePrefix)
+	klog.Infof("DNSPropagationProbeNamespaceIndex %v", DNSPropagationProbeNamespaceIndex)
+
+	var DNSPropagationProbeNamespace string
+	if err == nil {
+		DNSPropagationProbeNamespace = fmt.Sprintf("%s-%d", namespacePrefix, DNSPropagationProbeNamespaceIndex)
+	} else {
+		DNSPropagationProbeNamespace = namespacePrefix
+	}
```
tosi3k, line 84: I don't like this logic - what if someone sets the index in the measurement's params, yet there's a single test namespace, without any index?

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
```diff
@@ -63,6 +63,15 @@ spec:
...
       labels:
         quantile: "0.50"
+    - expr: |
+        probes_dns_propagation_count
+      record: probes:dns_propagation_count
+    - expr: |
+        probes_dns_propagation_seconds
+      record: probes:dns_propagation_seconds
+    - expr: |
+        sum(max_over_time(probes_dns_propagation_seconds[1d]))/sum(max_over_time(probes_dns_propagation_count[1d]))
```
tosi3k, line 73: Why do we have hardcoded 1d here?

tosi3k (changes requested):

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
besher-massri: As the timestamp itself isn't relevant here, we need to aggregate the measurements in order to calculate the average. As Prometheus measurements are time series by nature, the idea is to make a query that would show the last value as the true value, so when shown on Grafana with a gauge chart, it will display the correct number. As such, we need to extend the time to cover the entire timespan of the experiment (i.e., from measurement start to measurement gather). 1d seemed sufficient; we can add a very high number (e.g., 10000d) to be sure to cover all possible experiments.

Please note that this record is only used in the context of Grafana; the one used in CL2 where the percentile values are calculated is probes:dns_propagation_seconds.

clusterloader2/pkg/measurement/util/propagation_metric.go
besher-massri: Done

clusterloader2/pkg/measurement/util/propagation_metric.go
besher-massri: Done

tosi3k (changes requested):

clusterloader2/pkg/measurement/util/propagation_metric.go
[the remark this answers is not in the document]
besher-massri: I put it here since the original latency metric function is also there.

clusterloader2/pkg/measurement/util/propagation_metric.go
```diff
@@ -0,0 +1,90 @@
...
+	var latencyMetric LatencyMetric
+	values := []float64{}
+	for _, sample := range samples {
+		values = append(values, float64(sample.Value))
+	}
+	sort.Float64s(values)
+
+	p50, _ := stats.Percentile(values, 50)
+	p90, _ := stats.Percentile(values, 90)
+	p99, _ := stats.Percentile(values, 99)
```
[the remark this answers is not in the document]
besher-massri: Done

clusterloader2/pkg/measurement/util/propagation_metric.go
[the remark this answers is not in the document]
tosi3k: What function do you have in mind exactly?

`InitializeTemplateMappingForDNSPropagationProbe` is super specific to the measurement - this should not be in the `util` directory.

Same applies to `NewPropagationMetricPrometheus`.

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
tosi3k: Thx for the rationale.

I believe we should be good with just `1h`, though - I wouldn't expect to measure the propagation latency for anything longer than 10 minutes, and, even if, `1h` should be more than enough.

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
```diff
@@ -63,6 +63,15 @@ spec:
       record: probes:dns_lookup_latency:histogram_quantile
       labels:
         quantile: "0.50"
+    - expr: |
+        probes_dns_propagation_count
+      record: probes:dns_propagation_count
+    - expr: |
+        probes_dns_propagation_seconds
+      record: probes:dns_propagation_seconds
```
tosi3k, line 71: Do we need these recording rules if they are essentially copies?

clusterloader2/pkg/measurement/util/propagation_metric.go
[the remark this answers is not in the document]
besher-massri: Yeah you're right. Moved it

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
besher-massri: Removed

clusterloader2/pkg/prometheus/manifests/prometheus-rules.yaml
besher-massri: Fixed

PR #2667, opened 2024-04-26
scrape network policies

aojea:
Copy and paste from ScrapeAnet, this is the install manifest

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/kube-network-policies/kube-network-policies.yaml

/assign @dlapcevic @wojtek-t


clusterloader2/pkg/config/cluster.go
```diff
@@ -88,6 +88,7 @@ type PrometheusConfig struct {
 	ScrapeMetricsServerMetrics bool
 	ScrapeNodeLocalDNS         bool
 	ScrapeAnet                 bool
+	ScrapeNetPol               bool
```
wojtek-t, line 91: nit: ScrapeNetworkPolicies [let's avoid shortcuts, especially in configs]

wojtek-t (commented):
Just one minor nit, other than that LGTM

clusterloader2/pkg/config/cluster.go
aojea: done

Issue #2704, opened 2024-06-05
Ability to specify image repos for those used with clusterloader2


rackeric:
**What would you like to be added**: The ability to use custom image repos with clusterloader2, specifically the deployment used by the exec_service and images used in template yamls for the load test, like deployments and statefulsets. 

**Why is this needed**: Some clusters may not have external Internet access to reach the upstream registry.k8s.io and gcr.io image repos, but do have access to an internal mirror hosting these images.

rackeric:
I've submitted a PR of the changes I use for the "load" test manifests. https://github.com/kubernetes/perf-tests/pull/2705

rackeric:
And for the exec_service image here is this change to allow it to be set with clusterloader2 CLI flag. https://github.com/kubernetes/perf-tests/pull/2706

PR #2705, opened 2024-06-05
allows for image repos to be set via Environment Variables

rackeric:
#### What type of PR is this?

/kind feature

Sometimes clusters will not have access to upstream image repos. This updates the manifests in the clusterloader2 "load" test to support setting the 2 image repos used via Environment Variables, like the others starting with '.CL', while still keeping the current repos set as defaults. This allows the use of mirrors image repos which may be available to clusters.

`.CL2_REGISTRY_K8S_REPO` for registry.k8s.io
`.CL2_GCR_REPO` for gcr.io

Fixes # https://github.com/kubernetes/perf-tests/issues/2704

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -74,8 +74,11 @@
 {{$EXEC_TIMEOUT := DefaultParam .CL2_EXEC_TIMEOUT "3600s"}}
 {{$SLEEP_AFTER_EXEC_DURATION := DefaultParam .CL2_SLEEP_AFTER_EXEC_DURATION "0s"}}
 
-{{$registry := DefaultParam .CL2_LATENCY_POD_REGISTRY "registry.k8s.io"}}
-{{$latencyPodImage := DefaultParam .CL2_LATENCY_POD_IMAGE (Concat $registry "/pause:3.9")}}
+## Image Repositories
+{{$registryK8sRepo := DefaultParam .CL2_REGISTRY_K8S_REPO "registry.k8s.io"}}
+{{$gcrRepo := DefaultParam .CL2_GCR_REPO "gcr.io"}}
+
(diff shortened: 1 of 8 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -74,7 +74,11 @@
 {{$EXEC_TIMEOUT := DefaultParam .CL2_EXEC_TIMEOUT "3600s"}}
 {{$SLEEP_AFTER_EXEC_DURATION := DefaultParam .CL2_SLEEP_AFTER_EXEC_DURATION "0s"}}
 
-{{$latencyPodImage := DefaultParam .CL2_LATENCY_POD_IMAGE "registry.k8s.io/pause:3.9"}}
+## Image Repositories
+{{$registryK8sRepo := DefaultParam .CL2_REGISTRY_K8S_REPO "registry.k8s.io"}}
```
vincepri, line 78:
```suggestion
{{$imageRegistry := DefaultParam .CL2_REGISTRY_K8S_REPO "registry.k8s.io"}}
```

clusterloader2/testing/load/config.yaml
```diff
@@ -74,7 +74,11 @@
...
+{{$gcrRepo := DefaultParam .CL2_GCR_REPO "gcr.io"}}
```
vincepri, line 79: This field and above seem different registries for different images, I'll defer to this repository's maintainers here, but maybe we should try to find some naming that clarifies a bit more where they're set?

clusterloader2/testing/load/config.yaml
rackeric: I gave this a more descriptive name, `registryK8sRepo`, since there are 2 image registries used and I thought it would help highlight which repo exactly was being specified.

clusterloader2/testing/load/config.yaml
rackeric: Same, I'll take any suggestion made from the code owners.

PR #2706, opened 2024-06-06
allows for the exec_service pod image to be pulled from a custom repo

rackeric:
#### What type of PR is this?

/kind feature

Some clusters may not be able to reach the upstream Kubernetes image repos, like registry.k8s.io, but can reach some internal mirror of these. This will allow to set a custom image repo for the exec_service that cluster loader uses to create the exec pods. You can supply the image repo via a clusterloader2 CLI flag `--registry-k8s-repo`

Fixes # https://github.com/kubernetes/perf-tests/issues/2704

### DIFF
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -61,7 +61,8 @@ type ClusterConfig struct {
 // ExecServiceConfig represents all flags used by service config.
 type ExecServiceConfig struct {
 	// Determines if service config should be enabled.
-	Enable bool
+	Enable        bool
+	ImageRegistry string
 }
 
 // ModifierConfig represent all flags used by test modification
--- a/clusterloader2/pkg/execservice/exec_service.go
+++ b/clusterloader2/pkg/execservice/exec_service.go
@@ -67,10 +67,16 @@ func InitFlags(c *config.ExecServiceConfig) {
 		true,
 		"Whether to enable exec service that allows executing arbitrary commands from a pod running in the cluster.",
 	)
+	flags.StringVar(
+		&c.ImageRegistry,
+		"registry-k8s-repo",
+		"registry.k8s.io",
+		"FQDN of registry.k8s.io image repo",
+	)
 }
 
 // SetUpExecService creates exec pod.
-func SetUpExecService(f *framework.Framework, _ config.ExecServiceConfig) error {
+func SetUpExecService(f *framework.Framework, c config.ExecServiceConfig) error {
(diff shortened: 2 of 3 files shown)


clusterloader2/pkg/config/cluster.go
```diff
@@ -61,7 +61,8 @@ type ClusterConfig struct {
 // ExecServiceConfig represents all flags used by service config.
 type ExecServiceConfig struct {
 	// Determines if service config should be enabled.
-	Enable bool
+	Enable 			bool
+	RegistryK8sRepo string
```
vincepri, line 65: Could we run gofmt on these?

clusterloader2/pkg/execservice/exec_service.go
```diff
@@ -67,6 +67,12 @@ func InitFlags(c *config.ExecServiceConfig) {
 		true,
 		"Whether to enable exec service that allows executing arbitrary commands from a pod running in the cluster.",
 	)
+	flags.StringVar(
+		&c.RegistryK8sRepo,
```
vincepri, line 71: Should we just call this `ImageRegistry` instead?

clusterloader2/pkg/config/cluster.go
vincepri, line 65:
```suggestion
	ImageRegistry string
```

vincepri:
@wojtek-t @mboersma would you be able to take a look?

mboersma (approved):
/lgtm

rackeric:
@wojtek-t @mboersma I rebased my changes and lost the LGTM tags. but also, these tests seem to be failing for things unrelated to my change. how could I troubleshoot this? CC @vincepri

clusterloader2/pkg/execservice/exec_service.go
```diff
@@ -82,6 +88,7 @@ func SetUpExecService(f *framework.Framework, _ config.ExecServiceConfig) error
 	mapping["Name"] = execDeploymentName
 	mapping["Namespace"] = execDeploymentNamespace
 	mapping["Replicas"] = execPodReplicas
+	mapping["ImageRegistry"] = c.ImageRegistry
```
wojtek-t, line 91: Test failures are relevant - there is no `c` variable here....

rackeric: ack, thanks. getting it fixed now.

rackeric: all tests are passing now.

PR #2732, opened 2024-06-17
Update go version & Minor fixes to unblock dependencies upgrade

wojtek-t:

Fixes #


wojtek-t:
Blocked on https://github.com/kubernetes/test-infra/pull/32772

tosi3k:
Would it make sense to squash the commits or do you prefer to have it the current way?

If latter, then feel free to unhold (treat my nits as non-blocking).

/lgtm
/hold

wojtek-t:
Leaving as is, in case we will need to revert some part.

/hold cancel

PR #2733, opened 2024-06-17
Fix linter errors

wojtek-t:


aojea:
interesting 

https://github.com/kubernetes/perf-tests/blob/945835f70d509f9bcf1eb64bdc0581801e61cdb3/.golangci.yml#L35

PR #2757, opened 2024-07-01
prometheus: add optional path with additional manifests

marseel:
#### What type of PR is this?

/kind feature

In the past, we added all service/pod monitors directly into manifests directory within cl2/prometheus. This is not really scalable solution if one wants to test other components. Let's add option to allow to specify additional monitors from different path.

Example use-case: https://github.com/marseel/perf-tests/commit/11075f78427c079801fb7b6ed53041dbad05a763
Example command:
```
CL2_PROMETHEUS_PVC_ENABLED=false go run ./cmd/clusterloader.go             -v=4             --testconfig=./testing/dns/config.yaml             --provider=kind             --enable-prometheus-server             --tear-down-prometheus-server=false             --nodes=2             --report-dir=./report                     --kubeconfig=$HOME/.kube/config --prometheus-additional-monitors-path=./testing/dns/manifests/ --prometheus-apiserver-scrape-port=6443
```

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -95,17 +95,12 @@ func InitFlags(p *config.PrometheusConfig) {
 	flags.BoolEnvVar(&p.ScrapeMastersWithPublicIPs, "prometheus-scrape-masters-with-public-ips", "PROMETHEUS_SCRAPE_MASTERS_WITH_PUBLIC_IPS", false, "Whether to scrape master machines using public ips, instead of private.")
 	flags.IntEnvVar(&p.APIServerScrapePort, "prometheus-apiserver-scrape-port", "PROMETHEUS_APISERVER_SCRAPE_PORT", 443, "Port for scraping kube-apiserver (default 443).")
 	flags.StringEnvVar(&p.SnapshotProject, "experimental-snapshot-project", "PROJECT", "", "GCP project used where disks and snapshots are located.")
-	flags.StringEnvVar(&p.ManifestPath, "prometheus-manifest-path", "PROMETHEUS_MANIFEST_PATH", "", "Path to the prometheus manifest files.")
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -342,17 +350,21 @@ func (pc *Controller) GetFramework() *framework.Framework {
...
+func (pc *Controller) applyManifests(path, manifestGlob string) error {
 	return pc.framework.ApplyTemplatedManifests(
-		pc.manifestsFS(), manifestGlob, pc.templateMapping, client.Retry(apierrs.IsNotFound))
+		pc.manifestsFS(path), manifestGlob, pc.templateMapping, client.Retry(apierrs.IsNotFound))
 }
 
-func (pc *Controller) manifestsFS() fs.FS {
-	if pc.clusterLoaderConfig.PrometheusConfig.ManifestPath != "" {
-		return os.DirFS(pc.clusterLoaderConfig.PrometheusConfig.ManifestPath)
+func (pc *Controller) manifestsFS(path string) fs.FS {
```
Argh4k, line 362: I feel like it would be more readable to keep it the way it is and in applyDefaultManifests just use pc.ManifestsFS() and in applyManifests jus use the path passed directly. Then in line 296 you could just run pc.ApplyManifests with os.DirFS(additionalPath)

The problem that I have is that when reading the code I had to remind myself that ManifestPath in prometheus config is deprecated and we want to default to embedded . But in the case of additionalMonitorsPath we do not want to default to embedded manifests and you need to do additional check in line 294.

marseel: ah, I missed it's actually deprecated. How about I just remove it in that case?

Argh4k: 👍  I've checked and it does not seem to be used in perf-tests/test-infra repositories.

marseel: pushed new commit with removal, PTAL

PR #2764, opened 2024-07-08
adds WatchList Latency to APIResponsivenessPrometheus

p0lyn0mial:
#### What type of PR is this?
/kind feature

`WatchList` latency is gathered for 50th, 90th and 99th duration quantiles for watch list requests broken down by group, resource, scope.

The new metric (https://github.com/kubernetes/kubernetes/pull/120490) allows for comparing watch-list requests with standard list requests and measuring performance of the new requests in general.

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
+++ b/clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
@@ -62,6 +62,11 @@ const (
 
 	countFastQuery = "sum(increase(%v_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
 
+	// watchListLatencyMetricName the name of the metrics as defined in the api server
+	watchListLatencyMetricName = "apiserver_watch_list_duration_seconds"
+	// watchListLatencyQuery placeholders must be replaced with (1) quantile (2) query window size
+	watchListLatencyQuery = "histogram_quantile(%.2f, sum(rate(%v_bucket{}[%v])) by (group, version, resource, scope, le))"
+
 	// exclude all buckets of 1s and shorter
 	filterGetAndMutating = `verb!~"WATCH|LIST", subresource!="proxy", le="1"`
 	// exclude all buckets below or equal 5s
@@ -186,6 +191,17 @@ func (a *apiResponsivenessGatherer) gatherAPICalls(executor common.QueryExecutor
 				sample.Metric["quantile"] = model.LabelValue(fmt.Sprintf("%.2f", q))
 			}
 			latencySamples = append(latencySamples, samples...)
+
+			query = fmt.Sprintf(watchListLatencyQuery, q, watchListLatencyMetricName, promDuration)
+			samples, err = executor.Query(query, endTime)
+			if err != nil {
+				return nil, err
+			}
+			for _, sample := range samples {
+				sample.Metric["quantile"] = model.LabelValue(fmt.Sprintf("%.2f", q))
+				sample.Metric["verb"] = "WATCHLIST"
+			}
+			latencySamples = append(latencySamples, samples...)
 		}
 	} else {
 		// Latency measurement is based on 5m window aggregation,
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
```diff
@@ -62,6 +62,11 @@ const (
 
 	countFastQuery = "sum(increase(%v_bucket{%v}[%v])) by (resource, subresource, scope, verb)"
 
+	// watchListLatencyMetricName the name of the metrics as defined in the api server
+	watchListLatencyMetricName = "apiserver_watch_list_duration_seconds"
+	// watchListLatencyQuery placeholders must be replaced with (1) quantile (2) query window size
+	watchListLatencyQuery = "histogram_quantile(%.2f, sum(rate(%v_bucket{}[%v])) by (group, version, resource, scope, le))"
+
```
wojtek-t, line 69: Looking into test results:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/2764/pull-perf-tests-clusterloader2/1810319004401668096/artifacts/APIResponsivenessPrometheus_simple_load_2024-07-08T14:50:01Z.json

I see only one entry for watchlist (pod list on namespace scope).

Is that expected? What is issueing this request?

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus_test.go
```diff
@@ -85,12 +85,26 @@ type summaryEntry struct {
 }
 
 type fakeQueryExecutor struct {
-	samples []*sample
+	metricSamples map[string][]*sample
 }
 
 func (ex *fakeQueryExecutor) Query(query string, _ time.Time) ([]*model.Sample, error) {
+
```
wojtek-t, line 92: nit: remove empty line

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus_test.go
```diff
@@ -85,12 +85,26 @@ type summaryEntry struct {
...
+	var foundSamples bool
+	var metricSamples []*sample
+	for k, v := range ex.metricSamples {
```
wojtek-t, line 95: I don't understand it - why you don't want to have:
```
metricSamples, ok := ex.metricsSamples[query]
if !ok {
  return nil, fmt.Errorf(...)
}
...
```

wojtek-t (commented):
Small comments - overall it looks the way I wanted it to do.

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
p0lyn0mial: good question.
do we have metrics so that i could execute the prom query manually ?
could it be that the run was using a server with enabled watchlist feature ?

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus_test.go
p0lyn0mial: it is a simplification - allows for defining a simpler query for the tests.

for example, a test definition can define output for `apiserver_request_sli_latency` query whereas the actual query used by the cl is: `quantile_over_time(0.99, apiserver:apiserver_request_sli_latency_1m:histogram_quantile{verb!="WATCH", subresource!~"log|exec|portforward|attach|proxy"}[60s])`

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
p0lyn0mial: > could it be that the run was using a server with enabled watchlist feature ?

it looks like this was our case, the most recent run doesn't have any entries of watchlist (the feature was turned off on the server some time ago)

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/perf-tests/2764/pull-perf-tests-clusterloader2/1818234814688399360/artifacts/APIResponsivenessPrometheus_simple_load_2024-07-30T11:06:18Z.json

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus_test.go
wojtek-t: OK - I think it's debatable what is easier to follow, but.. I'm not going to fight for it :)

clusterloader2/pkg/measurement/common/slos/api_responsiveness_prometheus.go
wojtek-t: So I was wondering why we've seen only one such entry, but I guess it's because it's disabled in client-go by default (and was enabled only in KCM). So that makes sense.

PR #2765, opened 2024-07-10
Enable DNS propagation measurement for huge-service test

besher-massri:
Enabling DNS propagation measurement in huge-service test.
Measures the DNS propagation of the statefulset created in huge-service.

Required new env variable:
- `CL2_ENABLE_DNS_PROPAGATION_MEASUREMENT` should be set to true to enable the measurement

Optional new env variable:
- `CL2_DNS_PROPAGATION_THRESHOLD` time threshold for DNS propagation, defaults to 10s.

In addition, this measurement assumes huge-service is run with large statefulset enabled, hence the following existing vars will be needed as well 
- `CL2_ENABLE_LARGE_STATEFULSET` should be set to true to enable adding large statefulset
- `CL2_STATEFULSET_ENDPOINTS` indicates the size of the statefulset

/kind feature

### DIFF
--- a/clusterloader2/testing/huge-service/modules/measurements.yaml
+++ b/clusterloader2/testing/huge-service/modules/measurements.yaml
@@ -1,5 +1,8 @@
 # Valid actions: "start", "gather"
 {{$action := .action}}
+{{$statefulsetEndpoints := .statefulsetEndpoints}}
+{{$serviceName := .serviceName}}
+{{$namespaceIdx := .namespaceIdx}}
 
 {{$ALLOWED_SLOW_API_CALLS := DefaultParam .CL2_ALLOWED_SLOW_API_CALLS 0}}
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/huge-service/modules/measurements.yaml
```diff
@@ -70,6 +78,19 @@ steps:
       defaultAllowedRestarts: {{$ALLOWED_CONTAINER_RESTARTS}}
       customAllowedRestarts: {{YamlQuote $CUSTOM_ALLOWED_CONTAINER_RESTARTS 4}}
 {{end}}
+{{if $ENABLE_DNS_PROPAGATION_MEASUREMENT}}
+  - Identifier: DnsPropagation
+    Method: DnsPropagation
+    Params:
+      action: {{$action}}
+      DNSPropagationProbeStatefulSet: {{.serviceName}}
```
jprzychodzen, line 86: Nit: Please create variables at the top (like with $statefulsetEndpoints)

besher-massri: Fixed, thanks!

PR #2773, opened 2024-07-20
Fix Kind pluralization.

jm-franc:
#### What type of PR is this?

/kind bug

Fixes a bug where tests using resources that are not properly pluralizes as described in https://github.com/kubernetes/client-go/issues/1082 systematically fail.

For example, any test creating a `Gateway` object will fail (as it's turned into 'gatewaies').


jm-franc:
Pinging @marseel for approval.

/assign @marseel

PR #2794, opened 2024-07-23
CL2: Add isFatal param to WaitForPods and enable failing measurements

learnitall:
#### What type of PR is this?

/kind feature

This PR introduces a set of commits that enable measurements to fail during their `Execute` method. Errors that are returned during a measurement's `Execute` method are passed up the stack to the simple test executor, which then passes them through a function called `isErrsCritical`. This function can control if the entire test should bail, based on the criticalness of the errors passed to it. Right now, it's a no-op:

https://github.com/kubernetes/perf-tests/blob/e7bdd15c9122e55d7dc267768190addc8fdf7b9e/clusterloader2/pkg/test/simple_test_executor.go#L386-L389

This PR modifies the function to check if any of the errors in the given list are fatal. A measurement can mark an error as fatal if it returns an error wrapped by a newly added `FatalError` type.

This PR also modifies the `WaitForPods` measurement to utilize this functionality. A new param named `isFatal` can control if a test should bail if the `WaitForPods` measurement is unable to successfully determine that a set of required pods are running and ready. This can be helpful for tests that require some infrastructure within the cluster to run. For example:

```yaml
name: is-fatal-test
namespace:
  number: 1
tuningSets:
- name: Uniform1qps
  qpsLoad:
    qps: 1
steps:
- name: Create test pod
  phases:
  - namespaceRange:
      min: 1
      max: 1
    replicasPerNamespace: 1
    tuningSet: Uniform1qps
    objectBundle:
    - basename: my-test-pod
      objectTemplatePath: ./my-test-pod.yaml
- name: Wait for test pod
  measurements:
  - Identifier: WaitForRunningPods
    Method: WaitForRunningPods
    Params:
      labelSelector: name=my-test-pod
      desiredPodCount: 1
      timeout: 60s
      isFatal: true
- name: Sleep
  measurements:
  - Identifier: Sleep
    Method: Sleep
    Params:
      duration: 10s
```

In this config, the `Sleep` measurement will only be executed if the test pod has been successfully deployed.

### DIFF
--- a/clusterloader2/pkg/errors/fatal_error.go
+++ b/clusterloader2/pkg/errors/fatal_error.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrCritical = errors.New("criticl error")
+
+// NewErrCritical wraps an error and marks it as fatal.
+func NewErrCritical(err error) error {
+	return fmt.Errorf("%w: %w", ErrCritical, err)
+}
--- a/clusterloader2/pkg/measurement/common/wait_for_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_pods.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"k8s.io/klog/v2"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
 	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
 	"k8s.io/perf-tests/clusterloader2/pkg/util"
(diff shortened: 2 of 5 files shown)


learnitall:
cc: @marseel

learnitall:
Gentle ping @mborsz and @wojtek-t. Do you mind taking a look when you have a moment? :pray:

clusterloader2/pkg/errors/fatal_error.go
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+package errors
+
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrFatal = errors.New("fatal error")
```
marseel, line 24: As we already have `isErrsCritical` let's call it `ErrCritical` as well.
Or rename `isErrsCritical` to `isErrsFatal`

clusterloader2/pkg/measurement/common/wait_for_pods.go
```diff
@@ -72,7 +79,16 @@ func (w *waitForRunningPodsMeasurement) Execute(config *measurement.Config) ([]m
 	if err != nil {
 		return nil, err
 	}
+
 	_, err = measurementutil.WaitForPods(ctx, podStore, options)
+	if err != nil {
```
marseel, line 84: nit:
how about something simpler:
```
if err != nil && isFatal {
    return nil, errors.NewErrFatal(err)
}
```

marseel (commented):
Looks good, just small nits.

clusterloader2/pkg/measurement/common/wait_for_pods.go
learnitall: Ah that's much better, thanks.

clusterloader2/pkg/errors/fatal_error.go
learnitall: Good idea, I replaced `fatal` with `critical`

learnitall:
Thanks for the review @marseel! I just pushed some changes addressing your feedback.

marseel:
Thanks!
/lgtm
/approve

PR #2795, opened 2024-07-24
Bump clusterloader deps

wojtek-t:
Step towards https://github.com/kubernetes/perf-tests/pull/2774

PR #2806, opened 2024-08-07
clusterloader2: Bump Kubernetes dependency to v1.29.7

mboersma:
#### What type of PR is this?

/kind cleanup

This updates the vendoring of k8s.io in clusterloader2 to Kubernetes version v1.29.7.

Refs #2795

I tried to update to v1.30.3 as well, but I'm running into an issue where the `Flag` "v" is defined by prometheus and then again by component-base, causing a runtime error. I'll investigate further when I have time, but this change stands on its own.


mboersma:
@mborsz @wojtek-t is this bump worthwhile? Just trying to keep vendoring relatively current now that v1.28 will be deprecated.

Getting to v1.31 is trickier, but I can revisit that if desired.

wojtek-t:
Yes - it's definitely desired - thanks!

/lgtm
/approve

PR #2861, opened 2024-09-02
Fix image parametrization in load config

Michcioperz:
#### What type of PR is this?

/kind bug

A previous change https://github.com/kubernetes/perf-tests/pull/2803 introduced a variable parametrizing the image with fallback to a default value, but instead used the variable that does not have a fallback. This can cause the tests to fail.

None that I know of


mboersma (approved):
/lgtm

marseel:
Thanks!
/lgtm
/approve

PR #2864, opened 2024-09-05
Rename AKS provider to Azure

Jont828:
/kind feature

Fixes #

### DIFF
--- a/clusterloader2/pkg/provider/provider.go
+++ b/clusterloader2/pkg/provider/provider.go
@@ -123,10 +123,10 @@ func NewProvider(initOptions *InitOptions) (Provider, error) {
 		configs[RootKubeConfigKey] = initOptions.KubemarkRootKubeConfigPath
 	}
 	switch initOptions.ProviderName {
-	case AKSName:
-		return NewAKSProvider(configs), nil
 	case AWSName:
 		return NewAWSProvider(configs), nil
+	case AzureName, AKSName: // AKSName is for backward compatibility.
+		return NewAzureProvider(configs), nil
 	case AutopilotName:
 		return NewAutopilotProvider(configs), nil
 	case EKSName:
--- a/clusterloader2/README.md
+++ b/clusterloader2/README.md
@@ -26,7 +26,7 @@ These flags are required for any test to be run.
  - `kubeconfig` - path to the kubeconfig file.
  - `testconfig` - path to the test config file. This flag can be used multiple times
 if more than one test should be run.
- - `provider` - Cluster provider. Options are: gce, gke, kind, kubemark, aws, local, vsphere, skeleton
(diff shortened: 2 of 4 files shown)


clusterloader2/pkg/provider/provider.go
```diff
@@ -108,10 +108,10 @@ func NewProvider(initOptions *InitOptions) (Provider, error) {
 		configs[RootKubeConfigKey] = initOptions.KubemarkRootKubeConfigPath
 	}
 	switch initOptions.ProviderName {
-	case AKSName:
-		return NewAKSProvider(configs), nil
 	case AWSName:
 		return NewAWSProvider(configs), nil
+	case AzureName:
```
mboersma, line 113:
```suggestion
	case AzureName, AKSName:
```

Should continue to handle `--provider=aks` for backward compatibility?

mboersma (commented):
Could you also update the [README](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2#required) for CL2 to mention `azure` as an option for `provider`?

jackfrancis:
cc @JohnRusk

Do you have any thoughts on this change?

JohnRusk:
I think the stated rationale makes sense.  

The only risk I can think of would be if in fact this statement wasn't _always_ correct: "since it works with non-AKS Azure clusters (such as CAPZ) as well". I.e. if in fact there are expected to be any time of Azure cluster that it does not support - is that a problem?  E.g. does it create an implied mainteance burden to maining this so that it _does_ support them?

But I think my concern there is very minor, and unlikely to any reason to reject the PR.  So, if you're happy with it as it stands, so am I @jackfrancis .

jackfrancis:
> The only risk I can think of would be if in fact this statement wasn't _always_ correct: "since it works with non-AKS Azure clusters (such as CAPZ) as well". I.e. if in fact there are expected to be any kind of Azure cluster that it does not support - is that a problem? E.g. does it create an implied mainteance burden to modify this so that it _does_ support them?

I think that is a valid point, and I would argue that if we discover this to be true in the future (novel Kubernetes + Azure infra use case) that the provider doesn't support, we want to support it. I'm committed to that level of support.

JohnRusk:
Sounds good to me @jackfrancis

clusterloader2/pkg/provider/provider.go
jackfrancis: I'm +1 on this (let's keep the rename of the provider type to `AzureProvider`)

mboersma:
lgtm, but needs a rebase.

mboersma:
@Jont828 this needs to be rebased.

Jont828:
@mboersma Done!

mboersma (approved):
/lgtm

Jont828:
@mboersma Just rebased it, shouldn't have changed anything.

mboersma (approved):
/lgtm

PR #2898, opened 2024-09-16
CL2: Add refreshInterval param to WaitForNodes and WaitForRunningPods measurements

alyssa1303:
#### What type of PR is this?

/kind feature

This PR introduces a new param called `refreshInterval` to `WaitForNodes` and `WaitForRunningPods` measurements. The current value of refresh interval in both measurements are currently fixed. However, for test that includes a small number of nodes, 30 seconds for refreshing is too long, which can lead to inaccurate result (5 nodes that usually take 10 - 15 seconds to get ready almost always end up being reported as taking 30 seconds due to the fixed interval). At the same time, for test that includes a large number of pods and numbers, 5 seconds for pod's refresh interval might be too frequent.

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_nodes.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_nodes.go
@@ -61,6 +61,12 @@ func (w *waitForNodesMeasurement) Execute(config *measurement.Config) ([]measure
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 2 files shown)


mboersma (approved):
/ok-to-test
/lgtm

Looks like we're already using `refreshInterval` similarly here:
https://github.com/kubernetes/perf-tests/blob/5fad5f752fcfece3d0b330304b45a912f102563b/clusterloader2/pkg/measurement/common/wait_for_generic_k8s_object.go#L66-L69

marseel:
/approve

thanks!

PR #2914, opened 2024-09-27
Add option to tune sync interval

marseel:
/kind feature
/kind bug

Add an option to change the synchronization interval.
Also, fixes build regressions from #2734


tosi3k:
/lgtm

Thanks!

PR #2981, opened 2024-11-07
Add a sleep after service delete to allow NEGs get garbage collected.

sawsa307:
#### What type of PR is this?
/kind bug

Currently NEG scale tests fail during test suite cleanup because the cluster network cannot be deleted while there are NEGs referencing it. NEG garbage collect happens asynchronously after k8s services are deleted, so we need to add a wait and allow NEGs to be cleaned up before we attempt to delete the network.

Fixes ci-kubernetes-e2e-gke-rapid-300-neg-performance.Overall


clusterloader2/testing/neg/config.yaml
```diff
@@ -118,3 +118,9 @@ steps:
     Method: WaitForControlledPodsRunning
     Params:
       action: gather
+- name: Wait 15 minutes
+  measurements:
+    - Identifier: Wait
+      Method: Sleep
+      Params:
+        duration: 15m
```
dlapcevic, line 126: I suggest parametrizing the sleep duration value, so that separate tests can have a custom wait at the end. This will allow more accurate test configurations, and won’t require making changes to the test manifests in this repo.

Check out the CL2 params at top of the file, and add one such.
For example:
```
{{$waitAfterDeletion := DefaultParam .CL2_WAIT_AFTER_DELETION “15m”}}
...
- name: Wait after deletion
  measurements:
    - Identifier: Wait
      Method: Sleep
      Params:
        duration: {{$waitAfterDeletion}}
```

Please retest it before submitting.

clusterloader2/testing/neg/config.yaml
```diff
@@ -118,3 +118,9 @@ steps:
     Method: WaitForControlledPodsRunning
     Params:
       action: gather
+- name: Wait 15 minutes
```
dlapcevic, line 121: nit: rename to "Wait after deletion" or similar

dlapcevic (changes requested):

clusterloader2/testing/neg/config.yaml
sawsa307: Updated, thank you!

clusterloader2/testing/neg/config.yaml
sawsa307: Done.

sawsa307:
Addressed the comments, please take another look! @dlapcevic

dlapcevic (approved):

dlapcevic:
@mborsz, could you please approve this to be merged?

PR #3032, opened 2024-11-29
Add request count to watchlist tests

marseel:
#### What type of PR is this?

/kind feature

Let's add request count to watchlist tests as well. 
Additionally switch local testing of perfdash from using githubConfig to local config.

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See deployment.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG?=2.50
+TAG?=2.51
 
 REPO?=gcr.io/k8s-staging-perf-tests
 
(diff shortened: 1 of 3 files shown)


perfdash/config.go
```diff
@@ -658,6 +658,11 @@ var (
 				OutputFilePrefix: "APIResponsivenessPrometheus_simple",
 				Parser:           parsePerfData,
 			}},
+			"LoadRequestCount_PrometheusSimple": []TestDescription{{
```
p0lyn0mial, line 661: that's awesome, thanks!

perfdash/Makefile
```diff
@@ -19,7 +19,7 @@ run: perfdash
 		--address=0.0.0.0:8080 \
 		--builds=20 \
 		--force-builds \
-		--githubConfigDir=https://api.github.com/repos/kubernetes/perf-tests/contents/perfdash/test
+		--configPath=test/jobs.yaml
```
p0lyn0mial, line 22: ok, this changes the cfg for a local binary.

marseel: yes, just for testing easier :)

PR #3047, opened 2024-12-09
Tear down CL2 also on test config compilation failures

ali-a-a:
#### What type of PR is this?
/kind cleanup
When we fail to compile the config file (e.g., validation errors), the namespace created by CL2 will remain untouched. This PR tries to tear down created resources on any failures during the setup.
Fixes #


ali-a-a:
Would appreciate your opinion on this change

@wojtek-t @mborsz

PR #3050, opened 2024-12-13
Bump clusterloader deps

wojtek-t:

PR #3051, opened 2024-12-13
Remove dependency on kubernetes test/e2e in clusterloader2

wojtek-t:
Ref https://github.com/kubernetes/perf-tests/pull/3050#discussion_r1884019995

/assign @tosi3k

### DIFF
--- a/clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
+++ b/clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
@@ -59,7 +59,6 @@ type Grabber struct {
 	grabFromClusterAutoscaler         bool
 	masterName                        string
 	registeredMaster                  bool
-	waitForControllerManagerReadyOnce sync.Once
 }
 
 // deprecatedMightBeMasterNode returns true if given node is a registered master.
--- a/clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
+++ b/clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
@@ -20,16 +20,11 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"sync"
-	"time"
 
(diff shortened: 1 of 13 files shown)


tosi3k:
I think the signatures of the `SSH` method are a bit different in sshutil versions of k/k and CL2, so we need to adjust its usages in CL2 as well:

k/k:
https://github.com/kubernetes/kubernetes/blob/b1f2af04328936c2fa79db4af14f5c6ad9160748/test/e2e/framework/ssh/ssh.go#L203-L206

CL2:
https://github.com/kubernetes/perf-tests/blob/c92f79f36e86b72919ded10c7c574a04268e877a/clusterloader2/pkg/measurement/util/ssh.go#L43-L44

wojtek-t:
> I think the signatures of the SSH method are a bit different in sshutil versions of k/k and CL2, so we need to adjust its usages in CL2 as well:

Sorry - I didn't do "git add" of the file :) Should be fixed now.

clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
```diff
@@ -192,27 +190,6 @@ func (g *Grabber) GrabFromControllerManager() (ControllerManagerMetrics, error)
 		return ControllerManagerMetrics{}, fmt.Errorf("master's Kubelet is not registered. Skipping ControllerManager's metrics gathering")
 	}
 
-	var err error
```
wojtek-t, line 195: We didn't have anything like that in kube-scheduler or in any other components.

To fully eliminate dependency on test/e2e from k/k I'm taking the simplest approach of removing it here - to unblock https://github.com/kubernetes/perf-tests/pull/3050

tosi3k: SG

tosi3k:
/lgtm
/approve
/hold

I'm holding just as a precaution - feel free to unhold and merge the change on Monday :)

PR #3076, opened 2024-12-18
clusterloader2: Add missing templated var for prometheus apiserver port in master-ip service

devodev:
#### What type of PR is this?

/kind bug

When using an alternative port for the API server, the master-ip service template is not using the `PROMETHEUS_APISERVER_SCRAPE_PORT` set by the `--prometheus-apiserver-scrape-port` flag and defaults to 443.

This PR mirrors the behaviour seen in `master-endpoints.yaml` into `master-service.yaml` and fixes the ` Waiting for Prometheus stack to become healthy...` step:

```
util.go:101] 7/9 targets are ready, example not ready target: {map[endpoint:apiserver instance:172.18.0.2:443 job:master namespace:monitoring service:master] down}
```

I've also noticed an error in the ServiceMonitor template where the label selector was missing `matchLabels:` and added a commit to fix.

None

None


devodev:
Opened the PR as draft until I complete the CLA process.

devodev:
Waiting on this PR to get merged: https://github.com/cncf/gitdm/pull/562

devodev:
Not sure what I did wrong at first, but playing more with this, it seems this is not needed.

PR #3077, opened 2024-12-18
Don't overwrite ResourceUsageSummary output for different namespaces

devodev:
#### What type of PR is this?

/kind bug

The `ResourceUsageSummary` measurement implementation does not take into account the namespace when creating output files, which can lead to files being overwritten if multiple measurements are used for different namespaces.

This PR adds the namespace to the output filename to avoid collisions.

None

None


devodev:
Waiting on this PR to get merged: https://github.com/cncf/gitdm/pull/562

devodev:
I think I misunderstood this measurement and don't think this is required anymore.

PR #3078, opened 2024-12-19
clusterloader2: Avoid logging and awaiting waitgroup when chaosmonkey is disabled

devodev:
#### What type of PR is this?

/kind bug

Not really a bug but I noticed that the waitgroup returned from chaosMonkey.Init() is always non-nil and therefore we always log and await the waitgroup, which can be confusing when chaosmonkey nodekiller is disabled. I also updated to avoid emitting an empty log statement when the summary is empty.

```
I1219 14:48:15.653932   23514 simple_test_executor.go:80] Waiting for the chaos monkey subroutine to end...
I1219 14:48:15.653947   23514 simple_test_executor.go:82] Chaos monkey ended.
I1219 14:48:15.657172   23514 simple_test_executor.go:102] 
```

None

None

### DIFF
--- a/clusterloader2/pkg/chaos/monkey.go
+++ b/clusterloader2/pkg/chaos/monkey.go
@@ -17,7 +17,6 @@ limitations under the License.
 package chaos
 
 import (
-	"fmt"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


marseel:
Thanks for contribution!

/ok-to-test
/lgtm
/approve

PR #3087, opened 2025-01-08
Bump golang in perf-test to 1.23.4

wojtek-t:


wojtek-t:
I opened https://github.com/kubernetes/test-infra/pull/34095 that should hopefully fix the failures.

PR #3098, opened 2025-01-13
Update ClusterLoader2's temp KubeConfig PKI for ServiceAccount

sambdavidson:
Populating a ServiceAccount's Secret with CA and JWT data is a legacy flow [replaced in 1.24+](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#urgent-upgrade-notes) with the `kube-root-ca.crt ConfigMap` and `token` resources.  Public documentation now encourages using `tokens` instead: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#manual-secret-management-for-serviceaccounts

#### What type of PR is this?
/kind bug
/kind deprecation

Clusters using GKE's [CPA](https://cloud.google.com/kubernetes-engine/docs/concepts/about-control-plane-authority) (and many other modern clusters) no longer populate a SA's secret with CA and token information.

### DIFF
--- a/clusterloader2/run-e2e.sh
+++ b/clusterloader2/run-e2e.sh
@@ -88,26 +88,13 @@ cluster_loader_crb_exists=$(kubectl --kubeconfig "${KUBECONFIG}" get clusterrole
 if [[ "$cluster_loader_crb_exists" -eq 0 ]]; then
 	kubectl --kubeconfig "${KUBECONFIG}" create clusterrolebinding cluster-loader --clusterrole=cluster-admin --serviceaccount=default:cluster-loader
 fi
-cluster_loader_secret_exists=$(kubectl --kubeconfig "${KUBECONFIG}" get secret cluster-loader --ignore-not-found | wc -l)
-if [[ "$cluster_loader_secret_exists" -eq 0 ]]; then
-   cat << EOF | kubectl --kubeconfig "${KUBECONFIG}" create -f -
-apiVersion: v1
-kind: Secret
-metadata:
-  name: cluster-loader
-  namespace: default
-  annotations:
-    kubernetes.io/service-account.name: cluster-loader
-type: kubernetes.io/service-account-token
-EOF
-fi
 
 
 # Create a kubeconfig to use the above service account.
 kubeconfig=$(mktemp)
 server=$(kubectl --kubeconfig "${KUBECONFIG}" config view -o jsonpath='{.clusters[0].cluster.server}')
-ca=$(kubectl --kubeconfig "${KUBECONFIG}" get secret cluster-loader -o jsonpath='{.data.ca\.crt}')
-token=$(kubectl --kubeconfig "${KUBECONFIG}" get secret cluster-loader -o jsonpath='{.data.token}' | base64 --decode)
+ca=$(kubectl --kubeconfig "${KUBECONFIG}" get configmap kube-root-ca.crt -o jsonpath='{.data.ca\.crt}' | base64 -w 0)
+token=$(kubectl --kubeconfig "${KUBECONFIG}" --duration=8760h create token cluster-loader)
 echo "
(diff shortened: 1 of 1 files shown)


jprzychodzen:
I'm not exactly following why we should use another way.

Per upgrade notes (I guess it strictly related to https://github.com/kubernetes/kubernetes/pull/108309) we are creating a Service account, as per linked documentation https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets and https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount .

Could you elaborate why it's considered legacy?

liggitt:
> Could you elaborate why it's considered legacy?

Putting non-expiring credentials into API objects where they can be read by parties other than the ones intended to *use* the credentials is poor security practice.

That's why we recommend switching to using TokenRequest to get the token directly in the docs [you linked to](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-long-lived-api-token-for-a-serviceaccount)

> You can still manually create a service account token Secret; for example, if you need a token that never expires. However, using the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) subresource to obtain a token to access the API is recommended instead.

It also means you have to wait for a controller to populate a Secret with credentials, which is racy (the current script in this repo is just getting lucky that a controller is ~usually doing that ~fast before it gets to the line where it tries to read out the token / CA).

jprzychodzen:
I guess we need token refresh to keep current behavior (unlimited CL2 testing time)? Otherwise we are going to hit some interesting bug which will only manifest after 24h of testing.

I'm unsure if there is current such use case in use, however if we are changing this let's make it in a correct way.

sambdavidson:
@jprzychodzen Happy to update it to request 1 year tokens or whatever arbitrarily large size you want. Though keep in mind the length will be clamped by the `--service-account-max-token-expiration` flag. So there are already some bounds in place.

clusterloader2/run-e2e.sh
```diff
@@ -88,26 +88,13 @@ cluster_loader_crb_exists=$(kubectl --kubeconfig "${KUBECONFIG}" get clusterrole
...
-fi
 
 
 # Create a kubeconfig to use the above service account.
 kubeconfig=$(mktemp)
 server=$(kubectl --kubeconfig "${KUBECONFIG}" config view -o jsonpath='{.clusters[0].cluster.server}')
-ca=$(kubectl --kubeconfig "${KUBECONFIG}" get secret cluster-loader -o jsonpath='{.data.ca\.crt}')
-token=$(kubectl --kubeconfig "${KUBECONFIG}" get secret cluster-loader -o jsonpath='{.data.token}' | base64 --decode)
+ca=$(kubectl --kubeconfig "${KUBECONFIG}" get configmap kube-root-ca.crt -o jsonpath='{.data.ca\.crt}' | base64 -w 0)
+token=$(kubectl --kubeconfig "${KUBECONFIG}"  --namespace default --duration=8760h create token cluster-loader)
```
tosi3k, line 97: Nit: drop `--namespace default` since it's a noop?

Also, that would make it consistent with other `kubectl` invocations in this script.

sambdavidson: Done

sambdavidson:
For the record this is what happens when you request a duration greater-than what is supported:

```bash
$  token=$(kubectl --duration=9999h create token test-sa)
Warning: requested expiration of 35996400 seconds shortened to 172800 seconds
```

A warning is printed to stderr but it doesn't affect the contents of $token

```bash
$ echo $token
eyJhbGciOiJSU.........
```

So its a no-op more or less

PR #3138, opened 2025-01-22
Prometheus server should use SA token when talking to apiserver

mengqiy:
#### What type of PR is this?

/kind bug

Prometheus server should use service account token when talking to kube-apiserver. We should not rely on anonymous auth.

Fixes #

We currently create ClusterRole and ClusterRoleBinding for anonymous auth. https://github.com/kubernetes/perf-tests/blob/6b7d01a3b54dc671e92af23fa4e45b144ec4913c/clusterloader2/pkg/prometheus/prometheus.go#L370-L388


tosi3k:
FWIW this is addressed already in https://github.com/kubernetes/perf-tests/pull/3214.

PR #3141, opened 2025-01-27
Rewrite list benchmark to validate large list

serathius:
/kind feature

Removed not needed things and configured it to setup for large LIST:
* Removed secrets, they don't differ at all from testing lists configmaps. For future we would better support pods and CRs instead of secrets.
* Removed list scoped list, they overcomplicate the setup without providing benefits
* Added resourceVersion=0 in test. For future, add sub-scenarios for testing different LIST options.
* Removed automatic resource estimation. They are totally inaccurate, list for this setup they estimated 30k CPU and TBs of memory. I think we should have the default setup have highest sensible values and allow users to lower it for their local environment.
* Added collecting CPU and Memory via `TestMetrics`, found it to return higher values.

/assign @mborsz
cc @mengqiy


serathius:
cc @tosi3k

PR #3142, opened 2025-01-28
List benchmark perfdash type

serathius:
/kind feature
/assign @tosi3k @mborsz


perfdash/config.go
```diff
@@ -666,15 +666,33 @@ var (
 		},
 	}
 
+	benchmarkListDescription = TestDescriptions{
+		"E2E": {
+			"Resources": []TestDescription{{
+				Name:             "List benchmark",
```
mborsz, line 672: All the places seems to use name with different-no-space-naming-schema. You are including spaces etc. It's not clear to me if this will not cause any problems.

mborsz:
You likely need to increase image version in Makefile (see README file)

serathius:
Done

PR #3201, opened 2025-02-19
Implemented LIST benchmark for different content type

ayushpatil2122:
What type of PR is this?
/kind feature

Implement LIST benchmark for different content types (Json, Proto, CBOR, YAML)

Which issue(s) this PR fixes:
#130169

Fixes #


serathius:
Can you properly reference the main issue and other PRs created ? Please note that PRs and Issue are in different repositories so you need to use full path.

util-images/request-benchmark/main.go
```diff
@@ -38,19 +38,58 @@ const (
...
-	namespace = flag.String("namespace", "", "Replace %namespace% in URI with provided namespace")
-	URI       = flag.String("uri", "", "Request URI")
-	verb      = flag.String("verb", "GET", "A verb to be used in requests.")
-	qps       = flag.Float64("qps", -1, "The qps limit for all requests")
+	inflight    = flag.Int("inflight", 1, "Benchmark inflight (number of parallel requests being made to the apiserver")
+	namespace   = flag.String("namespace", "", "Replace %namespace% in URI with provided namespace")
+	URI         = flag.String("uri", "", "Request URI")
+	verb        = flag.String("verb", "GET", "A verb to be used in requests.")
+	qps         = flag.Float64("qps", -1, "The qps limit for all requests")
+	contentType = ContentType("")
```
jprzychodzen, line 56: What was the default before this change? Json or proto? Can we set is a default?

serathius: Empty content type defaults to JSON.

jprzychodzen:
After running this binary with `$ go run ./ --inflight=1 --uri=/api/v1/pods` I have following error:

```
Invalid content type: unsupported content type: 
```

serathius:
> After running this binary with `$ go run ./ --inflight=1 --uri=/api/v1/pods` I have following error:
> 
> ```
> Invalid content type: unsupported content type: 
> ```

We can drop the requirement to fail without providing content type. @ayushpatil2122

jprzychodzen:
> > After running this binary with `$ go run ./ --inflight=1 --uri=/api/v1/pods` I have following error:
> > ```
> > Invalid content type: unsupported content type: 
> > ```
> 
> We can drop the requirement to fail without providing content type. @ayushpatil2122

I would rather set reasonable default.

serathius:
ping @ayushpatil2122

serathius:
Please rebase instead of merging.

PR #3214, opened 2025-02-27
Remove usage of system:anonymous from ClusterRoleBinding used to scra…

emerbe:
…p apiserver metrics

#### What type of PR is this?

/kind bug

Binding any permissions to system:anonymous and system:unauthenticated is dangerous.
This PR removes usage of system:anonymous in CLusterROleBinding used to talk to apiserver. 

What is more, service account needs to be used when talking to apiserver.

Fixes #

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -378,10 +378,10 @@ func (pc *Controller) exposeAPIServerMetrics() error {
 	}
 	createClusterRoleBinding := func() error {
 		_, err := clientSet.GetClient().RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
-			ObjectMeta: metav1.ObjectMeta{Name: "system:anonymous"},
+			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer-binding"},
 			RoleRef:    rbacv1.RoleRef{Kind: "ClusterRole", Name: "apiserver-metrics-viewer"},
 			Subjects: []rbacv1.Subject{
-				{Kind: "User", Name: "system:anonymous"},
+				{Kind: "User", Name: "prometheus-k8s", Namespace: "monitoring"},
 			},
 		}, metav1.CreateOptions{})
 		return err
--- a/clusterloader2/pkg/prometheus/manifests/master-ip/master-serviceMonitor.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/master-ip/master-serviceMonitor.yaml
@@ -36,6 +36,7 @@ spec:
   {{else}}
     interval: 5s
   {{end}}
+    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
   {{if $PROMETHEUS_SCRAPE_MASTER_KUBELETS}}
   - interval: 5s
     port: kubelet


tosi3k:
/lgtm
/approve

Thanks!

tosi3k:
Thanks for adding auxiliary logging.

I checked the `pull-perf-tests-clusterloader2` presubmit and it seems that attempts to query the `/metrics` endpoint of `kube-scheduler` (KS) and `kube-controller-manager` (KCM) fail with the following error message in the presubmit:

```
forbidden: User "system:anonymous" cannot get path "/metrics"
```

At a glance, this looks weird given we bind the `cluster-admin` role to ClusterLoader2's Service Account in run-e2e-sh (which should get access to every single resource and non-resource URL) but:
1. Those failed requests are proxied by kube-apiserver,
2. `kube-apiserver` strips an HTTP request from the `Authorization` header after it's successfully authenticated and before it's further processed ([code](https://github.com/kubernetes/kubernetes/blob/0556b20d3d4aa378b080c1b9375bc59f799464fd/staging/src/k8s.io/apiserver/pkg/endpoints/filters/authentication.go#L87-L88)),
   - See https://github.com/kubernetes/kubernetes/issues/38775 for rationale.
3. Because of that, proxied requests from `kube-apiserver` do not contain CL2's bearer token (nor any other token),
4. Simultaneously, both KS and KCM delegate the authorization of requests coming their way to `kube-apiserver`:
   - See the [NewDelegatingAuthorizationOptions](https://github.com/kubernetes/kubernetes/blob/0556b20d3d4aa378b080c1b9375bc59f799464fd/cmd/kube-scheduler/app/options/options.go#L103) function being called in scheduler's code and the [definition](https://github.com/kubernetes/kubernetes/blob/0556b20d3d4aa378b080c1b9375bc59f799464fd/staging/src/k8s.io/apiserver/pkg/server/options/authorization.go#L78-L94) of that function.
   - Same thing happens for KCM.
5. Because of 3. and 4., the proxied requests are treated as "anonymous" by `kube-apiserver` and are eventually rejected :(.

To work around this, we could override the `--authorization-always-allow-paths` flag for KS and KCM (which defaults to `/healthz,/readyz,/livez`) to also contain the `/metrics` path in our tests.

Easiest way to achieve this workaround would be to open a separate PR to the https://github.com/kubernetes/test-infra repository in which you'd go over every occurrence of the `CONTROLLER_MANAGER_TEST_ARGS` and `SCHEDULER_TEST_ARGS` env vars in [this](https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes/sig-scalability) directory and append `--authorization-always-allow-paths=/healthz,/readyz,/livez,/metrics` to the value of those env vars there.

After that, we can rerun the presubmits.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -371,17 +371,41 @@ func (pc *Controller) exposeAPIServerMetrics() error {
 		_, err := clientSet.GetClient().RbacV1().ClusterRoles().Create(context.TODO(), &rbacv1.ClusterRole{
 			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer"},
 			Rules: []rbacv1.PolicyRule{
-				{Verbs: []string{"get"}, NonResourceURLs: []string{"/metrics"}},
+		            {
+		                Verbs:     []string{"get"},
+		                NonResourceURLs: []string{"/metrics"},
+		            },
+		            {
+		                Verbs:     []string{"get", "connect"},
```
vinayakankugoyal, line 379: this seems like a significant increase in the scope of this role. Could you add here or in the PR description why this is and the other newly added permissions are needed for this change?

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -371,17 +371,41 @@ func (pc *Controller) exposeAPIServerMetrics() error {
...
+		                Verbs:     []string{"get"},
+		                Resources: []string{"pods/metrics"},
+		                APIGroups: []string{""},
+		            },
+		            {
+		                Verbs:     []string{"list", "watch", "get"},
+		                Resources: []string{"pods"},
+		                APIGroups: []string{""},
+		            },
+		            {
```
vinayakankugoyal, line 393: this role is named apiserver-metrics-viewer so having nodes/metrics and pods/metrics is not accurate. Do we need to rename this role or add new ones for pods and nodes metrics?

clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
```diff
@@ -276,6 +276,8 @@ func (g *Grabber) getMetricsFromPod(client clientset.Interface, podName string,
 		Suffix("metrics").
 		Do(context.TODO()).Raw()
 	if err != nil {
+		klog.Errorf("Getting metrics failed, error: %v", err)
+		klog.Errorf("Getting metrics failed, rawOutput: %s", string(rawOutput))
```
tosi3k, line 280: Nit: let's clean this up.

clusterloader2/pkg/measurement/common/scheduler_latency.go
```diff
@@ -310,6 +310,7 @@ func (s *schedulerLatencyMeasurement) sendRequestToScheduler(c clientset.Interfa
 
 		if err != nil {
 			klog.Errorf("Send request to scheduler failed with err: %v", err)
+			klog.Errorf("Getting metrics failed, body: %s", string(body))
```
tosi3k, line 313: Nit: let's clean this up.

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubeProxy.yaml
```diff
@@ -16,6 +16,7 @@ spec:
     # (# nodes <= 1000) and 1min interval in big clusters (# nodes > 1000)
     # TODO(mm4tt): Once we prove the interval works in big clusters, simplify the expression.
     - interval: {{MinInt 60 (MultiplyInt 30 (AddInt 1 (DivideInt .Nodes 1001)))}}s
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
tosi3k, line 19: Nit: since it was not needed, then let's clean this up.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -371,17 +371,41 @@ func (pc *Controller) exposeAPIServerMetrics() error {
...
 	}
 	createClusterRoleBinding := func() error {
 		_, err := clientSet.GetClient().RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
-			ObjectMeta: metav1.ObjectMeta{Name: "system:anonymous"},
+			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer-binding"},
 			RoleRef:    rbacv1.RoleRef{Kind: "ClusterRole", Name: "apiserver-metrics-viewer"},
 			Subjects: []rbacv1.Subject{
-				{Kind: "User", Name: "system:anonymous"},
+				{Kind: "ServiceAccount", Name: "prometheus-k8s", Namespace: "monitoring"},
+				{Kind: "ServiceAccount", Name: "cluster-loader", Namespace: "default"},
```
tosi3k, line 408: Let's drop the `cluster-loader` SA from the list of subjects - it's already assigned the `cluster-admin` role in `run-e2e.sh` script, so this is redundant.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -371,17 +371,41 @@ func (pc *Controller) exposeAPIServerMetrics() error {
 		_, err := clientSet.GetClient().RbacV1().ClusterRoles().Create(context.TODO(), &rbacv1.ClusterRole{
 			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer"},
 			Rules: []rbacv1.PolicyRule{
-				{Verbs: []string{"get"}, NonResourceURLs: []string{"/metrics"}},
+		            {
+		                Verbs:     []string{"get"},
+		                NonResourceURLs: []string{"/metrics"},
+		            },
```
tosi3k, line 377: 1) Nit: please fix the indentation.
2) Let's just leave the access to `/metrics`, the other rules are not needed.

tosi3k (changes requested):
Please also squash all the commits into a single one.

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Cleared up

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Cleared up

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubeProxy.yaml
emerbe: Done

clusterloader2/pkg/measurement/common/scheduler_latency.go
emerbe: Done

clusterloader2/pkg/measurement/common/metrics/metrics_grabber.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,7 +362,70 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
+		return err
+	}
+	tokenResponse, err := retryCreateFunctionWithResponse(token)
+	if err != nil {
+		return err
+	}
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "monitoringserviceaccounttoken",
+			Namespace: "monitoring",
```
tosi3k, line 414: Nit: let's use the constant defined at the beginning of the file instead of the `"monitoring"` literal.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,7 +362,70 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
+		},
+		Data: map[string][]byte{
+			"token": []byte(tokenResponse),
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+	_, err = k8sClient.CoreV1().Secrets("monitoring").Create(context.TODO(), secret, metav1.CreateOptions{})
```
tosi3k, line 421: Nit: let's use the constant defined at the beginning of the file instead of the `"monitoring"` literal.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -378,10 +448,11 @@ func (pc *Controller) exposeAPIServerMetrics() error {
 	}
 	createClusterRoleBinding := func() error {
 		_, err := clientSet.GetClient().RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
-			ObjectMeta: metav1.ObjectMeta{Name: "system:anonymous"},
+			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer-binding"},
 			RoleRef:    rbacv1.RoleRef{Kind: "ClusterRole", Name: "apiserver-metrics-viewer"},
 			Subjects: []rbacv1.Subject{
-				{Kind: "User", Name: "system:anonymous"},
+				{Kind: "ServiceAccount", Name: "prometheus-k8s", Namespace: "monitoring"},
```
tosi3k, line 454: Nit: let's use the constant defined at the beginning of the file instead of the `"monitoring"` literal.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -378,10 +448,11 @@ func (pc *Controller) exposeAPIServerMetrics() error {
...
+				{Kind: "ServiceAccount", Name: monitoringServiceAccount, Namespace: "default"},
```
tosi3k, line 455: Nit: let's use the constant defined at the beginning of the file instead of the `"default"` literal.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,7 +362,70 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
 		os.DirFS(path), manifestGlob, pc.templateMapping, client.Retry(apierrs.IsNotFound))
 }
 
-// exposeAPIServerMetrics configures anonymous access to the apiserver metrics.
+// create ServiceAccount and token used by Prometheus to fetch data from testing cluster
```
tosi3k, line 365: Since this method's logic is pretty complicated, I suggest documenting it more extensively above its definition.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,7 +362,70 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
+func (pc *Controller) createServiceAccountInTestingCluster(k8sClient kubernetes.Interface) error {
+	klog.V(2).Info("Creating ServiceAccount in testing cluster")
+	// We need to get a client to the cluster where the test is being executed on,
+	// not the cluster that the prometheus is running in. Usually, there is only
+	// once cluster, but in case of kubemark we have two and thus we need to
+	// create a new client here.
+	clientSet, err := framework.NewMultiClientSet(
```
tosi3k, line 372: We already create the test cluster's clientset in the `exposeAPIServerMetrics` method, so we're duplicating this logic here.

I propose creating that clientset at the beginning of the `SetUpPrometheusStack` method and reusing it in other places in this file, including this one.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,7 +362,70 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
+	if err := retryCreateFunction(serviceAccount); err != nil {
+		return err
+	}
+	tokenResponse, err := retryCreateFunctionWithResponse(token)
+	if err != nil {
+		return err
+	}
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "monitoringserviceaccounttoken",
```
tosi3k, line 413: Nit: WDYT about simply `"prometheus-token"` instead?

That said, the naming change would also need to be included in `master-serviceMonitor.yaml`.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,7 +362,70 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
 		os.DirFS(path), manifestGlob, pc.templateMapping, client.Retry(apierrs.IsNotFound))
 }
 
-// exposeAPIServerMetrics configures anonymous access to the apiserver metrics.
+// create ServiceAccount and token used by Prometheus to fetch data from testing cluster
+func (pc *Controller) createServiceAccountInTestingCluster(k8sClient kubernetes.Interface) error {
```
tosi3k, line 366: Nit: WDYT about `s/createServiceAccountInTestingCluster/createToken/g`?

tosi3k (changes requested):

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -224,10 +227,25 @@ func NewController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *Control
 func (pc *Controller) SetUpPrometheusStack() error {
 	k8sClient := pc.framework.GetClientSets().GetClient()
 
+	// We need to get a client to the cluster where the test is being executed on,
+	// not the cluster that the prometheus is running in. Usually, there is only
+	// once cluster, but in case of kubemark we have two and thus we need to
+	// create a new client here.
```
tosi3k, line 233: That comment deserves a slight update - its shape made sense in context of the piece of code from which it was copied but some of its parts do not make sense here.

Please refresh it a bit, keeping in mind we need both clientsets in this method for kubemark-based clusters.

emerbe: TBH it still sounds reasonable to me.
Saying that, I've tried to polish it a bit, PTAL.

emerbe: Taken offline.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -224,10 +227,25 @@ func NewController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *Control
...
+	clientSet, err := framework.NewMultiClientSet(
```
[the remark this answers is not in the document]
emerbe: Taken offline.

tosi3k:
FYI @marseel @giorio94 - since you were interested in this.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -377,11 +452,12 @@ func (pc *Controller) exposeAPIServerMetrics() error {
 		return err
 	}
 	createClusterRoleBinding := func() error {
-		_, err := clientSet.GetClient().RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
-			ObjectMeta: metav1.ObjectMeta{Name: "system:anonymous"},
+		_, err := testClusterClientSet.RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer-binding"},
 			RoleRef:    rbacv1.RoleRef{Kind: "ClusterRole", Name: "apiserver-metrics-viewer"},
```
giorio94, line 457: The `apiserver-metrics-viewer` cluster role does not grant permission to access the kubelet metrics, causing the scraping to fail (while it was succeeding when using the prometheus service account). The difference seems to be due to the lack of permissions on `nodes/metrics`, previously granted by the prometheus-k8s cluster role:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: prometheus-k8s
rules:
- apiGroups:
  - ""
  resources:
  - nodes/metrics
  - services
  - endpoints
  - pods
  verbs:
  - get
  - list
  - watch
- nonResourceURLs:
  - /metrics
  verbs:
  - get
```

tosi3k: Yup, we noticed that yesterday as well and fixed properly today in https://github.com/kubernetes/perf-tests/pull/3240 :).

Sorry for the noise.

giorio94: Oh, that was super quick, I didn't expect the updated version to have already been merged. Thanks!

PR #3213, opened 2025-02-26
Implemented LIST benchmark for different content types

ayushpatil2122:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Implement LIST benchmark for different content types (Json, Proto, CBOR, YAML)
#### Which issue(s) this PR fixes:
#https://github.com/kubernetes/kubernetes/issues/130169

Fixes #https://github.com/kubernetes/kubernetes/issues/130169

PR #3215, opened 2025-02-27
Implemented LIST benchmark for different content types

serathius:
/kind feature

Took https://github.com/kubernetes/perf-tests/pull/3213 and fixed gofmt.

### DIFF
--- a/util-images/request-benchmark/main.go
+++ b/util-images/request-benchmark/main.go
@@ -38,6 +38,7 @@ const (
 	HealthCheckRequests = 10
 	NamespaceTmpl       = "%namespace%"
 )
+
 type ContentType string
 
 const (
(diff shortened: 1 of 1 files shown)


util-images/request-benchmark/main.go
```diff
@@ -39,18 +39,58 @@ const (
...
+	flag.Var(&contentType, "content-type", "Content type for requests (required). Valid values: [json, proto, cbor, yaml]")
 	flag.Parse()
 }
 
+func getContentTypeHeader(ct ContentType) (string, error) {
+	switch ct {
+	case JSONContentType:
+		return "application/json", nil
+	case ProtoContentType:
+		return "application/vnd.kubernetes", nil
```
tosi3k, line 84: Shouldn't this be `application/vnd.kubernetes.protobuf`?

Also, you might want to use constants from [here](https://github.com/kubernetes/kubernetes/blob/a18b4a8d97151a25c4d8088839d77e6ba482cac3/staging/src/k8s.io/apimachinery/pkg/runtime/types.go#L46-L49) instead to prevent typos like that in the future.

tosi3k (changes requested):

util-images/request-benchmark/main.go
serathius: Done.

serathius:
PTAL @tosi3k

PR #3240, opened 2025-03-10
Remove system anonymous from apis metrics v2

emerbe:
…p apiserver metrics

#### What type of PR is this?

/kind bug

Binding any permissions to system:anonymous and system:unauthenticated is dangerous.
This PR removes usage of system:anonymous in CLusterROleBinding used to talk to apiserver. 

What is more, service account needs to be used when talking to apiserver.

Fixes #

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"golang.org/x/sync/errgroup"
+	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
@@ -59,6 +60,7 @@ const (
 	nodeExporterPod              = "exporters/node_exporter/node-exporter.yaml"
 	windowsNodeExporterManifests = "exporters/windows_node_exporter/*.yaml"
 	pushgatewayManifests         = "pushgateway/*.yaml"
+	monitoringServiceAccount     = "monitoringserviceaccount"
 )
 
 //go:embed manifests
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,33 +378,87 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
+	}
+	_, err = k8sClient.CoreV1().Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// exposeAPIServerMetrics configures access to the apiserver metrics.
+func (pc *Controller) exposeAPIServerMetrics(testClusterClientSet kubernetes.Interface) error {
```
tosi3k, line 443: Nit: let's rename this method (and update the comment above) since now it gives access to more than just kube-apiserver's metrics.

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,33 +378,87 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
 	}
 	createClusterRoleBinding := func() error {
-		_, err := clientSet.GetClient().RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
-			ObjectMeta: metav1.ObjectMeta{Name: "system:anonymous"},
+		_, err := testClusterClientSet.RbacV1().ClusterRoleBindings().Create(context.TODO(), &rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer-binding"},
 			RoleRef:    rbacv1.RoleRef{Kind: "ClusterRole", Name: "apiserver-metrics-viewer"},
 			Subjects: []rbacv1.Subject{
-				{Kind: "User", Name: "system:anonymous"},
+				{Kind: "ServiceAccount", Name: "prometheus-k8s", Namespace: namespace},
```
tosi3k, line 460: Actually, this can be dropped - this ServiceAccount gets bound to the `prometheus-k8s` ClusterRole here:

https://github.com/kubernetes/perf-tests/blob/1aeeed9edf92d427cb338bdedcfe160f498067db/clusterloader2/pkg/prometheus/manifests/prometheus-clusterRoleBinding.yaml#L2-L12

and that ClusterRole already contains those permissions:

https://github.com/kubernetes/perf-tests/blob/1aeeed9edf92d427cb338bdedcfe160f498067db/clusterloader2/pkg/prometheus/manifests/prometheus-clusterRole.yaml#L2-L20

tosi3k (changes requested):

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
emerbe: Done

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -355,33 +378,87 @@ func (pc *Controller) applyManifests(path, manifestGlob string) error {
...
+}
+
+// configureRBACForMetrics creates a ClusterRole and ClusterRoleBinding
+// to allow the monitoringServiceAccount (used by Prometheus) to scrape metrics.
+func (pc *Controller) configureRBACForMetrics(testClusterClientSet kubernetes.Interface) error {
+	klog.V(2).Info("Exposing kube-apiserver metrics in the cluster")
 	createClusterRole := func() error {
-		_, err := clientSet.GetClient().RbacV1().ClusterRoles().Create(context.TODO(), &rbacv1.ClusterRole{
+		_, err := testClusterClientSet.RbacV1().ClusterRoles().Create(context.TODO(), &rbacv1.ClusterRole{
 			ObjectMeta: metav1.ObjectMeta{Name: "apiserver-metrics-viewer"},
```
tosi3k, line 448: Nit: maybe `s/apiserver-metrics-viewer/metrics-viewer`?

Ditto below for the roleRef and the name of the ClusterRoleBinding.

emerbe: Done

PR #3291, opened 2025-04-12
Increase Reporter extensibility

tylerFowler:
#### What type of PR is this?

/kind feature

This change exposes the SubstepResult struct in order to make `test.Reporter` a fully extensible interface. Currently the only out-of-the-box Reporter implementation invokes its own `ReportTestStepFinish` method, which does expose this data, but seemingly only to itself (it probably shouldn't be in the main interface for this reason). It relies on a package-private method from `ReportTestStep` to fill out this data, and the default Executor only invokes that top level method – this prevents users from implementing custom reporters with any level of test-step visibility.

The only way to access this newly public substep data is via the `StepResult.GetAllResults` method which is already mutex protected so simply exposing this seems safe. To that end only getter methods are added to ensure data mutation is limited.

Fixes #

### DIFF
--- a/clusterloader2/pkg/test/simple_reporter.go
+++ b/clusterloader2/pkg/test/simple_reporter.go
@@ -71,7 +71,7 @@ func (str *simpleReporter) ReportTestStepFinish(duration time.Duration, stepName
 }
 
 func (str *simpleReporter) ReportTestStep(result *StepResult) {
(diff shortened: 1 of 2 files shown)


tylerFowler:
Extra context: I'm creating CL2 tests fully programmatically, I'm skipping the `cmd` package entirely and loading each step manually, offering myself the ability to implement a custom provider & reporter.

PR #3305, opened 2025-04-25
Fix Linux Node Exporter

MikeZappa87:

Fixes #


clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/node-exporter-daemonset.yaml
```diff
@@ -0,0 +1,94 @@
...
+    metadata:
+      labels:
+        app: node-exporter
+    spec:
+      serviceAccountName: node-exporter
+      hostNetwork: true
+      hostPID: true
+      containers:
+        - name: node-exporter
+          image: quay.io/prometheus/node-exporter:latest
```
MikeZappa87, line 59: Find a tag, not latest

clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/node-exporter-daemonset.yaml
```diff
@@ -0,0 +1,94 @@
...
+          args:
+            - "--path.procfs=/host/proc"
+            - "--path.sysfs=/host/sys"
```
MikeZappa87, line 62: Figure out what people want to collect

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -395,46 +394,6 @@ func (pc *Controller) exposeAPIServerMetrics() error {
...
-	nodes, err := client.ListNodes(kubemarkFramework.GetClientSets().GetClient())
-	if err != nil {
-		return err
-	}
-
-	var g errgroup.Group
-	numMasters := 0
-	for _, node := range nodes {
-		node := node
-		if util.LegacyIsMasterNode(&node) || util.IsControlPlaneNode(&node) {
```
MikeZappa87, line 418: Do cloud environments even give access to the control plane nodes?

MikeZappa87: I can understand the motivation for this to be a static pod however should we maintain this requirement? It's not hard to keep the daemonset and the static pods but they should be enabled with separate flags.

MikeZappa87:
@wojtek-t do you have any thoughts here?

PR #3369, opened 2025-05-28
Fix: Correct netperf ReplicaSet flags in launch.go to allow pod startup

### LINKED ISSUE #2452: netperf documentation
Can we add a "Get started" section to the README on the netperf tool to describe how to set up the components?

**Why is this needed**:
I get the following error in the orchestrator's logs:

> flag provided but not defined: -testFrom

and the pod doesn't start

merdincz:
#### What type of PR is this?
/kind bug

The `launch.go` file currently attempts to create a **ReplicaSet** for the **netperf** application using the `--testFrom` and `--testTo `flags. These flags are not recognized by the netperf image, causing an error that prevents the Pod from starting successfully.

This PR removes the unsupported `--testFrom` and `--testTo` flags from the ReplicaSet definition. With this change, the netperf Pods can start correctly and achieve a healthy state.

Fixes #2452


vkhomchakim:
It is not code issue. It is caused by old, not rebuild, image which used by default sirot/netperf-latest

merdincz:
I agree that the code and image should be compatible. In this case, I’ve updated the code to match the existing image. If you’d rather adjust the image instead, feel free to open another PR for that.

PR #3377, opened 2025-06-05
[Draft] Add node scaling measurement, nodescaling test

hakuna-matatah:
#### What type of PR is this?

/kind feature

<todo-  edit> 

1. CL2 able to support tests around node scale up for different providers if they choose to, this will measure node scale throughput to begin with and we can add in future measurements such node_join latency, node_readiness latencies etc

2. This PR also includes a node test cl2 config that can be used to scale the nodes through cl2 framework.

Fixes #


clusterloader2/pkg/provider/scalers/types.go
```diff
@@ -0,0 +1,26 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scalers
+
+// NodeScaler defines interface for scaling nodes in a cluster
+type NodeScaler interface {
+	// ScaleNodes scales the cluster to the target number of nodes at the specified rate
+	// rate: number of nodes to add per interval
```
mengqiy: Maybe calling it `batchSize` is better.

How about deleting nodes?

clusterloader2/pkg/provider/scalers/types.go
```diff
@@ -0,0 +1,26 @@
...
+	// intervalMins: minutes to wait between scaling operations
```
mengqiy: Maybe using intervalSecond is better for all vendors.

clusterloader2/pkg/provider/scalers/types.go
```diff
@@ -0,0 +1,26 @@
...
+	// targetNodes: desired final number of nodes
+	ScaleNodes(rate, intervalMins, targetNodes int) error
```
mengqiy: I'd suggesting passing a context which allow users to cannel the operation if needed (e.g. timeout)

clusterloader2/pkg/provider/scalers/aws/node_scaler.go
```diff
@@ -0,0 +1,230 @@
...
+
+		// Calculate total current nodes
+		var totalCurrentNodes int32
+		for _, asg := range asgs {
+			totalCurrentNodes += *asg.DesiredCapacity
+		}
+
+		klog.Infof("Current total nodes: %d, target nodes: %d", totalCurrentNodes, targetNodes)
+
+		if totalCurrentNodes >= int32(targetNodes) {
```
mengqiy: Do we want to support scale in?

clusterloader2/pkg/provider/scalers/aws/node_scaler.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		newSize := smallestSize + nodesToAdd
+		klog.Infof("Scaling ASG %s from %d to %d nodes (adding %d nodes)",
+			*smallestASG.AutoScalingGroupName, smallestSize, newSize, nodesToAdd)
+
+		input := &autoscaling.UpdateAutoScalingGroupInput{
+			AutoScalingGroupName: smallestASG.AutoScalingGroupName,
+			DesiredCapacity:      awssdk.Int32(newSize),
+			MinSize:              awssdk.Int32(newSize),
```
mengqiy: Why do we need to update MinSize?

clusterloader2/pkg/provider/scalers/aws/node_scaler.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// waitForASGCapacity waits until the ASG reaches the desired capacity or times out
+func (s *nodeScaler) waitForASGCapacity(ctx context.Context, asgName string, desiredCapacity int32, timeout time.Duration) error {
+	klog.Infof("Waiting for ASG %s to reach capacity %d within interval of %v", asgName, desiredCapacity, timeout)
+
+	startTime := time.Now()
+	deadline := startTime.Add(timeout)
+
+	for {
```
mengqiy, line 189: A better way to implement this might be using  `select` with `ticker` and timeout can be controlled with context.
```
    ticker := time.NewTicker(interval)
    defer ticker.Stop()

    for {
        select {
        case <-ctx.Done():
            // Context canceled (or timed out); clean up and return.
            // fmt.Println("context canceled, exiting loop")
            return
        case <-ticker.C:
            // do something
        }
    }
```

clusterloader2/pkg/provider/scalers/aws/node_scaler.go
```diff
@@ -0,0 +1,230 @@
...
+	client := autoscaling.NewFromConfig(cfg)
+
+	return &nodeScaler{
+		region:      region,
+		clusterName: clusterName,
+		asgClient:   client,
+	}, nil
+}
+
+func (s *nodeScaler) ScaleNodes(rate, intervalMins, targetNodes int) error {
```
mengqiy: We should be able to abort the operation with context.

clusterloader2/pkg/provider/scalers/types.go
hakuna-matatah: I was initially only planning to support scale-out.

PR #3378, opened 2025-06-06
Fix for the coredns configmap update: the corefile was not updated, e…

Tanzp01:
…ven though a different cache size is set up in the test parameters.

When test GKE DNS performance with the auto creation of dns-perf-server, the coredns configmap corefile was not updated even if different cache size was set as parameters. This change should fix the coredns configmap update, and enable the test can also be run on python3.

Fixes #

### DIFF
--- a/dns/py/params.py
+++ b/dns/py/params.py
@@ -160,7 +160,11 @@ class CorednsCache(Param):
   def set(self, inputs, value):
     if value > 0:
       cf = inputs.configmap_yaml['data']['Corefile']
-      cfList = cf.decode().split("\n")
+      # fix the cf type so that it works also for python3
+      if(type(cf) == str):
+        cfList = cf.split("\n")
+      else:
+        cfList = cf.decode().split("\n")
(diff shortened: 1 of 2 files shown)


bowei:
You need to do the EasyCLA.

Tanzp01:
Hi, I have updated the necessary CLA, could you take a look?

dns/py/params.py
```diff
@@ -160,7 +160,11 @@ def is_relevant(self, attributes):
   def set(self, inputs, value):
     if value > 0:
       cf = inputs.configmap_yaml['data']['Corefile']
-      cfList = cf.decode().split("\n")
+      # fix the cf type so that it works also for python3
```
bowei: should we just update the code for Python 3? Isn't Python 2 deprecated?

Tanzp01: Thanks Bowei for the feedbacks! I have removed the line for python2, now only python3 supported for cleaner code

PR #3389, opened 2025-06-11
feat: support cluster-scoped objs in wait_for_generic_k8s_object

jwtty:
#### What type of PR is this?
/kind feature

Current `wait_for_generic_k8s_object` measurement does not support cluster-scoped objects. This simple PR enables it by allowing `namespaceRange` to NOT be specified and treating such objs as cluster-scoped.

Fixes #


jwtty:
Hi, @mborsz, @wojtek-t, may I ask for a review on this simple PR?

PR #3390, opened 2025-06-12
add ability to declare Dependency in tests

alaypatel07:
#### What type of PR is this?
/kind feature
/kind api-change

This PR introduces a new Dependency system. CL2 users can now declare a dependency to the tests, as an example, install the DRA driver for running DRA tests.

Fixes #

This blocks the work in #3368


clusterloader2/api/validation.go
```diff
@@ -297,6 +301,20 @@ func (v *ConfigValidator) validatePoissonLoad(rl *PoissonLoad, fldPath *field.Pa
 	return allErrs
 }
 
+func (v *ConfigValidator) validateDependency(d *Dependency, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if d.Name == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), "name is required"))
+	}
+	if d.Method == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("method"), "method is required"))
```
wojtek-t, line 310: I would like to eventually have a better validation like ensuring that proper methods exist (same for measurements). But I don't want to block on it - if you would like to help with a follow-up it would be great.

clusterloader2/pkg/dependency/manager.go
```diff
@@ -0,0 +1,192 @@
...
+	})
+
+	return nil
+}
+
+func (dm *dependencyManager) teardownDependencyInstance(depInstance *dependencyInstance) error {
+	conf := &Config{
+		ClusterFramework:    dm.clusterFramework,
+		PrometheusFramework: dm.prometheusFramework,
+		Params:              make(map[string]interface{}), // Empty params for teardown
```
wojtek-t: I'm not sure this will be enough.
Once we get to say two dri-drivers that you need to configure differently (and result in different daemonsets or sth like that), then params effectively identify the right daemonset.
So I'm wondering if you won't need the original params here too.

clusterloader2/api/validation.go
alaypatel07: I thought this was easy but I am running into a bunch of cyclic dependencies when I try to import dependency package into validation.go in api package, and couple ways I tried to solve it did not resolve those cycles. I assume there will be some re-org needed to get this validation implemented.

lets unblock DRA stuff for now.

clusterloader2/pkg/dependency/manager.go
alaypatel07: fixed it.

wojtek-t:
/lgtm
/approve

Thanks a lot!

clusterloader2/api/validation.go
alaypatel07: @wojtek-t https://github.com/kubernetes/perf-tests/pull/3391 follow-up PR for method validation for both dependency and measurements

PR #3397, opened 2025-06-18
update thresholds for dra test Churn phase

alaypatel07:
#### What type of PR is this?

Based on the job run here: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-azure-dra-with-workload-scalability/1935231328828002304, with the 100 node cluster, the time taken for pods to start in churn phase seems to be longer.

This PR is needed to test run to completion

Fixes #

### DIFF
--- a/clusterloader2/testing/dra/config.yaml
+++ b/clusterloader2/testing/dra/config.yaml
@@ -62,8 +62,6 @@ steps:
     Method: SchedulingMetrics
     Params:
       action: start
-      token: {{ $token }}
-      endpoint: "localhost:10259"
   - Identifier: FastFillPodStartupLatency
     Method: PodStartupLatency
     Params:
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/dra/config.yaml
```diff
@@ -128,20 +122,18 @@ steps:
...
-        endpoint: "localhost:10259"
     - Identifier: ChurnPodStartupLatency
       Method: PodStartupLatency
       Params:
         action: start
         labelSelector: job-type = short-lived
-        threshold: 20s
+        perc50Threshold: 40s
+        perc90Threshold: 60s
+        perc99Threshold: 80s
```
wojtek-t, line 136: Seems pretty high, but let's see where we're even at.

PR #3398, opened 2025-06-18
use GenericPrometheusQuery to get scheduler numbers instead of Schedu…

alaypatel07:
…lingMetrics

#### What type of PR is this?

This PR removes using SchedulingMetrics Measurement for DRA test job.

It instead uses GenericPrometheusQuery to get the same numbers as SchedulingMetrics

Fixes #


alaypatel07:
Alternative implementation here: #3399

wojtek-t:
Let's proceed with the other approach in https://github.com/kubernetes/perf-tests/pull/3399

PR #3399, opened 2025-06-19
implement a PrometheusSchedulingMetrics Measurement and Gatherer

alaypatel07:
#### What type of PR is this?
This PR implements a PrometheusSchedulingMetrics measurement that collects the
scheduling numbers using prometheus rather then querying scheduler pod via apiserver

This is needed for getting numbers from the DRA job

Part of fixes for: #https://github.com/kubernetes/kubernetes/issues/131198

### DIFF
--- a/clusterloader2/pkg/measurement/common/prometheus_scheduler_latency.go
+++ b/clusterloader2/pkg/measurement/common/prometheus_scheduler_latency.go
@@ -1,0 +1,152 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/prometheus_scheduler_latency.go
```diff
@@ -0,0 +1,136 @@
+package common
```
wojtek-t, line 17: boilerplate missing

clusterloader2/pkg/measurement/common/prometheus_scheduler_latency.go
```diff
@@ -0,0 +1,136 @@
...
+func init() {
+	create := func() measurement.Measurement {
+		return CreatePrometheusMeasurement(&prometheusSchedulerLatencyGatherer{})
+	}
+	if err := measurement.Register(prometheusSchedulingMetricsMeasurementName, create); err != nil {
+		klog.Fatalf("Cannot register %s: %v", prometheusSchedulingMetricsMeasurementName, err)
+	}
+}
+
+// prometheusSchedulerLatencyGatherer implements the Gatherer interface and generates the same
```
wojtek-t, line 45: We're relying on prometheus stack even for our basic SLO metrics (api call latencies) so that's fine.

Once we prove it works, I would be willing to deprecate and remove the original implementation of scheduler latency measurement.

clusterloader2/pkg/measurement/common/prometheus_scheduler_latency.go
alaypatel07: added

clusterloader2/pkg/measurement/common/prometheus_scheduler_latency.go
alaypatel07: +1, I like this approach better. Local tests work, lets see how the numbers look on 100 node test.

Corresponding test-infra PR: https://github.com/kubernetes/test-infra/pull/35012

PR #3429, opened 2025-06-25
for 100 node dra test bump fast fill pod startup latency

alaypatel07:
#### What type of PR is this?
/kind cleanup

Increasing the timeout to large enough value so we can get what the upper limit is.

```

I0625 21:27:21.786579   42173 dra.go:99] DRATestDriver: DRA example driver uninstalled successfully
I0625 21:27:21.786618   42173 simple_test_executor.go:418] All dependencies torn down successfully, cleanup time: 20.521201818s
E0625 21:27:21.786632   42173 clusterloader.go:253] --------------------------------------------------------------------------------
E0625 21:27:21.786639   42173 clusterloader.go:254] Test Finished
E0625 21:27:21.786646   42173 clusterloader.go:255]   Test: testing/dra/config.yaml
E0625 21:27:21.786651   42173 clusterloader.go:256]   Status: Fail
E0625 21:27:21.786657   42173 clusterloader.go:258]   Errors: [measurement call PodStartupLatency - FastFillPodStartupLatency error: pod startup: too high latency 99th percentile: got 5.223934749s expected: 5s]
E0625 21:27:21.786662   42173 clusterloader.go:260] --------------------------------------------------------------------------------

```

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-azure-dra-with-workload-scalability/1937971768484433920

Fixes #

### DIFF
--- a/clusterloader2/testing/dra/config.yaml
+++ b/clusterloader2/testing/dra/config.yaml
@@ -63,6 +63,9 @@ steps:
       Params:
         action: start
         labelSelector: job-type = long-running
+        perc50Threshold: 8s
+        perc90Threshold: 16s
+        perc99Threshold: 32s
     - Identifier: FastFillSchedulingMetrics
       Method: PrometheusSchedulingMetrics
       Params:
@@ -109,6 +112,9 @@ steps:
       Method: PodStartupLatency
       Params:
         action: gather
+        perc50Threshold: 8s
+        perc90Threshold: 16s
+        perc99Threshold: 32s
 - name: reset metrics for steady state churn
   measurements:
     - Identifier: ChurnSchedulingMetrics


wojtek-t:
This is the thing we should really be careful about and is exactly why we needed this test.

Pod startup is one of the very few official SLOs that we're trying to protect:
https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos

The results of this test show that we're not really meeting the goal.
The might be good reasons (the communication path with dra driver is something we don't necessarily fully control, and work done by dra driver can be time-consuming etc.), but I don't think we should just bump it and let it go.

I would like this to be brough more widely to all DRA folks, be understood and if we decide that it works as intended appropriate update the SLO to reflect that.
[I will be out for next 2 months so may not be particularly active in those discussions if we want them to happen now-ish]

@johnbelamaric  @pohly  - please include other folks too

pohly:
We are not aiming to achieve the same SLO for pods using claims as for traditional, stateless pods.

The rationale is the same as for [stateful pods](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md#other-notes):
> Starting a stateful pod requires attaching and mounting volumes, that takes non-negligible amount of time and doesn't even fully depend on Kubernetes.

The same argument applies to pods using GPUs. Now that DRA is a thing, it might be worth updating that section.

> I would like this to be brough more widely to all DRA folks

I think this is understood already. We can discuss it again when @alaypatel07 presents his results in the next WG Device Management meeting.

alaypatel07:
> This is the thing we should really be careful about and is exactly why we needed this test.
>
> Pod startup is one of the very few official SLOs that we're trying to protect:
> https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos
> 
> The results of this test show that we're not really meeting the goal.
> The might be good reasons (the communication path with dra driver is something we don't necessarily fully control, and work done by dra driver can be time-consuming etc.), but I don't think we should just bump it and let it go.

I agree that we would not be meeting the SLOs, but there are some nuances:
1. The test has not stabilized, so we dont know what the upper limit is
1. The test is lacking the mechanisms for debugging the longer times. I see the following missing:
   1. logs from plugin
   1. mesasurements about allocation time (this was discussed in the last wg-device-management) meeting, I will prepare a PR for it.
   
@wojtek-t While we work on those things, this bump in the limits is temporary to find out what numbers are we actually achieving with DRA.

alaypatel07:
We have triaged the slowness in pod creation -> running, we don't need this anymore.

/close

PR #3432, opened 2025-07-04
Support for pod, CR resource type for benchmark Listing

GunaKKIBM:
#### What type of PR is this?
Add one of the following kinds:
/kind feature

Adding support for benchmarking list request for pods and CR

https://github.com/kubernetes/kubernetes/issues/130169

Couple of things that aren't covered in this PR
- Support for different list types - paginated, exactRV as mentioned in this comment https://github.com/kubernetes/kubernetes/issues/130169#issuecomment-3031286753
- Support for listing CR (ClusterRoles), will add it based on the review of the existing changes for pods.

### DIFF
--- a/clusterloader2/testing/list/config.yaml
+++ b/clusterloader2/testing/list/config.yaml
@@ -2,7 +2,7 @@
 {{$configMapBytes := DefaultParam .CL2_LIST_CONFIG_MAP_BYTES 100000}}
 {{$configMapNumber := DefaultParam .CL2_LIST_CONFIG_MAP_NUMBER 10000}}
 {{$configMapGroup := DefaultParam .CL2_LIST_CONFIG_MAP_GROUP "list-configmap"}}
-{($podsNumber := DefaultParam .CL2_LIST_PODS_NUMBER 100000)}
+{{$podsNumber := DefaultParam .CL2_LIST_PODS_NUMBER 100000}}
 {{$podGroup := DefaultParam .CL2_LIST_POD_GROUP "list-pods"}}
 {{$podBytes := DefaultParam .CL2_LIST_POD_BYTES 10000}}
 


GunaKKIBM:
Respective PR in test-infra: https://github.com/kubernetes/test-infra/pull/35086

clusterloader2/testing/list/pod.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: {{.Name}}
+  labels:
+    app: {{$group}}
+spec:
+  containers:
+  - name: nginx-container
+    image: nginx
+    ports:
+    - containerPort: 80
```
kishen-v:
```suggestion
    - containerPort: 80

```

GunaKKIBM:
@mborsz @serathius @wojtek-t Please review this PR. If the changes look good for Pod resource type, I can do the same for ClusterRole as well. Thanks!

serathius:
This looks good, one thing to address is that we would like to have control over object size, so we can test cases where pod objects are larger than simple nginx pod. We could add it in another iteration.

clusterloader2/testing/list/pod.yaml
```diff
@@ -0,0 +1,14 @@
+{{$group := DefaultParam .group .Name}}
+
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{.Name}}
+  labels:
+    app: {{$group}}
+spec:
```
serathius, line 10: The goal is to measure the cost of Listing objects from apiserver. Pods are a different from configmaps as they cannot be passively created without side effects. Immediately after pod creation K8s controllers will start processing them for container to run, scheduler will assign node, Kubelet will pull it and start updating status etc. We would like to avoid that.

Could you propose some way to avoid those pods being processed? For example adding a node selector that prevents scheduling or find other ways to disable scheduler?

GunaKKIBM:
> This looks good, one thing to address is that we would like to have control over object size, so we can test cases where pod objects are larger than simple nginx pod. We could add it in another iteration.

Just wanted get some clarification here, here object size maps to requests for cpu and memory for the pod right? for which I can add arguments in the job to make in configurable.

serathius:
> object size maps to requests for cpu and memory for the pod right?

No, the size of manifest in bytes. See the configmap. We just load random bytes.

clusterloader2/testing/list/pod.yaml
GunaKKIBM: After checking how and where perfdash is deployed  which led me to this link -https://github.com/kubernetes/k8s.io/blob/main/running-in-community-clusters.md

Adding nodeaffinity/antiaffinity didn't seem like the right option, hence chose to put a schedulerName which is simpler.

GunaKKIBM:
> > object size maps to requests for cpu and memory for the pod right?
> 
> No, the size of manifest in bytes. See the configmap. We just load random bytes.

The simplest way I could think of increasing the pod manifest size was to set an env variable, whose value is random data whose size is configurable. I have updated the PR accordingly, please take a look.

clusterloader2/testing/list/pod.yaml
serathius: Sounds good

serathius:
Will take a look tomorrow.

serathius:
Running the test locally fails for me with:
```
F0820 12:10:01.400825  366852 clusterloader.go:388] Test compilation failed: [config reading error: parsing error: template: :96: undefined variable "$podsNumber"]
exit status 1
```

Please follow https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/testing/list to test it locally yourself.

GunaKKIBM:
> podsNumber

It was an issue with brackets where podsNumber is defined, fixed now. Thank you!

serathius:
> It was an issue with brackets where podsNumber is defined, fixed now. Thank you!

Great find! I totally missed that

serathius:
ok-to-test

clusterloader2/testing/list/config.yaml
```diff
@@ -83,3 +86,35 @@ steps:
     params:
       namePrefix: "list-configmaps-"
       replicas: 0
+
+- name: Create pods
```
serathius, line 90: Think you should delete configmaps and wait couple of minutes so that they are not counted in memory used when collecting metrics for pods.

clusterloader2/testing/list/config.yaml
```diff
@@ -83,3 +86,35 @@ steps:
...
+    params:
+      namePrefix: "list-pods-"
+      replicas: {{$listReplicas}}
+      uri: /api/v1/pods?resourceVersion=0
+      namespaced: false
+      contentType: {{$contentType}}
+- module: *startMeasurements
+- name: Wait 5 minutes
+  measurements: *waitMeasurements
+- module: *gatherMeasurements
```
serathius, line 115: Doing measurements this way will override file with configmap measurements. We need to put them in separate files.

cc @mborsz for ideas.

clusterloader2/testing/list/config.yaml
GunaKKIBM: I might need a little help with some reference or example within in this repo. Does the action - "delete" help? I tried looking into any existing examples, but not quite sure.

GunaKKIBM: @mborsz , could you please help me here?

GunaKKIBM: @mborsz @wojtek-t I am still looking for some help here!

PR #3438, opened 2025-07-10
add latency breakdown for DRA with workload tests

alaypatel07:
#### What type of PR is this?
/kind feature

The existing DRA tests are having very high scheduling latencies. See this for example:
```
    {
      "data": {
        "Perc50": 30227.696,
        "Perc90": 54776.237,
        "Perc99": 60934.231
      },
      "unit": "ms",
      "labels": {
        "Metric": "create_to_schedule"
      }
    },
```
https://storage.googleapis.com/kubernetes-ci-logs/logs/ci-kubernetes-e2e-azure-dra-with-workload-scalability/1942321392846376960/artifacts/PodStartupLatency_ChurnPodStartupLatency_dra-steady-state_2025-07-07T21:26:59Z.json

This PR adds additional measurements such that resource allocation claim latency is available when test is run.

This PR also adds latency measurements for kubelet operations.

Fixes #


alaypatel07:
cc @pohly @shyamjvs

clusterloader2/testing/dra/config.yaml
```diff
@@ -63,10 +63,30 @@ steps:
...
+        metricName: fastfill_kubelet_latencies
+        metricVersion: v1
+        unit: s
+        queries:
+          - name: p99_dra_prepare_resources
+            query: histogram_quantile(0.99, sum(rate(dra_operations_duration_seconds_bucket{operation_name="PrepareResources"}[%v])) by (le))
+          - name: p99_dra_unprepare_operations
+            query: histogram_quantile(0.99, sum(rate(dra_operations_duration_seconds_bucket{operation_name="UnprepareResources"}[%v])) by (le))
+          - name: p99_dra_grpc_node_prepare_resources
+            query: histogram_quantile(0.99, sum(rate(dra_grpc_operations_duration_seconds_bucket{method_name="NodePrepareResources"}[%v])) by (le))
```
pohly: It's not clear from [the documentation](https://github.com/kubernetes/kubernetes/blob/2bdcad7419c7a584da4174513e36f867e88424df/pkg/kubelet/metrics/metrics.go#L1010-L1032) (cc @bart0sh), but "method_name", in contrast to "operation_name", includes the full gRPC method name. Currently that is `/k8s.io.kubelet.pkg.apis.dra.v1beta1.DRAPlugin/NodePrepareResources` and `/k8s.io.kubelet.pkg.apis.dra.v1beta1.DRAPlugin/NodeUnprepareResources`.

@bart0sh: is that correct? Do we have any tests for these? Can you update the docs?

pohly: Below you are using `=~".` but not here.

clusterloader2/pkg/measurement/common/slos/resourceclaim_allocation_latency.go
```diff
@@ -0,0 +1,254 @@
...
+
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/perf-tests/clusterloader2/pkg/errors"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement"
+	measurementutil "k8s.io/perf-tests/clusterloader2/pkg/measurement/util"
+	"k8s.io/perf-tests/clusterloader2/pkg/measurement/util/informer"
+	"k8s.io/perf-tests/clusterloader2/pkg/util"
+)
+
+const resourceClaimAllocationLatencyName = "ResourceClaimAllocationLatency"
```
pohly: Some comment what this measures would be good.

clusterloader2/testing/dra/config.yaml
alaypatel07: updated to use `=~".` at both places.

> but "method_name", in contrast to "operation_name", includes the full gRPC method name. Currently that is /k8s.io.kubelet.pkg.apis.dra.v1beta1.DRAPlugin/NodePrepareResources and /k8s.io.kubelet.pkg.apis.dra.v1beta1.DRAPlugin/NodeUnprepareResources.

This is what I have found, +1 to doc update.

clusterloader2/pkg/measurement/common/slos/resourceclaim_allocation_latency.go
alaypatel07: updated

clusterloader2/pkg/measurement/common/slos/resourceclaim_allocation_latency.go
```diff
@@ -0,0 +1,257 @@
...
+	return informer.StartAndSync(inf, m.stopCh, informerSyncTimeout)
+}
+
+func (m *resourceClaimAllocationLatencyMeasurement) addEvent(_, obj interface{}) {
+	event := &claimEventData{obj: obj, recvTime: time.Now()}
+	m.queue.Add(event)
+}
+
+func (m *resourceClaimAllocationLatencyMeasurement) processEvents() {
+	for m.processNextWorkItem() {
```
alaypatel07, line 202: @pohly I really think the latency numbers here will be influenced by how fast the client is here. We need the following two things:
1. Need to make this concurrent, atleast at the same level as the default resource claim controller.
2. Need to have some kind of timestamp in API object either resourceClaim.status or an event where timestamp from the server is read rather than assumed from the event processing in the client.

pohly: "Make this concurrent" - what is "this" in that item?

"timestamp in API object" - what is the semantic of it?

You want to measure "how long does it take for a claim to be allocated, since the time that the claim was created", right? The Binding Conditions KEP is adding an "allocation timestamp" in the result = the time at which the scheduler stored the allocation result. Is that sufficient?

Note that this latency doesn't really answer *why* it takes long, it just shows that it does. It may be perfectly normal, for example because there are no resources. It's only useful in a controlled scenario, like the benchmark, where it is known that the scheduler should be able to schedule immediately.

alaypatel07: > "Make this concurrent" - what is "this" in that item?

By this I mean the for loop at line 160. It could be something like:
```
	for i := 0; i < 10; i++ {
		go wait.UntilWithContext(ctx, m.runWorker, time.Second)
	}
	...
	
func (m * resourceClaimAllocationLatencyMeasurement) runWorker(ctx context.Context) {
	for m.processNextWorkItem(ctx) {
	}
}
```
https://github.com/kubernetes/kubernetes/blob/9822e51403e46cc0aeae49cf64eae653eff29f9e/pkg/controller/resourceclaim/controller.go#L427-L429

> You want to measure "how long does it take for a claim to be allocated, since the time that the claim was created", right? The Binding Conditions KEP is adding an "allocation timestamp" in the result = the time at which the scheduler stored the allocation result. Is that sufficient?

yes this should be sufficient. 

> Note that this latency doesn't really answer why it takes long, it just shows that it does. It may be perfectly normal, for example because there are no resources. It's only useful in a controlled scenario, like the benchmark, where it is known that the scheduler should be able to schedule immediately.

Yes I understand it doesnt show "why", but because the scheduler metrics currently suffer from the granularity bug, this is the best we could do to further break down if the latency in scheduler is due to dra plugin. Please correct me here if this is not true.

pohly: I understand that it would be useful in this case. I just wouldn't like to add a dedicated API field for a niche use case. The "time of allocation" seems useful enough to add it.

alaypatel07:
@pohly can you PTAL at this again, TIA

mborsz:
/approve

Please rebase

alaypatel07:
@mborsz 

```
# cat /tmp/clusterloader2-results/ResourceClaimAllocationLatency_ChurnClaimAllocationLatency_dra-steady-state_2025-07-29T17:19:22-04:00.json | head -n 10
{
  "version": "1.0",
  "dataItems": [
    {
      "data": {
        "Perc50": 652.252595,
        "Perc90": 810.603496,
        "Perc99": 930.503008
      },
      "unit": "ms",
 # cat /tmp/clusterloader2-results/'GenericPrometheusQuery churn_kubelet_latencies_dra-steady-state_2025-07-29T17:19:22-04:00.json' | head -n 20
{
  "version": "v1",
  "dataItems": [
    {
      "data": {
        "p99_dra_grpc_node_prepare_resources": 0.099,
        "p99_dra_grpc_node_unprepare_resources": 0.099,
        "p99_prepare_operations": 0.099,
        "p99_unprepare_operations": 0.09899999999999999
      },
      "unit": "s"
    }
  ]
}
```

I have:
1. removed some of the debugging commits to make this PR focus only on latency breakdown
2. rebased and
3. tested locally, it seems to work as expected

PTAL, thanks

PR #3439, opened 2025-07-10
add dra-baseline tests with workloads without resourceclaims

alaypatel07:
#### What type of PR is this?
/kind feature

This PR adds test such that 90% of available CPU in the cluster is first saturated and then 10% remaining CPU is used by pods that are churned for 10 times. 

This creates a baseline test to compare how cluster performs with and without DRA

Fixes #

### DIFF
--- a/clusterloader2/testing/dra-baseline/README.md
+++ b/clusterloader2/testing/dra-baseline/README.md
@@ -1,0 +1,64 @@
+(8000 m – 80 m) / 8 ≈ 990 m CPU
+```
+
+##### 2  Run the test
+
+```bash
+# Ensure a Prometheus stack is running so metric-based measurements succeed.
+
+./run-e2e.sh cluster-loader2 \
+  --provider=kind \
+  --kubeconfig=$HOME/.kube/config \
+  --report-dir=/tmp/clusterloader2-results \
+  --testconfig=testing/dra-baseline/config.yaml \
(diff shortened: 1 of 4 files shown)


clusterloader2/testing/dra-baseline/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+##### What the test does
+
+1. Calculates per-pod CPU from node capacity and `CL2_PODS_PER_NODE`.
+2. Fills each node to ~90 % CPU utilisation with long-running Jobs.
+3. Waits until all fill pods are running, then gathers startup & scheduler metrics.
+4. Resets metrics and runs short-lived Jobs (churn) that consume the remaining capacity.
+5. Gathers the same metrics for the churn phase.
+
+Collected measurements include PodStartupLatency and Prometheus-based scheduler metrics, allowing direct comparison to the DRA test (`testing/dra/config.yaml`).
```
pohly, line 26: This explains the purpose, but as it stands there's nothing specific to DRA in *this* scenario. For the sake of iterating quickly we can merge this as-is, but long term it might be better to rename this to "steady-state-90-percent" or something like this and and then over in the DRA variant explain that it is similar to this one.

pohly: /lgtm

My comment was not a blocker, just a random observation.

PR #3481, opened 2025-07-28
add resource quota for dra plugin needed for gke test

alaypatel07:
#### What type of PR is this?
/kind bug

The dra-example-driver sets the priority class of the plugin pods to be `system-node-critical`. Due to this the gce tests were failing as the priority class did not have enough quota.

In order to fix this issue, this PR:
1. adds a quota for the priority class equal to the number of nodes in the cluster
2. implements a pending TODO which allows the tests to determine the Namespace of dra-example-driver by providing a param for it

Fixes #

### DIFF
--- a/clusterloader2/pkg/dependency/dra/manifests/resourceQuota.yaml
+++ b/clusterloader2/pkg/dependency/dra/manifests/resourceQuota.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: dra-example-driver-kubeletplugin
+  namespace: {{.Namespace}}
+  labels:
+    helm.sh/chart: dra-example-driver-0.1.3
+    app.kubernetes.io/name: dra-example-driver
+    app.kubernetes.io/instance: dra-example-driver
+    app.kubernetes.io/version: "v0.1.0"
+    app.kubernetes.io/component: kubeletplugin
+spec:
+  hard:
+    pods: {{.WorkerNodeCount}}
+  scopeSelector:
+    matchExpressions:
+      - operator: In
+        scopeName: PriorityClass
+        values:
+          - system-node-critical
--- a/clusterloader2/pkg/dependency/dra/dra.go
+++ b/clusterloader2/pkg/dependency/dra/dra.go
@@ -33,10 +33,11 @@ import (
 const (
 	draDependencyName = "DRATestDriver"
 	//TODO: this needs to be converted into a parameter. Will will not need this until parititionable devices test
-	draNamespace          = "dra-example-driver"
-	draDaemonsetName      = "dra-example-driver-kubeletplugin"
-	checkDRAReadyInterval = 30 * time.Second
-	defaultDRATimeout     = 10 * time.Minute
+	draNamespace           = "dra-example-driver"
+	defaultWorkerNodeCount = "100"
+	draDaemonsetName       = "dra-example-driver-kubeletplugin"
+	checkDRAReadyInterval  = 30 * time.Second
+	defaultDRATimeout      = 10 * time.Minute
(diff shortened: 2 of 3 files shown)


alaypatel07:
```
# k get resourcequota -n dra-example-driver   dra-example-driver-kubeletplugin -oyaml
apiVersion: v1
kind: ResourceQuota
metadata:
  creationTimestamp: "2025-07-28T14:14:32Z"
  labels:
    app.kubernetes.io/component: kubeletplugin
    app.kubernetes.io/instance: dra-example-driver
    app.kubernetes.io/name: dra-example-driver
    app.kubernetes.io/version: v0.1.0
    helm.sh/chart: dra-example-driver-0.1.3
  name: dra-example-driver-kubeletplugin
  namespace: dra-example-driver
  resourceVersion: "2118"
  uid: 27c93b51-7b35-4053-a77c-2aec775bf49d
spec:
  hard:
    pods: "4"
  scopeSelector:
    matchExpressions:
    - operator: In
      scopeName: PriorityClass
      values:
      - system-node-critical
status:
  hard:
    pods: "4"
  used:
    pods: "4"
 ```
 
 tested this locally in a 4 node kind cluster and it works as expected.

clusterloader2/pkg/dependency/dra/manifests/resourceQuota.yaml
lmktfy, line 1: This doesn't look right.

- it's a _dynamically allocated_ resource
- how can the resource, or any Pod etc that supports it, be considered node critical?

A Pod at node critical priority means preempting that Pod would make the Node unhealthy.

If we are using `system-node-critical` as a placeholder for "as high as it can be" then:
- could we use a very high Pod priority instead, eg 1000000000?
- if not, let's add a comment to the ResourceQuota manifest / template to explain that
- maybe annotate the ResourceQuota with `kubernetes.io/description` as well

Something's not right here, surely?

lmktfy (commented):
This is related to example code. What we do in our examples, people will copy.

It's important that we set a _good_ example.

clusterloader2/pkg/dependency/dra/manifests/resourceQuota.yaml
lmktfy: Maybe we really need to change the example driver to use a different PriorityClass. Well, if so: we should.

alaypatel07: @lmktfy can you please create an issue on the https://github.com/kubernetes-sigs/dra-example-driver with the right direction on what needs to be done?

Meanwhile, would you consider this a blocker to merge the PR? This blocks the scalability test runs, it would be nice to solve this in dra-example-driver and come back here to fix it rather then fixing it here. WDYT?

johnbelamaric: I don't think this is relevant to this scale test. Also, if I were paying for 8 H100s and none were available on the node because the driver got evicted, I would consider it "unhealthy"...seems use-case dependent, with this being most common. The driver is needed for the node to provide its primary function.

nojnhuh: I opened https://github.com/kubernetes-sigs/dra-example-driver/issues/117 to track this for the example driver.

nojnhuh (approved):
/lgtm

We can follow up on the priority class of the example driver in https://github.com/kubernetes-sigs/dra-example-driver/issues/117.

nojnhuh:
@alaypatel07 It might be useful to backfill the PR description with some context in case there's more that isn't captured by the title.

alaypatel07:
@nojnhuh thanks for the review, updated the description for posterity

PR #3539, opened 2025-08-26
Bump version of go compiler that bootstraps golang tip test to 1.25.0

konryd:
This change updates the golang version preemptively, before golang1.25 will become un-buildable with 1.24

This is a rhythmical update. Prev instance: https://github.com/kubernetes/perf-tests/pull/3238

/assign @mborsz

### DIFF
--- a/golang/Makefile
+++ b/golang/Makefile
@@ -15,7 +15,7 @@
 .PHONY: build-go build push
 
 export GCS_BUCKET?=k8s-infra-scale-golang-builds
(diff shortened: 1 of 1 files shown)


tosi3k:
/ok-to-test

@konryd https://dl.google.com/go/go1.25.6.linux-amd64.tar.gz doesn't actually exist right now, does it? I think you'd need to reduce the patch number in the `1.25.6` SemVer version to achieve what you intend by this PR.

PR #3540, opened 2025-08-26
Add wait for ingress deletion in NEG test

panslava:
What type of PR is this?
/kind bug
/kind flake

What this PR does / why we need it
- Ensures NEG tests wait until all Ingress resources are actually deleted before proceeding, reducing end-of-test flakiness.
- Handles immediate deletions where `DeletionTimestamp` may be nil by recording deletion phases using a local timestamp fallback, so deletion latency is tracked consistently.

Changes
- Add explicit wait for Ingress deletion in the NEG test flow:
  - `clusterloader2/testing/neg/config.yaml:125`
    - Add module step invoking `ingress-measurements.yaml` with `action: waitForDeletion`.
- Record delete phases even when `DeletionTimestamp` is nil:
  - `clusterloader2/pkg/measurement/common/service_creation_latency.go:365`
    - Service deletion: set `deleting` from `DeletionTimestamp` when present; otherwise use `time.Now()`. Always set `deleted` to `time.Now()`.
  - `clusterloader2/pkg/measurement/common/service_creation_latency.go:383`
    - Ingress deletion: same behavior as Services.

Which issue(s) this PR fixes
Fixes #

- This aligns NEG cleanup with L4LB tests, which already wait for deletion, making behavior consistent.
- Mixed time sources: creation uses server time; deletion fallback uses `time.Now()`. Negative latencies are already clamped to zero in `phase_latency.go`, so this remains safe, but be aware of potential minor clock skew when interpreting results.

Release note
```release-note
ClusterLoader2: Ensure NEG tests wait for Ingress deletions before finalizing. ServiceCreationLatency now records deletion phases even when DeletionTimestamp is nil, improving reliability of deletion tracking and reducing test flakiness.
```

### DIFF
--- a/clusterloader2/pkg/measurement/common/service_creation_latency.go
+++ b/clusterloader2/pkg/measurement/common/service_creation_latency.go
@@ -337,7 +337,7 @@ func (s *serviceCreationLatencyMeasurement) handleIngressObject(oldObj, newObj i
 		return
 	}
 	newIngress, ok = newObj.(*networkingv1.Ingress)
-	if newIngress != nil && !ok {
+	if newObj != nil && !ok {
 		klog.Errorf("%s: uncastable new object: %v", s, newObj)
 		return
 	}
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/service_creation_latency.go
```diff
@@ -337,7 +337,7 @@ func (s *serviceCreationLatencyMeasurement) handleIngressObject(oldObj, newObj i
 		return
 	}
 	newIngress, ok = newObj.(*networkingv1.Ingress)
-	if newIngress != nil && !ok {
+	if newObj != nil && !ok {
```
panslava, line 340: This is just fixes mistaken code I noticed, we should compare newObj, not newIngress (look how we do the same above this for services)

PR #3592, opened 2025-09-09
Add parser for etcd API benchmarks

kishen-v:
#### What type of PR is this?
/kind feature

This PR adds a new option to parse the etcd-benchmark related data, which is done as a part of defining a performance suite for etcd. Ref: https://github.com/etcd-io/etcd/issues/16467.

image
This is one of the many etcd API/Operations that will be supported on the perf-dash.

None

### DIFF
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -695,10 +695,10 @@ var (
 	}
 
 	etcdAPIBenchmarkDescription = TestDescriptions{
-		"EtcdAPIBenchmark": {
-			"PUT ": []TestDescription{{
-				Name:             "put",
-				OutputFilePrefix: "etcd_perf",
+		"EtcdAPIPerformance": {
+			"Benchmark ": []TestDescription{{
+				Name:             "benchmark",
+				OutputFilePrefix: "EtcdAPI",
 				Parser:           parsePerfData,
 			}},
 		},
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -694,16 +694,27 @@ var (
 		},
 	}
 
+	etcdAPIBenchmarkDescription = TestDescriptions{
+		"EtcdAPIBenchmark": {
+			"PUT ": []TestDescription{{
+				Name:             "put",
+				OutputFilePrefix: "etcd_perf",
+				Parser:           parsePerfData,
+			}},
+		},
+	}
+
 	jobTypeToDescriptions = map[string]TestDescriptions{
-		"performance":   performanceDescriptions,
-		"benchmark":     benchmarkDescriptions,
-		"networking":    networkingDescriptions,
-		"dnsBenchmark":  dnsBenchmarkDescriptions,
-		"storage":       storageDescriptions,
-		"throughput":    throughputDescriptions,
-		"windows":       windowsDescriptions,
-		"watchlist":     watchListDescriptions,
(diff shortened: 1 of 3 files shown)


perfdash/config.go
```diff
@@ -694,16 +694,27 @@ var (
 		},
 	}
 
+	etcdAPIBenchmarkDescription = TestDescriptions{
+		"EtcdAPIBenchmark": {
+			"PUT ": []TestDescription{{
```
serathius: Perfdash requires manual building and push by its maintainers. I don't think it's sustainable to depend on them to make a change everytime we add a new benchmark.

 Please see the other descriptors. They don't specify which method they are testing. Only which metric.

kishen-v: Hey @serathius,

Thanks for your review. Currently, the generated perfdash report is hardcoded to set the metric as “APIResponsiveness”, which was done during the initial implementation - [Ref: Link](https://github.com/etcd-io/etcd/blob/3945526b8ca9735d5af8d7fd56b56c7d429c5258/pkg/report/perfdash.go#L64)
To give more control over viewing other benchmarks, I’m looking to add a label such as "Operation": "PUT/WATCH/etc" based on the benchmark operation performed while removing the existing label.
For eg:
```
{
 "version": "v1",
 "dataItems": [
  {
   "data": {
    "Perc50": 5.4992,
    "Perc90": 9.0704,
    "Perc99": 18.8455
   },
   "labels": {
    "Operation": "PUT"
   },
   "unit": "ms"
  }
 ]
}
```
This approach keeps the existing logic intact while allowing multiple perf-dash–formatted benchmarks to be stored in a single file and support parsing the same, as discussed in yesterday's call.

On the same topic, currently we generate a single perfdash file per benchmark operation. Should we plan to have more generic file name for the generated report such as  `EtcdAPI_benchmark_<date>.json` to accomodate other kinds of benchmark from the existing naming format ([etcd_perf_{bencmark_operation}_{date}.json)](https://github.com/etcd-io/etcd/blob/3945526b8ca9735d5af8d7fd56b56c7d429c5258/pkg/report/perfdash.go#L76)?  This would help us collate the changes before submitting a supporting PR in perfdash repository.

Thanks!

serathius: Yes `EtcdAPI_benchmark_<date>.json` makes more snse, I think we should plan for adding more operations without needing to change perfdash.

kishen-v:
/hold
For adding the labels first to the jobs on test-infra: https://github.com/kubernetes/test-infra/pull/35529

kishen-v:
/unhold
A local run against the fresh tags:
image

hwdef:
Hey, now that we have this PR, why isn't it showing up at https://perf-dash.k8s.io?

kishen-v:
Hey @hwdef,

While PR #3652 updated the deployment spec to pick up the Prow configurations, the existing deployment hasn't been updated is what I feel. We should be able to view the metrics on https://perf-dash.k8s.io/ once the new configurations are loaded. 

Thanks!

hwdef:
@kishen-v 
Thanks for your replay. Do you know how to update the deployment? #3652  has been merged for a month

serathius:
Last push made by me caused the dashboard to go down :(. Will need to ask @mborsz nicely for help.

PR #3630, opened 2025-09-29
add count for errors and error timestamps for dnspropagation measurement

Tanzp01:
Added count for errors and error timestamps for dnspropagation measurement, it will skip the "no such host" error during startup but only record the real errors with timestamp
Also added will record the all the latencies recorded with the timestamps

#### What type of PR is this?
/kind feature

To add more information of the DNS performance

Tested locally, able to see the latency logs with timestamp

### DIFF
--- a/util-images/probes/pkg/dnspropagation/probe.go
+++ b/util-images/probes/pkg/dnspropagation/probe.go
@@ -21,9 +21,13 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"log/slog"
 	"math"
 	"math/rand"
 	"net"
+
+	"os"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -153,12 +170,33 @@ func runSinglePod(client kubernetes.Interface, url string, podName string, names
...
 	tick := time.NewTicker(interval)
 	defer tick.Stop()
+	var lookupErrorLogged = false
 	for {
 		select {
 		case <-tick.C:
 			klog.V(4).Infof("DNS lookup %s", url)
 			if err := lookup(url); err != nil {
-				klog.Warningf("DNS lookup error: %v", err)
+				if strings.Contains(err.Error(), "no such host") {
```
marqc: nit: https://pkg.go.dev/net#DNSError DNSError struct contains isNotFound boolean flag that might be more reliable and convenient than string comparison

marqc (approved):

util-images/probes/pkg/dnspropagation/probe.go
Tanzp01: Thanks Marek for the suggestion! I have updated the code to use DNSError.IsNotFound instead.

PR #3631, opened 2025-09-30
Add logging for dns lookups with timestamps

Tanzp01:
Added 2 structured loggings for: 
1. dns lookup latency with timestamp
2. dns error with timestamp

#### What type of PR is this?
/kind feature

To better analyse the performance of DNS over time

There were existing loggings, but not structured. I added these 2 opt-in structured logging while keeping the previous loggings, so that any existing analysis based on the logging won't be affected

### DIFF
--- a/dns/dnsperfgo/main.go
+++ b/dns/dnsperfgo/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io"
 	"log"
 	"log/slog"
 	"net"
@@ -80,6 +81,8 @@ func main() {
 	flag.Parse()
 	if config.enableLatencyLogging {
 		infoLogger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
+	} else {
+		infoLogger = slog.New(slog.NewJSONHandler(io.Discard, nil))
 	}
 	errorLogger = slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelError}))
 	log.Printf("Starting dnstest with config parameters - %+v", config)
@@ -235,12 +238,11 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
 		startTime := time.Now()
 		_, err := lookupFunc(name)
 		latency := time.Since(startTime)
-		if infoLogger != nil && err == nil {
-			infoLogger.Info("DNS lookup successful, latency recorded",
-				"hostname", name,
-				"latency_seconds", latency.Seconds(),
-			)
-		}
+		infoLogger.Info("DNS lookup finished",
+			"name", name,
+			"latency_ms", latency.Milliseconds(),
+			"err", err,
+		)
 		dnsLatency.Observe(latency.Seconds())
 		resultChan <- err
 	}(resultChan)
@@ -254,8 +256,8 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
 		}
 		if err != nil {
 			errorLogger.Error("Failed DNS lookup",
-				"hostname", name,
-				"error", err.Error(),
+				"name", name,
+				"error", err,
 				"timed_out", timedOut,
 			)
 		}


Tanzp01:
cc @bowei  the owner of this directory

dns/dnsperfgo/main.go
```diff
@@ -237,7 +265,7 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
 		if err != nil {
 			log.Printf("Failed DNS lookup of name %q, err - %v\n", name, err)
```
marqc: This log message is duplicated by structured errorLogger. Maybe structured errorLogger should always be enabled and replace the existing log message? I don't know if this is common for any automation to traverse these logs and changing log message can lead to some failures.

Tanzp01: Thanks, I have enabled the errorLogger by default and remove the old log.

dns/dnsperfgo/main.go
```diff
@@ -223,6 +235,12 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
 		startTime := time.Now()
 		_, err := lookupFunc(name)
 		latency := time.Since(startTime)
+		if infoLogger != nil && err == nil {
```
bowei: why not route the infoLogger to a dummy logger that doesn't output instead of having `if` statements?

bowei: I also recommend logging for all lookups, irrespective of err == nil or not.

dns/dnsperfgo/main.go
```diff
@@ -223,6 +235,12 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
...
+			infoLogger.Info("DNS lookup successful, latency recorded",
```
bowei: suggest:

infoLogger("lookup", "name", name, "latency_ms", latency.Milliseconds(), "err", err)

Latency we want in milliseconds
Record the error.

dns/dnsperfgo/main.go
```diff
@@ -235,7 +253,11 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
 			log.Printf("DNS lookup of name %q, err - %v\n", name, err)
 		}
 		if err != nil {
-			log.Printf("Failed DNS lookup of name %q, err - %v\n", name, err)
+			errorLogger.Error("Failed DNS lookup",
+				"hostname", name,
+				"error", err.Error(),
```
bowei: Do you need to call Error? Isn't it extraneous?

dns/dnsperfgo/main.go
```diff
@@ -235,7 +253,11 @@ func (c *dnsClient) runQuery(name string, timeout time.Duration, lookupFunc Look
 			log.Printf("DNS lookup of name %q, err - %v\n", name, err)
 		}
 		if err != nil {
-			log.Printf("Failed DNS lookup of name %q, err - %v\n", name, err)
+			errorLogger.Error("Failed DNS lookup",
+				"hostname", name,
```
bowei: make it easier to maintain -- "name", name

bowei (commented):
Unrelated comment:

* There is what looks like abandoned code in comments at the bottom -- I would send another PR to delete it.
* The timedOut variable looks weird to me but don't fix it in this change.

dns/dnsperfgo/main.go
Tanzp01: Thanks. 
After my change, now the infoLogger will now record the successful lookup and error (e.g. no such host). To measure latency accurately, the time out will not reach the infoLogger. But the errorLogger will record error (e.g. no such host) and time out. I think it is enough to cover all the information we need.

dns/dnsperfgo/main.go
```diff
@@ -39,10 +41,10 @@ import (
...
 
 type Config struct {
-	qps                                      int
-	testDuration, idleDuration, queryTimeout time.Duration
-	hostnameFile                             string
-	queryClusterNames, logQueries            bool
+	qps                                                 int
+	testDuration, idleDuration, queryTimeout            time.Duration
+	hostnameFile                                        string
+	queryClusterNames, logQueries, enableLatencyLogging bool
```
bowei, line 47: [minor] prefer to have these on individual lines otherwise the gofmt will indent the type decl quite a bit to the right.

bowei (approved):
seems like there are a few more minor comments that are open.

bowei:
/approve
/lgtm

Seems like there are some minor comments that are still open

dns/dnsperfgo/main.go
Tanzp01: Thanks, I will fix it in another PR with your other comments.

dns/dnsperfgo/main.go
Tanzp01: Already done.

dns/dnsperfgo/main.go
Tanzp01: Already done.

dns/dnsperfgo/main.go
Tanzp01: Already done.

PR #3642, opened 2025-10-15
Render version of perfdash on UI

kishen-v:
#### What type of PR is this?
/kind feature

This PR allows the version of the perfdash to be displayed on the UI and reduces the complexity around identifying the image used in production.

Fixes #

The UI now renders as follows:
Performance Dashboard


perfdash/Makefile
```diff
@@ -1,6 +1,7 @@
 all: push
 
 # See deployment.yaml for the version currently running-- bump this ahead before rebuilding!
+# Additionally please ensure to update the same in www/VERSION
```
serathius, line 4: Could you add a static analysis that validate that VERSION file contents are equal the TAG in make file?

kishen-v: Added a check before we build the container. 
Thanks for the review!

perfdash/Makefile
```diff
@@ -24,7 +26,11 @@ run: perfdash
 # Use buildkit to have "COPY --chmod=" support (availability of it in "regular" docker build depends on docker version).
 .PHONY: container
 container:
+ifneq ($(TAG),$(WWW_VERSION))
+	$(error Version mismatch! Makefile=$(TAG), www/VERSION=$(WWW_VERSION). Please align before building)
```
serathius: This should be a check during testing.

kishen-v: Sorry about that. Missed the fact that presubmits can catch the version mismatch much earlier. 
Thanks for pointing out.

perfdash/Makefile
```diff
@@ -1,12 +1,19 @@
...
 TAG?=2.56
+WWW_VERSION=$(shell cat www/VERSION)
 
 REPO?=gcr.io/k8s-staging-perf-tests
 
 .PHONY: test
+test: WWW_VERSION = $(shell cat www/VERSION)
+
 test: perfdash.go parser.go config.go metrics-downloader-helper.go metrics-downloader.go gcs_metrics_bucket.go s3_metrics_bucket.go local_metrics_dir.go
+ifneq ($(TAG),$(WWW_VERSION))
```
serathius: Maybe after go test so we run the tests even if the files don't align?

kishen-v: Addressed, I've also made a small change to use `shell` for validating the version, as `ifneq` is evaluated globally at parse time. 

Using `ifneq` brought in a dependency for all make targets to ensure that the versions matched even during development, and if at all we've to use different tags.

kishen-v:
Hey @serathius, 

I've rebased the changes against the latest master and resolved the merge conflicts. Can you PTAL when time permits?

Thanks!

BenTheElder:
/remove-lifecycle rotten
/lgtm
[echoing the LGTM above since the bot missed it]

PR #3644, opened 2025-10-15
Allow increases object size via environment variable

serathius:
Skipped Job type as environment variable is immutable, causing reconciliation to break. Will address this in the future.

/kind feature

Ref https://github.com/kubernetes/kubernetes/issues/134375

### DIFF
--- a/clusterloader2/testing/load/statefulset.yaml
+++ b/clusterloader2/testing/load/statefulset.yaml
@@ -1,6 +1,7 @@
 {{$HostNetworkMode := DefaultParam .CL2_USE_HOST_NETWORK_PODS false}}
 {{$EnablePVs := DefaultParam .CL2_ENABLE_PVS true}}
 {{$RUN_ON_ARM_NODES := DefaultParam .CL2_RUN_ON_ARM_NODES false}}
+{{$podPayloadSize := DefaultParam .CL2_STATEFULSET_POD_PAYLOAD_SIZE 0}}
 {{$Image := DefaultParam .Image "registry.k8s.io/pause:3.9"}}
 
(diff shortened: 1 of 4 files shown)


clusterloader2/testing/load/statefulset.yaml
```diff
@@ -1,6 +1,7 @@
 {{$HostNetworkMode := DefaultParam .CL2_USE_HOST_NETWORK_PODS false}}
 {{$EnablePVs := DefaultParam .CL2_ENABLE_PVS true}}
 {{$RUN_ON_ARM_NODES := DefaultParam .CL2_RUN_ON_ARM_NODES false}}
+{{$podPayloadSize := DefaultParam .CL3_STATEFULSET_POD_PAYLOAD_SIZE 0}}
```
mborsz: CL3?

mborsz (commented):
Do we know if anyone is using the pervious ".Env" field?

serathius:
> Do we know if anyone is using the pervious ".Env" field?

No, strangely it's set in `reconcile-objects.yaml` but for `daemonSet` which doesn't use `.Env`

https://cs.k8s.io/?q=Env%3A&i=nope&literal=nope&files=clusterloader2%2Ftesting%2Fload&excludeFiles=&repos=kubernetes/perf-tests

clusterloader2/testing/load/statefulset.yaml
serathius: Fixed

serathius:
> No, strangely it's set in reconcile-objects.yaml but for daemonSet which doesn't use .Env

No, my brain mistook resources. This is important. Good catch!

PR #3645, opened 2025-10-15
Make randomized data in pods deterministic to avoid pod recreation

serathius:
This allows us to set resource size for Jobs which pod spec is immutable.

/kind feature

Followup to https://github.com/kubernetes/perf-tests/pull/3644


mborsz:
please fix linter, LGTM otherwise

PR #3436, opened 2025-07-08
Add DRA jobs to perf dash

nojnhuh:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:

Moving the DRA scalability jobs into their own `DRA` subdirectory in https://github.com/kubernetes/test-infra/pull/35033 made their results disappear from perf-dash. This PR adds that subdirectory as a source from which to pull data.

#### Which issue(s) this PR fixes:

Fixes #

PR #3647, opened 2025-10-16
Revert "Add DRA jobs to perf dash"

nojnhuh:
#### What type of PR is this?

/kind bug

This PR reverts #3436 which seems like the most likely reason why https://perf-dash.k8s.io/ looks broken at the moment:
image

Fixes #


nojnhuh:
@wojtek-t Are there any logs I can check to see why this might've broken things? I tried running perf-dash locally but couldn't reproduce this exact issue when I point to the same `githubConfigDir`s.

PR #3653, opened 2025-10-22
fix missing project id and delete deprecated LegacyIsMasterNode function

upodroid:
#### What type of PR is this?

/kind cleanup

Scale jobs using kops aren't reading the project id correctly because the GCP project id is not explicitly set. I changed it to read node providerID set by the cloud controller manager.

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-kops-gce-5000-ipalias-using-cl2/1980979471863255040

I also cleaned up a function deprecated in k8s 1.19

Fixes #


upodroid:
friendly bump

upodroid:
/hold cancel

thanks

PR #3654, opened 2025-10-23
Parametrize DRA test config

emerbe:
#### What type of PR is this?

/kind cleanup

This PR modifies DRA test logic a bit so it's parametrized to simplify running test with different drivers and with different scale.

Fixes #

### DIFF
--- a/clusterloader2/pkg/dependency/dra/dra.go
+++ b/clusterloader2/pkg/dependency/dra/dra.go
@@ -167,12 +167,6 @@ func (d *draDependency) isDRADriverReady(config *dependency.Config, daemonsetNam
 }
 
 func isResourceSlicesPublished(config *dependency.Config, namespace string) (bool, error) {
-	// Get a list of all nodes
-	// nodes, err := getReadyNodesCount(config)
-	// if err != nil {
-	// 	return false, fmt.Errorf("failed to list nodes: %v", err)
-	// }
-
 	driverPluginPods, err := getDriverPluginPods(config, namespace, draDaemonsetName)
 	if err != nil {
 		return false, fmt.Errorf("failed to list driverPluginPods: %v", err)
--- a/clusterloader2/testing/dra/config.yaml
+++ b/clusterloader2/testing/dra/config.yaml
@@ -2,28 +2,47 @@
 {{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .CL2_NODES_PER_NAMESPACE 100)}}
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$STEADY_STATE_QPS := DefaultParam .CL2_STEADY_STATE_QPS 5}}
+{{$RESOURCE_SLICES_PER_NODE := DefaultParam .CL2_RESOURCE_SLICES_PER_NODE 1}}
+{{$LONG_LIVED_JOBS_STARTUP_PERC50_THRESHOLD := DefaultParam .CL2_LONG_LIVED_JOBS_STARTUP_PERC50_THRESHOLD "5s"}}
+{{$LONG_LIVED_JOBS_STARTUP_PERC90_THRESHOLD := DefaultParam .CL2_LONG_LIVED_JOBS_STARTUP_PERC90_THRESHOLD "5s"}}
+{{$LONG_LIVED_JOBS_STARTUP_PERC100_THRESHOLD := DefaultParam .CL2_LONG_LIVED_JOBS_STARTUP_PERC100_THRESHOLD "5s"}}
+{{$CHURN_JOBS_WAIT_PERC50_THRESHOLD := DefaultParam .CL2_CHURN_JOBS_WAIT_PERC50_THRESHOLD "5s"}}
+{{$CHURN_JOBS_WAIT_PERC90_THRESHOLD := DefaultParam .CL2_CHURN_JOBS_WAIT_PERC90_THRESHOLD "5s"}}
+{{$CHURN_JOBS_WAIT_PERC99_THRESHOLD := DefaultParam .CL2_CHURN_JOBS_WAIT_PERC99_THRESHOLD "5s"}}
+{{$SHORT_LIVED_JOBS_WAIT_THRESHOLD := DefaultParam .CL2_SHORT_LIVED_JOBS_WAIT_THRESHOLD "10m"}}
+{{$LONG_LIVED_JOBS_WAIT_THRESHOLD := DefaultParam .CL2_LONG_LIVED_JOBS_WAIT_THRESHOLD "10m"}}
+{{$RUNNING_JOBS_OPERATION_THRESHOLD := DefaultParam .CL2_RUNNING_JOBS_OPERATION_THRESHOLD "120s"}}
 {{$token := .CL2_TOKEN }}
 
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
 
+# dra
+{{$draManifests := DefaultParam .CL2_DRA_MANIFESTS "dra-example-driver"}}
+{{$draDaemonsetName := DefaultParam .CL2_DRA_DAEMONSET_NAME "dra-example-driver-kubeletplugin"}}
+
 # Node resource configuration
 {{$gpusPerNode := DefaultParam .CL2_GPUS_PER_NODE 8}}
+{{$resourceSlicesPerNode := DefaultParam .CL2_RESOURCE_SLICES_PER_NODE 1}}
+{{$totalResourceSliceCount := MultiplyInt $resourceSlicesPerNode .Nodes}}
 {{$totalGPUs := MultiplyInt $gpusPerNode .Nodes}}
 
 # fast fill job configuration - for initial fill up
 {{$fillPercentage := DefaultParam .CL2_FILL_PERCENTAGE 90}}
 {{$fillPodsCount := DivideInt (MultiplyInt $totalGPUs $fillPercentage) 100}}
 {{$fillPodsPerNamespace := DivideInt $fillPodsCount $namespaces}}
-{{$longJobSize := 1}}
-{{$longJobRunningTime := DefaultParam .CL2_LONG_JOB_RUNNING_TIME "1h"}}
+{{$longLivedJobSize := 1}}
+{{$longLivedJobRunningTime := DefaultParam .CL2_LONG_LIVED_JOB_RUNNING_TIME "1h"}}
 
 # churn job configuration for steady state
-{{$smallJobPodsCount := SubtractInt $totalGPUs (MultiplyInt $fillPodsPerNamespace $namespaces)}}
-{{$smallJobsPerNamespace := DivideInt $smallJobPodsCount $namespaces}}
-{{$smallJobSize := 1}}
-{{$smallJobCompletions := 10}}
-{{$jobRunningTime := DefaultParam .CL2_JOB_RUNNING_TIME "30s"}}
+{{$shortLivedJobPodsCount := SubtractInt $totalGPUs (MultiplyInt $fillPodsPerNamespace $namespaces)}}
+{{$calculatedSJPN := DivideInt $shortLivedJobPodsCount $namespaces}}
+{{$maxSJPN := DefaultParam .CL2_MAX_SHORT_LIVED_JOBS_PER_NAMESPACE 999999}}
+{{$shortLivedJobsPerNamespace := MinInt $calculatedSJPN $maxSJPN}}
+{{$shortLivedJobSize := 1}}
+{{$shortLivedJobCompletions := DefaultParam .CL2_SHORT_LIVED_JOB_COMPLETIONS 10}}
+{{$shortLivedJobRunningTime := DefaultParam .CL2_SHORT_LIVED_JOB_RUNNING_TIME "30s"}}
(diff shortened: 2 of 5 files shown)


clusterloader2/testing/dra/config.yaml
```diff
@@ -164,9 +195,9 @@ steps:
       Params:
         action: start
         labelSelector: job-type = short-lived
-        perc50Threshold: 40s
-        perc90Threshold: 60s
-        perc99Threshold: 80s
+        perc50Threshold: {{$CHURN_POD_STARTUP_PERC50_THRESHOLD}}
```
alaypatel07: @emerbe lets move this change out, ideally, the purpose of these scale test is to weed out issues that cause the latency to increase. Infact, I suggest we should just remove the 40s, 60s and 80s configuration and fall back to 5s default

clusterloader2/testing/dra/config.yaml
```diff
@@ -2,12 +2,29 @@
 {{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .CL2_NODES_PER_NAMESPACE 100)}}
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$STEADY_STATE_QPS := DefaultParam .CL2_STEADY_STATE_QPS 5}}
+{{$RESOURCE_SLICES_PER_NODE := DefaultParam .CL2_RESOURCE_SLICES_PER_NODE 1}}
+{{$UPSIZE_THRESHOLD := DefaultParam .CL2_UPSIZE_THRESHOLD "10m"}}
```
alaypatel07: This value does not have any impact except from printing the threshold value. Why is the change needed?

clusterloader2/testing/dra/config.yaml
```diff
@@ -2,12 +2,29 @@
...
 
+# dra
+{{$draNamespace := DefaultParam .CL2_DRA_NAMESPACE "dra-example-driver"}}
+{{$draManifests := DefaultParam .CL2_DRA_MANIFESTS "dra-example-driver"}}
+{{$draDaemonsetName := DefaultParam .CL2_DRA_DAEMONSET_NAME "dra-example-driver-kubeletplugin"}}
+
 # Node resource configuration
 {{$gpusPerNode := DefaultParam .CL2_GPUS_PER_NODE 8}}
+{{$resourceSlicesPerNode := DefaultParam .CL2_RESOURCE_SLICES_PER_NODE 1}}
+{{$workerNodeCount := MultiplyInt $resourceSlicesPerNode .Nodes}}
```
alaypatel07: nit: should this be named `$totalResourceSliceCount`?

clusterloader2/testing/dra/config.yaml
```diff
@@ -2,12 +2,29 @@
...
+{{$CHURN_POD_STARTUP_PERC99_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC99_THRESHOLD “80s”}}
+{{$FINISHED_JOBS_THRESHOLD := DefaultParam .CL2_FINISHED_JOBS_THRESHOLD "10m"}}
+{{$RUNNING_JOBS_THRESHOLD := DefaultParam .CL2_RUNNING_JOBS_THRESHOLD "10m"}}
+{{$RUNNING_JOBS_OPERATION_THRESHOLD := DefaultParam .CL2_RUNNING_JOBS_OPERATION_THRESHOLD "120s"}}
 {{$token := .CL2_TOKEN }}
 
 {{$namespaces := DivideInt .Nodes $NODES_PER_NAMESPACE}}
 
+# dra
+{{$draNamespace := DefaultParam .CL2_DRA_NAMESPACE "dra-example-driver"}}
```
alaypatel07: whats the point of exposing these as env variables?

emerbe: Good point. I removed it and made namepsace name in `dra.go` generic

clusterloader2/testing/dra/config.yaml
emerbe: OK, done.

clusterloader2/testing/dra/config.yaml
emerbe: I found it beneficial to be able to parametrize these values, especially for large-scale tests. For example, I had to increase them when testing 5,000 nodes with limited resources.

I propose to set the default to 5 seconds to variables which would align with your suggestion while still allowing test users the flexibility to adjust it.

Have you been running those test on a bigger scale?

I've tried and for 5k nodes I had to set:
      - --env=CL2_CHURN_POD_STARTUP_PERC50_THRESHOLD=5m
      - --env=CL2_CHURN_POD_STARTUP_PERC90_THRESHOLD=6m
      - --env=CL2_CHURN_POD_STARTUP_PERC99_THRESHOLD=8m

clusterloader2/testing/dra/config.yaml
emerbe: Those thresholds define the acceptable limits for the pod startup times.

The initial values in this PT were too stretched for 100 nodes but in bigger scale I observed that those need to be higher. Even for 100 nodes in GKE 5 seconds is not enough:

E1102 21:24:30.207908   25461 clusterloader.go:258]   Errors: [measurement call PodStartupLatency - FastFillPodStartupLatency error: pod startup: too high latency 99th percentile: got 8.792346318s expected: 5s]

I've been using this test with different than drivers and observed that times needed for different measurements may vary.

clusterloader2/testing/dra/config.yaml
alaypatel07: > I propose to set the default to 5 seconds to variables which would align with your suggestion while still allowing test users the flexibility to adjust it.

If you dont change the defaults to the upstream tests then parameterizing it sounds good to me.

> Have you been running those test on a bigger scale? 

Unfortunately the test-infra for higher scale is just getting set up, so we will have the numbers reflected here: https://testgrid.k8s.io/sig-scalability-dra#gce-dra-with-workload-master-scalability-5000

but we have run simulated tests as part of GA'ing DRA results are here: https://docs.google.com/document/d/1YzPR3j0gxDNwL6iaY28trUsKVb2xyhrwHY-DqVcVaUE/edit?tab=t.0

clusterloader2/pkg/dependency/dra/dra.go
```diff
@@ -167,12 +167,6 @@ func (d *draDependency) isDRADriverReady(config *dependency.Config, daemonsetNam
 }
 
 func isResourceSlicesPublished(config *dependency.Config, namespace string) (bool, error) {
-	// Get a list of all nodes
```
alaypatel07, line 170: not sure why changes in this file are needed.

emerbe: I've noticed that I haven't deleted commented code in the previous PR :/

clusterloader2/testing/dra/config.yaml
emerbe: So, the current values are the same as default.

Since 5k tests are being set up this will be a good opportunity to use those parameters ;)
I will mark comment this conversation as resolved then.

clusterloader2/pkg/dependency/dra/dra.go
```diff
@@ -34,7 +34,7 @@ import (
 
 const (
 	draDependencyName      = "DRATestDriver"
-	draNamespace           = "dra-example-driver"
+	draNamespace           = "dra-driver"
```
alaypatel07: I dont think we should change the name of the driver. 

The "dra-example-driver" has a very clear reference to this: https://github.com/kubernetes-sigs/dra-example-driver/tree/main.

Are there any other reasons why we need to change the name here that I am missing?

emerbe: The test now allows different drivers (provided their YAMLs are copied into the manifests directory), but the draNamespace is not parametrized.

I've made the namespace more generic. If someone wants to reuse this test (as I am doing) or merge more drivers into OSS, it is confusing to create those drivers pods in a namespace named dra-example-driver.

alaypatel07: Parameterising the draNamespace makes sense to me.

Changing the namespace of dra-example-driver doesn't, if nothing breaks for you with this name please keep it as such. To me it's more clarifying than confusing.

@mortent thoughts?

emerbe: This is actually pretty minor change in my opinion.

I'm fine with changing it back to `dra-example-driver`.

We can come back to this discussion in the future if/when this becomes confusing.

Will mark it as `Resolved` since I've addressed initial comment.

clusterloader2/testing/dra/config.yaml
```diff
@@ -2,12 +2,28 @@
...
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$STEADY_STATE_QPS := DefaultParam .CL2_STEADY_STATE_QPS 5}}
+{{$RESOURCE_SLICES_PER_NODE := DefaultParam .CL2_RESOURCE_SLICES_PER_NODE 1}}
+{{$UPSIZE_THRESHOLD := DefaultParam .CL2_UPSIZE_THRESHOLD "5s"}}
+{{$UPSIZE_PERC50_THRESHOLD := DefaultParam .CL2_UPSIZE_PERC50_THRESHOLD "5s"}}
+{{$UPSIZE_PERC90_THRESHOLD := DefaultParam .CL2_UPSIZE_PERC90_THRESHOLD "5s"}}
+{{$CHURN_POD_STARTUP_PERC50_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC50_THRESHOLD "5s"}}
+{{$CHURN_POD_STARTUP_PERC90_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC90_THRESHOLD "5s"}}
+{{$CHURN_POD_STARTUP_PERC99_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC99_THRESHOLD "5s"}}
+{{$FINISHED_JOBS_THRESHOLD := DefaultParam .CL2_FINISHED_JOBS_THRESHOLD "10m"}}
```
alaypatel07: I think CL2_LONG_LIVED_JOBS_WAIT_THRESHOLD and CL2_CHURN_JOBS_WAIT_THRESHOLD will good env variable names for this.

Also other places where we refer to these jobs in variable names can be changed for example:

```
{{$smallJobCompletions := DefaultParam .CL2_SMALL_JOB_COMPLETIONS 10}}
```

clusterloader2/testing/dra/config.yaml
```diff
@@ -46,10 +65,12 @@ tuningSets:
 
 
 dependencies:
-- name: Install dra-example-driver for test
+- name: Install dra-driver for test
```
alaypatel07: nit: this was leftover from previous version of the PR

emerbe: Fixed

clusterloader2/testing/dra/config.yaml
```diff
@@ -2,12 +2,28 @@
 {{$NODES_PER_NAMESPACE := MinInt .Nodes (DefaultParam .CL2_NODES_PER_NAMESPACE 100)}}
 {{$LOAD_TEST_THROUGHPUT := DefaultParam .CL2_LOAD_TEST_THROUGHPUT 10}}
 {{$STEADY_STATE_QPS := DefaultParam .CL2_STEADY_STATE_QPS 5}}
+{{$RESOURCE_SLICES_PER_NODE := DefaultParam .CL2_RESOURCE_SLICES_PER_NODE 1}}
+{{$UPSIZE_THRESHOLD := DefaultParam .CL2_UPSIZE_THRESHOLD "5s"}}
```
mengqiy: It's a bit unclear about the meaning of this when we have UPSIZE_PERC50_THRESHOLD and UPSIZE_PERC90_THRESHOLD

clusterloader2/testing/dra/config.yaml
```diff
@@ -2,12 +2,28 @@
...
+{{$UPSIZE_PERC50_THRESHOLD := DefaultParam .CL2_UPSIZE_PERC50_THRESHOLD "5s"}}
+{{$UPSIZE_PERC90_THRESHOLD := DefaultParam .CL2_UPSIZE_PERC90_THRESHOLD "5s"}}
+{{$CHURN_POD_STARTUP_PERC50_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC50_THRESHOLD "5s"}}
+{{$CHURN_POD_STARTUP_PERC90_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC90_THRESHOLD "5s"}}
+{{$CHURN_POD_STARTUP_PERC99_THRESHOLD := DefaultParam .CL2_CHURN_POD_STARTUP_PERC99_THRESHOLD "5s"}}
```
mengqiy: It doesn't make a lot of sense when we use the same default value for P50, P90 and P99.

clusterloader2/testing/dra/config.yaml
emerbe: Done

clusterloader2/testing/dra/config.yaml
emerbe: Those are defaults for 100 nodes that can be changed for larger scale.

clusterloader2/testing/dra/config.yaml
emerbe: That's absolute threshold.

I've changed name to: `LONG_LIVED_JOBS_STARTUP_PERC100_THRESHOLD`

PR #3702, opened 2025-12-02
Add requests to exec deployment

serathius:
/kind feature

Adding requests to the execservice to prevent OOMs in execservice. 

On the 5k scalability tests the nodes are running close the memory limit. This causes issues for resource-size experiments where we get  `api_availability_measurement.go:83] execservice issue: problem with RunCommand(): output="", err=signal: killed` which are sign of OOMs.

To mitigate it I assigned enough requests to ensure that exec_deployment is scheduled on heapster node to ensure enough resources. 

Alternatively I could make it templated to avoid requiring too much resources in base case.

/assign @mborsz

### DIFF
--- a/clusterloader2/pkg/execservice/exec_service.go
+++ b/clusterloader2/pkg/execservice/exec_service.go
@@ -84,7 +84,10 @@ func SetUpExecService(f *framework.Framework, c config.ExecServiceConfig) error 
 		klog.V(3).Infof("%s: service already running!", execServiceName)
 	}
 	klog.V(2).Infof("%v: setting up service!", execServiceName)
-	mapping := make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


mborsz:
I think they are too large and doesn't fit in heapster node, at least for 100 nodes scenario (presubmit):

> F1202 09:49:41.612333   14832 clusterloader.go:343] Error while setting up exec service: got context deadline exceeded while waiting for 3 pods to be running in namespace(cluster-loader), labelSelector(feature = exec) - summary of pods : Pods: 3 out of 3 created, 2 running (2 updated), 0 pending scheduled, 1 not scheduled, 0 inactive, 0 terminating, 0 unknown, 0 runningButNotReady

PR #3770, opened 2026-01-06
benchmark: fix GCS bucket path for fetching CI job logs

dims:
The benchmark tool was using the old "kubernetes-jenkins" GCS bucket to fetch job logs, but all CI jobs have been migrated to the "kubernetes-ci-logs" bucket as part of the k8s-infra migration.

This was causing the benchmark job (ci-perf-tests-kubemark-100-benchmark) to fail with "No matching dirs were found" errors when trying to fetch metrics from ci-kubernetes-e2e-gci-gce-scalability and ci-kubernetes-kubemark-100-gce jobs.

- https://storage.googleapis.com/k8s-metrics/failures-latest.html?include=perf&sort=days&asc=0
- https://testgrid.k8s.io/sig-scalability-perf-tests#kubemark-100-benchmark

#### What type of PR is this?

/kind bug

Fixes #

### DIFF
--- a/benchmark/pkg/metricsfetcher/util/util.go
+++ b/benchmark/pkg/metricsfetcher/util/util.go
@@ -46,10 +46,18 @@ type GCSLogUtils struct {
 	googleGCSBucketUtils *utils.Utils
 }
 
+// GCS bucket and directory for Kubernetes CI logs.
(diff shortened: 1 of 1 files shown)


dims:
/assign mborsz
cc @hakman @hakuna-matatah

PR #3781, opened 2026-01-16
watch-list: support core resources via api-version/resource flags

p0lyn0mial:
#### What type of PR is this?
/kind feature

Fixes #

### DIFF
--- a/util-images/watch-list/main.go
+++ b/util-images/watch-list/main.go
@@ -18,12 +18,16 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
+	"fmt"
 	"os"
+	"strconv"
 	"time"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
-	v1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
@@ -36,6 +40,8 @@ var (
 	informerCount               int
 	testTimeout                 time.Duration
 	enableWatchListAlphaFeature bool
+	apiVersion                  string
+	resource                    string
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


util-images/watch-list/main.go
```diff
@@ -66,38 +71,72 @@ func main() {
...
 func registerFlags() {
 	klog.InitFlags(flag.CommandLine)
 
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "path to kubeconfig.")
-	flag.StringVar(&targetNamespace, "namespace", "huge-secrets-1", "namespace that host secrets to list. If empty a default (huge-secrets-1) value will be used.")
+	flag.StringVar(&targetNamespace, "namespace", "huge-secrets-1", "namespace that hosts the target resource. If empty a default (huge-secrets-1) value will be used.")
 	flag.IntVar(&informerCount, "count", 4, "the number of informers per targetNamespace to run. If empty a default (4) value will be used.")
 	flag.DurationVar(&testTimeout, "timeout", time.Minute, "timeout duration for the test")
 	flag.BoolVar(&enableWatchListAlphaFeature, "enableWatchListFeature", false, "whether to set KUBE_FEATURE_WatchListClient env var")
+	flag.StringVar(&apiVersion, "api-version", "v1", "apiVersion of the target resource (e.g. v1, apps/v1). If empty a default (v1) value will be used.")
```
p0lyn0mial, line 112: api-version and resource could be combined into one flag but then we need some way of splitting combined input. 

i don't think there is a convention in kube.

util-images/watch-list/main.go
```diff
@@ -66,38 +71,72 @@ func main() {
...
-func startInformersFor(ctx context.Context, client kubernetes.Interface, count int, namespace string) []cache.InformerSynced {
+func startInformersForResource(ctx context.Context, client kubernetes.Interface, gvr schema.GroupVersionResource, count int, namespace string) ([]cache.InformerSynced, error) {
 	var informersSynced []cache.InformerSynced
+	factories := make([]informers.SharedInformerFactory, 0, count)
+
 	for i := 0; i < count; i++ {
-		inf := v1.NewSecretInformer(client, namespace, time.Duration(0), cache.Indexers{})
-		informersSynced = append(informersSynced, inf.HasSynced)
-		go inf.Run(ctx.Done())
+		factory := informers.NewSharedInformerFactoryWithOptions(
```
p0lyn0mial, line 121: I think that for a test binary it is okay to startup multiple factories.
Doing it that way makes the code easy.

util-images/watch-list/main.go
serathius: Right, one thing that we will need to think is how to create informer per core resource. For now having a one for pods should be good enough.

util-images/watch-list/main.go
serathius: SGTM, I think Shared Informer would reuse informer per resource.

p0lyn0mial:
once this PR merges I think that we need to release a new img, xref: https://github.com/p0lyn0mial/perf-tests/blob/67eba5a31de567343e72d2af56f7fe461456ed52/util-images/watch-list/README.md#building-and-releasing

PR #3782, opened 2026-01-16
Add a pod continiously re-initializing informer to load test to test LIST requests

serathius:
/kind feature

cc @mborsz @p0lyn0mial

### DIFF
--- a/clusterloader2/testing/load/modules/informer/deployment.yaml
+++ b/clusterloader2/testing/load/modules/informer/deployment.yaml
@@ -1,0 +1,27 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{.Name}}
+  labels:
+    group: informer
+spec:
+  selector:
+    matchLabels:
+      group: informer
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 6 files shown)


clusterloader2/testing/load/modules/informer/deployment.yaml
```diff
@@ -0,0 +1,27 @@
...
+          image: gcr.io/k8s-staging-perf-tests/watch-list:v0.0.1
+          resources:
+            requests:
+              memory: "16Gi"
+              cpu: "6"
+            limits:
+              memory: "16Gi"
+              cpu: "6"
+          command: [ "watch-list" ]
+          args: [ "--alsologtostderr=true", "--v=4", "--timeout=100h", "--count=1", "--enableWatchListFeature=false"]
```
p0lyn0mial, line 27: by default the binary will list `secrets` in `huge-secrets-1` namespace. we need to provide additional flags for changing the resource type and ns.

serathius: Right, we decided to move forward knowing that.

PR #3791, opened 2026-01-26
Fix 'Move reconcile-objects manifests into the module directory'

Qqkyu:
#### What type of PR is this?

/kind bug
/kind regression

Two symlinks were broken after https://github.com/kubernetes/perf-tests/pull/3778


Qqkyu:
cc: @serathius

PR #3801, opened 2026-02-09
Adding mapEventsByOrder option for PodStartupLatency

YahiaBadr:
This option (if enabled) changes the way the PodStartupLatency calculates the latency, intead of matching each pod creation time with the same pod shceduled time, it will matches the first creation time with the first schedule time whatever the pod is the same or not.

#### What type of PR is this?
/kind feature

This option (if enabled) changes the way the PodStartupLatency calculates the latency, intead of matching each pod creation time with the same pod scheduled time, it will matches the first creation time with the first schedule time whatever the pod is the same or not.

Fixes #


Qqkyu (commented):
/ok-to-test
/lgtm
/approve

PR #3802, opened 2026-02-10
feat: Add an ability to measure clusters with dns only control plane …

matix522:
…access

This update allows ClusterLoader2 to target clusters using a DNS endpoint for the control plane. It includes support for the MASTER_DNS_ENDPOINT environment variable and relevant configuration changes. Changed logging to show all of the Node adresses to allow for dualstack clusters.

#### What type of PR is this?

What type of PR is this?
/kind feature

This PR is needed by the GKE team to test the performance of clusters that use DNS address instead of IP address for cluster control plane

### DIFF
--- a/clusterloader2/pkg/prometheus/prometheus.go
+++ b/clusterloader2/pkg/prometheus/prometheus.go
@@ -156,10 +156,14 @@ func NewController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *Control
 	if errList != nil {
 		return nil, errList
 	}
-	mapping["MasterIps"], err = getMasterIps(clusterLoaderConfig.ClusterConfig, clusterLoaderConfig.PrometheusConfig.ScrapeMastersWithPublicIPs)
-	if err != nil {
-		klog.Warningf("Couldn't get master ip, will ignore manifests requiring it: %v", err)
-		delete(mapping, "MasterIps")
+	if clusterLoaderConfig.ClusterConfig.MasterDNSEndpoint != "" {
+		klog.Infof("Cluster uses DNS master enpoint will not provide MasterIps")
+	} else {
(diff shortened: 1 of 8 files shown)


clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -649,6 +655,9 @@ func getMasterIps(clusterConfig config.ClusterConfig, usePublicIPs bool) ([]stri
 		klog.V(2).Infof("Using internal master ips (%s) to monitor master's components", clusterConfig.MasterInternalIPs)
 		return clusterConfig.MasterInternalIPs, nil
 	}
+	if clusterConfig.MasterDNSEndpoint != "" {
+		return nil, fmt.Errorf("cluster has no master IPs, but has master DNS endpoint %s", clusterConfig.MasterDNSEndpoint)
+	}
```
Qqkyu: Two related questions here

1st: Is this check position intended? In other words, even when the `master-endpoint` is set the following combinations:
- `usePublicIPs` (using `--prometheus-scrape-masters-with-public-ips`) and having > 0 master IPs
- \> 0 master internal IPs

Shouldn't error here? I'm a bit confused because I'd assume that we'd add this check at the beginning of the `getMasterIps` function and/or not invoke this function at all when `MasterDNSEndpoint` is present. 

2nd: Should we invoke this function if we have `clusterConfig.MasterDNSEndpoint`? AFAIU currently it'd go like this:

1. `NewController` function invokes the `getMasterIps` **always**
2. `clustersConfig.MasterDNSEndpoint` is specified, so we return error
3. We display the `Couldn't get master ip, will ignore manifests requiring it: %v`

Probably not the end of the world, but should it really warn? It's a valid path after all?

matix522: Good points, I moved the check to the top of the function (to disallow calling in future), as well as not calling this function in the first place when DNS endpoint is provided

Qqkyu (commented):
/lgtm

PR #3803, opened 2026-02-10
Filter events to tracked pods in PodStartupLatency

YahiaBadr:
When mapEventsByOrder is enabled, gatherScheduleTimes now filters events to only include those belonging to tracked pods before performing the count verification and ordering. This makes the measurement more robust against extra events for untracked pods in the same namespace.

#### What type of PR is this?
/kind feature

When mapEventsByOrder is enabled, gatherScheduleTimes now filters events to only include those belonging to tracked pods before performing the count verification and ordering. This makes the measurement more robust against extra events for untracked pods in the same namespace.

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -304,17 +304,17 @@ func (p *podStartupLatencyMeasurement) gatherScheduleTimes(c clientset.Interface
 		return err
 	}
 
-	if p.mapEventsByOrder {
-		// Filter events to only include those that belong to pods we are tracking.
-		var filteredEvents []corev1.Event
-		for _, event := range schedEvents.Items {
-			key := createMetaNamespaceKey(event.InvolvedObject.Namespace, event.InvolvedObject.Name)
-			if _, exists := p.podStartupEntries.Get(key, createPhase); exists {
-				filteredEvents = append(filteredEvents, event)
-			}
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -305,6 +305,16 @@ func (p *podStartupLatencyMeasurement) gatherScheduleTimes(c clientset.Interface
...
 	if p.mapEventsByOrder {
+		// Filter events to only include those that belong to pods we are tracking.
+		var filteredEvents []corev1.Event
+		for _, event := range schedEvents.Items {
+			key := createMetaNamespaceKey(event.InvolvedObject.Namespace, event.InvolvedObject.Name)
+			if _, exists := p.podStartupEntries.Get(key, createPhase); exists {
+				filteredEvents = append(filteredEvents, event)
+			}
+		}
+		schedEvents.Items = filteredEvents
```
tosi3k: Nit: we can move this part above the `if p.mapEventsByOrder` statement and simplify the `else` branch's content on the occasion.

YahiaBadr: Sure, i was thinking about that but wanted to test it first on my test as editing the else part would put all tests on risk. I will open a followup-PR to clean it and merge it once tested.

YahiaBadr: fixed

PR #3805, opened 2026-02-13
fetch metrics correctly from control plane instances

upodroid:
#### What type of PR is this?

/kind bug

When we run scale tests with kops, we get the following error:

```
W0213 14:15:54.305851   57465 metrics_grabber.go:105] Master node is not registered. Grabbing metrics from Scheduler, ControllerManager and ClusterAutoscaler is disabled.
```

This happens because the controlplane instances are no longer called `master-*`, now we'll look for the control plane labels/taints to detect the control plane instances

Fixes #


Qqkyu (commented):
/lgtm

PR #3806, opened 2026-02-14
drop managing storage classes

upodroid:
#### What type of PR is this?

/kind cleanup

The cluster provisioner should install a valid CSI driver for the cluster used for testing. Therefore, this should no longer be managed by clusterloader2, also in-tree CSI drivers don't exist anymore and should be removed.

Fixes #


Qqkyu (commented):
/lgtm

Will we remove the `PROMETHEUS_STORAGE_CLASS_PROVISIONER` after this from the [sig-scalability-presets.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-presets.yaml)?

PR #2909, opened 2024-09-19
Use `krel push` instead of the legacy push-build.sh

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
The script has been deprecated for a while and we should stick to use krel only.

cc @kubernetes/release-managers 
#### Which issue(s) this PR fixes:
None

#### Special notes for your reviewer:

Requires https://github.com/kubernetes/release/pull/3768

/hold

PR #3845, opened 2026-02-27
Use krel push instead of the legacy push-build.sh

saschagrunert:
#### What type of PR is this?

/kind cleanup

The `push-build.sh` script has been deprecated for a while and we should use `krel push` instead. This is a follow-up to #2909 which got bot-closed due to inactivity.

None

cc @mm4tt @tosi3k @kubernetes/release-managers

### DIFF
--- a/golang/push.sh
+++ b/golang/push.sh
@@ -17,10 +17,11 @@
 set -euxo pipefail
 
 # Push Kubernetes build to GCS.
-cd $ROOT_DIR/k8s.io/kubernetes
-../release/push-build.sh \
+cd "$ROOT_DIR/k8s.io/release"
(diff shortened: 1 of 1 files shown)


rf232:
I believe that this pr broke the build-and-push-k8s-at-golang-tip build: https://testgrid.k8s.io/sig-scalability-golang#build-and-push-k8s-at-golang-tip

saschagrunert:
> I believe that this pr broke the build-and-push-k8s-at-golang-tip build: https://testgrid.k8s.io/sig-scalability-golang#build-and-push-k8s-at-golang-tip

Working towards a fix in https://github.com/kubernetes/release/pull/4310

PR #3873, opened 2026-03-04
Add flags for errorLogger and latencyLogger

Tanzp01:
#### What type of PR is this?
/kind feature

To gain more detailed information in propagation latency and errors

Will update the test result of this PR


Tanzp01:
I believe the flag need to be changed together with image: registry.k8s.io/perf-tests/probes:v0.0.7.

PR #3890, opened 2026-03-14
clusterloader2: create pod-startup-latency test

justinsb:
This is similar to node-throughput, but focuses on the latency of pod startup instead of the throughput.

* We reduce the rate at which we create pods to 1/s,
  so that we can measure the startup latency without creating a backlog of pending pods.

* We set the imagePullPolicy to Never,
  so that we can measure pod startup latency without image pull time.

* We set automountServiceAccountToken to true,
  to be more representative of real workloads.

### DIFF
--- a/clusterloader2/testing/pod-startup-latency/config.yaml
+++ b/clusterloader2/testing/pod-startup-latency/config.yaml
@@ -1,0 +1,93 @@
+# ASSUMPTIONS:
+# - This test is designed for 1-node cluster.
+# - Pods take less than 1 second to start, so we can launch them at a rate of 1 pod per second without creating a backlog.
+
+#Constants
+{{$POD_COUNT := DefaultParam .POD_COUNT 30}}
+{{$POD_THROUGHPUT := DefaultParam .POD_THROUGHPUT 1}} # We expect the pod to launch in less than one second, so launching more than 1 pod per second would create a backlog and test a different behaviour.
+{{$CONTAINER_IMAGE := DefaultParam .CONTAINER_IMAGE "registry.k8s.io/pause:3.9"}}
+{{$POD_STARTUP_LATENCY_THRESHOLD := DefaultParam .POD_STARTUP_LATENCY_THRESHOLD "5s"}}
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/pod-startup-latency/config.yaml
```diff
@@ -0,0 +1,92 @@
...
+# - This test is designed for 1-node cluster.
+
+#Constants
+{{$POD_COUNT := DefaultParam .POD_COUNT 30}}
+{{$POD_THROUGHPUT := DefaultParam .POD_THROUGHPUT 1}} # We expect the pod to launch in less than one second, so launching more than 1 pod per second would create a backlog and test a different behaviour.
+{{$CONTAINER_IMAGE := DefaultParam .CONTAINER_IMAGE "registry.k8s.io/pause:3.9"}}
+{{$POD_STARTUP_LATENCY_THRESHOLD := DefaultParam .POD_STARTUP_LATENCY_THRESHOLD "5s"}}
+{{$OPERATION_TIMEOUT := DefaultParam .OPERATION_TIMEOUT "15m"}}
+
+name: node-throughput
```
Qqkyu: This should be updated to `pod-startup-latency`, right?

justinsb: Good point - fixed!

clusterloader2/testing/pod-startup-latency/config.yaml
```diff
@@ -0,0 +1,93 @@
+# ASSUMPTIONS:
+# - This test is designed for 1-node cluster.
+# - Pods take less than 1 second to start, so we can launch them at a rate of 1 pod per second without creating a backlog.
+
+#Constants
+{{$POD_COUNT := DefaultParam .POD_COUNT 30}}
+{{$POD_THROUGHPUT := DefaultParam .POD_THROUGHPUT 1}} # We expect the pod to launch in less than one second, so launching more than 1 pod per second would create a backlog and test a different behaviour.
```
Qqkyu, line 7: Would testing a different behaviour using this test still make sense? If not, we might want to remove this parameter and just hardcode 1 for the `POD_THROUGHPUT`?

Issue #3889, opened 2026-03-14
clusterloader2: create_to_schedule etc are based on second-granularity measurements


justinsb:
**What happened**:

When I run tests, I see e.g. create_to_schedule has zero values:

```
    {
      "data": {
        "Perc50": 0,
        "Perc90": 0,
        "Perc99": 0
      },
      "unit": "ms",
      "labels": {
        "Metric": "create_to_schedule"
      }
    },
```

**What you expected to happen**:

Some measurements that we can work from.

**How to reproduce it (as minimally and precisely as possible)**:

```
kind create cluster
go run ./cmd/clusterloader.go --nodes=1 --provider=kind --report-dir=artifacts/ --testconfig=testing/node-throughput/config.yaml --testoverrides=./testing/overrides/node_containerd.yaml --kubeconfig=${HOME}/.kube/config --v=2
```

**Anything else we need to know?**:

justinsb:
So I believe the issue here is that createPhase is based on metadata.creationTimestamp, and schedulePhase is based on the event timestamp.  Both have per-second granularity.  This means that for small measurements (like scheduling) the values are dominated by randomness around whether we "cross a second".

I think the fix is to record "precise" values based on when we see them in the watch stream.  It's not perfect (there's an extra variable which is watch latency), but we don't have to deal with the second-rounding problem.

I'm happy to send a PR to fix this, just wanted to anchor it in an issue.

isaac-dasan:
Hi @justinsb,
Looks like you moved pod_startup_latency to nanosecond watch-stream timestamps here in https://github.com/kubernetes/perf-tests/pull/3891

`resourceclaim_allocation_latency.go` still mixes second-granularity CreationTimestamp for create/pod_create with nanosecond `recvTime` for allocate - can I fix it?

Gaurav598:
Hi, @justinsb @isaac-dasan I’d like to work on this issue.
I’ve investigated the remaining second-granularity timestamp usage in `resourceclaim_allocation_latency.go` and found that the ResourceClaim creation and Pod creation timestamps are still derived from `CreationTimestamp`, while the allocation path already uses the informer receive time.

I plan to align these paths with the watch-stream timestamp approach used in #3891 and add regression tests covering sub-second precision.
I’ll follow up with a PR once the implementation and tests are ready.

PR #3891, opened 2026-03-14
fix: capture higher-precision metrics for pod_startup_latency

justinsb:
Issue #3889

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -19,17 +19,14 @@ package slos
 import (
 	"context"
 	"fmt"
-	"sort"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
```diff
@@ -205,6 +228,10 @@ var podStartupTransitions = map[string]measurementutil.Transition{
 		From: createPhase,
 		To:   schedulePhase,
 	},
+	"create_to_schedule_precise": {
```
justinsb: Not sure whether we want to just replace the existing metrics or create new ones....

Qqkyu: From what I saw there was leaning towards replacing the existing metric, so let's do that and we'll merge this PR. I'll monitor the tests and revert if anything breaks

justinsb:
Sorry this took so long.  I've updated it to just switch to using the watch stream; agree that seemed to be the consensus, and I also found pkg/measurement/common/slos/resourceclaim_allocation_latency.go which already uses watch events, so I think there's already precedent for using the watch stream (it is simpler).

Qqkyu (commented):
/lgtm
/approve

Only suggestion: can we add some UTs for this? Especially covering the stream-based logic

justinsb:
/test pull-perf-tests-ec2-master-scale-performance-100

Odd TLS issue, I'm _guessing_ it's because I fired off two tests in parallel

justinsb:
@Qqkyu - added some tests.  I actually restored the unit test file that I had previously removed, but now it uses events.  Good idea!

Qqkyu (commented):
/lgtm
/approve

PR #3898, opened 2026-03-17
Adjust check probes timeout for huge clusters

kamilWyszynski1:
This PR adjusts prometheus probes check interval for clusters with 5k or more nodes.

For such clusters, we observed `InClusterNetworkLatency` to be failing due to timeouts in individual probe's checks, what's more even if the response from prometheus was successful, check was time outing during parsing.


kamilWyszynski1:
@marseel could you have a look please?

clusterloader2/pkg/measurement/common/probes/probes.go
```diff
@@ -291,6 +291,10 @@ func (p *probesMeasurement) waitForProbesReady(config *measurement.Config) error
 	if err != nil {
 		return err
 	}
+	checkProbesReadyInterval, err := util.GetDurationOrDefault(config.Params, "checkProbesReadyInterval", defaultCheckProbesReadyInterval)
+	if err != nil {
+		return err
+	}
 	return wait.Poll(
```
marseel: nit: in case you want to clean up deprecated `wait.Poll` and replace it with PollUntilContextTimeout

kamilWyszynski1: Sure, done.

PR #3900, opened 2026-03-18
clusterloader2: add CL2_TOLERATION override to all workload templates

wonderyl:
Allows CL2 to run on tainted nodes.

##### Details
Adds a new CL2_TOLERATION parameter that, when set to a non-empty string, adds a toleration, with operator: Exists and effect: NoSchedule, using that string as the key. This allows directing all workloads to tainted nodes.
This feature unblocks the use of Kwok virtual nodes.

Applies to: Deployment, StatefulSet, DaemonSet, Job (reconcile-objects) and simple-deployment (scheduler-throughput / huge-service).

/kind feature


wonderyl:
@mborsz , @wojtek-t 
I would appreciate if you could take some time to take a look. This is a relatively straight forward PR.

PR #3926, opened 2026-03-25
Add operation timeout buffer variables to load scenario

kamilWyszynski1:
This PR adds operation timeout buffer variables in load scenario. 

Users might want to adjust timeout for each section of `reconcile-objects`, for now without changing number of pods or `LOAD_TEST_THROUGHPUT` it's not possible. Default value is not changed.

### DIFF
--- a/clusterloader2/testing/load/config.yaml
+++ b/clusterloader2/testing/load/config.yaml
@@ -69,6 +69,10 @@
 {{$schedulerThroughputNamespaces := IfThenElse $IS_SMALL_CLUSTER 1 $schedulerThroughputNamespaces}}
 # END scheduler-throughput section
 
+{{$CREATE_OPERATION_TIMEOUT_BUFFER := DefaultParam .CL2_CREATE_OPERATION_TIMEOUT_BUFFER 900}}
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/config.yaml
```diff
@@ -384,7 +388,7 @@ steps:
       operationTimeout: 15m
       {{else}}
       tuningSet: default
-      operationTimeout: {{AddInt $deletionTime 900}}s
+      operationTimeout: {{AddInt $deletionTime $DELETE_OPERATION_TIMEOUT_BUFFER}}s
```
Qqkyu: Do we want to have the `DELETE_OPEARTION_TIMEOUT_BUFFER` only if the `RATE_LIMIT_POD_CREATION` is false? Wouldn't it make sense to add it to both if cases?

Same for the `CREATE_OPERATION_TIMEOUT_BUFFER` and `UPDATE_OPERATION_TIMEOUT_BUFFER`

kamilWyszynski1: Good point, makes sense. Added buffer also in case of `RATE_LIMIT_POD_CREATION` is set.

PR #3931, opened 2026-03-31
watch-list/load-test: wait for the informer pods to be running before starting the load test

p0lyn0mial:
#### What type of PR is this?
/kind bug

While looking into the latency difference between the WatchList ON and the WatchList OFF (standard LIST) for the scalability test ([xref](https://github.com/kubernetes/kubernetes/issues/134375#issuecomment-3992839042))  I realized the comparison isn’t valid because they run under different conditions. 

Both pods (WatchList ON/OFF) are created early in the test, but right after that the test creates a large number of pods without waiting for them to start. Probably this overloads the kubelet on the node. As a result, the OFF pod starts very late (about 71–73 minutes), missing the entire "scale up" phase it’s supposed to measure. The ON pod starts almost immediately and runs during the full "scale up".

This PR adds a step after the informer deployment creation to ensure both pods are running before the "scale up" begins.

An alternative approach would be to start both pods after the "scale up" phase.

Fixes #

Totally untested. Is there a way to test it before merging ?

### DIFF
--- a/clusterloader2/testing/load/modules/informer/config.yaml
+++ b/clusterloader2/testing/load/modules/informer/config.yaml
@@ -33,3 +33,24 @@ steps:
           objectTemplatePath: /modules/informer/deployment.yaml
           templateFillMap:
             EnableWatchListFeature: "true"
+{{if eq $action "create"}}
+- name: Starting measurement for waiting for informer pods to be running
+  measurements:
+  - Method: WaitForControlledPodsRunning
+    Instances:
+    - Identifier: WaitForRunningInformerDeployments
+      Params:
+        apiVersion: apps/v1
+        kind: Deployment
+    Params:
+      action: start
+      labelSelector: group = informer
+      operationTimeout: 2m
+- name: Waiting for informer pods to be running
+  measurements:
+  - Method: WaitForControlledPodsRunning
+    Instances:
+    - Identifier: WaitForRunningInformerDeployments
+    Params:
+      action: gather
+{{end}}


serathius:
The `pull-perf-tests-gce-master-scale-performance-5000` job runs, but it doesn't test the informers :(

p0lyn0mial:
> The pull-perf-tests-gce-master-scale-performance-5000 job runs, but it doesn't test the informers :(

@serathius gave me info on how to update that job. once i do that we should be able to test the changes by running the CI job.

serathius:
```
E0331 12:18:44.201990   23936 wait_for_controlled_pods.go:629] WaitForControlledPodsRunning: error for informer-0/informer-watch-list-off-0: got context deadline exceeded while waiting for 1 pods to be running in namespace(informer-0), controlledBy(informer-watch-list-off-0) - summary of pods : Pods: 1 out of 1 created, 0 running (0 updated), 0 pending scheduled, 1 not scheduled, 0 inactive, 0 terminating, 0 unknown, 0 runningButNotReady 
E0331 12:18:44.202013   23936 wait_for_controlled_pods.go:641] WaitForControlledPodsRunning: informer-0/informer-watch-list-off-0 timed out
E0331 12:18:44.202035   23936 wait_for_controlled_pods.go:372] Timed out Deployments: informer-0/informer-watch-list-off-0
W0331 12:18:44.202116   23936 simple_test_executor.go:186] Got errors during step execution: [measurement call WaitForControlledPodsRunning - WaitForRunningInformerDeployments error: 1 objects timed out: Deployments: informer-0/informer-watch-list-off-0]
```

PR #3935, opened 2026-04-07
Add l4lb-class test scenario

### LINKED ISSUE #3934: [Enhancement] Add ClusterLoader2 test scenario for custom loadBalancerClass
I would like to add a new test scenario in ClusterLoader2 under `testing/l4lb-class/` to support testing of Services with a custom `loadBalancerClass` specified. 

**Why is this needed**:
This is needed to enable scalability testing for the `loadBalancerClass` feature. Currently, we lack a standard test scenario in the OSS `perf-tests` repository to benchmark and validate the performance of the control plane when handling a large number of services utilizing custom load balancer classes.

08volt:
#### What type of PR is this?

/kind feature

This PR adds a new test scenario for `l4lb-class` in ClusterLoader2. It includes configuration for testing LoadBalancers allowing to specify a `loadBalancerClass`, along with corresponding deployment and service templates. 

Fixes #3934 

This test scenario is designed to validate the behavior of custom load balancer classes in large scale clusters.

### DIFF
--- a/clusterloader2/testing/l4lb-class/config.yaml
+++ b/clusterloader2/testing/l4lb-class/config.yaml
@@ -1,0 +1,61 @@
+{{$LARGE_BACKEND_LB_SERVICE_COUNT := DefaultParam .CL2_LARGE_BACKEND_LB_SERVICE_COUNT 2}}
+{{$MEDIUM_BACKEND_LB_SERVICE_COUNT := DefaultParam .CL2_MEDIUM_BACKEND_LB_SERVICE_COUNT 2}}
+{{$SMALL_BACKEND_LB_SERVICE_COUNT := DefaultParam .CL2_SMALL_BACKEND_LB_SERVICE_COUNT 2}}
+{{$lbQPS := DefaultParam .CL2_LB_TEST_QPS 20}}
+{{$LOAD_BALANCER_CLASS := DefaultParam .CL2_LOAD_BALANCER_CLASS "invalid.scalability.gke.io"}}
+
+name: l4lbload
+namespace:
+  number: 1
+tuningSets:
+- name: LBConstantQPS
+  qpsLoad:
+    qps: {{$lbQPS}}
+steps:
+- module:
+    path: /modules/measurements.yaml
+    params:
+      action: start
+- name: Start measurement for running pods
+  measurements:
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: start
+      apiVersion: apps/v1
+      kind: Deployment
+      labelSelector: test = l4lb-load
+      operationTimeout: 15m
+- module:
+    path: /modules/services.yaml
+    params:
+      actionName: Create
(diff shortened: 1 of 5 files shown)


08volt:
Hi @mborsz @wojtek-t could you please have a look? thank you

clusterloader2/testing/l4lb-class/service.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{.Name}}
+  labels:
+    test: l4lb-load
+    size: {{.LBSizeLabel}}
+  annotations:
+    networking.gke.io/load-balancer-type: {{.LoadBalancerType}}
```
Qqkyu, line 9: There is no `LoadBalancerType` passed from the `modules/services.yaml` - is this expected?

clusterloader2/testing/l4lb-class/config.yaml
```diff
@@ -0,0 +1,63 @@
...
+      largeBackendLbServiceCount: 0
+      mediumBackendLbServiceCount: 0
+      smallBackendLbServiceCount: 0
+- name: Waiting for objects deletion to be completed
+  measurements:
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
+- module:
```
Qqkyu, line 60: Don't we need a `waitForDeletion` step before jumping to `action: gather` for `modules/measurements.yaml`? Similar to [here](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/l4lb/config.yaml#L95)

clusterloader2/testing/l4lb-class/config.yaml
```diff
@@ -0,0 +1,63 @@
...
+      actionName: Delete
+      largeBackendLbServiceCount: 0
+      mediumBackendLbServiceCount: 0
+      smallBackendLbServiceCount: 0
+- name: Waiting for objects deletion to be completed
+  measurements:
+  - Identifier: WaitForRunningDeployments
+    Method: WaitForControlledPodsRunning
+    Params:
+      action: gather
```
Qqkyu, line 59: Related to the `waitForDeletion` - the step above deleted the deployments so there will be 0 running pods.  It won't wait for them to finish terminating, so we should exchange this for `waitForDeletion` blocks I think?

clusterloader2/testing/l4lb-class/service.yaml
08volt: I missed removing it from here as we don't need it. thanks

clusterloader2/testing/l4lb-class/config.yaml
08volt: fixed

clusterloader2/testing/l4lb-class/config.yaml
08volt: fixed thanks

PR #3947, opened 2026-04-17
Exemplar pod

serathius:

Taking a small step to avoid too big PR and risky changes.

Added 10 realistic labels and 10 realistic annotations to pod.
Added 2 init containers and 2 containers.
Added environment variables but just for main container to avoid repetition for now.
Resulting pod should be <10KB which is the current size of pod with `CL2_DEPLOYMENT_POD_PAYLOAD_SIZE=6000` in default test.

/cc @jprzychodzen @mborsz @wojtek-t


serathius:
cc @BenTheElder 
```
message: The zone 'projects/k8s-infra-e2e-boskos-scale-09/zones/us-east1-b' does not
  have enough resources available to fulfill the request.  '(resource type:compute)'
```

PR #3976, opened 2026-04-22
Add pagination to collecting events

### LINKED ISSUE #138415: Establish "Pod Shape" as a part of Scalability Envelope
### What would you like to be added?

Following the proposal to introduce Resource Size as a new Scalability Dimension ([#134375](https://github.com/kubernetes/kubernetes/issues/134375)), I would like to propose establishing "Pod Shape" as its own formal scalability envelope.

We should clearly separate the concepts of:

* **Shape:** Which fields are set and their proportions (e.g., number of init containers, sidecars, volumes).
* **Size:** The size of the field values themselves, which can be adjusted independently.

The goal is to align as a community on an "Exemplar Pod Shape" that we guarantee to be scalable. Once aligned, we should formally document this shape envelope in the community configurations and limits ([thresholds.md](https://github.com/kubernetes/community/blob/main/sig-scalability/configs-and-limits/thresholds.md)).

### Why is this needed?

While addressing resource size limits is crucial, [recent analyses](https://github.com/kubernetes/kubernetes/issues/134375#issuecomment-3952087283) have shown that the types of fields set in a Pod spec can have a disproportionate, non-linear impact on resulting Pod size and control plane load. 

Currently, our tests generate a specific total size in etcd by using a "flat" pod shape and stuffing a large random payload into a single environment variable. However, real-world users achieve the same overall pod size by utilizing complex architectural patterns (like sidecars and init containers), which generate drastically different cascading overheads:
* **ManagedFields:** These are automatically generated to track field ownership and [often double the size](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/555-server-side-apply#scalability) of the entire pod specification. It's size depends on number of fields set which is not tested at all. 
* **Init Containers:** Adding just one init container causes the Kubelet to send two additional status updates. This increases the number of writes in a standard pod scheduling cycle by 50% (from 4 to 6 writes), generating significantly more watch events for all controllers and inflating watch cache memory.
* **Sidecars:** Each sidecar container requires its own container status entry, which adds approximately 1KB of overhead due to image digests and sandbox metadata.

# Current Test Shape vs. Intended Exemplar Shape

To illustrate the gap, here is a comparison between the current test pod shape and the intended exemplar pod shape. Both have a total pod size in etcd of roughly 10KB, but achieve it in completely different ways:

| Field                 | Current Test Pod Shape | Proposed Exemplar Pod Shape     |
| --------------------- | ---------------------- | ------------------------------- |
| Annotations           | 0                      | 10                              |
| Labels                | 1                      | 10                              |
| Init Containers       | 0                      | 2                               |
| Containers            | 1                      | 2                               |
| Volumes               | 2                      | 10 per container/init container                             |
| Envs                  | 1                      | 10 per container/init container |
| Env Value             | 6KB                    | Minimal to hit intended pod size|
| Total Pod Size (etcd) | ~10KB                  | ~10KB                           |

## Proposed Plan

1. **Align on the Exemplar Pod Shape:** Finalize the proposed shape outlined above as the target baseline envelope to ensure realistic size overhead and API usage.
2. **Update Scalability tests and Benchmarks** Update K8s scalability tests and benchmarks to reflect exemplar pod.
3. **Fuzzing & Investigation:** Conduct fuzzing on other pod fields to identify any remaining areas with non-linear performance impacts.
2. **Document** Add this defined shape envelope to [sig-scalability/configs-and-limits/thresholds.md](https://github.com/kubernetes/community/blob/main/sig-scalability/configs-and-limits/thresholds.md) to formally guarantee its scalability.

/sig scalability /sig api-machinery
/cc @wojtek-t @mborsz @liggitt

serathius:
Fix https://github.com/kubernetes/kubernetes/issues/138415#issuecomment-4295852866

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/pod_startup_latency.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/pager"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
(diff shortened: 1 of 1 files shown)


Qqkyu (commented):
/lgtm

PR #3978, opened 2026-04-23
Add environment variables to sidecar to increase pod size from 9KB to 10KB to match exactly the non-realistic scenario

serathius:

/cc @Qqkyu 

This time verified the exact output size.
```
$ kubectl proxy &
$ curl -s -H "Accept: application/vnd.kubernetes.protobuf" http://localhost:8001/api/v1/namespaces/test-vzywmm-1/pods/small-deployment-1-57cb6867cd-v46ws | wc -c
10285
```


Qqkyu (commented):
/lgtm

PR #3980, opened 2026-04-24
Add exemplar fields to other workload types

serathius:
With https://github.com/kubernetes/perf-tests/pull/3976 we got a have successful run with realistic pods in deployments. Let's now add them to other workloads types.

image

/cc @Qqkyu


Qqkyu (commented):
/lgtm

PR #3982, opened 2026-04-24
watch-list/load-test: Add option to run informer watch-list pods on control-plane node

p0lyn0mial:
Add `CL2_INFORMER_RUN_ON_CONTROL_PLANE` toggle to create additional watch-list deployments pinned to the control-plane node via nodeSelector and toleration.

I will open a new PR to test-infra which will set `CL2_INFORMER_RUN_ON_CONTROL_PLANE`  for `pull-perf-tests-gce-master-scale-performance-5000` job (later on we will add the new env var for all related jobs).

This ^ won't break anything and will allow us to test this PR.

### DIFF
--- a/clusterloader2/testing/load/modules/informer/deployment.yaml
+++ b/clusterloader2/testing/load/modules/informer/deployment.yaml
@@ -1,4 +1,5 @@
 {{$TolerateControlPlane := DefaultParam .CL2_INFORMER_TOLERATE_CONTROL_PLANE false}}
+{{$RunOnControlPlane := DefaultParam .RunOnControlPlane false}}
 
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/load/modules/informer/deployment.yaml
```diff
@@ -15,11 +16,15 @@ spec:
       labels:
         group: informer
     spec:
-{{if $TolerateControlPlane}}
+{{if or $TolerateControlPlane $RunOnControlPlane}}
```
p0lyn0mial, line 19: `TolerateControlPlane` is unused, we could remove it.

p0lyn0mial, line 19: This PR allows running the watch-list binary both inside and outside the cluster at the same time. Alternatively, we could restrict it to a single option (either inside or outside).

PR #4019, opened 2026-05-08
watchlist binary on the master node connects to the API server via localhost

p0lyn0mial:
earlier we added an option to run watch-list pods on control-plane nodes (#3982).

this pr forces these pods to connect to the kube api over localhost.

this is important when debugging scalability issues where networking is the usual suspect.

### DIFF
--- a/clusterloader2/testing/load/modules/informer/deployment.yaml
+++ b/clusterloader2/testing/load/modules/informer/deployment.yaml
@@ -24,6 +24,7 @@ spec:
           effect: NoSchedule
 {{end}}
 {{if $RunOnControlPlane}}
+      hostNetwork: true
       nodeSelector:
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/modules/informer/deployment.yaml
```diff
@@ -24,6 +24,9 @@ spec:
           effect: NoSchedule
 {{end}}
 {{if $RunOnControlPlane}}
+      # Use the node's network namespace so that 127.0.0.1 reaches the
```
serathius: Drop comments. We still write code for humans not AI, yet

p0lyn0mial: done.

PR #4020, opened 2026-05-10
util-images: pin gcb-docker-gcloud image to current digest

Paramesh324:
#### What type of PR is this?

/kind cleanup

Updates util-images/cloudbuild.yaml to use gcr.io/k8s-staging-test-infra/gcb-docker-gcloud pinned by digest (sha256:ff388e0dc16351e96f8464e2e185b74a7578a5ccb7a112cf3393468e59e6e2d2) with a tag comment (v20260205-38cfa9523f). The previous tag was removed by staging registry retention, which causes image push jobs to fail with manifest unknown; digest pinning also reduces the chance of repeat breakage when old tags are GC’d.

Related : kubernetes/kubernetes#138936

None; single-line image reference change only. Matches the pattern recommended in the issue (digest pin + tag comment).

### DIFF
--- a/util-images/cloudbuild.yaml
+++ b/util-images/cloudbuild.yaml
@@ -6,7 +6,7 @@ options:
   machineType: 'N1_HIGHCPU_8'
 
 steps:
-  - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20210917-12df099d55'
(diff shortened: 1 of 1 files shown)


Qqkyu (commented):
/ok-to-test
/lgtm
/approve

PR #4026, opened 2026-05-11
WIP: debug-watchlist-latency SKILL

p0lyn0mial:
I used AI while debugging https://github.com/kubernetes/kubernetes/issues/138670 and asked it to turn our conversations into a reusable skill. It’s still a work in progress, but the goal is for it to eventually help others troubleshoot WatchList/List latency issues more efficiently.

### DIFF
--- a/.claude/skills/SKILL.md
+++ b/.claude/skills/SKILL.md
@@ -1,0 +1,465 @@
+---
+name: debug-watchlist-latency
+description: >
+  Debug watchlist or list latency from a scale-performance-5000 test run.
+  Triggers: "watchlist latency", "scale-performance-5000",
+  "watch_list_duration_seconds", GCS artifact URLs from perf-tests
+  scale-performance jobs, watch cache performance, goroutine scheduling
+  contention, APF throttling during scale tests.
+argument-hint: "<gcs-artifact-url>"
+arguments: [url]
+allowed-tools: Bash Read WebFetch
+---
+
+# Debug Watchlist Latency at Scale
+
+Systematically debug WATCHLIST and LIST latency from a `scale-performance-5000` test run.
+Given a GCS artifact URL, walk through watchlist pod logs, Prometheus metrics, apiserver
+logs, and pprof profiles to identify the root cause of latency regressions.
+
+Reference issue: https://github.com/kubernetes/kubernetes/issues/138670
+
+The test creates a real 5000-node cluster and runs approximately 165K pods (this number
+may change across test revisions — check the actual test configuration for the current
+value). The high pod count drives significant watch event churn, especially during
+scale-up and scale-down phases, which is where latency issues typically surface.
+
+## Locate Test Artifacts
+
+Resolve the GCS URL `$url` into browsable artifact paths. Use `gsutil ls` to enumerate:
+
+```
+$url/
+  artifacts/
+    <master-name>/                    # e.g. ci-master-scale-master
+      kube-apiserver-audit-logs/      # audit logs per apiserver instance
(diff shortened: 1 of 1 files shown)


serathius:
High level, debugging scalability requires multiple sub-skills like downloading logs, metrics etc. Let's split this into skills focused on single task. Also this is not a general skill for whole repo. Would be good to put relevant skills into relevant directories. 

/cc @AwesomePatrol

.claude/skills/SKILL.md
```diff
@@ -0,0 +1,465 @@
...
+**Signature**: `apiserver_watch_events_sizes` skewed toward large buckets.
+CPU profile shows hot path in `Encode` or `CacheEncode`.
+
+**Mechanism**: Large objects (ConfigMaps, Secrets, CRDs) cause high per-event serialization
+cost, multiplied by watcher count. `CachingObject` avoids re-serialization across watchers
+using the same format, but mixed protobuf/JSON clients negate this.
+
+**Next steps**: Identify large objects. Consider server-side field selection or object trim.
+
+## Metrics Reference
```
serathius, line 324: Maybe we could just reference https://github.com/kubernetes/kubernetes/blob/b6b2ab9ded421bba720e86fc17e8146074e84287/hack/tools/instrumentation/testdata/stable-metrics-list.yaml ?

.claude/skills/SKILL.md
```diff
@@ -0,0 +1,465 @@
+---
+name: debug-watchlist-latency
+description: >
+  Debug watchlist or list latency from a scale-performance-5000 test run.
```
serathius, line 4: This scenario is specific to `./clusterloader/testing/load` . Don't think we should put it in top directory. It should be moved there, we can can reference it from here.

serathius: Or if we want to keep it in `skills` directory. Have a dedicated skill called `load-test`.

mengqiy: +1
I'm leaning towards the latter.

.claude/skills/SKILL.md
```diff
@@ -0,0 +1,465 @@
...
+| `go_sched_goroutines_goroutines` | Gauge | Label `state` = runnable, running |
+| `go_sched_latencies_seconds` | Histogram | Go scheduler latency, high P99 = scheduling delay |
+| `go_sync_mutex_wait_total_seconds_total` | Counter | Total time waiting on mutexes |
+| `process_cpu_seconds_total` | Counter | Rate gives CPU utilization |
+
+## Code Pointers
+
+| Component | File | Key locations |
+|-----------|------|---------------|
+| Event dispatch | `staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go` | `dispatchEvents()`:886, `dispatchEvent()`:978 |
```
mengqiy, line 373: Do we have to provide line numbers here? It can be fragile when the line number is out of date.

mengqiy (commented):
It's a good move to start check in SKILLs.
Some general sub-skills (e.g. fetching logs, prom snapshot, profiling data) can be split and reused later by other debugging skills. e.g. scheduler throughput

Jefftree:
I wonder if we're over-indexing on making these skills generic enough to cover all analysis cases. The mechanical parts (like download CI logs) are the kind of thing an agent can figure out on its own. With most harnesses that have memory, they'll converge on the optimal path eventually. (not against making this part faster, but imo has lower impact on the analysis).

My 2c is that the non generic part is where the value is in this skill, because we can repeat this analysis chain in a consistent manner across runs, and its hard to keep that consistent without a skill.

PR #4027, opened 2026-05-12
add watch cache initialization metric to restart test

Jefftree:
Verified via testing with default=on.

Metric sample: https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/perf-tests/4027/pull-perf-tests-clusterloader2/2054337704006520832/artifacts/GenericPrometheusQuery%20Watch%20Cache%20Initialization%20Duration_load_2026-05-12T23:24:53Z.json

Uses metric added in https://github.com/kubernetes/kubernetes/pull/138767

### DIFF
--- a/clusterloader2/testing/load/modules/restart-apiserver.yaml
+++ b/clusterloader2/testing/load/modules/restart-apiserver.yaml
@@ -20,6 +20,23 @@ steps:
     Params:
       action: pauseSystemPodMetrics
 
+- name: Start WatchCacheInitializationDuration measurement
+  measurements:
+  - Identifier: WatchCacheInitializationDuration
+    Method: GenericPrometheusQuery
+    Params:
+      action: start
+      metricName: Watch Cache Initialization Duration
+      metricVersion: v1
+      unit: s
+      dimensions:
+      - resource
+      queries:
+      - name: Avg
+        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
+      - name: Count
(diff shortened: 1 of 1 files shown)


clusterloader2/testing/load/modules/restart-apiserver.yaml
```diff
@@ -20,6 +20,23 @@ steps:
...
+      action: start
+      metricName: Watch Cache Initialization Duration
+      metricVersion: v1
+      unit: s
+      dimensions:
+      - resource
+      queries:
+      - name: Avg
+        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
+      - name: Count
```
Jefftree, line 37: Not strictly necessary but verifies a watch cache isn't flapping and we only have one init event.

clusterloader2/testing/load/modules/restart-apiserver.yaml
```diff
@@ -20,6 +20,23 @@ steps:
...
+    Params:
+      action: start
+      metricName: Watch Cache Initialization Duration
+      metricVersion: v1
+      unit: s
+      dimensions:
+      - resource
+      queries:
+      - name: Avg
+        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
```
Jefftree, line 36: it's possible (though unlikely) watch cache isn't initialized for certain resources. This prevents metrics from breaking (divide by 0) if initialization is 0.

PR #4030, opened 2026-05-14
perfdash: grey out label dropdown values with no data

Jefftree:
#### What type of PR is this?

/kind feature

As someone new to using perfdash, I found it really confusing that we have a whole matrix of dropdown options with most of them being noop because there is no data populated.

Want to propose a qol improvement to gray out the the options that have no data and disable clicking them.

Before: 

After:

I tested this locally with just the benchmark proto list data.


Jefftree:
/cc @serathius @mborsz 

It seems like no one's touched this code for years. Is this something valuable that we can land?

wojtek-t:
This is awesome :)

/lgtm
/approve

serathius:
> This is awesome :)

Can't agree more.

PR #4057, opened 2026-05-18
clusterloader2: add opt-in periodic heap profiling

Jefftree:
Adds an opt-in `--heap-profile-interval` (`CL2_HEAP_PROFILE_INTERVAL`) flag to periodically write clusterloader2 heap profiles to the report dir for diagnosing runner-pod OOMs (kubernetes/test-infra#37056).

Test run with CL2 profiles: https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/perf-tests/4057/pull-perf-tests-gce-master-scale-performance-100/2056475433968340992/artifacts/

### DIFF
--- a/clusterloader2/pkg/heapprofile/heapprofile.go
+++ b/clusterloader2/pkg/heapprofile/heapprofile.go
@@ -1,0 +1,59 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package heapprofile periodically dumps the clusterloader2 process heap
+// profile into the report dir for diagnosing runner-pod OOMs. Profiles follow
+// the same <name>_<RFC3339>.pprof naming convention as the KCP profiles
+// collected by pkg/measurement/common/profile.go, so existing artifact tooling
+// picks them up the same way.
+package heapprofile
+
+import (
+	"os"
+	"path"
+	"runtime/pprof"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+// Start launches periodic heap profiling.
+func Start(reportDir string, interval time.Duration) {
+	klog.Infof("Heap profiling enabled: writing to %q every %s", reportDir, interval)
+	go func() {
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for range ticker.C {
+			writeHeapProfile(reportDir, "periodic")
+		}
+	}()
+}
+
+func writeHeapProfile(reportDir, tag string) {
+	ts := time.Now().Format(time.RFC3339)
+	filePath := path.Join(reportDir, "ClusterLoader2_MemoryProfile_"+tag+"_"+ts+".pprof")
+	f, err := os.Create(filePath)
+	if err != nil {
+		klog.Errorf("Cannot create heap profile %q: %v", filePath, err)
+		return
+	}
+	defer f.Close()
+	if err := pprof.Lookup("heap").WriteTo(f, 0); err != nil {
+		klog.Errorf("Cannot write heap profile %q: %v", filePath, err)
+		return
+	}
+	klog.V(2).Infof("Wrote heap profile %q", filePath)
+}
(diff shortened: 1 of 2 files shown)


clusterloader2/cmd/heapprofile.go
```diff
@@ -0,0 +1,80 @@
...
+	if reportDir == "" {
+		klog.Warningf("heap-profile-interval is set but --report-dir is empty; heap profiling disabled")
+		return
+	}
+	dir := path.Join(reportDir, "pprof")
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		klog.Errorf("Cannot create heap profile directory %q: %v; heap profiling disabled", dir, err)
+		return
+	}
+
```
serathius: Use normal function

serathius (changes requested):

clusterloader2/cmd/heapprofile.go
```diff
@@ -0,0 +1,80 @@
...
+	dump := func(tag string) {
+		ts := time.Now().UTC().Format(time.RFC3339)
+		name := path.Join(dir, fmt.Sprintf("cl2-heap-%s-%s.pb.gz", tag, ts))
```
serathius: We already collect profiles from KCP, can we reuse the code/path construction?

clusterloader2/cmd/heapprofile.go
Jefftree: Fixed.

clusterloader2/cmd/heapprofile.go
Jefftree: It's a bit difficult to refactor because it operates on the [test scope](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/test/simple_test_executor.go#L95-L111). Going with the approach of following the same naming convention `ClusterLoader2_MemoryProfile_*` instead of sharing the exact code path.

Jefftree:
@serathius Confirmed we obtain the profile in https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/perf-tests/4057/pull-perf-tests-gce-master-scale-performance-100/2056475433968340992/artifacts/

clusterloader2/pkg/heapprofile/heapprofile.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}()
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		s := <-c
+		klog.V(2).Infof("Received signal: %v", s)
+		writeHeapProfile(reportDir, "sigterm")
+		os.Exit(1)
```
serathius: How adding a signal handler impacts the process shutdown? I'm also worried about having signal handler in some random subpackage instead of in main. Process lifecycle management should not be spread out the codebase.

Jefftree: Duplicated it from https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/experimental.go#L259-L269 which does something similar.

I'm going to remove this since we increased the memory and won't be exercising this path for a while.

clusterloader2/pkg/heapprofile/heapprofile.go
```diff
@@ -0,0 +1,67 @@
...
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for range ticker.C {
+			writeHeapProfile(reportDir, "periodic")
+		}
+	}()
+}
+
+func writeHeapProfile(reportDir, tag string) {
+	var buf bytes.Buffer
```
serathius: Do we need to store profile in memory? Could we just pipe it directly to avoid bloating memory further? If needed we could use temporary file and move.

clusterloader2/pkg/heapprofile/heapprofile.go
```diff
@@ -0,0 +1,67 @@
...
+	"path"
+	"runtime/pprof"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+// Start launches periodic heap profiling. It is a no-op when interval <= 0.
+func Start(reportDir string, interval time.Duration) {
+	if interval <= 0 {
```
serathius: I think we should move this check to main, the call `heapprofile.Start(clusterLoaderConfig.ReportDir, heapProfileInterval)` seems like it always collect profile.

serathius: We can just panic/error if the interval is incorrect.

Jefftree:
```
 if heapProfileInterval > 0 {
         heapprofile.Start(clusterLoaderConfig.ReportDir, heapProfileInterval)
 }
```

Moved the validation to main as well since it seems quite confusing for this function to exit on config error.

PR #4059, opened 2026-05-18
request-benchmark: add table and json-pretty content types

Jefftree:
Adds table (Accept: application/json;as=Table;v=v1;g=meta.k8s.io) and json-pretty (Accept: application/json + ?pretty=true) content types so LIST benchmarks can cover them. Part of kubernetes/kubernetes#130169.

### DIFF
--- a/util-images/request-benchmark/main.go
+++ b/util-images/request-benchmark/main.go
@@ -42,10 +42,14 @@ const (
 type ContentType string
 
 const (
-	JSONContentType  ContentType = "json"
-	ProtoContentType ContentType = "proto"
-	CBORContentType  ContentType = "cbor"
-	YAMLContentType  ContentType = "yaml"
+	JSONContentType       ContentType = "json"
+	ProtoContentType      ContentType = "proto"
+	CBORContentType       ContentType = "cbor"
+	YAMLContentType       ContentType = "yaml"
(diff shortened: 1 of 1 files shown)


Jefftree:
Curosity more than anything on how these perform.

/cc @serathius

util-images/request-benchmark/main.go
```diff
@@ -86,6 +90,12 @@ func getContentType(ct ContentType) (string, error) {
 		return "application/cbor", nil
 	case YAMLContentType:
 		return "application/yaml", nil
+	case TableContentType:
+		return tableAccept, nil
+	case JSONPrettyContentType:
+		// Pretty-printing is controlled by the ?pretty=true query
+		// parameter. The response is still application/json.
+		return "application/json", nil
```
serathius:
```suggestion
case JSONContentType, JSONPrettyContentType:
		// Pretty-printing is controlled by the ?pretty=true query
		// parameter. The response is still application/json.
return "application/json", nil
case ProtoContentType:
return "application/vnd.kubernetes.protobuf", nil
case CBORContentType:
return "application/cbor", nil
case YAMLContentType:
return "application/yaml", nil
	case TableContentType:
		return tableAccept, nil
```

Jefftree: Done

PR #4062, opened 2026-05-20
agents/skills: add download-ci-artifacts skill

p0lyn0mial:
Add a general purpose `download-ci-artifacts` skill to `.agents/skills`.

The skill is a dumb downloader. No opinions about what's useful, no hardcoded output path, no test specific logic. It accepts a URI, downloads what the user asks for, and reports what it got.

The domain knowledge about what to download (which artifacts matter) will live in a separate "explore-ci-artifacts" skill or in a coordinator.

Tested with Claude by
1. adding a `.claude/skills` symlink pointing to `.agents/skills`
2. typing `/download-ci-artifacts Download junit.xml from https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/138327/pull-kubernetes-gce-master-scale-performance-5000/2056034689277235200` in the prompt. 
3. got:
```
❯ /download-ci-artifacts Download junit.xml from https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/138327/pull-kubernetes-gce-master-scale-performance-5000/2056034689277235200

⏺ Downloading junit.xml from the specified Prow job.

⏺ Bash(mkdir -p /Users/lukasz/go/src/k8s.io/perf-tests/2056034689277235200)
  ⎿  Done

⏺ Bash(gcloud storage cp 'gs://kubernetes-ci-logs/pr-logs/pull/138327/pull-kubernetes-gce-master-scale-performance-5000/2056034689277235200/**/junit*.xml' /Users/luka…)
  ⎿  Copying gs://kubernetes-ci-logs/pr-logs/pull/138327/pull-kubernetes-gce-master-scale-performance-5000/2056034689277235200/artifacts/junit_runner.xml to file:///Users/lukasz/go/src/k8s.io/perf-tests/2056034689277235200
     /junit_runner.xml

     .....

  Listed 1 directory (ctrl+o to expand)

⏺ Downloaded 1 file to 2056034689277235200/:

  ┌──────────────────┬───────┐
  │       File       │ Size  │
  ├──────────────────┼───────┤
  │ junit_runner.xml │ 585 B │
  └──────────────────┴───────┘
  ```

### DIFF
--- a/.agents/skills/download-ci-artifacts/SKILL.md
+++ b/.agents/skills/download-ci-artifacts/SKILL.md
@@ -1,0 +1,21 @@
+---
+name: download-ci-artifacts
+description: Download artifacts from a CI run's GCS bucket to local disk.
+---
+# Task
+Download artifacts from a CI run's GCS bucket to a local directory. Accept a URI and a list of files or patterns to download. Do not decide what to download. Download only what the user requests.
+
+# Workflow
+## 1. Parse URI
+- **URI normalization**: Accept gcsweb URLs (`https://gcsweb.k8s.io/gcs/...`), raw `gs://` paths, or Prow job URLs (`https://prow.k8s.io/view/gs/...`).
+  Convert to GCS paths by replacing `https://gcsweb.k8s.io/gcs/` or `https://prow.k8s.io/view/gs/` with `gs://`.
+
+## 2. Create Local Directory
+- **Output path**: If an output directory is provided, use it. Otherwise create a local directory using an identifier extracted from the URI (e.g., the Prow build ID).
+
+## 3. Download
+- **Targeted download**: Download only the files or patterns the user specified (e.g., `kube-apiserver.log*`, `*.json`, `junit.xml`).
+- **Parallel transfer**: Use `gcloud storage cp` for downloads.
+
+## 4. Report
+- **Summary**: List the files downloaded and their sizes.

### CODE REVIEW

File: clusterloader2/testing/load/skills/download-ci-logs.md [hunk 1]
```diff
@@ -0,0 +1,34 @@
+# Download CI Logs
```
Jefftree: Have you tested this compared to an actual `SKILL.md`? I tried the approach described by adding a note to reference this directory and it consistently takes a lot more turns to reach the file (part of it could be due to `Skills are located in clusterloader2/testing/load/skills/` not being a good trigger). If we want to frame this as an actual skill rather than general documentation, I think it makes more sense to follow the skill format and location.

p0lyn0mial: Hey, thanks for the comment! Based on the feedback from [#4026](https://github.com/kubernetes/perf-tests/pull/4026) the skills in `clusterloader2/testing/load/skills/` should not be tied to any specific AI assistant like Claude or Codex. That’s why we don’t want to put files in `.claude/skills/`

For the same reason we can't use Claude specific formatter. But we could adopt a more generic structured format that hopefully any AI assistant would understand. I actually slightly change the format see if you like it.

Try putting the following  `"When the user asks to download CI logs or provides a GCS/gcsweb/Prow URL, follow the instructions in clusterloader2/testing/load/skills/download-ci-logs.md"` in the CLAUDE.md. 
Hopefully it takes now less turns to reach the skill file.

serathius: Based on my research we should put skills in `.agents/skills/` directory. Within directories like this, we can still reference the skills by name.

Jefftree: [Gemini](https://geminicli.com/docs/cli/skills/), [Codex](https://developers.openai.com/codex/skills) use `.agents/skills` (and a few other harnesses). [Claude](https://code.claude.com/docs/en/skills) seems fixated on `.claude/skills`. The convention I've seen across repos is have `.agents/skills` as source of truth and `.claude/skills` symlink to it.

serathius: Having .agents + symlink sounds good to me.

p0lyn0mial: ok, putting skills under `.agents/skills/` at the repo root makes them auto-discovered by harnesses. When skills live in a shared location, consistency matters. I think that we should then enforce the format of skills proposed to this repo. Do you agree ?

We could enforce this in CI with a simple linter script, as a manual review step, or both. We could even write the linter itself as a skill. A `skill-lint/SKILL.md` that validates other skills' structure.

WDYT ?

serathius: Let's not overthink it. It's just markdown file, we can change it later. We were just discussing location.

p0lyn0mial: My research shows that consistency is what makes skills composable. 

For example, a coordinator (our `debug-scalability-test` skill) could fan out to 6 skills in parallel and merge their results. If 5 return `{ "status": "ok", "findings": [...] }` and the 6th returns a free text paragraph then the merge breaks. The same applies to inputs. Each skill needs to declare what it expects (REPORT_DIR, PROMETHEUS_URL, etc.) so the coordinator knows what to pass. 

Consistent format means every skill has predictable input and output contracts. The coordinator can compose them all the same way, without needing custom handling for each one.

serathius: We are not building a rocket. We can still change it after merging.

PR #4064, opened 2026-05-22
cl2: strip managedFields from pod informers to cut memory

Jefftree:
Strip `.metadata.managedFields` from CL2's pod informers, matching kube-apiserver/KCM/scheduler.


Qqkyu (commented):
/lgtm
/approve

PR #4067, opened 2026-05-26
Bump probes image tag

Tanzp01:
#### What type of PR is this?
/kind cleanup
/kind feature

The tag hasn't been bumped for some time, thus does not include some changes such as change from https://github.com/kubernetes/perf-tests/pull/3630

Fixes that the previous PR #3630 was not released with a tag


Qqkyu (commented):
/lgtm
/approve

PR #4068, opened 2026-05-27
CL2: Optimize memory usage for pods

mborsz:
#### What type of PR is this?
Attempt to optimize memory usage for CL2: The current state:
* Most of the allocations are done by temporary allocations done by runtime.DefaultUnstructuredConverter.FromUnstructured in getIsPodUpdatedPodPredicateFromUnstructured
* Most of the inuse_space is pods' spec.

We cannot simply drop pod spec as getIsPodUpdatedPodPredicateFromUnstructured requires it, but we can drop it for pods that are not managed by us (e.g. in non-managed namespaces).

Two optimizations:
* Drop pod.Spec for pods in non-managed namespaces (the ones where we never isPodUpdated), e.g. daemonsets that can be massive for large clusters -- hope to help with inuse memory
* Cache DefaultUnstructuredConverter.FromUnstructured results to avoid constant calling in a hot loop. This aims to reduce an alloc rate and gc pressure.

/kind feature

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
@@ -19,6 +19,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -65,6 +66,8 @@ const (
 
 var podIndexerFactory = &sharedPodIndexerFactory{}
 
+var automanagedNamespaceID = regexp.MustCompile(`^test-[a-z0-9]+-[0-9]+$`)
+
 func init() {
 	if err := measurement.Register(waitForControlledPodsRunningName, createWaitForControlledPodsRunningMeasurement); err != nil {
 		klog.Fatalf("Cannot register %s: %v", waitForControlledPodsRunningName, err)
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -86,7 +89,26 @@ func (s *sharedPodIndexerFactory) PodsIndexer(c clientset.Interface) (*measureme
...
-	informerFactory := informers.NewSharedInformerFactoryWithOptions(c, 0, informers.WithTransform(informer.TrimManagedFields))
+	trimFn := func(obj interface{}) (interface{}, error) {
+		if accessor, err := meta.Accessor(obj); err == nil && accessor.GetManagedFields() != nil {
+			accessor.SetManagedFields(nil)
+		}
+		pod, ok := obj.(*v1.Pod)
+		if !ok {
+			return obj, nil
+		}
+		if !automanagedNamespaceID.MatchString(pod.Namespace) {
```
serathius: Why we drop fields for automanaged? Why not all. Don't know the motivation, but I'm worried about having a mix of pods objects with different fields. At least leave a comment.

mborsz: Changed code to more readable form and added comment.

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -100,6 +102,11 @@ func (lsde *lazySpecDiffError) Error() string {
 	return fmt.Sprintf("Not matching templates, diff: %v", gocmp.Diff(lsde.templateSpec, lsde.podSpec))
 }
 
+type podCacheKey struct {
+	uid types.UID
+	rv  string
+}
+
 func getIsPodUpdatedPodPredicateFromUnstructured(obj *unstructured.Unstructured) (func(_ *corev1.Pod) error, error) {
```
serathius, line 110: Why do we even use dynamic client for pods? That's seems ultra inefficient.

clusterloader2/pkg/measurement/common/wait_for_controlled_pods.go
```diff
@@ -86,7 +89,29 @@ func (s *sharedPodIndexerFactory) PodsIndexer(c clientset.Interface) (*measureme
...
+	trimFn := func(obj interface{}) (interface{}, error) {
+		if accessor, err := meta.Accessor(obj); err == nil && accessor.GetManagedFields() != nil {
+			accessor.SetManagedFields(nil)
+		}
+		pod, ok := obj.(*v1.Pod)
+		if !ok {
+			return obj, nil
+		}
+		if automanagedNamespaceID.MatchString(pod.Namespace) {
+			// For pods managed by the clusterloader, we need to keep the full pod.Spec for checkIfPodsAreUpdated feature.
```
serathius, line 101: What's the percentage of clusterloader pods vs other?  5-10k vs 140k?

mborsz: Depends on the cluster size. in huge clusters, this can be like 50% (daemonsets).

clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
mborsz: This function is not for pods, it is for controllers -- CL2 supports arbitrary type of controllers: statefulsets, daemonsets, some random CRDs and stuff. This function takes as a parameter the controller object (unstrctured) and converts it to the typed PodTemplateStruct to be able to compare with typed podStruct.

PR #4071, opened 2026-05-28
CL2: Partially revert 4068

mborsz:
Breaks namespaceList feature which can watch pods in  non-managed namespaces.


serathius:
Would be great to prepare a unit test to prevent such regressions in the future.

PR #4072, opened 2026-05-29
fix(request-benchmark): use resilient Content-Type parsing to ignore charset

Jefftree:
Fixes a CI timeout failure by making `request-benchmark`'s `Content-Type` validation resilient to optional parameters like `charset=utf-8`.

In GCE CI runs, intermediate network proxies or the Go runtime can append `; charset=utf-8` to JSON response headers. The previous strict string comparison failed in these cases, causing the tool to panic and crash-loop during its startup health check.

This change uses `mime.ParseMediaType` to validate only the core media type, ignoring extra parameters.

Ref: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-benchmark-list-json-pretty/2060011128372596736

Gemini figured this out, I've verified the issue in the logs, but didn't find exactly how the charset gets added. I believe it's through the proxy.

### DIFF
--- a/util-images/request-benchmark/main.go
+++ b/util-images/request-benchmark/main.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"mime"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


Jefftree:
ping @serathius

PR #4078, opened 2026-06-03
clusterloader2: make PROMETHEUS_SCRAPE_APISERVER_ONLY overridable

dims:
#### What type of PR is this?
/kind feature

Adds `--prometheus-scrape-apiserver-only` / `PROMETHEUS_SCRAPE_APISERVER_ONLY` to override the per-provider `Features().ShouldPrometheusScrapeApiserverOnly` default. Empty keeps the provider default (no change for existing jobs); kops self-managed AWS jobs can set `false` to also scrape kube-scheduler and kube-controller-manager, which the `aws` provider otherwise drops (leaving `PrometheusSchedulingMetrics` empty, unlike GCE).

Default is preserved for all providers; a companion test-infra change opts the EC2 DRA jobs in.

### DIFF
--- a/clusterloader2/pkg/config/cluster.go
+++ b/clusterloader2/pkg/config/cluster.go
@@ -76,6 +76,7 @@ type ModifierConfig struct {
 
 // PrometheusConfig represents all flags used by prometheus.
 type PrometheusConfig struct {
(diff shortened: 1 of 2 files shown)


dims:
thanks @shyamjvs !

PR #4079, opened 2026-06-03
cl2/restart-apiserver: report watch cache init duration per apiserver instance

Jefftree:
/kind cleanup

Group watch cache init duration by `(resource, instance)` so the restarted apiserver is reported separately instead of averaged across all instances. Otherwise it reports a single blended number (54.36 ms) instead of the three below.

([prow run](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/139476/pull-kubernetes-gce-master-scale-performance-100-apiserver-restart/2062247362662764544)), numbers for watch cache init for pods:
- `10.128.0.102`: 0.53 ms
- `10.128.0.104`: 0.65 ms
- `10.128.0.105`: 161.89 ms (restarted)

### DIFF
--- a/clusterloader2/testing/load/modules/restart-apiserver.yaml
+++ b/clusterloader2/testing/load/modules/restart-apiserver.yaml
@@ -31,11 +31,12 @@ steps:
       unit: s
       dimensions:
       - resource
+      - instance
       queries:
-      - name: Avg
-        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
-      - name: Count
-        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count)
+      - name: InitDuration
+        query: sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
+      - name: InitCount
+        query: sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count)
 
 - name: Restart kube-apiserver and wait for /readyz
   measurements:


clusterloader2/testing/load/modules/restart-apiserver.yaml
```diff
@@ -31,11 +31,12 @@ steps:
...
       dimensions:
       - resource
+      - instance
       queries:
-      - name: Avg
-        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
-      - name: Count
-        query: sum by (resource) (apiserver_watch_cache_initialization_duration_seconds_count)
+      - name: InitDuration
+        query: sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
```
serathius, line 37: Shouldn't we just take max?

serathius: max_over_time(histogram_quantile(1, rate(apiserver_watch_cache_initialization_duration_seconds_bucket[1m]))[1h:])

serathius: To test it we could get a prometheus snapshot from 5k test and load it into P&G stack

Jefftree: We would only look for the max but I think having the additional data gives us more info and helps us debug things in case the numbers are super off. (eg: What if watch cache initialization happened twice due to delay).

Also, we want to sum because histogram only reports the bucket, which is a lot less precise than having the exact value.

serathius: Not sure you can integrate it into perfdash that way.

Jefftree: Ah wasn't really thinking about perfdash. I'll rethink this a bit.

serathius: That's ok, we shouldn't be blocking on perfdash support, however it's always good to have in mind.

PR #4080, opened 2026-06-03
cl2: scrape etcd pprof over the insecure client-http port

Jefftree:
#### What type of PR is this?
/kind bug

kubernetes/kops#18370 moved etcd /debug/pprof to a separate client-http listener, so scrape it on ETCD_PPROF_PORT (default 2385) instead of 2379.

Otherwise we don't get pprof data (eg: https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-gce-scale-performance-5000/2062218224480555008/artifacts/)

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

### DIFF
--- a/clusterloader2/pkg/measurement/common/profile.go
+++ b/clusterloader2/pkg/measurement/common/profile.go
@@ -232,9 +232,7 @@ func (p *profileMeasurement) getProfileCommand(config *measurement.Config) (stri
 
 	var command string
 	if p.config.componentName == "etcd" {
-		etcdCert := config.ClusterFramework.GetClusterConfig().EtcdCertificatePath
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/profile.go
```diff
@@ -232,9 +232,7 @@ func (p *profileMeasurement) getProfileCommand(config *measurement.Config) (stri
 
 	var command string
 	if p.config.componentName == "etcd" {
-		etcdCert := config.ClusterFramework.GetClusterConfig().EtcdCertificatePath
```
Jefftree, line 235: cert not necessary for insecure port

PR #4081, opened 2026-06-03
dra: bump testing/dra config timeouts for 5k scale

dims:
#### What type of PR is this?
/kind bug

Fixes `WaitForFinishedJobs` undercounting at scale. When churn jobs are deleted by `ttlSecondsAfterFinished` before the measurement gathers, `handleObject` read completion from `newJob` (nil on delete) and dropped the key, so the finished count never reached the target and the measurement timed out at 5k. Use `handleJob` (the surviving object) and keep already-finished entries on delete. Also bumps the DRA churn `WaitForFinishedJobs` timeout 15m→30m.

Lands the fix from `alaypatel07:dra-scalability-bump-timeout` so the GCE/EC2 5k DRA jobs can use `kubernetes/perf-tests@master` instead of the fork branch.

### DIFF
--- a/clusterloader2/pkg/measurement/common/wait_for_jobs.go
+++ b/clusterloader2/pkg/measurement/common/wait_for_jobs.go
@@ -217,13 +217,14 @@ func (w *waitForFinishedJobsMeasurement) handleObject(oldObj, newObj interface{}
 		klog.Errorf("Failed obtaining meta key for Job: %v", err)
 		return
 	}
-	completed, condition := finishedJobCondition(newJob)
+	completed, condition := finishedJobCondition(handleJob)
 
 	w.lock.Lock()
 	defer w.lock.Unlock()
 	if completed {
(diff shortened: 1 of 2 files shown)


dims:
This lands the DRA 5k timeout bumps in perf-tests master, replacing the fork branch the GCE 5k DRA periodic (ci-kubernetes-e2e-kops-gce-5000-node-dra-with-workload-ipalias-using-cl2) currently points at for these same timeouts: https://github.com/alaypatel07/perf-tests/tree/dra-scalability-bump-timeout  Once this merges, that job's perf-tests extra_ref can switch back to kubernetes/perf-tests@master.

alaypatel07 (approved):
Changes make sense to me. 

@dims Please remove hold when ready 

/lgtm
/hold

alaypatel07:
Ohh well we still need the approval tag so hold is not necessary 

/hold cancel

PR #4082, opened 2026-06-05
cl2/load: add max watch cache init duration metric

Jefftree:
/kind feature

Add a max watch cache initialization duration per resource measurement to the apiserver-restart load module. xref https://github.com/kubernetes/perf-tests/pull/4079#discussion_r3353731298

Using a new metric because I'd like the keep the old data as debugging data in case we see weird numbers.

### DIFF
--- a/clusterloader2/testing/load/modules/restart-apiserver.yaml
+++ b/clusterloader2/testing/load/modules/restart-apiserver.yaml
@@ -37,6 +37,18 @@ steps:
         query: sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count) > 0)
       - name: InitCount
         query: sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count)
+  - Identifier: WatchCacheInitializationDurationMax
+    Method: GenericPrometheusQuery
+    Params:
+      action: start
+      metricName: Watch Cache Initialization Duration Max
+      metricVersion: v1
+      unit: s
+      dimensions:
+      - resource
+      queries:
+      - name: InitDuration
+        query: max by (resource) (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count) > 0))
 
 - name: Restart kube-apiserver and wait for /readyz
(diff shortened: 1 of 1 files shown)


perfdash/config.go
```diff
@@ -215,6 +215,11 @@ var (
 				OutputFilePrefix: "MetricsForE2E",
 				Parser:           parseApiserverInitEventsCount,
 			}},
+			"LoadWatchCacheInitializationDurationMax": []TestDescription{{
```
Jefftree: not looking forward to figuring out how to deploy perfdash

Jefftree:
/cc @serathius 

Haven't fully tested the perfdash part but this will be the general shape with your max suggestion.

clusterloader2/testing/load/modules/restart-apiserver.yaml
```diff
@@ -37,6 +37,18 @@ steps:
...
+    Params:
+      action: start
+      metricName: Watch Cache Initialization Duration Max
+      metricVersion: v1
+      unit: s
+      dimensions:
+      - resource
+      queries:
+      - name: InitDuration
+        query: max by (resource) (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_sum) / (sum by (resource, instance) (apiserver_watch_cache_initialization_duration_seconds_count) > 0))
```
serathius, line 51: `>0` ?

perfdash/config.go
serathius: cc @kishen-v

clusterloader2/testing/load/modules/restart-apiserver.yaml
Jefftree: See https://github.com/kubernetes/perf-tests/pull/4027#discussion_r3230815853

perfdash/config.go
kishen-v: Hey @Jefftree, 
To test the changes locally, I followed the guide from [here](https://github.com/kubernetes/perf-tests/tree/master/perfdash#local-development) and modified [this file](https://github.com/kubernetes/perf-tests/blob/master/perfdash/test/jobs.yaml) to point to the job whose data required parsing.

serathius:
/hold

For manual tests

Jefftree:
Splitting out the perfdash change to https://github.com/kubernetes/perf-tests/pull/4090 to be able to test this with an actual prow job. I've verified locally already

PR #4089, opened 2026-06-08
Do not log prometheus response for failed targets check

kamilWyszynski1:
For huge clusters, reading and logging the Prometheus response on the `api/v1/targets` call in the `CheckTargetsReady function` might cause memory issues, because prometheus dumps whole configuration during this call. It also bloats logs, making them unreadable or too large for some tools to consume or display.


Qqkyu (commented):
/lgtm
/approve

PR #4091, opened 2026-06-09
dra tests: revert higher SLO times for churn job

alaypatel07:
#### What type of PR is this?

/kind cleanup

Recent 5k node job shows the churn pod latency is well under 5s.

This commit reverts the higher slo times that were put in place just for debugging.

Fixes #


alaypatel07:
cc @dims @pohly

PR #4094, opened 2026-06-09
clusterloader2: limit kubelet scrape cardinality and record Prometheus RSS

dims:
At 5000 nodes the DRA jobs can OOM Prometheus: the kubelet `/metrics` endpoint ingests per-node `go_*/process_*/rest_client_*/workqueue_*/apiserver_*` series across every node, none of which the DRA measurements query (they use `dra_*` and `scheduler_*`).

- Add an **opt-in** `CL2_PROMETHEUS_KUBELET_KEEP_ONLY_DRA` that keeps only `dra_*|up` on the kubelet `/metrics` endpoint (default off, so other kubelet-scraping uses are unaffected).
- cAdvisor scraping is **unchanged** — the existing `test-*` drop already removes workload-pod series, and control-plane container metrics are kept.
- Add a `PrometheusMemory` calibration measurement (peak RSS, peak TSDB head series, ingested samples by job) to size the memory factor from data.

#### What type of PR is this?
/kind failing-test

Bounds DRA kubelet-scrape cardinality (opt-in) and records Prometheus memory/cardinality so the memory factor can be set from data.

NONE

### DIFF
--- a/clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubelet.yaml
+++ b/clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubelet.yaml
@@ -1,4 +1,5 @@
 {{$PROMETHEUS_SCRAPE_KUBELETS := DefaultParam .PROMETHEUS_SCRAPE_KUBELETS false}}
+{{$KEEP_ONLY_DRA := DefaultParam .CL2_PROMETHEUS_KUBELET_KEEP_ONLY_DRA false}}
 
 {{if $PROMETHEUS_SCRAPE_KUBELETS}}
 apiVersion: monitoring.coreos.com/v1
@@ -19,6 +20,16 @@ spec:
     bearerTokenSecret:
       name: prometheus-token
       key: token
+    {{if $KEEP_ONLY_DRA}}
+    # DRA scale tests only query the kubelet DRA-plugin metrics (dra_*). Keeping just those
+    # drops the per-node go_*/process_*/rest_client_*/workqueue_*/apiserver_* series
+    # (x node count), which bounds Prometheus memory at 5k+ nodes. DRA-specific: enabled
+    # only via CL2_PROMETHEUS_KUBELET_KEEP_ONLY_DRA so other kubelet-scraping uses are unaffected.
+    metricRelabelings:
+      - sourceLabels: [__name__]
+        regex: 'dra_.*|up'
+        action: keep
+    {{end}}
   - port: https-metrics
     scheme: https
     path: /metrics/cadvisor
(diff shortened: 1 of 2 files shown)


clusterloader2/testing/dra/config.yaml
```diff
@@ -100,6 +100,19 @@ steps:
             query: histogram_quantile(0.99, sum(rate(dra_grpc_operations_duration_seconds_bucket{method_name=~".*NodePrepareResources"}[%v])) by (le))
           - name: p99_dra_grpc_node_unprepare_resources
             query: histogram_quantile(0.99, sum(rate(dra_grpc_operations_duration_seconds_bucket{method_name=~".*NodeUnprepareResources"}[%v])) by (le))
+    # Calibration: peak Prometheus RSS over the whole run. Uses prometheus self-scrape
+    # (not cAdvisor), so it survives the cardinality-limit relabeling. Lets us right-size
+    # CL2_PROMETHEUS_KUBELET_MEMORY_SCALE_FACTOR from data instead of guessing.
+    - Identifier: PrometheusMemory
```
alaypatel07, line 106: +1

alaypatel07 (approved):
/lgtm

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubelet.yaml
```diff
@@ -19,6 +19,13 @@ spec:
     bearerTokenSecret:
       name: prometheus-token
       key: token
+    # Keep only the kubelet /metrics series CL2 queries (DRA plugin op & gRPC latencies).
+    # Drops the per-node go_*/process_*/rest_client_*/workqueue_*/apiserver_* series
+    # (x node count), which are never queried, to bound Prometheus memory at scale.
+    metricRelabelings:
+      - sourceLabels: [__name__]
+        regex: 'dra_.*|up'
```
serathius, line 30: dra specific config should be under DRA specific feature flag.

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubelet.yaml
```diff
@@ -29,9 +36,11 @@ spec:
     bearerTokenSecret:
       name: prometheus-token
       key: token
+    # cAdvisor container series are not queried by any CL2 measurement; drop them all.
```
serathius: I'm not convinced. Majority of metrics should be already dropped by dropping `test-*` namespace. What else is there? 

I expect KCP and built-in daemonsets. We should keep at least the KCP.

clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubelet.yaml
dims: Keep-list now behind a DRA-specific flag (CL2_PROMETHEUS_KUBELET_KEEP_ONLY_DRA, default off) — generic kubelet scraping is untouched.

PR #4097, opened 2026-06-11
perfdash: make /www traversable by nonroot (use 0755)

dims:
COPY --chmod=0644 leaves the /www directory without the execute bit. perfdash runs as nonroot (distroless:nonroot, uid 65532), and a nonroot process cannot traverse a directory lacking the execute bit, so every static asset under /www returns 403 and the dashboard fails to load. Use 0755 so the directory is traversable.

#### What type of PR is this?

/kind cleanup

Fixes #


Qqkyu (commented):
/lgtm
/approve

PR #4099, opened 2026-06-11
perfdash: build for linux/amd64 by default (override with PLATFORM)

dims:
#### What type of PR is this?

/kind cleanup

The perfdash image build was not pinned to an architecture: `make container` built for the host platform and `make push` used the deprecated `gcloud docker ... push`. Building from a non-amd64 host (for example an arm64 laptop) therefore produced an arm64 image, which cannot run on the amd64 nodes the perfdash deployment lands on.

This pins the image platform explicitly and defaults it to `linux/amd64` via a `PLATFORM` variable, applied to **both** the `container` (build) and `push` targets, and replaces `gcloud docker push` with `docker push`. Another architecture can be built on demand:

    make push PLATFORM=linux/arm64

### DIFF
--- a/perfdash/Makefile
+++ b/perfdash/Makefile
@@ -6,6 +6,9 @@ TAG?=2.59
 WWW_VERSION=$(shell cat www/VERSION)
 
 REPO?=gcr.io/k8s-staging-perf-tests
+# Image platform. Defaults to linux/amd64 (perfdash runs on amd64 nodes); override
+# to build another arch, e.g. make push PLATFORM=linux/arm64
+PLATFORM?=linux/amd64
 
 .PHONY: test
 test: WWW_VERSION = $(shell cat www/VERSION)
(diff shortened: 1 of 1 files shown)


perfdash/Makefile
```diff
@@ -35,9 +36,14 @@ run: perfdash
...
+# Build and push a multi-arch (amd64+arm64) manifest straight to the registry.
+# buildx can only --push a multi-platform image (not --load it locally) and needs
+# a container-driver builder, so create a dedicated one if it is missing.
+# --provenance=false keeps the manifest list to just the arch images (no attestations).
 .PHONY: push
-push: container
-	gcloud docker -s $(REPO) -- push $(REPO)/perfdash:$(TAG)
+push:
+	docker buildx inspect perfdash-multiarch >/dev/null 2>&1 || docker buildx create --name perfdash-multiarch --driver docker-container
+	docker buildx build --builder perfdash-multiarch --provenance=false --pull --platform $(PLATFORMS) -t $(REPO)/perfdash:$(TAG) --push .
```
serathius: push target is for release, don't people that released perfdash before reported issue with arch. Do we really need it?

I would prefer to keep things simple. KISS, buildx is not simple.

dims: understood. i used this for local testing as well. will make it single arch, but can flip the linux/arm64 if someone needs it.

PR #4100, opened 2026-06-11
.agents/skills: add cl2-artifacts-reference guide

p0lyn0mial:
#### What type of PR is this?
/kind documentation

Add a reference guide for ClusterLoader2 CI artifacts as a skill.
Note that is not a skill per se. It is rather documentation that describes the artifact directory structure.

Starts with two sections but more sections can be added in the future PRs.

Fixes #

### DIFF
--- a/.agents/skills/cl2-artifacts-reference/SKILL.md
+++ b/.agents/skills/cl2-artifacts-reference/SKILL.md
@@ -1,0 +1,22 @@
+---
+name: cl2-artifacts-reference
+description: Reference guide for ClusterLoader2 CI artifacts and their GCS paths.
+---
+# Goal
+Describe the artifacts produced by a ClusterLoader2 scalability test run.
+Use this reference to understand what each file is and where it lives
+in the artifact tree.
+
+# Reference
+## 1. Artifact Directory Structure
+- **Root**: `artifacts/` under the CI job run's output path.
+- **Subdirectories**:
+  - `control-plane-*/` — control-plane node logs.
+  - `nodes-*/` — a subset of worker node logs (not all nodes).
+  - `cluster-info/` — resource YAML dumps (cluster-scoped at root, namespaced under per-namespace subdirectories).
+  - `addons-*/` — addons node logs.
+  - `otlp/` — OpenTelemetry traces.
+
+## 2. Test Results
+- `junit.xml` — CL2 test results with pass/fail per step and measurement violation messages.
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
@serathius @dims @Jefftree please let me know your thoughts on this PR.

.agents/skills/cl2-artifacts-reference/SKILL.md
```diff
@@ -0,0 +1,22 @@
+---
+name: cl2-artifacts-reference
+description: Reference guide for ClusterLoader2 CI artifacts and their GCS paths.
```
p0lyn0mial, line 3: This could be the place where the domain knowledge is defined.

.agents/skills/cl2-artifacts-reference/SKILL.md
```diff
@@ -0,0 +1,22 @@
...
+---
+# Goal
+Describe the artifacts produced by a ClusterLoader2 scalability test run.
+Use this reference to understand what each file is and where it lives
+in the artifact tree.
+
+# Reference
```
p0lyn0mial, line 10: Maybe we should be more specific about what files can be found in the directory structure.

.agents/skills/cl2-artifacts-reference/SKILL.md
serathius: That's doesn't sound like skill rather reference

serathius: https://agentskills.io/specification#file-references

.agents/skills/cl2-artifacts-reference/SKILL.md
```diff
@@ -0,0 +1,22 @@
...
+## 1. Artifact Directory Structure
+- **Root**: `artifacts/` under the CI job run's output path.
+- **Subdirectories**:
+  - `control-plane-*/` — control-plane node logs.
```
Jefftree, line 14: Realistically, would an agent really need these kinds of descriptions to navigate? (this is a question, not asking you to change)

.agents/skills/cl2-artifacts-reference/SKILL.md
Jefftree: +1, I feel that a lot of these descriptions are evident from the name. The specific nuanced things are the worthwhile things that an agent can't easily derive from directory name alone.

.agents/skills/cl2-artifacts-reference/SKILL.md
Jefftree: +1

PR #4104, opened 2026-06-14
perfdash: preserve pod name prefixes during normalization

immanuwell:
#### What type of PR is this?
/kind bug

perfdash can strip the whole pod prefix when the first segment looks like a version or hash.

before:
`v2-controller-6d4f5d4b7f-zx9pl/controller` -> `/controller`
`cdbf-agent-6d4f5d4b7f-zx9pl/agent` -> `/agent`

that's a real edge case, not just theory. Kubernetes object names can start with alphanumeric chars, and controllers append generated suffixes to valid prefixes, so names like `v2-controller-...` are legal and reachable.

this change only starts stripping version or hash looking pieces after we already kept one stable prefix segment. existing cases like `kube-dns-74dbf45884-7gkmp/dnsmasq` still normalize the same.

None

repro:
`go test -run TestRemoveDisambiguationInfixes -v`

new test cases cover the broken outputs above. full `go test ./...` also passes in `perfdash`.

### DIFF
--- a/perfdash/pod_names.go
+++ b/perfdash/pod_names.go
@@ -31,7 +31,8 @@ func RemoveDisambiguationInfixes(podAndContainer string) string {
 	pieces := strings.Split(pod, "-")
 	var last string
 	for i, piece := range pieces {
-		if looksLikeHash(piece) || versionRegexp.MatchString(piece) {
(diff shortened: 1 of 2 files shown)


Qqkyu (commented):
/ok-to-test
/lgtm
/approve

PR #4141, opened 2026-06-16
prometheus-operator: make memory request scale with number of nodes

supahero1:
#### What type of PR is this?

/kind bug

For a 10k cluster, this component was OOMing. It's weird that it requests a static amount of memory that's not scaled with number of nodes above 500 or 1k.

`100 + x/2` is a rough estimate that more or less fits with what was stated before.


clusterloader2/pkg/prometheus/manifests/0prometheus-operator-deployment.yaml
```diff
@@ -46,10 +46,10 @@ spec:
         resources:
           limits:
             cpu: 200m
-            memory: {{IfThenElse (lt .Nodes 500) 200 400}}Mi
+            memory: {{AddInt 100 (DivideInt .Nodes 2)}}Mi
```
mborsz: High level the memory usage of prometheus-operator shouldn't depend on the cluster size, rather on number of prometheus instances it runs.

I'm afraid that if we merge it this way, we will break some tests by requesting a lot of memory for this pod and making it pending.

Maybe we can just add an override, keep the default behavior as it was and give us a way to set this value arbitrarily based on our specific test conditions?

supahero1: Indeed, changed

supahero1:
Seems like prometheus-opeartor version used in CL2 is outdated and a newer version allows for a selector that doesn't worry about all the things going on the cluster except Prometheus related things. Going to make a PR for that instead of this.

PR #4143, opened 2026-06-16
Add optional Pods LIST benchmark scenario

adityaRaj369:
#### What type of PR is this?

/kind feature

This adds optional LIST benchmark coverage for Pods, following the existing ConfigMaps benchmark structure.

The Pods benchmark is gated behind `CL2_LIST_POD_NUMBER`, so existing LIST benchmark jobs keep their current behavior by default.

This avoids YAML/Table/json-pretty content-type benchmark work because those jobs are being removed in kubernetes/test-infra#37268.

Addresses part of kubernetes/kubernetes#130169

I could not run the Go test suite locally because Go is not installed on my machine. I did run `git diff --check`.

### DIFF
--- a/clusterloader2/testing/list/README.md
+++ b/clusterloader2/testing/list/README.md
@@ -4,11 +4,14 @@ List load test perform the following steps:
 - Create configmaps.
   - The namespaces used here are managed by CL2.
   - The size and number of configmaps can be specified using `CL2_LIST_CONFIG_MAP_BYTES` and `CL2_LIST_CONFIG_MAP_NUMBER`.
+- Optionally create pods.
+  - The number of pods can be specified using `CL2_LIST_POD_NUMBER`.
+  - The pod list benchmark is skipped by default unless `CL2_LIST_POD_NUMBER` is greater than 0.
 - Create RBAC rules to allow lister pods to access these configmaps
-- Create lister pods using deployment in a separate namespace to list configmaps and secrets.
+- Create lister pods using deployment in a separate namespace to list configured resources.
   - The namespace is created using `namespace.yaml` as a template, but its lifecycle is managed by CL2.
   - The number of replicas for the lister pods can be specified using `CL2_LIST_BENCHMARK_PODS`.
-- Measurement uses [`APIResponsivenessPrometheusSimple`](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/README.md) to meansure API latency for list configmaps and secrets calls.
+- Measurement uses [`APIResponsivenessPrometheusSimple`](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/README.md) to measure API latency for list calls.
 
 The lister pods leverage https://github.com/kubernetes/perf-tests/tree/master/util-images/request-benchmark to create in-cluster list load.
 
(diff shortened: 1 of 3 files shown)


adityaRaj369:
CLA is signed now. Could a Kubernetes member please run /ok-to-test when convenient?

serathius:
/assign @Jefftree 
Would be good to use https://github.com/kubernetes/kubernetes/issues/138415

Jefftree:
+1, the `pause` pod is too simple to measure realistic per-pod list cost. Could we base `pod.yaml` on the exemplar shape pinned behind `CL2_REALISTIC_POD`: https://github.com/kubernetes/perf-tests/blob/8a76880121c78a7683bbe05234bf08bc268af886/clusterloader2/testing/load/modules/reconcile-objects/deployment.yaml#L40-L168?

adityaRaj369:
@Jefftree 
Updated pod.yaml to use the exemplar pod shape when CL2_REALISTIC_POD is enabled, while keeping the simple pause pod as the default. I also left the configmap/secret volume references optional so the list benchmark can create these Pods without requiring per-Pod ConfigMap/Secret objects.

adityaRaj369:
@Jefftree gentle ping — CI is green, and the CL2_REALISTIC_POD update from your earlier feedback is included. Please let me know if there is anything else I should adjust to move this toward review.

clusterloader2/testing/list/README.md
```diff
@@ -4,11 +4,15 @@ List load test perform the following steps:
 - Create configmaps.
   - The namespaces used here are managed by CL2.
   - The size and number of configmaps can be specified using `CL2_LIST_CONFIG_MAP_BYTES` and `CL2_LIST_CONFIG_MAP_NUMBER`.
+- Optionally create pods.
+  - The number of pods can be specified using `CL2_LIST_POD_NUMBER`.
+  - The pod list benchmark is skipped by default unless `CL2_LIST_POD_NUMBER` is greater than 0.
+  - The exemplar pod shape can be enabled using `CL2_REALISTIC_POD`.
```
Jefftree: no need to make a toggle, always realistic pod

clusterloader2/testing/list/config.yaml
```diff
@@ -56,6 +58,40 @@ steps:
...
+      max: 1
+    tuningSet: Sequence
+    replicasPerNamespace: {{$podNumber}}
+    objectBundle:
+    - basename: {{$podGroup}}
+      objectTemplatePath: pod.yaml
+      templateFillMap:
+        group: {{$podGroup}}
+
+- name: Waiting for pods to be running
```
Jefftree: no need to wait, we're just testing the apiserver.

clusterloader2/testing/list/config.yaml
```diff
@@ -83,3 +119,10 @@ steps:
     params:
       namePrefix: "list-configmaps-"
       replicas: 0
+{{if $podNumber}}
```
Jefftree: hm, configmaps and pod tests running concurrently is not ideal. can we do one after the other?

adityaRaj369:
@Jefftree Updated. Pods now always use the realistic pod shape, the pod readiness wait was removed, and the configmap/pod list benchmarks now run sequentially instead of concurrently.

clusterloader2/testing/list/config.yaml
```diff
@@ -83,3 +85,46 @@ steps:
     params:
       namePrefix: "list-configmaps-"
       replicas: 0
+
+{{if $podNumber}}
```
Jefftree, line 89: Actually, let's clean up and delete configmaps before starting the pod test.

Jefftree (commented):
Looks great, once you address the last comment, respond in a rhyme will LGTM

clusterloader2/testing/list/config.yaml
adityaRaj369: ConfigMaps now clear away,
before the Pod LIST gets its day.

PR #4146, opened 2026-06-17
cl2: allow multiple objects in one file

supahero1:
#### What type of PR is this?

/kind feature

Enables multiple yamls in one yaml. I'm not planning or saying anybody should be rewriting existing configs to switch to this, but I think it's nice to have in some cases (especially when the objects are just a couple of lines). QoL.


mborsz:
/ok-to-test
/lgtm
/approve

Thank you! Lack of this was quite annoying.

PR #4150, opened 2026-06-19
Customise OOM tracker to not fail on OOMs

emerbe:

This PR introduces a new clusterOOMsFailureEnabled parameter (defaulting to true) to the ClusterOOMsTracker measurement. When set to false, the measurement will log a warning instead of returning an error when OOMs are detected.

Why we need it: Currently, any OOM event that is not explicitly ignored via the exact-match clusterOOMsIgnoredProcesses list causes the entire test to fail. This makes it difficult to globally track and record OOMs for observability without destabilizing CI jobs. This flag allows test owners to run the measurement in a "record-only" mode, ensuring all OOMs are still properly tracked and exported to the ClusterOOMsTracker.json artifact without explicitly failing the Prow job.

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/common/ooms_tracker.go
+++ b/clusterloader2/pkg/measurement/common/ooms_tracker.go
@@ -42,6 +42,7 @@ const (
 	clusterOOMsTrackerEnabledParamName   = "clusterOOMsTrackerEnabled"
 	clusterOOMsTrackerName               = "ClusterOOMsTracker"
 	clusterOOMsIgnoredProcessesParamName = "clusterOOMsIgnoredProcesses"
+	clusterOOMsFailureEnabledParamName   = "clusterOOMsFailureEnabled"
 	informerTimeout                      = time.Minute
 	oomEventReason                       = "OOMKilling"
 	initialListPageSize                  = 10000
@@ -107,7 +108,7 @@ func (m *clusterOOMsTrackerMeasurement) Execute(config *measurement.Config) ([]m
 	case "gather":
 		m.lock.Lock()
 		defer m.lock.Unlock()
-		return m.gather()
+		return m.gather(config)
 	default:
 		return nil, fmt.Errorf("unknown action %v", action)
 	}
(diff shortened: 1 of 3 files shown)


clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -42,6 +42,7 @@ const (
 	clusterOOMsTrackerEnabledParamName   = "clusterOOMsTrackerEnabled"
 	clusterOOMsTrackerName               = "ClusterOOMsTracker"
 	clusterOOMsIgnoredProcessesParamName = "clusterOOMsIgnoredProcesses"
+	clusterOOMsFailureEnabledParamName   = "clusterOOMsFailureEnabled"
```
Qqkyu, line 45: Can we update the https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/docs/oom_detection.md to document this new parameter?

clusterloader2/pkg/measurement/common/ooms_tracker.go
```diff
@@ -229,7 +230,7 @@ func (m *clusterOOMsTrackerMeasurement) stop() {
 	}
 }
 
-func (m *clusterOOMsTrackerMeasurement) gather() ([]measurement.Summary, error) {
+func (m *clusterOOMsTrackerMeasurement) gather(config *measurement.Config) ([]measurement.Summary, error) {
```
Qqkyu, line 233: Consider adding some basic UTs to `ooms_tracker_test.go` to cover behaviour when ooms are recorded and failure is enabled/disabled

Qqkyu (commented):
/ok-to-test

clusterloader2/pkg/measurement/common/ooms_tracker.go
emerbe: Done

clusterloader2/pkg/measurement/common/ooms_tracker.go
emerbe: Done

PR #4151, opened 2026-06-20
Shrink exemplar pod: drop sidecar env padding added in #3978

Jefftree:
/kind cleanup

Reduces the load test exemplar pod from 10KB back to 9KB, undoing the sidecar env padding added in #3978. The goal is to stabilize 5k scale tests.

### DIFF
--- a/clusterloader2/testing/load/modules/reconcile-objects/deployment.yaml
+++ b/clusterloader2/testing/load/modules/reconcile-objects/deployment.yaml
@@ -127,41 +127,6 @@ spec:
           name: secret
       - name: sidecar
         image: registry.k8s.io/pause:3.9
-        env:
-        - name: POD_NAME
-          valueFrom:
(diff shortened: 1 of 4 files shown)


clusterloader2/testing/load/modules/reconcile-objects/deployment.yaml
serathius, line 1: What about other other controllers? DeamonSet, Statefulset etc

Jefftree: whoops, done for daemonset, job, statefulset.

serathius:
Ke?

alvaroaleman:
/lgtm

(repeating the lgtm from https://github.com/kubernetes/perf-tests/pull/4151#issuecomment-4758635024 to see if adding it works now)

PR #4158, opened 2026-06-24
Provide NodeHealthTracker measurement for CL2

marcinus:
#### What type of PR is this?

/kind feature

This PR introduces a Node Health tracker measurement. This measurement subscribes (opens WATCH) to the Node resources in Kubernetes. When the number of unhealthy nodes goes above `max(4, 1% of all nodes)`, then a message printed to the log (at most once a minute) and the test is marked as failed during `gather` phase.

### DIFF
--- a/clusterloader2/pkg/measurement/common/node_health_tracker.go
+++ b/clusterloader2/pkg/measurement/common/node_health_tracker.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


clusterloader2/pkg/measurement/common/node_health_tracker.go
```diff
@@ -0,0 +1,243 @@
...
+		m.checkThresholdAndLog()
+	}
+}
+
+func (m *nodeHealthTrackerMeasurement) checkThresholdAndLog() {
+	if m.nodeCount == 0 {
+		return
+	}
+	unhealthyNodes := m.nodeCount - m.runningNodes
+	threshold := math.Max(4, float64(m.nodeCount)*0.01)
```
mborsz: 4 and 0.01 should be configurable. The defaults seem reasonable

PR #4159, opened 2026-06-24
Support label selector for WaitForGenericK8sObjects

4xo0:
/kind feature

Adds labelSelector support to WaitForGenericK8sObjects.

Tests:
- go test ./pkg/measurement/util
- go test ./pkg/measurement/common -run 'TestGetNamespaces_AllNamespaces|TestGetLabelSelector'

### DIFF
--- a/clusterloader2/pkg/measurement/util/wait_for_conditions.go
+++ b/clusterloader2/pkg/measurement/util/wait_for_conditions.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


clusterloader2/pkg/measurement/util/wait_for_conditions.go
mborsz: Can we actually use labelSelector on the watch level? This would greatly reduce a pressure on the informer and watch itself in some scenarios?

4xo0: Done.

mborsz:
/ok-to-test
/lgtm
/approve

Thank you!

PR #4160, opened 2026-06-25
Add custom resource LIST benchmark scenario

yuminn-k:
#### What type of PR is this?

/kind feature
/sig api-machinery

Adds an optional Custom Resource LIST scenario to the existing ClusterLoader2 `testing/list` benchmark.

The scenario is disabled by default and can be enabled with `CL2_LIST_CUSTOM_RESOURCE_NUMBER`. When enabled, it:

- creates a namespaced `ListBenchmarkResource` CRD for the benchmark,
- uses a larger, Kueue-like nested CRD schema and custom resource shape,
- creates benchmark custom resources with configurable payload size,
- starts a request-benchmark deployment that LISTs the custom resource endpoint,
- runs the custom-resource LIST benchmark as a separate sequential benchmark segment,
- cleans up the custom resources and CRD after the benchmark.

Part of kubernetes/kubernetes#130169.

This focuses on the custom-resource slice of kubernetes/kubernetes#130169 and is rebased on top of the Pods LIST benchmark that is now in `master`.

AI disclosure: This PR was prepared in part with the assistance of an AI coding assistant. I reviewed the generated changes, understand the code being submitted, and verified it with the tests listed below.

Validation:

- `go test ./api ./pkg/config ./pkg/test` from `clusterloader2` passes.
- `git diff --check` passes.
- Dry-run config rendering was checked for the default path, the CR-enabled path, and the CR+Pods-enabled path.
  - The default rendered config contains no custom-resource or pod LIST steps.
  - `CL2_LIST_CUSTOM_RESOURCE_NUMBER=1` renders the CRD, custom-resource, custom-resource lister, and cleanup steps.
  - `CL2_LIST_CUSTOM_RESOURCE_NUMBER=1 CL2_LIST_POD_NUMBER=1` renders the custom-resource segment followed by the pod segment.
- The rendered CRD and a sample rendered custom resource pass `kubectl apply --dry-run=server` against a local kind apiserver.

### DIFF
--- a/clusterloader2/testing/list/README.md
+++ b/clusterloader2/testing/list/README.md
@@ -1,14 +1,17 @@
 # List load test
 
-List load test perform the following steps:
-- Create configmaps.
+List load test performs the following steps:
+- Create configured resources.
   - The namespaces used here are managed by CL2.
-  - The size and number of configmaps can be specified using `CL2_LIST_CONFIG_MAP_BYTES` and `CL2_LIST_CONFIG_MAP_NUMBER`.
-- Optionally create pods.
+  - Configmaps are always created. The size and number of configmaps can be specified using `CL2_LIST_CONFIG_MAP_BYTES` and `CL2_LIST_CONFIG_MAP_NUMBER`.
+  - The number of custom resources can be specified using `CL2_LIST_CUSTOM_RESOURCE_NUMBER`.
+  - The size of each custom resource payload can be specified using `CL2_LIST_CUSTOM_RESOURCE_BYTES`.
+  - The custom resource list benchmark is skipped by default unless `CL2_LIST_CUSTOM_RESOURCE_NUMBER` is greater than 0.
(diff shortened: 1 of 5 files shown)


Jefftree:
request: Can we make the CRD larger, like much larger. We'd like to be able to catch things like https://github.com/kubernetes/kubernetes/issues/136950#issuecomment-3889610473. Create something similar to that size

yuminn-k:
Thanks for the pointer. I rebased the PR on current master and updated the custom-resource benchmark to use a larger, Kueue-like nested CRD schema and custom resource shape instead of the previous single payload string.

I also moved the custom-resource LIST load into its own sequential benchmark segment so it does not run mixed with the configmap segment.

Validation after the update:
- `go test ./api ./pkg/config ./pkg/test` from `clusterloader2`
- `git diff --check`
- dry-run config rendering for default, CR-enabled, and CR+Pods-enabled paths
- server-side dry-run of the rendered CRD and a sample rendered custom resource against a local kind apiserver

PR #4161, opened 2026-06-25
clusterloader2: collect etcd-events pprof profiles in TestMetrics

Jefftree:
#### What type of PR is this?
/kind feature

TestMetrics only profiles the main etcd. This also collects CPU/heap/block pprof from etcd-events, via a new --etcd-events-pprof-port flag (default 2386). The etcd-events pprof endpoint is already exposed by the kops scalability scenario: https://github.com/kubernetes/kops/blob/7a0fb7e7c2672507c73a89113e7eeba532cdd35b/tests/e2e/scenarios/scalability/run-test.sh#L90-L92

/assign @serathius


Jefftree:
Tested. etcd events pprof show up here now: https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/perf-tests/4161/pull-perf-tests-gce-master-scale-performance-100/2070228371333189632/artifacts/

PR #4164, opened 2026-06-29
clusterloader2/testing/watch-list: increase QPS to 2000 and informer count to 2

p0lyn0mial:
it will speed up https://github.com/kubernetes/test-infra/pull/37346

### DIFF
--- a/clusterloader2/testing/watch-list/job.yaml
+++ b/clusterloader2/testing/watch-list/job.yaml
@@ -21,5 +21,5 @@ spec:
               memory: "16Gi"
               cpu: "6"
           command: [ "watch-list" ]
-          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=1", "--resource=pods", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
+          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=2", "--resource=pods", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
       restartPolicy: Never
--- a/clusterloader2/testing/watch-list/config.yaml
+++ b/clusterloader2/testing/watch-list/config.yaml
@@ -9,9 +9,9 @@ namespace:
   number: {{$namespaces}}
   prefix: "watch-list"
 tuningSets:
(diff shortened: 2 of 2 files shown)


clusterloader2/testing/watch-list/job.yaml
```diff
@@ -21,5 +21,5 @@ spec:
               memory: "16Gi"
               cpu: "6"
           command: [ "watch-list" ]
-          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=1", "--resource=pods", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
+          args: [ "--alsologtostderr=true", "--v=4", "--timeout={{.Duration}}", "--count=2", "--resource=pods", "--enableWatchListFeature={{.EnableWatchListFeature}}"]
```
p0lyn0mial, line 24: I think that `bench list` runs 10 concurrent clients. 

I will update to 10 clients in the future too.

serathius: Don't think we do. Don't understand the changes here. Did we get at least 1 successful run?

p0lyn0mial: > Don't understand the changes here

QPS controls the rate of creating the pods.
I think the relevant code is at https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/tuningset/simple_tuning_set_factory.go#L54

> Did we get at least 1 successful run?

yes, i tested many times on a remote machine and it worked. I collected the following results:

Screenshot 2026-06-29 at 16 36 06

p0lyn0mial: I wrapped the CL2 binary with /usr/bin/time to measure memory usage.  the peak mem usage reported in the above table. it seems that setting `k8s-clients-number` helped to decrease the mem usage.

p0lyn0mial: also, I think that the benchmark list runs 10 replicas oof the request benchmark pod

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-scalability/sig-scalability-periodic-jobs.yaml#L1013

PR #4168, opened 2026-06-30
WaitForGenericK8sObjects: add support for empty conditions

supahero1:
#### What type of PR is this?

/kind feature

Adds support for simply waiting for the given pods to become created, without anything else.


clusterloader2/pkg/measurement/common/wait_for_generic_k8s_object.go
```diff
@@ -70,17 +70,25 @@ func (w *waitForGenericK8sObjectsMeasurement) Execute(config *measurement.Config
 	}
 	successfulConditions, err := util.GetStringArray(config.Params, "successfulConditions")
```
mborsz: Let's add util.GetStringArrayDefault helper and use it here and below

supahero1: Done

PR #4169, opened 2026-06-30
Do not log HTTP status code in case of error

marcinus:
#### What type of PR is this?

/kind cleanup

Due to a bug in the HTTP client library, the HTTP status code is not returned correctly and should not be logged in case of an error.


clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -112,12 +112,14 @@ func (a *apiAvailabilityMeasurement) pollHost(hostIP string) (string, error) {
...
 	status := 0
-	result.StatusCode(&status)
-	availability := status == http.StatusOK
-	if !availability {
-		klog.Warningf("cluster not available; HTTP status code: %d", status)
-		if err := result.Error(); err != nil {
-			klog.Warningf("request error: %v", err)
+	availability := false
+	if err := result.Error(); err != nil {
+		klog.Warningf("cluster probably not available; request error: %v", err)
```
mborsz: Let's consider error as "not available", drop "probably" :)

PR #4173, opened 2026-07-02
Exclude shutdown ready check on ApiAvailability

marcinus:
#### What type of PR is this?

/kind feature

This PR excludes the shutdown healthcheck from APIAvailability measurement to accept cases when we hit the "old" VM during upgrade which is shutting down. The rationale behind it is that VMs have graceful degradation, meaning that the crucial components are being shut down at the latest stage, enabling the VM to respond to requests properly even during shutdown.

### DIFF
--- a/clusterloader2/pkg/measurement/common/api_availability_measurement.go
+++ b/clusterloader2/pkg/measurement/common/api_availability_measurement.go
@@ -110,6 +110,12 @@ func (a *apiAvailabilityMeasurement) pollHost(hostIP string) (string, error) {
 }
 
 func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/measurement/common/api_availability_measurement.go
```diff
@@ -110,7 +110,7 @@ func (a *apiAvailabilityMeasurement) pollHost(hostIP string) (string, error) {
 }
 
 func (a *apiAvailabilityMeasurement) updateClusterAvailabilityMetrics(c clientset.Interface) {
-	result := c.CoreV1().RESTClient().Get().AbsPath("/readyz").Do(context.Background())
+	result := c.CoreV1().RESTClient().Get().AbsPath("/readyz").Param("exclude", "shutdown").Do(context.Background())
```
mborsz, line 119: Add some comment why we set this param

PR #4174, opened 2026-07-03
Allow customizing registry.k8s.io across ClusterLoader2

apendua:
#### Summary

*   Introduced the `registry-k8s-repo` CLI flag and `REGISTRY_K8S_REPO` environment variable (mapped to `ImageRegistry` in template mapping) to allow replacing the default `registry.k8s.io` image repository.
*   Updated all test templates and measurement manifests  to use the new `{{.ImageRegistry}}` template variable.
*   Added `ImageRegistry` into the `probes` measurement local template mapping.
*   Updated `docs/GETTING_STARTED.md` documentation to instruct users on how to use the `ImageRegistry` template variable and flag overrides.

#### What type of PR is this?

/kind feature

Rate-limiting or network isolation often requires pulling test images from a registry mirror rather than the upstream registry.k8s.io. Exposing this customization option via a ClusterLoader2 flag allows users to seamlessly point all benchmark workloads, metrics exporters, probes to their local registry.

This change fails to address some deployed dependencies (such as the GCE PD CSI driver), due to additional complexities, but I am planning to prepare a followup PR to take care of that.

Fixes #

### DIFF
--- a/clusterloader2/pkg/config/template_provider.go
+++ b/clusterloader2/pkg/config/template_provider.go
@@ -264,6 +264,7 @@ func GetMapping(clusterLoaderConfig *ClusterLoaderConfig, testOverridePaths []st
 		return nil, errors.NewErrorList(fmt.Errorf("mapping creation error: %v", err))
 	}
 	mapping["Nodes"] = clusterLoaderConfig.ClusterConfig.Nodes
+	mapping["ImageRegistry"] = clusterLoaderConfig.ImageRegistry
 	envMapping, err := LoadCL2Envs()
 	if err != nil {
 		return nil, errors.NewErrorList(goerrors.Errorf("mapping creation error: %v", err))
--- a/clusterloader2/cmd/clusterloader.go
+++ b/clusterloader2/cmd/clusterloader.go
@@ -138,6 +138,7 @@ func initFlags() {
 	flags.IntVar(&port, "port", 8000, "Port to be used by http server with pprof.")
 	flags.DurationEnvVar(&heapProfileInterval, "heap-profile-interval", "CL2_HEAP_PROFILE_INTERVAL", 0, "Interval between clusterloader2 heap profiles. 0 represents disabled.")
 	flags.BoolVar(&dryRun, "dry-run", false, "Whether to skip running test and only compile test config")
+	flags.StringEnvVar(&clusterLoaderConfig.ImageRegistry, "registry-k8s-repo", "REGISTRY_K8S_REPO", "registry.k8s.io", "FQDN of registry.k8s.io image repo")
 
 	initClusterFlags()
 	execservice.InitFlags(&clusterLoaderConfig.ExecServiceConfig)
@@ -287,6 +288,7 @@ func main() {
 	if err := flags.Parse(); err != nil {
 		klog.Exitf("Flag parse failed: %v", err)
 	}
+	clusterLoaderConfig.ExecServiceConfig.ImageRegistry = clusterLoaderConfig.ImageRegistry
 
 	// Start http server with pprof.
 	go func() {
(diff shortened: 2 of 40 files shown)


clusterloader2/pkg/config/template_provider.go
```diff
@@ -264,6 +264,7 @@ func GetMapping(clusterLoaderConfig *ClusterLoaderConfig, testOverridePaths []st
 		return nil, errors.NewErrorList(fmt.Errorf("mapping creation error: %v", err))
 	}
 	mapping["Nodes"] = clusterLoaderConfig.ClusterConfig.Nodes
+	mapping["ImageRegistry"] = clusterLoaderConfig.ImageRegistry
```
Qqkyu: I think this should be set only if not already defined. Otherwise we might override `ImageRegistry` specified inside a test override file (see line 262):

```go
	mapping, err := LoadTestOverrides(clusterLoaderConfig.OverridePaths, testOverridePaths)
```

```suggestion
	if _, ok := mapping["ImageRegistry"]; !ok {
        mapping["ImageRegistry"] = clusterLoaderConfig.ImageRegistry
    }
```

clusterloader2/pkg/prometheus/prometheus.go
```diff
@@ -231,6 +226,7 @@ func NewController(clusterLoaderConfig *config.ClusterLoaderConfig) (pc *Control
 	mapping["PROMETHEUS_MEMORY_REQUEST"] = clusterLoaderConfig.PrometheusConfig.PrometheusMemoryRequest
 	snapshotEnabled, _ := pc.isEnabled()
 	mapping["RetainPD"] = snapshotEnabled
+	mapping["ImageRegistry"] = clusterLoaderConfig.ImageRegistry
```
Qqkyu: I don't think this is needed here since we've got:

```go
	mapping, errList := config.GetMapping(clusterLoaderConfig, nil)
```

above?

apendua: Good catch! Removing...

clusterloader2/pkg/config/template_provider.go
apendua: Thanks for pointing this out! I will refactor this a bit.

apendua: I took a slightly different but equivalent approach. Can take a second look?

clusterloader2/pkg/config/template_provider.go
```diff
@@ -259,7 +262,11 @@ func LoadTestOverrides(paths []string, testOverridePaths []string) (map[string]i
 
 // GetMapping returns template variable mapping for the given ClusterLoaderConfig.
 func GetMapping(clusterLoaderConfig *ClusterLoaderConfig, testOverridePaths []string) (map[string]interface{}, *errors.ErrorList) {
-	mapping, err := LoadTestOverrides(clusterLoaderConfig.OverridePaths, testOverridePaths)
+	// Default mappings, which can be overridden by test overrides.
+	defaults := make(map[string]interface{})
+	defaults["ImageRegistry"] = clusterLoaderConfig.ImageRegistry
```
Qqkyu: I think with this new logic users can override CLI flags with values in overrides files which is backward, we override values in override files with CLI flags

Also the depenndencies (for example `draDependency` in `dra.go`) and measurements (for example `probes` in `probes.go`) fetch the image registry from the `ClusterLoaderConfig`, which remains unmodified (and continue to use the default registry `registry.k8s.io`)

apendua: My bad. I assumed that CLI flag would specify the default value but it seems like this is the other way around.

Qqkyu (commented):
/lgtm
/approve

PR #4175, opened 2026-07-04
fix(benchmark): exclude failed runs from benchmark run selection

immanuwell:
#### What type of PR is this?
/kind bug

`GetLastNJobRuns` and `GetJobRunsFromLastNHours` were accepting any readable `finished.json`, even when the run result was not `SUCCESS`.
That lets failed jobs slip into benchmark samples, which is kinda off for perf comparisons.
This patch only keeps runs where `GetJobRunFinishedStatus(...)` is `true`.

Repro:
1. In `benchmark/pkg/metricsfetcher/runselector/schemes`, use `MockFinishedStatuses: map[int]bool{1:true, 2:false, 3:true}`.
2. Before this patch `GetLastNJobRuns("job", 3, utils)` returns `[3 2 1]`.
3. After this patch it returns `[3 1]`.

None found.

I re-ran `go test ./pkg/metricsfetcher/runselector/schemes` and `./verify/test.sh`.
I also checked the public GCS layout for `ci-kubernetes-kubemark-100-gce`, it does use root `finished.json`, so this edge case is real, not just mock stuff.

### DIFF
--- a/benchmark/pkg/metricsfetcher/runselector/schemes/last_n_hours.go
+++ b/benchmark/pkg/metricsfetcher/runselector/schemes/last_n_hours.go
@@ -41,7 +41,8 @@ func GetJobRunsFromLastNHours(job string, n int, utils util.JobLogUtils) ([]int,
 			if (currentTime - startTimestamp) >= uint64(3600*n) {
 				break
 			}
-			if _, err := utils.GetJobRunFinishedStatus(job, buildNumber); err == nil {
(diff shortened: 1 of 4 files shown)


Qqkyu (commented):
/ok-to-test

immanuwell:
@Qqkyu friendly reminder: all tests are green

Qqkyu (commented):
/lgtm
/approve

PR #4180, opened 2026-07-09
Source load init-events from Prometheus for HA and restart safety

Jefftree:
MetricsForE2E reads init-events from one apiserver as a point-in-time counter, which under-counts under HA, so source it from Prometheus instead. (for kubernetes/test-infra#37360)

Landing https://github.com/kubernetes/test-infra/pull/37360 will let us test this premerge.

### DIFF
--- a/clusterloader2/testing/load/modules/measurements.yaml
+++ b/clusterloader2/testing/load/modules/measurements.yaml
@@ -220,6 +220,19 @@ steps:
       - name: Watch cache reinitializations
         query: sum(increase(apiserver_watch_cache_initializations_total[%v:])) by (resource)
 {{end}}
+  - Identifier: InitEvents
+    Method: GenericPrometheusQuery
+    Params:
+      action: {{$action}}
+      metricName: Init Events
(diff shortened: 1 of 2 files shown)


serathius:
I don't have much context on LoadInitEventsCount, and why was it implemented this way. Hard to say if changing it to prometheus will break something without testing t manually.

Have you looked into original PR https://github.com/kubernetes/perf-tests/pull/605?

Jefftree:
> I don't have much context on LoadInitEventsCount, and why was it implemented this way. Hard to say if changing it to prometheus will break something without testing t manually.
> 
> Have you looked into original PR #605?

So the raw metric is the same as prometheus scrapes the same `/metrics` endpoint. I dug into the previous PR, I believe that we didn't have the prometheusquery measurement integration and had to roll them by hand at that time. I think we have quite a few of these instances that ends up with custom logic to parse the metric that looks hard to maintain. eventually would like to transition them to prometheus based in the future.

I've tested this with a local perfdash deployment.

PR #4183, opened 2026-07-10
benchmark-list: expose Go runtime metrics to disambiguate memory regressions

### LINKED ISSUE #4140: benchmark-list: an allocation improvement shows up as a memory regression
The recent ~100 MiB step in benchmark-list apiserver memory ([perfdash](https://perf-dash.k8s.io/#/?jobname=benchmark%20list&metriccategoryname=E2E&metricname=Resources&Resource=memory&PodName=kube-apiserver-benchmark-list-master%2Fkube-apiserver)) turned out to be benign GC headroom, but proving that took a manual dig into Prometheus snapshots because the dashboard only carries one number. Filing this to improve what we monitor so the next step is self-diagnosing.

**What happened**

Memory stepped 2685 to 2785 MiB and stayed there. It bisects cleanly to https://github.com/kubernetes/kubernetes/pull/139617 (JSON list encoder), which reuses one buffer instead of allocating per item, about 600x less garbage per list. The RSS went up because of that: less allocation means GC fires ~6x less often, so under GOGC=100 the heap sits higher between collections.

Runs: [good](https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-benchmark-list/2065489118564978688) (alpha.1.110) to [bad](https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-benchmark-list/2065549768825769984) (alpha.1.127).

Image

The live heap never moved, which is the whole point and is invisible on the current dashboard:

| 8-min avg, good vs bad | good | bad |
|---|---|---|
| allocation rate | 612 MB/s | 105 MB/s |
| GC frequency | 28.5/min | 4.9/min |
| **heap_inuse (live)** | **1785 MB** | **1778 MB** |
| working_set (what we chart) | up ~100 MiB | |

**The gap**

For this job perfdash carries a single apiserver memory series, `container_memory_working_set_bytes` from `ResourceUsageSummary`, which is essentially RSS. working_set moves with GC pacing, not just real memory use, so on its own it cannot distinguish:

- a real leak or live-heap growth (the thing we want to catch), from
- GC headroom shifting because allocation rate changed (this episode, benign).

It fails the other way too: a real leak can hide under a flat working_set if GC compensates. So every allocation-pattern change risks a false alarm, and real regressions can be masked.

**Ideas**

We can surface additional metrics that are already scraped:

- live heap (`go_memstats_heap_inuse_bytes`), flat here
- GC rate (`rate(go_gc_duration_seconds_count[5m])`), dropped 6x
- allocation rate (`rate(go_memstats_alloc_bytes_total[5m])`), dropped 6x

Sidenote: the proto variant dashboard also shows a "regression" from an allocation improvement (https://github.com/kubernetes/kubernetes/pull/139350), so this is a recurring pattern of allocation improvements reading as memory regressions on the dashboard.

Gaurav598:
What type of PR is this?

/kind bug

This change exposes additional Go runtime metrics for the benchmark-list benchmark using the existing `GenericPrometheusQuery` infrastructure.

The benchmark currently primarily surfaces `container_memory_working_set_bytes`, which can increase when allocation rate decreases and Go GC runs less frequently. This can make allocation improvements appear as memory regressions even when the live heap remains unchanged.

This PR adds GenericPrometheusQuery measurements for:
- `go_memstats_heap_inuse_bytes`
- `rate(go_gc_duration_seconds_count[%v])`
- `rate(go_memstats_alloc_bytes_total[%v])`
and registers these metrics in Perfdash so they are displayed alongside the existing benchmark results.

Files changed:
- `clusterloader2/testing/list/modules/measurements.yaml` – add GenericPrometheusQuery measurements.
- `perfdash/config.go` – register GenericMeasurements for the benchmark-list dashboard.
- `perfdash/parser_test.go` – add coverage for GenericPrometheusQuery parsing.

The implementation is additive and does not modify the existing `ResourceUsageSummary` pipeline.

Which issue(s) this PR fixes:
Fixes #4140

This change intentionally uses the existing GenericPrometheusQuery pipeline instead of extending ResourceUsageSummary.
Relevant tests executed:
- `go test ./...` (perfdash)
- `go test ./pkg/measurement/common` (clusterloader2)

The implementation only adds new benchmark metrics and Perfdash registration without changing existing benchmark behavior.

### DIFF
--- a/clusterloader2/testing/list/modules/measurements.yaml
+++ b/clusterloader2/testing/list/modules/measurements.yaml
@@ -42,4 +42,50 @@ steps:
         threshold: {{$APISERVER_MEMORY_THRESHOLD_MEGABYTES}}
         requireSamples: false
         lowerBound: false
+  - Identifier: APIServerHeapInuse
+    Method: GenericPrometheusQuery
+    Params:
+      action: {{$action}}
+      metricName: kube-apiserver heap inuse
+      metricVersion: v1
+      unit: MiB
+      queries:
+      - name: MaxHeapInuse
+        query: |
+          max_over_time(
+          sum(
+            go_memstats_heap_inuse_bytes{job="master",endpoint="apiserver"}
+            / 1024 / 1024
+          )[%v:])
+        requireSamples: false
+  - Identifier: APIServerGCRate
+    Method: GenericPrometheusQuery
+    Params:
+      action: {{$action}}
+      metricName: kube-apiserver GC rate
+      metricVersion: v1
+      unit: count/s
+      queries:
+      - name: GCRate
+        query: |
+          sum(
+            rate(
+              go_gc_duration_seconds_count{job="master", endpoint="apiserver"}[%v])
+          )
+        requireSamples: false
+  - Identifier: APIServerAllocationRate
+    Method: GenericPrometheusQuery
+    Params:
+      action: {{$action}}
+      metricName: kube-apiserver allocation rate
+      metricVersion: v1
+      unit: MiB/s
+      queries:
+      - name: AllocationRate
+        query: |
+          sum(
+            rate(go_memstats_alloc_bytes_total{job="master",endpoint="apiserver"}[%v])
+            / 1024 / 1024
+          )
+        requireSamples: false
   {{end}}
--- a/perfdash/config.go
+++ b/perfdash/config.go
@@ -750,6 +750,13 @@ var (
 				Parser:           parsePerfData,
 			}},
 		},
+		"GenericMeasurements": {
+			"GenericMeasurements": []TestDescription{{
+				OutputFilePrefix:            GenericPrometheusQueryMeasurementName,
+				Parser:                      parsePerfData,
+				FetchMetricNameFromArtifact: true,
+			}},
+		},
 	}
 
 	etcdAPIBenchmarkDescription = TestDescriptions{


Jefftree:
Please test this with a local perfdash instance.

Gaurav598:
@Jefftree 
I have validated the Perfdash configuration locally.
To avoid fabricating synthetic data, I pointed a local Perfdash instance at the live public GCS artifacts for ci-kubernetes-benchmark-list using a temporary local config.

As shown in the attached screenshot, the GenericMeasurements category is discovered correctly and Perfdash successfully parses and renders the existing GenericPrometheusQuery artifact (kube-apiserver memory usage) for the benchmark-list job.
The new heap in-use, GC rate, and allocation rate metrics introduced by this PR are not present in the current public artifacts because they will only be produced after benchmark-list runs with this change.

This confirms that the Perfdash parser/configuration wiring added in this PR is working as expected. Please let me know if you'd like me to validate anything else.

Perf-dash

clusterloader2/testing/list/modules/measurements.yaml
```diff
@@ -42,4 +42,49 @@ steps:
...
+      action: {{$action}}
+      metricName: kube-apiserver heap inuse
+      metricVersion: v1
+      unit: MiB
+      queries:
+      - name: MaxHeapInuse
+        query: |
+          max_over_time(
+          sum(
+            go_memstats_heap_inuse_bytes{namespace="monitoring", pod=~"kube-apiserver.*", container="kube-apiserver"}
```
Jefftree: `pod=` selector doesn't exist because this is an apiserver `/metrics` metric. Please verify with data from `prometheus_snapshot.tar` in the artifacts https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/logs/ci-kubernetes-benchmark-list/2076696781722226688/artifacts/. 

Please also disclose AI usage and get it to verify its work. Otherwise these hallucinations are very difficult to review. The new test also doesn't really offer anything and is just made up names.

Gaurav598: Hi @Jefftree, thanks for the review and for catching the selector issue.

I also want to disclose that I used AI to assist with navigating the codebase and drafting the initial implementation. It incorrectly inferred the selectors for these Go runtime metrics and suggested a synthetic test that wasn't providing meaningful coverage. I verified your feedback against the prometheus_snapshot.tar artifact and updated the implementation accordingly.

Changes made:
Updated the PromQL queries to use the correct selectors (job="master", endpoint="apiserver"), matching the labels present in the Prometheus snapshot.
Removed the synthetic unit test from perfdash/parser_test.go.
Re-verified the Perfdash wiring locally against the public ci-kubernetes-benchmark-list artifacts (screenshot attached in my previous comment).

I've pushed the fixes. Thanks again for the detailed review and guidance.

Gaurav598:
Hi @Jefftree, just a gentle ping on this PR.
I have updated the PromQL selectors based on the TSDB snapshot and removed the synthetic test as requested. The CI is fully green. Please let me know if everything looks good or if there is anything else you need from me!

Jefftree:
Task looks good. I'm still trying to reason if this gives us the signal we want, sorry will need a bit more time thinking about it. I had thought these metrics would give us a clear diff but it doesn't seem like it.

Gaurav598:
Hi @Jefftree, no worries at all, take your time! I'm glad the implementation itself looks good.
If you come to the conclusion that different metrics (perhaps go_memstats_heap_alloc_bytes or specific percentiles) might provide a clearer signal to distinguish between benign GC shifts and actual memory regressions, please let me know.
I'm happy to update the queries or explore other options whenever you are ready.

Gaurav598:
Hi @Jefftree, hope you've been doing well!

Just a gentle follow-up on this PR since it's been a few weeks. Have you had a chance to think about whether these metrics will provide the clear signal you were looking for?
Let me know if you'd like to proceed with this implementation or if we should explore a different approach. Still happy to help either way!

Gaurav598:
Hi @Jefftree, just following up on this PR. It’s been a couple of weeks since the last update, so I wanted to check if you’ve had a chance to reach a conclusion on whether these metrics provide the signal we’re looking for.

If you’d prefer a different approach or different metrics, I’m happy to make the necessary changes. Thanks!

Gaurav598:
Hi @Jefftree, just following up on this PR. It’s been a couple of weeks since the last update, so I wanted to check if you’ve had a chance to reach a conclusion on whether these metrics provide the signal we’re looking for.

If you’d prefer a different approach or different metrics, I’m happy to make the necessary changes. Thanks!

PR #4192, opened 2026-07-14
fix: validate stepped load delays

immanuwell:
#### What type of PR is this?

/kind bug

Rejects zero or negative steppedLoad stepDelay values. Zero made bursts run back to back, kinda defeating stepped load.

N/A

Repro: omit stepDelay from a steppedLoad config. It passed validation and ran bursts immediately.

Tests: go test ./...


Qqkyu (commented):
/lgtm
/approve

PR #4016, opened 2026-05-06
Drop year from copyright header

Jefftree:
Sync boilerplate verification with k/k upstream's "no year in copyright header for new files".

xref: https://github.com/kubernetes/kubernetes/pull/134835

PR #4194, opened 2026-07-15
fix(clusterloader2): preserve test suite config artifacts

immanuwell:
#### What type of PR is this?

/kind bug

Keeps dry-run configs from every test-suite scenario. Reused configs used to overwrite earlier artifacts. 
Now each scenario gets its own file

None.

Repro: run `testing/density/scheduler-suite.yaml` in dry-run mode. 
Before, 4 dumps were logged but only 1 file stayed. Now all 4 are kept

Tests: `go test ./...`

### DIFF
--- a/clusterloader2/cmd/clusterloader_test.go
+++ b/clusterloader2/cmd/clusterloader_test.go
@@ -1,0 +1,64 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


clusterloader2/cmd/clusterloader_test.go
```diff
@@ -0,0 +1,64 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
Qqkyu: I think after the https://github.com/kubernetes/perf-tests/pull/4016 we don't want to include the year in copyright header, `verify/verify-boilerplate.sh` will fail with this

other than this lgtm

Qqkyu (commented):
/ok-to-test

clusterloader2/cmd/clusterloader_test.go
immanuwell: done

immanuwell:
@Qqkyu done, PTAL

Qqkyu (commented):
/lgtm
/approve

PR #4196, opened 2026-07-18
fix(perfdash): paginate S3 metric listings

immanuwell:
#### What type of PR is this?

/kind bug

S3 returns at most 1000 keys per listing, current code drops later pages. This can hide recent builds or metric files. Paginate both listings

None.

Repro: return `IsTruncated=true` with a continuation token from a mock S3 endpoint. Before this change only page one is returned, now both pages are read

Tests: `make test`, `go test -race ./...`, `go build ./...`


Qqkyu (commented):
/ok-to-test
/lgtm
/approve

PR #4200, opened 2026-07-20
request-benchmark: introduce http subcommand with backward compatibility

p0lyn0mial:
#### What type of PR is this?
/kind feature

This is the first step in unifying [`watch-list`](https://github.com/kubernetes/perf-tests/tree/master/util-images/watch-list) and `request-benchmark` into a single binary. Today `watch-list` runs informers.  The upcoming `informer` subcommand will bring that capability into `request-benchmark`.

This PR introduces `http` subcommand and also preserves backward compatibility: callers passing flags directly (no subcommand) are silently routed to http mode.

This PR allows to run:
```
#backward compatible case
request-benchmark --uri=/api/v1/pods --inflight=4

# new subcommand case
request-benchmark http --uri=/api/v1/pods --inflight=4
 ``` 

Fixes #

### DIFF
--- a/util-images/request-benchmark/main.go
+++ b/util-images/request-benchmark/main.go
@@ -53,15 +53,6 @@ const (
 	tableAccept = "application/json;as=Table;v=v1;g=meta.k8s.io"
 )
 
-var (
-	inflight    = flag.Int("inflight", 1, "Benchmark inflight (number of parallel requests being made to the apiserver")
-	namespace   = flag.String("namespace", "", "Replace %namespace% in URI with provided namespace")
-	URI         = flag.String("uri", "", "Request URI")
-	verb        = flag.String("verb", "GET", "A verb to be used in requests.")
-	qps         = flag.Float64("qps", -1, "The qps limit for all requests")
-	contentType = ContentType("json")
-)
-
 func (c ContentType) String() string {
 	return string(c)
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: util-images/request-benchmark/main.go [hunk 1]
```diff
@@ -99,24 +85,56 @@ func getContentType(ct ContentType) (string, error) {
...
+			}
+			return
+		default:
+			log.Fatalf("unknown subcommand %q, valid: http", args[0])
+		}
+	}
+	// legacy: no subcommand means http mode
+	if err := runHTTP(args); err != nil {
+		log.Fatal(err)
+	}
```
serathius, line 102:
```suggestion
	args := os.Args[1:]
	mode  := "http"
	if len(args) > 0 && !strings.HasPrefix(args[0], "-") {
		mode = args[0]
	}
	switch args[0] {
	case "http":
		if err := runHTTP(args[1:]); err != nil {
			log.Fatal(err)
		}
		return
	default:
		log.Fatalf("unknown subcommand %q, valid: http", args[0])
	}
```

serathius:
Overall looks good with one comment.

Jefftree:
LGTM. Out of curiosity how do we test these kinds of changes? Image is only built on postsubmit, and we need to manually bump all the usages after. Do we just test locally with a kind cluster? Wonder if there's something we can improve around the current process.

serathius:
> Out of curiosity how do we test these kinds of changes?

Manually based on trust at least until investing in automation would help us accelerate work instead of slowing down.

util-images/request-benchmark/main.go
p0lyn0mial: i like the suggestion. thanks.

p0lyn0mial:
> Out of curiosity how do we test these kinds of changes?

yes, for now I'm testing the binary locally against a local kind cluster.

PR #4202, opened 2026-07-21
fix(dns): preserve pod names in histogram rows

immanuwell:
Keeps `pod_name` on DNS histogram rows.

The default client deployment has 4 pods. With `--db`, histogram rows got a NULL pod name, so joins with result rows duplicated data

Repro: run `cd dns && python py/run_perf.py --dns-server coredns --params params/coredns/test.yaml --db out/db`, then join `histograms` and `results` on run and subrun. 
Before this fix, every histogram row joins to every client result row

Test: `make verify-all-dev` and DNS Python unit tests

### DIFF
--- a/dns/README.md
+++ b/dns/README.md
@@ -254,6 +254,7 @@ CREATE TABLE results (
 CREATE TABLE histograms (
   run_id,
   run_subid,
+  pod_name,
   rtt_ms,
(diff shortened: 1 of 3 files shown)


immanuwell:
@bowei @shyamjvs please take a look

PR #4204, opened 2026-07-23
request-benchmark: add informer subcommand

p0lyn0mial:
Add informer subcommand to request-benchmark

The new command is almost a direct copy of the [watch-list binary](https://github.com/kubernetes/perf-tests/tree/master/util-images/watch-list), adapted to fit the subcommand structure. It is meant to be a drop-in replacement. All flags exposed by the watch-list binary are preserved


util-images/request-benchmark/mode_informer.go
```diff
@@ -0,0 +1,146 @@
+/*
```
p0lyn0mial, line 1: compare with https://github.com/kubernetes/perf-tests/blob/master/util-images/watch-list/main.go

util-images/request-benchmark/mode_informer_test.go
```diff
@@ -0,0 +1,57 @@
+/*
```
p0lyn0mial, line 1: compare with https://github.com/kubernetes/perf-tests/blob/master/util-images/watch-list/main_test.go

Jefftree:
can you update the README for request-benchmark to document this new behavior? 

Otherwise lgtm

PR #4208, opened 2026-07-24
Injest prometheus snapshot and visualize apiserver latency metrics on dashboard.

serathius:
/kind feature

image

Downloading via golang is not perfect, but was to lazy to rewrite it.

/cc @Jefftree

### DIFF
--- a/perflens/config/grafana/dashboards/apiserver_metrics.json
+++ b/perflens/config/grafana/dashboards/apiserver_metrics.json
@@ -1,0 +1,772 @@
+{
+  "annotations": {
+    "list": []
+  },
+  "editable": true,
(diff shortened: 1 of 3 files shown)


perflens/config/grafana/dashboards/apiserver_metrics.json
```diff
@@ -0,0 +1,772 @@
...
+        "query": "label_values(apiserver_request_total, verb)",
+        "refresh": 1,
+        "regex": "",
+        "skipUrlSync": false,
+        "sort": 1,
+        "type": "query"
+      }
+    ]
+  },
+  "time": {
```
Jefftree, line 751: this looks hard coded but we can iterate.

PR #4242, opened 2026-07-27
Download files using gcloud

serathius:
/kind feature

/cc @michaelasp


michaelasp:
/lgtm

Looks misformatted, can lgtm again after gofmt fix since I don't believe it'll merge now.

PR #4243, opened 2026-07-27
Split perflens into subfiles

serathius:
/kind cleanup

/cc @michaelasp


perflens/cmd/ingest/main.go
```diff
@@ -133,7 +39,7 @@ type options struct {
 }
 
 func parseFlags() options {
-	buildID := flag.String("build-id", "", "Prow Build ID or run identifier")
+	buildID := flag.String("build-id", "", "Prow Build ID, run identifier, or run directory path")
```
michaelasp, line 42: This feels like we are starting to overload what a build id is.

michaelasp (commented):
/lgtm

Non-blocking comment since this is mostly a cleanup PR, but wondering for future improvements.

PR #4246, opened 2026-07-28
Add node exporter dashboard

serathius:
/kind feature

image


michaelasp:
/lgtm

Looks good from the attached view, haven't looked too closely at the json but let's iterate from here.

PR #4248, opened 2026-07-30
Fix DRATestDriver readiness passing before drivers and ResourceSlices exist

alaypatel07:
#### What type of PR is this?

/kind bug
/kind flake

DRATestDriver readiness could pass while no driver DaemonSet pods or ResourceSlices existed yet (`DesiredNumberScheduled == NumberReady == 0`, and `0` ResourceSlices matching `0` running plugin pods). FastFill then starts before the DRA driver is usable, which shows up as scheduler "cannot allocate all claims" retries and inflated `create_to_schedule` / FastFillPodStartup p99.

This change waits until `WorkerNodeCount` driver pods are Desired/Ready and at least that many ResourceSlices are published, and uses one typed `getWorkerCount` helper for both manifest templating and readiness checks.

Fixes # https://github.com/kubernetes/kubernetes/issues/141046

Observed on aws-dra-500Nodes-with-workload FastFillPodStartup p99 spikes where scheduler logs repeatedly report "500 cannot allocate all claims" before binding succeeds.

### DIFF
--- a/clusterloader2/pkg/dependency/dra/dra.go
+++ b/clusterloader2/pkg/dependency/dra/dra.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"embed"
 	"fmt"
-	"strconv"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -35,7 +34,7 @@ const (
 	draDependencyName      = "DRATestDriver"
 	draNamespace           = "dra-example-driver"
 	draManifests           = "dra-example-driver"
-	defaultWorkerNodeCount = "100"
+	defaultWorkerNodeCount = 100
 	draDaemonsetName       = "dra-example-driver-kubeletplugin"
 	checkDRAReadyInterval  = 30 * time.Second
 	defaultDRATimeout      = 10 * time.Minute
@@ -78,9 +77,14 @@ func (d *draDependency) Setup(config *dependency.Config) error {
 		return err
 	}
 
+	workerCount, err := getWorkerCount(config)
+	if err != nil {
+		return err
+	}
+
 	mapping := map[string]interface{}{
 		"Namespace":       namespace,
-		"WorkerNodeCount": getWorkerCount(config),
+		"WorkerNodeCount": workerCount,
 		"ImageRegistry":   config.ClusterLoaderConfig.ImageRegistry,
 	}
 
@@ -150,7 +154,7 @@ func (d *draDependency) waitForDRADriverToBeHealthy(config *dependency.Config, t
 }
 
 func (d *draDependency) isDRADriverReady(config *dependency.Config, daemonsetName string, namespace string) (done bool, err error) {
-	expectedWorkers, err := getExpectedWorkerCount(config)
+	expectedWorkers, err := getWorkerCount(config)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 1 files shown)


clusterloader2/pkg/dependency/dra/dra.go
```diff
@@ -159,66 +168,45 @@ func (d *draDependency) isDRADriverReady(config *dependency.Config, daemonsetNam
...
+	if len(resourceSlices.Items) < expectedWorkers {
+		klog.V(2).Infof("waiting for resourceslices to be available, expected: %d, available: %d",
+			expectedWorkers, len(resourceSlices.Items))
 		return false, nil
 	}
 	return true, nil
 }
 
-func getDriverPluginPods(config *dependency.Config, namespace string, namePrefix string) (int, error) {
-	pods, err := config.ClusterFramework.GetClientSets().GetClient().CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{})
```
alaypatel07, line 197: From what I recall, this logic was put in place because at larger scale, some nodes might go to NotReady and it will block the test from moving forward.  To avoid that, only running pods are counted against resource slices.

However, I see two issues with this:
1. the logic is in-correct in finding terminal state. The resource slice count will be equal to running pods at many different points of the driver daemonset rollout not just the terminal state.
2. If the nodes are NotReady, you do not have enough capacity and the test is going to fail anyway. So user running the tests should find other ways of dealing with NotReady Nodes, like cordoning/tainting them and subtracting them from node count, so it is ignored by the test.

johnbelamaric:
@alaypatel07 is the failure related?

alaypatel07:
@johnbelamaric no the specific test doesnt exercise DRA path: 
```
kubetest2: Test expand_less23m1s{  exit status 255} | kubetest2: Test expand_less | 23m1s | {  exit status 255}
kubetest2: Test expand_less | 23m1s
{  exit status 255}
```

Failure seems to be unrelated.

alaypatel07:
```
I0730 21:23:58.534542   21321 framework.go:285] Applying templates for "manifest/exec_deployment.yaml"
I0730 21:24:08.705996   21321 framework.go:285] Applying templates for "manifests/daemonset.yaml"
I0730 21:25:18.842531   21321 clusterloader.go:494] Test config successfully dumped to: /logs/artifacts/generatedConfig_dra-steady-state.yaml
I0730 21:25:18.842544   21321 clusterloader.go:265] --------------------------------------------------------------------------------
I0730 21:25:18.842547   21321 clusterloader.go:266] Running /home/prow/go/src/k8s.io/perf-tests/clusterloader2/testing/dra/config.yaml
I0730 21:25:18.842548   21321 clusterloader.go:267] --------------------------------------------------------------------------------
I0730 21:25:19.036291   21321 framework.go:285] Applying templates for "manifests/dra-example-driver/*.yaml"
E0730 21:30:34.074327   21321 clusterloader.go:275] --------------------------------------------------------------------------------
E0730 21:30:34.074348   21321 clusterloader.go:276] Test Finished
E0730 21:30:34.074350   21321 clusterloader.go:277]   Test: /home/prow/go/src/k8s.io/perf-tests/clusterloader2/testing/dra/config.yaml
E0730 21:30:34.074352   21321 clusterloader.go:278]   Status: Fail
E0730 21:30:34.074354   21321 clusterloader.go:280]   Errors: [dependency Install dra-example-driver for test setup timed out after 5m0s]
E0730 21:30:34.074355   21321 clusterloader.go:282] --------------------------------------------------------------------------------
```

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/perf-tests/4248/pull-perf-tests-ec2-500-node-dra-with-workload-amazonvpc-using-cl2/2082934316282679296

this is relevant failure, but I dont see any logs to debug. I will try it locally if I can reproduce this scale.

PR #4249, opened 2026-07-31
watch-list: switch informer job to request-benchmark

p0lyn0mial:
tested locally.


p0lyn0mial:
failing `pull-perf-tests-gce-master-scale-performance-100` should be fixed by https://github.com/kubernetes/perf-tests/pull/4253

PR #4252, opened 2026-08-04
Parametrize OOM restarts and ContainerRestarts to be optionally faili…

emerbe:
…ng tests. The default behavior does not change if new params are not used in test configuration

#### What type of PR is this?

/kind feature

This PR introduces parameters that allow users to toggle whether container restarts and cluster OOM events cause tests to fail. This provides flexibility for custom test configurations while preserving the existing default behavior of failing on these events.

Fixes #


Qqkyu (commented):
/ok-to-test
/lgtm
/approve

PR #4253, opened 2026-08-04
Add write throughput benchmark

serathius:
1. Deploy 10k realistic pod manifests.
2. Setup 10 concurrent watchers
3. Setup writer doing 100 PATCH QPS

In the future:
* Measure API server resources and watch latency to detect when things break.
* Increase the dimensions to find the K8s limit.

/kind feature

/assign @p0lyn0mial


p0lyn0mial:
I didn't have time to look at the failures but it seems the the `performance-100` got broken - https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-perf-tests-gce-master-scale-performance-100

PR #4263, opened 2026-08-13
Raise etcd backend quota for write-throughput kind cluster

Jefftree:
etcd's 2GiB default is too small for this benchmark, the 15m run hits NOSPACE around minute 7.

### DIFF
--- a/clusterloader2/testing/write-throughput/kind.yaml
+++ b/clusterloader2/testing/write-throughput/kind.yaml
@@ -18,3 +18,4 @@ nodes:
       local:
         extraArgs:
           listen-metrics-urls: "http://0.0.0.0:2382"
+          quota-backend-bytes: "8589934592"


clusterloader2/testing/write-throughput/kind.yaml
```diff
@@ -18,3 +18,5 @@ nodes:
       local:
         extraArgs:
           listen-metrics-urls: "http://0.0.0.0:2382"
+          # etcd's 2GiB default is too small for this benchmark, the 15m run hits NOSPACE around minute 7.
```
serathius:
```suggestion
```

Jefftree: was intentional since we might tune parameters in the future and it's not super obvious that this would need to changed. Removed.

PR #4265, opened 2026-08-13
Drop GVM from ClusterLoader2 getting-started prerequisites

### LINKED ISSUE #3401: ClusterLoader dependency on GVM
I just wanted to flag that the GVM dependency of cluster loader is suffering from massive, fundamental issues like e.g.

* https://github.com/moovweb/gvm/issues/483#issuecomment-2210527684 - dead on arrival, fails to install
* breaking `$PATH` variables: https://github.com/moovweb/gvm/issues/466
* multiple reports of it breaking the Linux `cd` command (it's injecting an alias which doesn't work with whitespaces) e.g. https://github.com/moovweb/gvm/issues/446 or https://github.com/moovweb/gvm/issues/439
* it's also fundamentally being unreasonable as it requires Go to install Go: https://github.com/moovweb/gvm/issues/453

On top of that, seems that GVM is also unmaintained at this point as none of the PRs community provided to fix it are being handled, repo maintainer seems to he unavailable.

My suggestion is to stop using GVM and remove it from perf-tests prerequisites.

Jefftree:
GVM is unmaintained and broken to install. It was only a docs instruction for getting Go, never a build dependency, and `gvm linkthis` is a no-op for a Go module.

Fixes #3401

### DIFF
--- a/clusterloader2/docs/GETTING_STARTED.md
+++ b/clusterloader2/docs/GETTING_STARTED.md
@@ -6,25 +6,14 @@ In this tutorial, we will:
 - Implement a simple CL2 test and run it
 - Run load test on 100 nodes cluster
 
+You need Go installed, see the [Go install] guide.
+
 ## Clone perf-tests repository
 
 Start with cloning perf-tests repository:
 ```bash
 git clone git@github.com:kubernetes/perf-tests.git
 cd perf-tests
-```
-
-## Install GVM
-Follow instructions on [GVM install].
-Install golang with specific version (1.26.4 was tested in this tutorial):
-```bash
-gvm install go1.26.4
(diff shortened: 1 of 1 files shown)


serathius:
Hmm, I'm still using GVM, recommend it :P

But I agree it's personal preference and not needed here.

/lgtm
/approve

weakcamel:
> @serathius: Hmm, I'm still using GVM, recommend it :P

When I raised it, it was specifically about installation on M3 chips: https://github.com/moovweb/gvm/issues/483#issuecomment-2210527684 - is that what you're on?

BTW, sadly that issue is still open a year later, last commit in GVM repo is from 2023 - totally abandonware at this point.

PR #4266, opened 2026-08-13
agents: add scale-run-metrics reference skill

Jefftree:
Catalogs which Prometheus series and profiles from a scale run are important. 

I've been doing a lot of agent assisted analysis and each time the shape comes out somewhat differently with new metrics requested by reviewers. Some metrics get missed or agents tunnel into certain metrics. This is an attempt to unify them into a reference so future analysis can loosely follow a template. 

Overall shape is up for discussion, so feel free to suggest different ways of representing this info if desired.

### DIFF
--- a/.agents/skills/scale-run-metrics/references/metrics.md
+++ b/.agents/skills/scale-run-metrics/references/metrics.md
@@ -1,0 +1,143 @@
+# Metrics of interest when analyzing a scale run
+
+The measurement JSONs cover the SLOs. Everything below comes from the Prometheus snapshot
+(`--experimental-prometheus-snapshot-to-report-dir=true`) and the profiles collected
+alongside it.
+
+## What gets scraped
+
+From `pkg/prometheus/manifests/master-ip/master-serviceMonitor.yaml`. All of it carries
+`job="master"`, so pick a component by its port.
+
+| component | port | interval | scraped when |
+|---|---|---|---|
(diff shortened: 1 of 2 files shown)


.agents/skills/scale-run-metrics/references/metrics.md
```diff
@@ -0,0 +1,143 @@
+# Metrics of interest when analyzing a scale run
```
serathius, line 1: I was hoping that we could make perflens the codified list of good metrics via dashboards, so humans can read them and AI agents can use them as documentation. What do you think about adding graphs for all the metrics below and have agent just read queries from dashboard?

Jefftree: Semantically I think having a holistic reference steers the agents better. 

I find purely using perflens to be quite token inefficient, and giving agents a guide lets them build out the dashboards too. This isn't intended to be an isolated skill, most runs will probably invoke both perflens and this skill. Also, the large json blobs of perflens is not really human readable, and I was hoping we can map out something simpler in this markdown file. 

Anyway this is just empirical evidence, I can encode them in perflens directly if you'd like but I don't think we're going to want to read the json for that and end up just going off the UI.

serathius: I'm looking for a single source of truth, not necessarily pushing everyone to single solution. If perflens is efficient let's discuss that separately. Human will not read json blob, agent can just use jq to take queries out via skill.

PR #4267, opened 2026-08-14
perflens: re-base run timestamps so runs share one axis

Jefftree:
- Ingest re-bases each run's samples to a common anchor, so one time range works for every run and selecting two runs overlays them instead of showing two spikes days apart. perflens_run_start_timestamp_seconds keeps the original start.
- Adds a Control Plane dashboard: process resources, Go runtime, etcd disk.
- Turns off the Grafana news feed and analytics calls.

Basically allows us to loosely compare two runs in the same dashboard and we don't need to hunt for the time as we align every run to start at 2000/1/1 0:00 UTC. 

image

image

### DIFF
--- a/perflens/README.md
+++ b/perflens/README.md
@@ -29,9 +29,19 @@ so the x-axis reads as elapsed time since run start instead of wall clock. Midni
 deliberate: Grafana has no duration axis mode, so an axis labelled `00:00`, `00:30`,
 `01:00` reads as elapsed time for free.
 
-The offset is computed once per run, from the earliest `minTime` across its Prometheus
-snapshot blocks, and every block of the run shifts by the same amount. Only the origin
-moves, values and durations are untouched.
+The origin, `t=0`, is **ClusterLoader2's first step**, read from the run's
+`build-log.txt`. That is the same event in every run, which is what makes two runs
+comparable. The start of the Prometheus data is not: it is when scraping began, which
+trails cluster bring-up by a run dependent amount, so anchoring on it would offset two
+overlaid runs by the difference in their bring-up times. Runs whose CL2 log was not
+collected fall back to the start of scraped data, and `perflens ingest` prints which
+anchor it used.
+
+Samples scraped before the test started therefore land *before* the origin, outside the
+default dashboard window. Widen the range to the left to see bring-up.
+
+The offset is computed once per run and every block of the run shifts by the same amount.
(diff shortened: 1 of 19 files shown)


Jefftree:
/assign @serathius

okay let's iterate this.

perflens/README.md
```diff
@@ -19,3 +19,39 @@
...
+---
+
+## Normalized time
+
+Ingest re-bases every sample of a run onto a common anchor of **2000-01-01T00:00:00Z**,
+so the x-axis reads as elapsed time since run start instead of wall clock. Midnight is
+deliberate: Grafana has no duration axis mode, so an axis labelled `00:00`, `00:30`,
+`01:00` reads as elapsed time for free.
+
+The offset is computed once per run, from the earliest `minTime` across its Prometheus
```
serathius: Nit: Can we compute minTime based on Cluster loader steps?

perflens/docker/docker-compose.yml
```diff
@@ -9,6 +9,10 @@ services:
     environment:
       - GF_AUTH_ANONYMOUS_ENABLED=true
       - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
+      - GF_NEWS_NEWS_FEED_ENABLED=false
```
serathius, line 12: What is this?

perflens/pkg/ingest/block.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/oklog/ulid/v2"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+)
+
+const twoHoursMS = int64(2 * 60 * 60 * 1000)
+
+// runSpanStepMS is the cadence used for values that hold for a whole run, such as
```
serathius, line 36: Not that I have high requirements for vibecoded project, but I cannot understand this comment at all.

serathius:
Overall I like the direction of automatically aligning runs on timeline, however I'm worried that rewriting prometheus snapshots adds complexity and performance implication. I used thanos to avoid rewriting blocks for exactly this reason. 

My ask is whether we can make this change semi reviewable so we can discuss prometheus setup and code architecture. Would splitting this PR into steps would be useful or we should merge without code review?

PR #4277, opened 2026-08-19
request-benchmark: add --content-type flag to informer mode

p0lyn0mial:
the first step toward migrating benchmark-list from http to informer mode (the base variant of the benchmark-list uses json).

tested manually on my local cluster.


serathius:
/lgtm
/approve

Love the change :P!

PR #4315, opened 2026-08-25
DNSPropagation: use watch to know when pods are ready

supahero1:
#### What type of PR is this?

/kind bug

Removes busy polling on pods, aggregates measurement over entire time window instead of last x minutes Prometheus window, increases QPS. Basically support for larger clusters where this currently does not work.

Fixes #

### DIFF
--- a/util-images/probes/pkg/dnspropagation/probe.go
+++ b/util-images/probes/pkg/dnspropagation/probe.go
@@ -18,7 +18,6 @@ package dnspropagation
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


util-images/probes/pkg/dnspropagation/probe.go
```diff
@@ -108,48 +111,86 @@ func Run() {
...
 func runProbe(client kubernetes.Interface) {
 	klog.V(1).Infof("DNS propagation probe started")
+
+	indices := selectSample(*podCount, *sampleCount)
+	targetPods := make(map[string]struct{}, len(indices))
+	for _, idx := range indices {
+		targetPods[fmt.Sprintf("%s-%d", *statefulSet, idx)] = struct{}{}
+	}
+
+	watcher, err := client.CoreV1().Pods(*namespace).Watch(context.TODO(), metav1.ListOptions{})
```
mborsz: You want to use informer. Using "naked" watches is a bad idea, it can be interrupted any time, you need some logic for recreating them, relists etc.

supahero1: Indeed indeed

PR #4316, opened 2026-08-26
clusterloader2: optimize Node informers with sharedNodeIndexerFactory and fix typed reflector logs

supahero1:
#### What type of PR is this?

/kind bug

In large scale tests creating multiple informers per node simply does not work

Fixes #

### DIFF
--- a/clusterloader2/pkg/imagepreload/imagepreload.go
+++ b/clusterloader2/pkg/imagepreload/imagepreload.go
@@ -19,27 +19,24 @@ package imagepreload
 import (
 	"context"
 	"embed"
+	"fmt"
 	"strings"
-	"sync"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/perf-tests/clusterloader2/pkg/flags"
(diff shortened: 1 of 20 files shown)


clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -134,21 +115,46 @@ func (c *controller) PreloadImages() error {
...
+			if !ok {
+				continue
+			}
+
+			if nodeCounter != nil {
+				match, err := nodeCounter.ShouldRun(node)
+				if err != nil || !match {
+					continue
+				}
+			} else if !util.IsNodeSchedulableAndUntainted(node) {
```
mborsz, line 147: Why IsNodeSchedulableAndUntainted? I cannot see this in previous code?

clusterloader2/pkg/measurement/common/node_health_tracker.go
```diff
@@ -128,66 +163,125 @@ func (m *nodeHealthTrackerMeasurement) start(config *measurement.Config) error {
...
 
-	ctx := context.Background()
-	client := config.ClusterFramework.GetClientSets().GetClient()
-
-	lw := &cache.ListWatch{
-		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-			options.LabelSelector = selector.LabelSelector
-			options.FieldSelector = selector.FieldSelector
-			return client.CoreV1().Nodes().List(ctx, options)
+	reg, err := nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
mborsz, line 184: Is it safe to call AddEventHandler on already started informer?

Do we need to remove EventHandler at some point or will we be running it forever?

clusterloader2/pkg/imagepreload/imagepreload.go
supahero1: Hm true unecessary fallback for when nodeCounter fails a cast

PR #4317, opened 2026-08-26
clusterloader2: optimize DaemonSet measurement with sharedNodeIndexerFactory

supahero1:
#### What type of PR is this?

/kind bug

Smaller chunk of #4316 

Fixes #

### DIFF
--- a/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
+++ b/clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
@@ -226,7 +226,7 @@ func getDaemonSetNumSchedulableNodes(c clientset.Interface, podSpec *corev1.PodS
 	podSpecCopy := podSpec.DeepCopy()
 	addOrUpdateDaemonPodTolerations(podSpecCopy)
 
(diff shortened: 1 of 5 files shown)


clusterloader2/pkg/measurement/util/runtimeobjects/runtimeobjects.go
```diff
@@ -223,11 +223,15 @@ func GetReplicasFromRuntimeObject(c clientset.Interface, obj runtime.Object) (Re
 
 // getDaemonSetNumSchedulableNodes returns the number of schedulable nodes matching both nodeSelector and NodeAffinity.
 func getDaemonSetNumSchedulableNodes(c clientset.Interface, podSpec *corev1.PodSpec) (ReplicasWatcher, error) {
-	selector, err := metav1.LabelSelectorAsSelector(metav1.SetAsLabelSelector(podSpec.NodeSelector))
+	podSpecCopy := podSpec.DeepCopy()
+	addOrUpdateDaemonPodTolerations(podSpecCopy)
```
mborsz, line 227: Why are we adding addOrUpdateDaemonPodTolerations here?

IIUC this may be an attempt to fix a bug that in GetReplicasFromRuntimeObject, we call it only for unstructured -> daemonset path and skip for typed daemonset?

Is this correct?
If so, should we remove redundant calls from unstructured -> daemonset path?

PR #4343, opened 2026-09-01
Add --use-existing-prometheus for in-cluster PromQL

nehadubey1338:
Measurements currently skip Prometheus unless CL2 deploys its own kube-prometheus stack. Allow querying a Service already in monitoring, with optional --prometheus-service and --prometheus-proxy-scheme. Defaults stay prometheus-k8s over http.

#### What type of PR is this?
/kind feature

Measurements currently skip Prometheus unless CL2 deploys its own kube-prometheus stack. Allow querying a Service already in monitoring, with optional --prometheus-service and --prometheus-proxy-scheme. Defaults stay prometheus-k8s over http.

Fixes #


alaypatel07:
@nehadubey1338 please follow the steps to get a CLA, so the EasyCLA tests passes.

PR #4344, opened 2026-09-02
benchmark-list: add informer mode

p0lyn0mial:
i created the following eval for an agent to make sure the new mode works:

```
to confirm same resource, scope (cluster-wide), verb i ask it to use `apiserver_request_total` and `httplogs`

to confirm same encoding (`proto`) i asked it to checked the pod's (client-side) logs

to confirm the http mode (existing) has not been broken i asked it to rendered the templates
```

**next** i'd like to run a CI job and check the artifacts.

### DIFF
--- a/clusterloader2/testing/list/config.yaml
+++ b/clusterloader2/testing/list/config.yaml
@@ -8,6 +8,10 @@
 {{$listReplicas := DefaultParam .CL2_LIST_BENCHMARK_PODS 1}}
 {{$contentType := DefaultParam .CL2_LIST_BENCHMARK_CONTENT_TYPE "json"}}
 
+{{$mode := DefaultParam .CL2_LIST_BENCHMARK_MODE "http"}}
+{{$informerCount := DefaultParam .CL2_LIST_INFORMER_COUNT 1}}
+{{$enableWatchList := DefaultParam .CL2_LIST_ENABLE_WATCHLIST false}}
+
 name: list-benchmark
 namespace:
   number: 1
@@ -58,6 +62,19 @@ steps:
         bytes: {{$configMapBytes}}
         group: {{$configMapGroup}}
 
+{{if eq $mode "informer"}}
+- module:
+    path: modules/list-benchmark.yaml
+    params:
+      namePrefix: "list-configmaps-"
+      replicas: {{$listReplicas}}
+      contentType: {{$contentType}}
+      mode: {{$mode}}
+      resource: configmaps
+      apiVersion: v1
+      count: {{$informerCount}}
+      enableWatchList: {{$enableWatchList}}
+{{else}}
 - module:
     path: modules/list-benchmark.yaml
     params:
@@ -66,6 +83,7 @@ steps:
       uri: /api/v1/configmaps?resourceVersion=0
       namespaced: false
       contentType: {{$contentType}}
+{{end}}
 - module:
     path: /modules/measurements.yaml
     params:
@@ -114,6 +132,19 @@ steps:
       templateFillMap:
         group: {{$podGroup}}
 
+{{if eq $mode "informer"}}
+- module:
+    path: modules/list-benchmark.yaml
+    params:
+      namePrefix: "list-pods-"
+      replicas: {{$listReplicas}}
+      contentType: {{$contentType}}
+      mode: {{$mode}}
+      resource: pods
+      apiVersion: v1
+      count: {{$informerCount}}
+      enableWatchList: {{$enableWatchList}}
+{{else}}
 - module:
     path: modules/list-benchmark.yaml
     params:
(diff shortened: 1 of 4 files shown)


clusterloader2/testing/list/modules/list-benchmark.yaml
```diff
@@ -2,6 +2,7 @@
 {{$replicas := DefaultParam .replicas 0}}
```
p0lyn0mial, line 2: at some point we could rename `list-benchmark` to `informer-benchmark`

p0lyn0mial: i didn't want to create a new file to show the diff.

serathius: makes sense

clusterloader2/testing/list/deployment.yaml
```diff
@@ -1,7 +1,14 @@
...
-{{$image := DefaultParam .CL2_BENCHMARK_IMAGE "gcr.io/k8s-staging-perf-tests/perf-tests-util/request-benchmark:v0.0.10"}}
 {{$cpu := DefaultParam .cpu "1250m"}}
+{{$mode := DefaultParam .Mode "http"}}
+{{$image := DefaultParam .CL2_BENCHMARK_IMAGE "gcr.io/k8s-staging-perf-tests/perf-tests-util/request-benchmark:v0.0.14"}}
+# http mode streams responses to io.Discard, so 32Mi is plenty. informer mode
+# caches the full result set (~1GB per informer for 10k x 100KB configmaps),
+# so it needs a much larger limit that scales with .Count.
 {{$memory := DefaultParam .memory "32Mi"}}
+{{if eq $mode "informer"}}
+{{$memory = DefaultParam .memory "2Gi"}}
```
serathius, line 10: Right, one of the perks of using direct HTTP client was ability to discard, it allowed us to avoid maintaining the resource requests/limits, however at the point we would prefer accurate e2e latency of initializing informer.

clusterloader2/testing/list/config.yaml
```diff
@@ -114,6 +134,20 @@ steps:
...
+    path: modules/list-benchmark.yaml
+    params:
+      namePrefix: "list-pods-"
+      replicas: {{$listReplicas}}
+      contentType: {{$contentType}}
+      mode: {{$mode}}
+      resource: pods
+      apiVersion: v1
+      count: {{$informerCount}}
+      timeout: {{$benchmarkTimeout}}
```
serathius: Does the timeout here matters? Duration of benchmark is dictated by the wait step below, so I think we should avoid encouraging passing and usage of informer timeout.

serathius:
Overall lgtm, would like to see impact on benchmark metrics on the presubmit.

clusterloader2/testing/list/config.yaml
p0lyn0mial: given that the `watch-list` periodics/benchmarks were removed (the informer mode was used by these tests) we could drop the timeout arg altogether, wdyt?

p0lyn0mial:
`pull-perf-tests-benchmark-list-informer` job produced only [50](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/perf-tests/4344/pull-perf-tests-benchmark-list-informer/2095835086019629056/artifacts/APIResponsivenessPrometheus_list-benchmark_2026-09-04T11:39:10Z.json) requests for cluster-scoped configmaps whereas pull-perf-tests-benchmark-list-proto` produced around [146](https://storage.googleapis.com/kubernetes-ci-logs/logs/ci-kubernetes-benchmark-list-proto/2095847097529012224/artifacts/APIResponsivenessPrometheus_list-benchmark_2026-09-04T12:26:06Z.json)

the issue is that the informer mode sleeps 5s before starting a new informer. i've opened a fix for it  https://github.com/kubernetes/perf-tests/pull/4346 . PTAL.

clusterloader2/testing/list/config.yaml
serathius: Yes

serathius:
Compared results:

informers:
```
 {
      "data": {
        "Perc50": 7609.307194,
        "Perc90": 9817.075068,
        "Perc99": 14244.150633
      },
      "unit": "ms",
      "labels": {
        "Count": "258",
        "Resource": "configmaps",
        "Scope": "cluster",
        "SlowCount": "0",
        "Subresource": "",
        "Verb": "LIST"
      }
    },
```
HTTP
```
  {
      "data": {
        "Perc50": 12371.721903,
        "Perc90": 14474.34438,
        "Perc99": 14947.434438
      },
      "unit": "ms",
      "labels": {
        "Count": "250",
        "Resource": "configmaps",
        "Scope": "cluster",
        "SlowCount": "0",
        "Subresource": "",
        "Verb": "LIST"
      }
    },
```

PR #4346, opened 2026-09-04
request-benchmark: informer continuous load

p0lyn0mial:
this pr changes request-benchmark to:

1. run informer mode until killed, like the other modes
2. re-sync informers back-to-back instead of every 5s

I hope this will be enough to produce a load comparable to the http mode. 
the http mode also disables QPS by setting it to -1.

xref: https://github.com/kubernetes/perf-tests/pull/4344


util-images/request-benchmark/mode_informer.go
```diff
@@ -46,7 +44,6 @@ func runInformer(args []string) error {
 	kubeconfig := fs.String("kubeconfig", "", "path to kubeconfig.")
 	targetNamespace := fs.String("namespace", "", "namespace to run informers for. If empty will open on all namespaces.")
 	informerCount := fs.Int("count", 4, "the number of informers per namespace to run.")
-	testTimeout := fs.Duration("timeout", time.Minute, "timeout duration for the test")
```
p0lyn0mial, line 49: i've also decided to drop unused `timeout` flag.

util-images/request-benchmark/mode_informer.go
```diff
@@ -100,29 +94,31 @@ func runInformer(args []string) error {
...
-		if err != nil {
-			return false, err
-		}
-
-		klog.Infof("Waiting for gvr = %v informers to sync", targetGVR)
-		if ok := cache.WaitForCacheSync(ctx.Done(), informersSynced...); !ok {
-			return false, fmt.Errorf("timed out waiting for gvr %v informers to sync: %w", targetGVR, ctx.Err())
+	ctx := context.Background()
+	for {
+		if err = func(ctx context.Context) error {
```
serathius, line 99: Don't think this anonymous function is needed.

p0lyn0mial: the anonymous function is providing a per sync scope (cancelInformers()) before the next loop iteration runs. i can create a new helper function if that would be easier.

PR #4358, opened 2026-09-08
clusterloader2: use precise timestamps for ResourceClaim latency

### LINKED ISSUE #3889: clusterloader2: create_to_schedule etc are based on second-granularity measurements

Gaurav598:
## What this PR does

Uses informer receive timestamps for ResourceClaim and Pod creation measurements in `resourceclaim_allocation_latency.go` to preserve sub-second precision.

## Why

`metadata.creationTimestamp` has second-level precision, which can make short ResourceClaim lifecycle measurements inaccurate.

This affects measurements such as:

* `claim_allocation`
* `pod_create_to_claim_create`

The allocation path already uses the informer receive time, so this change aligns the ResourceClaim creation and Pod creation paths with the same approach.

## Changes

* Inject a `clock.Clock` into the ResourceClaim allocation latency measurement.
* Capture the informer receive time when ResourceClaims are observed.
* Use the captured receive time for `createPhase`.
* Capture the informer receive time when Pods are observed.
* Use the cached Pod receive time for `podCreatePhase`.
* Keep the `fetchPodCreateTime` fallback based on `CreationTimestamp` unchanged.

## Tests

Added regression tests using `FakeClock` covering:

* sub-second ResourceClaim allocation latency
* sub-second Pod-to-ResourceClaim creation latency
* already-allocated ResourceClaims being ignored

Tests run:

```text
go test ./pkg/measurement/common/slos/... -v -run 'TestResourceClaim'
go test ./pkg/measurement/common/slos/... -count=1
```

Both pass.

Fixes #3889

### DIFF
--- a/clusterloader2/pkg/measurement/common/slos/resourceclaim_allocation_latency.go
+++ b/clusterloader2/pkg/measurement/common/slos/resourceclaim_allocation_latency.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
 
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/perf-tests/clusterloader2/pkg/errors"
@@ -64,6 +65,7 @@ func createResourceClaimAllocationLatencyMeasurement() measurement.Measurement {
 		entries:          measurementutil.NewObjectTransitionTimes(resourceClaimAllocationLatencyMeasurementName),
 		queue:            workqueue.NewTyped[*claimEventData](),
 		podCreationTimes: make(map[string]time.Time),
+		clock:            clock.RealClock{},
 	}
 }
 
@@ -90,6 +92,7 @@ type resourceClaimAllocationLatencyMeasurement struct {
 	podCreationTimes map[string]time.Time
 	podCacheLock     sync.RWMutex
 	podGetCalls      int64
+	clock            clock.Clock
 }
 
 func (m *resourceClaimAllocationLatencyMeasurement) Execute(cfg *measurement.Config) ([]measurement.Summary, error) {
@@ -179,7 +182,9 @@ func (m *resourceClaimAllocationLatencyMeasurement) start(c clientset.Interface)
 }
 
 func (m *resourceClaimAllocationLatencyMeasurement) addEvent(_, obj interface{}) {
-	event := &claimEventData{obj: obj, recvTime: time.Now()}
+	// Capture the receive time in the informer callback, before the event enters the
+	// workqueue, so queue latency does not affect the measurement.
+	event := &claimEventData{obj: obj, recvTime: m.clock.Now()}
 	m.queue.Add(event)
 }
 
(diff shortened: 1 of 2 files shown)


isaac-dasan (commented):
I ran into this same issue and had written an equivalent patch before I saw this PR. I'd commented on #3889 back on Aug 30 asking to take it but never actually opened anything. No point filing a duplicate, so here's some validation data instead. The two implementations are functionally identical, so all of this applies directly to your branch.

### Verification against this PR's head (4f57b14)

I took the FakeClock regression tests I'd written separately and dropped them onto your branch. Both pass:

* sub-second latency when every object shares the same `creationTimestamp` second
* already-allocated claims getting ignored on first observation

`go build ./...` and `go vet ./pkg/measurement/common/slos/...` are both clean.

### Before and after on a real cluster

kind v1.36.4, 3 nodes, `dra-example-driver`, 8 GPU-claiming Job pods. Same config, same cluster, only the binary changes (master `4f4d01ea3` vs this PR):

| Metric | before | after |
|---|---|---|
| `claim_allocation` P50 | 636 ms | **84 ms** |
| `claim_allocation` P90 | 1110 ms | **111 ms** |
| `pod_create_to_claim_create` P50 | 0 ms | **15.6 ms** |
| `pod_create_to_claim_create` P90 | 0 ms | **35.2 ms** |

`pod_create_to_claim_create` reproduced the exact all-zero symptom from #3889.

The per-claim `claim_allocation` numbers on master show it best: `186, 314, 522, 636, 770, 868, 953, 1110 ms`. That's eight samples spread almost uniformly across [0, 1.1s], for pods whose real latency was between 63 and 111 ms. Classic `measured = true + frac(create_time)` with `frac ~ Uniform[0, 1s)`. With this change those same eight come out as `63, 74, 76, 84, 90, 94, 99, 111 ms`.

`PodGetCalls` was 0 on both runs, so the `fetchPodCreateTime` fallback that this PR leaves on `CreationTimestamp` never actually gets hit. Agree it's fine to leave alone.

### Two notes, neither blocking

1. `pod_create_to_claim_create` now subtracts stamps coming from two different informers, so what it really measures is `true + (claim_watch_lag - pod_watch_lag)`. `addPodEvent` runs inline in the handler while claims go through a workqueue, so the pod stamps will drift late under churn.
2. Together with the clamp at `phase_latency.go:143` (negative goes to 0), that biases the metric downward at high load instead of surfacing the skew.

Neither one touches `claim_allocation`, which is the SLO metric and pulls both endpoints from the same informer. Might still be worth a comment saying the metric is really controller reaction plus inter-watch skew, just so nobody over-reads it later.

LGTM from me, though I'm not an org member so that doesn't carry any Prow weight.

/cc @justinsb @alaypatel07 - this needs `/ok-to-test` from a member before CI will run.

Gaurav598:
Thanks @isaac-dasan for the independent validation and the real-cluster before/after data. This is very helpful, especially the reproduction of the all-zero `pod_create_to_claim_create` behavior and the improvement in `claim_allocation` measurements.

I agree with the two notes around cross-informer timestamp skew and the existing clamp behavior. I'll keep them in mind as follow-up considerations; I don't think they block this fix.

Thanks again for the thorough validation!

Gaurav598:
Hi @justinsb @alaypatel07, just following up on this PR.
@isaac-dasan has independently validated the patch, including the FakeClock tests and a real-cluster before/after run, and gave an LGTM. The PR is still waiting on /ok-to-test from a Kubernetes member before CI can run.

Whenever you get a chance, could you please take a look and, if appropriate, run /ok-to-test?
Thanks!

PR #4361, opened 2026-09-10
Add exemplar pod and compaction to etcd write throughput benchmark

serathius:
/kind feature

/cc @p0lyn0mial


serathius:
Python thing looks like flake. Fix in https://github.com/kubernetes/perf-tests/pull/4362

serathius:
What? Why [pull-perf-tests-verify-all-python](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/perf-tests/4361/pull-perf-tests-verify-all-python/2097994097347268608) fails now while it passed on https://github.com/kubernetes/perf-tests/pull/4362

PR #4363, opened 2026-09-10
List benchmark exemplar pod

p0lyn0mial:
In this PR:

1. I removed existing pod template
2. I took the pod definition from https://github.com/kubernetes/perf-tests/pull/4361
3. I adapted the pod template from 2 to work with CL2 (tested locally)


serathius:
Those additional fields were set for etcd benchmark because normally those are added and managed by apiserver. For apiserver benchmarks there is no need to write them.

p0lyn0mial:
> Those additional fields were set for etcd benchmark because normally those are added and managed by apiserver. For apiserver benchmarks there is no need to write them.

ah, ok, so the prev template was good enough. thanks for the info.

PR #4378, opened 2026-09-18
Move write throughput benchmark to 10KB pod

serathius:
/kind feature
/cc @p0lyn0mial


clusterloader2/testing/write-throughput/modules/initial-pods/pod.yaml
```diff
@@ -123,12 +470,237 @@ spec:
...
+          - key: ca.crt
+            path: ca.crt
+          name: kube-root-ca.crt
+      - downwardAPI:
+          items:
+          - fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.namespace
+            path: namespace
+status:
```
Jefftree, line 673: Does this actually do what we want? https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/pod/strategy.go#L92-L94. afaik it's cleared

serathius: You are right, might be hard to guarantee size of objects without running a apiserver :(

PR #4383, opened 2026-09-21
Add pod size test

serathius:
Think we need to start testing pod size so we can have more customization and reuse. First adding just test showing the pod size in manifest. 

/kind feature

/cc @p0lyn0mial @Jefftree


clusterloader2/testing/write-throughput/modules/initial-pods/size_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	pod, ok := obj.(*corev1.Pod)
+	require.True(t, ok, "decoded object must be *corev1.Pod")
+
+	protoSerializer := protobuf.NewSerializer(scheme.Scheme, scheme.Scheme)
+	var buf bytes.Buffer
+	err = protoSerializer.Encode(pod, &buf)
+	require.NoError(t, err)
+
+	require.Equal(t, 2068, buf.Len())
```
p0lyn0mial, line 63: does it mean that the size of the pod from the server will be at least 2068 ? (proto, no compression)

p0lyn0mial:
Is the new `TestPodSize` run in a CI job ?

serathius:
> Is the new `TestPodSize` run in a CI job ?

It already does in pull-perf-tests-verify-test:
```
ok  	k8s.io/perf-tests/clusterloader2/testing/write-throughput/modules/initial-pods	0.053s
```

